PR #247, opened 2017-12-20
Add check for null webview; rename confusing function

liuche:
Also checks another place where webview could be null and has crashed.

### DIFF
--- a/app/src/main/java/org/mozilla/focus/activity/MainActivity.java
+++ b/app/src/main/java/org/mozilla/focus/activity/MainActivity.java
@@ -512,12 +512,12 @@ public class MainActivity extends LocaleAwareAppCompatActivity implements OnUrlE
         }
     }
 
-    public void toggleCursor(boolean enabled) {
+    public void setCursorEnabled(boolean toEnable) {
         final FragmentManager fragmentManager = getSupportFragmentManager();
         final BrowserFragment browserFragment = (BrowserFragment) fragmentManager.findFragmentByTag(BrowserFragment.FRAGMENT_TAG);
 
-        isCursorEnabled = enabled;
-        browserFragment.toggleCursor(enabled);
+        isCursorEnabled = toEnable;
+        browserFragment.setCursorEnabled(toEnable);
     }
 
     // todo: naming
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/mozilla/focus/activity/MainActivity.java
```diff
@@ -512,12 +512,12 @@ private void initAmazonFactory() {
         }
     }
 
-    public void toggleCursor(boolean enabled) {
+    public void setCursorEnabled(boolean toEnable) {
```
mcomella: nit: `setIsCursorEnabled(boolean isEnabled)` is a little clearer but this is fine.

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
```diff
@@ -206,7 +207,11 @@ public void onValueChanged(@NonNull Boolean loading) {
                 }
 
                 final MainActivity activity = (MainActivity)getActivity();
-                activity.setCursorEnabled(!getWebView().getUrl().contains("youtube.com/tv"));
+                final IWebView webView = getWebView();
+                // Enable the cursor if website is youtube.com/tv
```
mcomella: nit; comment is redundant to code

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
```diff
@@ -206,7 +207,11 @@ public void onValueChanged(@NonNull Boolean loading) {
...
+                final boolean enableCursor = webView == null ? false :
```
mcomella: nit: I think nested ternaries throw up a red flag in my mind but I'm not sure if it's warranted.

I think it's just my caution to use ternaries carefully because they can be confusing (since they're just a garble of syntax).

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
```diff
@@ -206,7 +207,11 @@ public void onValueChanged(@NonNull Boolean loading) {
...
+                        (webView.getUrl() == null ? false : !webView.getUrl().contains("youtube.com/tv"));
```
mcomella: nit: Add a comment explaining why this is necessary, or that this is a wack-a-mole fix that we don't fully understand.

mcomella: Since we could rewrite this, this is the only really important nit, I think.

PR #256, opened 2017-12-22
Add constant for new m.youtube.com link

liuche:
Based on discussion with @Sdaswani over slack, about possibly changing the url. Leaving this to @Sdaswani to merge, if needed.

You need to still do the following:
1) Update the autocomplete urls (in app/src/main/assets/domains/* for each locale) with the new locale
2) Verify that the new url shows the "remote control" navigation page, rather than the "cursor" mobile site - if the mobile site doesn't show the "remote control" navigation page, then the page won't work (users won't be able to play, pause, fast forward, full-screen, etc).

### DIFF
--- a/app/src/main/java/org/mozilla/focus/utils/AppConstants.java
+++ b/app/src/main/java/org/mozilla/focus/utils/AppConstants.java
@@ -37,4 +37,5 @@ public final class AppConstants {
     }
 
     public static final String YOUTUBE_TILE_LINK = "https://m.youtube.com";
+    public static final String YOUTUBE_MATCH_STRING = "m.youtube.com";
 }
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/mozilla/focus/utils/AppConstants.java
```diff
@@ -37,4 +37,5 @@ public static boolean supportsDownloadingFiles() {
     }
 
     public static final String YOUTUBE_TILE_LINK = "https://m.youtube.com";
```
mcomella: nit: it'd be great if this was constructed from the match string, but wfm.

Sdaswani:
thanks for the PR, will use if necessary.

liuche:
Not going to use this PR looks like, closing for now.

PR #298, opened 2018-01-09
Fixes #266 - Eases cursor movement and adds friction

### LINKED ISSUE #266: Moving the cursor and selecting things using the remote is clunky
Via Review:

The blue dot is a disaster. - The meeting of buttons is a matter of luck. The resolution of the control is not adapted to the HDMI resolution of the screen. The point is too big. If the buttons are unfavorable you can click 4 buttons at the same time. Have Firefox on my laptop. This app has nothing to do with the Windows browser (except the name). .... disappointing

https://www.amazon.de/review/ROKQLKUOFG780/ref=cm_cr_dp_title?ie=UTF8&ASIN=B078B5YMPD

boek:
This should allow you to have much more precision when pointing and have a more natural acceleration curve.

### DIFF
--- a/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
+++ b/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
@@ -57,7 +57,7 @@ public class BrowserFragment extends WebFragment implements View.OnClickListener
     private static final int ANIMATION_DURATION = 300;
 
     private static final String ARGUMENT_SESSION_UUID = "sessionUUID";
-    private static final int SCROLL_VELOCITY = 45;
+    private static final int SCROLL_MULTIPLIER = 45;
 
     public static BrowserFragment createForSession(Session session) {
         final Bundle arguments = new Bundle();
@@ -677,7 +677,7 @@ public class BrowserFragment extends WebFragment implements View.OnClickListener
 
     private int getScrollVelocity() {
         int speed = (int)cursor.getSpeed();
-        return speed * SCROLL_VELOCITY;
+        return speed * SCROLL_MULTIPLIER;
     }
 
     public void cursorHitEdge(Edge edge) {
--- a/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
+++ b/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
@@ -57,7 +57,7 @@ public class BrowserFragment extends WebFragment implements View.OnClickListener
     private static final int ANIMATION_DURATION = 300;
 
     private static final String ARGUMENT_SESSION_UUID = "sessionUUID";
-    private static final int SCROLL_VELOCITY = 1200;
+    private static final int SCROLL_VELOCITY = 45;
 
     public static BrowserFragment createForSession(Session session) {
         final Bundle arguments = new Bundle();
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/mozilla/focus/widget/Cursor.java
```diff
@@ -79,15 +79,19 @@ public void stopMoving(Direction direction) {
         if (activeDirections.size() == 0) {
             handler.removeCallbacks(tick);
             moving = false;
+            speed = 0;
         }
     }
 
     private void move() {
+        speed++;
+        speed *= FRICTION;
```
mcomella: nit: I'm not sure this is most natural way to write this – it seems strange to modify the velocity by 1 and then multiply the whole velocity by a constant. Typically, I think you'd achieve this affect by modifying the acceleration (the 1 value you're adding) instead.

That being said, I'm not sure if ^ that's just more poor understanding of physics and how you'd come up with a better solution anyway.

boek:
@mcomella yeah, I agree. I don't fully understand it myself. Let me tweak it a little bit and we'll get a build to everyone else

boek:
@mcomella Try now. I also fixed the scroll speed to be relative to the cursor speed. I think its starting to feel a bit better.

@Sdaswani @liuche want to take a look as well? Possibly @lime124 too!

Sdaswani:
100% I'll check this out, thanks @boek .

Sdaswani:
Thanks for the demo @boek , good progress!

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
```diff
@@ -59,7 +60,7 @@
     private static final int ANIMATION_DURATION = 300;
 
     private static final String ARGUMENT_SESSION_UUID = "sessionUUID";
-    private static final int SCROLL_VELOCITY = 1200;
+    private static final int SCROLL_VELOCITY = 45;
```
mcomella: nit: this isn't really a `VELOCITY` – it's a "velocity -> scroll velocity" multipler.

app/src/main/java/org/mozilla/focus/widget/Cursor.java
```diff
@@ -24,7 +24,7 @@
 public class Cursor extends View {
 
     private float CURSOR_SIZE = 45;
-    private final int MAX_SPEED = 15;
+    private final int MAX_SPEED = 25;
```
mcomella: I actually meant the initial presses felt slow (is it just the remote latency though?) but that being said, this still feels a little better, I think.

mcomella:
Also, just FYI that if you press multiple directions, we have funky behavior:
- A diagonal will go faster than a single direction (it's because we travel at max speed in each direction, but should travel only at our max speed on the diagonal – Goldeneye 64 actually has the same behavior! :)
- Opposing directions (left, right) will cause it to go perpendicular (e.g. up).

Not major, but just wanted you to know.

app/src/main/java/org/mozilla/focus/widget/Cursor.java
```diff
@@ -91,9 +91,11 @@ private void move() {
         speed++;
         speed *= FRICTION;
         speed = Math.min(MAX_SPEED, speed);
+        boolean isMovingDiagnol = activeDirections.size() > 1;
+        float moveSpeed = isMovingDiagnol ? speed / 2 : speed;
```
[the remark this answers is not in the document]
boek: @mcomella Ah that makes much more sense. I will take a stab at it

boek:
@mcomella also re:

> Opposing directions (left, right) will cause it to go perpendicular (e.g. up).

This is because the remote will also press the down or up key when pressing two sides

`org.mozilla.tv.firefox.debug E/button prssed: [RIGHT, LEFT, DOWN]`

PR #302, opened 2018-01-10
Closes #295: enable VoiceView to work in WebView

### LINKED ISSUE #295: [a11y] Support VoiceView on websites
Navigational pointer (on sites besides youtube) overrides key presses, preventing user from navigating with VoiceView.

To be explicit, we should kill the cursor in a11y mode.

mcomella:
#295

VoiceView is a screen reader for visually-imparied users and is the replacement for Talkback on Fire OS.

The functionality is not perfect but I think it might be good enough for now. I filed new bugs which we can triage for the other wonky behavior I've found.


boek:
The code makes sense to me. Looking into how to test now

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
```diff
@@ -15,13 +15,13 @@
 import android.support.annotation.Nullable;
 import android.support.v4.app.FragmentManager;
 import android.text.TextUtils;
+import android.util.Log;
```
boek: Looks like a Log import slipped through

mcomella:
> Looking into how to test now

You can enable accessibility in the Settings and they give you a tutorial (note that you'll need audio).

Roughly, in default mode:
- dpad to navigate spatially
- double-tab menu to open menu
- clicking menu will announce the page contents
- play/pause, fast forward, rewind are used by the system too to adjust announcements
- Hold menu to enter review mode
- Hold back+menu to enter/exist voiceview (once enabled in settings)

Review mode (note: this is the primary way to navigate a web page; sometimes you get stuck on the first item with "character, word, item" and if so, advance forward an item, and try up/down again to change the granularity to something like "link"):
- left/right to navigate between focuses
- up/down changes granularity of focuses

app/src/main/java/org/mozilla/focus/ext/Context.kt
```diff
@@ -17,3 +18,11 @@ val Context.appVersionName: String?
         val packageInfo = packageManager.getPackageInfo(packageName, 0)
         return packageInfo.versionName
     }
+
+val Context.isVoiceViewEnabled: Boolean
+    get() {
+        // In my understanding, isTouchExplorationEnabled should represent VoiceView being enabled,
+        // whereas isEnabled says if the accessibility manager is enabled for other services.
+        val am = this.getSystemService(Context.ACCESSIBILITY_SERVICE) as? AccessibilityManager ?: return false
```
liuche: I'm assuming that we wouldn't need to listen for if the accessibility changes - you probably can't do that in the middle of the app running.

app/src/main/java/org/mozilla/focus/ext/Context.kt
```diff
@@ -17,3 +18,11 @@ val Context.appVersionName: String?
         val packageInfo = packageManager.getPackageInfo(packageName, 0)
         return packageInfo.versionName
     }
+
+val Context.isVoiceViewEnabled: Boolean
+    get() {
+        // In my understanding, isTouchExplorationEnabled should represent VoiceView being enabled,
+        // whereas isEnabled says if the accessibility manager is enabled for other services.
```
liuche: `More details: if you are strictly interested in whether TalkBack is enabled, use am.isTouchExplorationEnabled(). (If Select to Speak is enabled and TalkBack disabled, am.isEnabled() will still return true.)`

https://stackoverflow.com/a/12362545/582004

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
```diff
@@ -268,6 +257,27 @@ public void onChanged(Integer progress) {
         return view;
     }
 
+    /**
+     * Gets the current state of the application and updates the cursor state accordingly.
+     *
+     * Note that this pattern could use some improvements:
+     * - It's a little weird to get the current state from globals, rather than get passed in relevant values.
+     * - BrowserFragment.setCursorEnabled should be called from this code path, but that's unclear
```
liuche: mmm that's a good point - I think that's one of the things we need to get right in our refactor.

app/src/main/java/org/mozilla/focus/activity/MainActivity.java
```diff
@@ -82,6 +86,18 @@
     private boolean isDrawerOpen = false;
     private boolean isCursorEnabled = true;
 
+    private final AccessibilityManager.TouchExplorationStateChangeListener voiceViewStateChangeListener = new AccessibilityManager.TouchExplorationStateChangeListener() {
+        @Override
+        public void onTouchExplorationStateChanged(final boolean enabled) {
```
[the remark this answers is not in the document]
mcomella: For correctness, I think so – earlier, for `isVoiceViewEnabled`, we check `isAccessibilityEnabled && isTouchExplorationEnabled` (i.e. both). However, managing two listeners with different state sounded :( and, as you pointed out earlier, `isTouchExplorationEnabled` represents Talkback (which should be VoiceView too) so I really think the one flag is necessary.

I'll update `isVoiceViewEnabled` to use only the one flag for consistency.

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
```diff
@@ -211,7 +212,12 @@ public void onValueChanged(@NonNull Boolean loading) {
                 // Bandaid null checks, underlying issue #249
                 final boolean enableCursor = webView != null &&
                         webView.getUrl() != null &&
-                        !webView.getUrl().contains("youtube.com/tv");
+                        !webView.getUrl().contains("youtube.com/tv") &&
+                        getContext() != null &&
+                        !ContextKt.isVoiceViewEnabled(getContext()); // VoiceView has its own navigation controls.
+                if (enableCursor) {
+                    Log.d("lol", "what");
```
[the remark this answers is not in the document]
mcomella: this was removed in a later commit, afaict.

Issue #38, opened 2017-12-13
Drop malformed url requests for non-main frame (was: AmazonWebView does not have WebResourceRequest)


ekager:
shouldInterceptRequest method takes a string instead of WebResourceRequest with AmazonWebView. Look into if we need to create a WebResourceRequest here.

mcomella:
The only method we use is `request.isForMainFrame`, which I think we can check with the `TrackingProtectionWebViewClient.currentPageUrl`:

```java
        if (!request.isForMainFrame() &&
                !scheme.equals("http") && !scheme.equals("https")) {
            // Block any malformed non-http(s) URIs. WebView will already ignore things like market: URLs,
            // but not in all cases (malformed market: URIs, such as market:://... will still end up here).
            // (Note: data: URIs are automatically handled by WebView, and won't end up here either.)
            // file:// URIs are disabled separately by setting WebSettings.setAllowFileAccess()
            return new AmazonWebResourceResponse(null, null, null);
        }
```

I didn't have time to investigate whether or not this would be sufficient, however.

Note that the UrlMatcher for tracking protection whitelists first-party requests (against the `currentPageUrl` so checking for main frame requests is actually unnecessary, afaict).

mcomella:
liuche found where this was initially implemented and it contains useful comments: https://github.com/mozilla-mobile/focus-android/issues/106#issuecomment-281780919

mcomella:
[triage] We're going to wait on the WebView upgrade to see if this API is added back in #379.

mcomella:
This is handled by components now: they're probably doing the right thing. :)

PR #318, opened 2018-01-12
Closes #311: enable tracking protection

### LINKED ISSUE #311: Add tracking protection backend
We're blocked on the UX in #234, but we can start with the backend here.

mcomella:
The `isBlockingEnabled` flag should still work when we add the UI.

From my commit summary:

I think request.isForMainFrame() is unnecessary because of this line in
UrlMatcher.matches:

```java
        if (pageHost != null && pageHost.equals(resourceHost)) {
            return false;
        }
```

Which whitelists first party requests: this is assuming the `pageHost`, which
comes from `TrackingProtectionWebViewClient.currentPageUrl`, actually
represents the current page loading (and thus the main frame).

I noted that my Android TV emulator seems to block the same numbers of trackers
as Focus.

### DIFF
--- a/app/src/webview/java/org/mozilla/focus/webview/TrackingProtectionWebViewClient.java
+++ b/app/src/webview/java/org/mozilla/focus/webview/TrackingProtectionWebViewClient.java
@@ -112,6 +112,7 @@ public class TrackingProtectionWebViewClient extends AmazonWebViewClient {
         // The matcher code could be better named to make this apparent.
         if (blockedSiteMatcher.matches(resourceUri, currentPageUri)) {
             if (callback != null) {
+                // TODO: the variable this value updates is inaccurate: see #317.
                 callback.countBlockedTracker();
             }
             return new AmazonWebResourceResponse(null, null, null);
--- a/app/src/webview/java/org/mozilla/focus/webview/TrackingProtectionWebViewClient.java
+++ b/app/src/webview/java/org/mozilla/focus/webview/TrackingProtectionWebViewClient.java
@@ -102,19 +102,20 @@ public class TrackingProtectionWebViewClient extends AmazonWebViewClient {
             return new AmazonWebResourceResponse(null, null, null);
         }
 
-        final UrlMatcher matcher = getMatcher(view.getContext());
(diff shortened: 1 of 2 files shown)


app/src/webview/java/org/mozilla/focus/webview/TrackingProtectionWebViewClient.java
```diff
@@ -102,19 +102,20 @@ public AmazonWebResourceResponse shouldInterceptRequest(final AmazonWebView view
...
         }
 
-        final UrlMatcher matcher = getMatcher(view.getContext());
+        final UrlMatcher blockedSiteMatcher = getMatcher(view.getContext());
 
         // Don't block the main frame from being loaded. This also protects against cases where we
         // open a link that redirects to another app (e.g. to the play store).
-        final Uri pageUri = Uri.parse(currentPageURL);
-
-//        if ((!request.isForMainFrame()) &&
```
liuche: I dug through some code and it looks like the isForMainFrame was added to prevent non-browser links (like market://) from crashing on this matcher check.

If you verify that it still works, then I'm okay with it, but otherwise I'd leave it in.

https://github.com/mozilla-mobile/focus-android/issues/106#issuecomment-281780919

mcomella: I can't use `isForMainFrame` because with this version of the webView we get a string, rather than the expected type we get in Focus.

That being said, I'll dig in on your research – thanks for doing that (good idea!)!

mcomella: > it looks like the isForMainFrame was added to prevent non-browser links (like market://) from crashing on this matcher check.

Looking at the linked issue, it looks like the http/http(s) checks were added to prevent crashing with malformed URLs like (`market:://`) but isForMainFrame is unrelated. #38 is open for fixing the http(s) check, which is out of scope of this issue (and we haven't seen any problems for it yet - maybe AmazonWebView is different because we're on a TV).

PR #321, opened 2018-01-16
Hides the cursor while idle

boek:


app/src/main/java/org/mozilla/focus/widget/Cursor.java
```diff
@@ -26,6 +28,8 @@
     private final int VELOCITY = 4;
     private float CURSOR_SIZE = 45;
     private final int MAX_SPEED = 15;
+    private final int CURSOR_ALPHA = 102;
```
liuche: Nit: CURSOR_VISIBLE_ALPHA

app/src/main/java/org/mozilla/focus/widget/Cursor.java
```diff
@@ -78,11 +92,14 @@ public void stopMoving(Direction direction) {
         activeDirections.remove(direction);
         if (activeDirections.size() == 0) {
             handler.removeCallbacks(tick);
+            cursorHandler.postDelayed(cursorTick, 3000);
             moving = false;
         }
     }
 
     private void move() {
+        if (getAlpha() != 1f) { setAlpha(1f); }
```
mcomella: For correctness, I think we'd also need to cancel the active ViewPropertyAnimator.

app/src/main/java/org/mozilla/focus/widget/Cursor.java
```diff
@@ -50,14 +54,23 @@ public void run() {
...
 
+    private Handler cursorHandler = new Handler();
+    private Runnable cursorTick = new Runnable() {
+        @Override
+        public void run() {
+            animate().alpha(0f).setDuration(CURSOR_ANIMATION_DURATION);
+        }
+    };
+
+
```
mcomella: nit: only one newlines between things

app/src/main/java/org/mozilla/focus/widget/Cursor.java
```diff
@@ -50,14 +54,23 @@ public void run() {
...
     public Cursor(Context context, AttributeSet attrs) {
         super(context, attrs);
 
         isInit = true;
         // create the Paint and set its color
         paint = new Paint();
         paint.setStyle(Paint.Style.FILL);
-        paint.setAlpha(102);
+        paint.setAlpha(CURSOR_ALPHA);
```
mcomella: nit: -> CURSOR_MAX_ALPHA or similar

app/src/main/java/org/mozilla/focus/widget/Cursor.java
```diff
@@ -67,6 +80,7 @@ public Point getLocation() {
 
     public void moveCursor(Direction direction) {
         activeDirections.add(direction);
+        cursorHandler.removeCallbacks(cursorTick);
```
mcomella: If we stick with Handlers, I think we might want to remove the callbacks in `onStop` too – otherwise:
- The Runnable gets added to the handler queue and could run while the fragment is in the background (waste of resources)
- The Runnable has a strong reference to the Context/Activity (being an anonymous class of the View), which is a large object, creating a memory leak if the cursor is recreated often (this can happen on "configuration change", when the device is rotated or keyboard added, etc., for which Android destroys the active activity and recreates it; note: I think we override this behavior for Focus so maybe this is less relevant). This is a common Android problem ([here's where it was first mentioned](https://android-developers.googleblog.com/2009/01/avoiding-memory-leaks.html) with static variables).
- Keeping a strong reference, the Runnable can call animate on a detached view hierarchy – I don't think this would crash but it's a waste of resources and could be confusing while debugging if there are multiple runnables.

You could fix the memory leak (which is incrementally better) by not using an anonymous class and creating a `WeakReference` to the view (we already do this for Activity and Context in various places in the code base). But this is probably more complex than just calling onStart/onStop.

But then again, hooking this up to the onStart/onStop callbacks is a pain in the ass and sort of breaks encapsulation, so maybe it's not worth the complexity. @liuche What do you think?

app/src/main/java/org/mozilla/focus/widget/Cursor.java
```diff
@@ -50,14 +54,23 @@ public void run() {
     };
 
 
+    private Handler cursorHandler = new Handler();
+    private Runnable cursorTick = new Runnable() {
```
mcomella: I think we can use `ViewPropertyAnimator.setStartDelay` instead so we don't need to deal with Handlers/Loopers. Don't forget to cancel the animator in `onStart/onStop`

app/src/main/java/org/mozilla/focus/widget/Cursor.java
```diff
@@ -78,11 +92,14 @@ public void stopMoving(Direction direction) {
         activeDirections.remove(direction);
         if (activeDirections.size() == 0) {
             handler.removeCallbacks(tick);
+            cursorHandler.postDelayed(cursorTick, 3000);
```
liuche: Nit: make this a constant too

app/src/main/java/org/mozilla/focus/widget/Cursor.java
```diff
@@ -50,14 +54,23 @@ public void run() {
     };
 
 
+    private Handler cursorHandler = new Handler();
+    private Runnable cursorTick = new Runnable() {
+        @Override
+        public void run() {
+            animate().alpha(0f).setDuration(CURSOR_ANIMATION_DURATION);
```
liuche: Nit: since we're mixing alphas here (the alphas of the Paint and View, which I think is Android is really bad at perf-wise), I wonder if we want to just "setVisibility(GONE/VISIBLE)". Are people going to see this animation? (Is it too jerky?)

boek:
@liuche You definitely notice the animation. It didn't feel as intentional without it. But maybe that's okay? Makes the code quite a bit simpler for sure. I'll record a couple of videos so you both can see what I mean.

mcomella:
@boek I think the animation would be nice – we could optimize it if we see visible perf issues (we don't have a battery to worry about and we have specific device models we can test on).

boek:
@mcomella 

So I was able to remove the runnable and just use  `animate()` and it works well. My question is this: Does `animate()` return a new `PropertyAnimator` every time? Or is it a single object thats owned by the view? Do I need to store a reference to it to be able to cancel that specific one?

mcomella:
> Does animate() return a new PropertyAnimator every time?

I don't know! Let's find out!

http://androidxref.com/8.0.0_r4/xref/frameworks/base/core/java/android/view/View.java#24707

Looks like it returns the same instance! Good idea.

boek:
@mcomella That is a great resource!

Ready for round 2.

app/src/main/java/org/mozilla/focus/widget/Cursor.java
boek: Since the cursor alpha never changes (It's the entire view that animates out). Does this still make sense to change?

mcomella: Oh, I see – it's fine the way it is. :)

app/src/main/java/org/mozilla/focus/widget/Cursor.java
```diff
@@ -80,19 +85,30 @@ public void moveCursor(Direction direction) {
...
             speed = 0;
+
+            animate().alpha(VIEW_MIN_ALPHA)
+                    .setDuration(CURSOR_ANIMATION_DURATION)
+                    .setStartDelay(CURSOR_HIDE_AFTER_TIME);
         }
     }
 
     private void move() {
+        if (getAlpha() != VIEW_MAX_ALPHA) {
```
mcomella: I feel like this should be in `moveCursor`, in the `!moving` branch, so we're not calling this every frame:
```java
if (!moving()) {
    isMoving = true;
    setAlpha(VIEW_MAX_ALPHA);
    animate().cancel();
}
```

- Is there a reason this is done in `move` (i.e. each frame)?
- Why do we check the alpha? It feels like the alpha comparison here is being used as a proxy for `if (!moving)` (in my code snippet)
- Why are `setAlpha` and `cancel()` are being called from different branches?

boek: > * Why are setAlpha and cancel() are being called from different branches?

I added a couple of comments to the code to clarify. I think it could probably be more clear with the code, but I was struggling on coming up with a better solution

app/src/main/java/org/mozilla/focus/widget/Cursor.java
```diff
@@ -20,12 +20,19 @@
...
 
 public class Cursor extends View {
 
     private float CURSOR_SIZE = 45;
+
     private final int MAX_SPEED = 25;
     private final double FRICTION = 0.98;
+    private final int CURSOR_ALPHA = 102;
+    private final int CURSOR_ANIMATION_DURATION = 250;
+    private final long CURSOR_HIDE_AFTER_MILLIS = TimeUnit.SECONDS.toMillis(3); // Hides cursor after 3 seconds
```
mcomella: nit: comment seems redundant

app/src/main/java/org/mozilla/focus/widget/Cursor.java
```diff
@@ -71,28 +77,44 @@ public float getSpeed() {
 
     public void moveCursor(Direction direction) {
         activeDirections.add(direction);
-        
+
+        // If the cursor isn't moving start the move loop
```
mcomella: nit: I think you could do without this comment if you changed `handler` -> `animationFrameHandler`

app/src/main/java/org/mozilla/focus/widget/Cursor.java
```diff
@@ -71,28 +77,44 @@ public float getSpeed() {
...
         if (!moving) {
+
```
mcomella: nit: this newline is unnecessary in java land

app/src/main/java/org/mozilla/focus/widget/Cursor.java
```diff
@@ -71,28 +77,44 @@ public float getSpeed() {
...
+            // If the cursor is hidden, show it.
+            // If it's currently visible cancel any pending animations
```
mcomella: But what if we're currently animating so we're not at VIEW_MAX_ALPHA? We need to both cancel the animation and setAlpha.

I feel like it'd be easier to just do both unconditionally (within `!moving`).

app/src/main/java/org/mozilla/focus/widget/Cursor.java
```diff
@@ -71,28 +77,44 @@ public float getSpeed() {
...
+            if (getAlpha() != VIEW_MAX_ALPHA) {
+                setAlpha(VIEW_MAX_ALPHA);
+            } else {
+
```
mcomella: nit: unnecessary newline

app/src/main/java/org/mozilla/focus/widget/Cursor.java
mcomella: To be explicit, I'm not too concerned about the performance of doing these operations if they're unnecessary because we only do them once. :)

PR #335, opened 2018-01-17
Closes #312 - Disable telemetry for dev builds instead of by default.

### LINKED ISSUE #312: Disable telemetry on dev builds
It's currently enabled in `TelemetryWrapper.isTelemetryEnabled`. fwiw, this is what focus does:
```kotlin
            return preferences.getBoolean(
                    resources.getString(R.string.pref_key_telemetry), isEnabledByDefault) && !AppConstants.isDevBuild()
```

liuche:
I realized I had some code un-added code hanging out, which is the rest of this PR. Did not -a my commit x_x

### DIFF
--- a/app/src/main/java/org/mozilla/focus/telemetry/TelemetryWrapper.kt
+++ b/app/src/main/java/org/mozilla/focus/telemetry/TelemetryWrapper.kt
@@ -39,10 +39,6 @@ object TelemetryWrapper {
 
     private const val MAXIMUM_CUSTOM_TAB_EXTRAS = 10
 
-    private val isEnabledByDefault: Boolean
-        // Telemetry is enabled by default for Focus video except for dev builds
-        get() = !AppConstants.isDevBuild()
-
     private object Category {
         val ACTION = "action"
         val ERROR = "error"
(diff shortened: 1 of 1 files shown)


liuche:
The getBoolean call gets the pref if it's set, or fetches the default value if it's not set. If someone has a dev build and they've toggled the setting (thereby setting the preference), then their dev build could send telemetry.

mcomella:
> If someone has a dev build and they've toggled the setting (thereby setting the preference), then their dev build could send telemetry.

I'm concerned this will lead to accidental usage – e.g.: I toggle my dev build once, I move on to work on something else, never clearing my data, and I've been accidentally uploading the whole time – especially because I don't think we have a separate dev channel.

Given the lack of a dev channel, I'd prefer if developers needed to explicitly change code in order to enable these settings, e.g. a `DebugConstants` file (which could also do things like start the app on the Setting screen for faster feedback loops!).

@liuche What do you think?

mcomella:
Oh, seems there is a `setUpdateChannel`.

If someone is trying to debug telemetry on the server, this could be distracting if there are random devices uploading (but then again, they should probably try to limit it to themselves anyway because not having this precaution doesn't stop it from happening)...

I guess I'm concerned with how implicit it is and difficult to figure out whether or not the preference is set, as a developer.

I'll let you decide if you want to keep this approach but please document it (in the wiki) if you do.

Issue #325, opened 2018-01-17
Add telemetry for drawer navigation


mcomella:
Meta #319:
- Back
- forward
- home

Presumably we'd want refresh and settings too

PR #339, opened 2018-01-19
Add menu navigation telemetry

mcomella:
One thing that feels silly to me is that the enums passed into my events
reference external Strings for their values, but I like that all of the
possible Category, Action, Method, Values, and Extras are listed in one place
so I kept it in.

Let me know if you think it's silly, liuche!

---

- How often do users navigate back via the controller?
- How often do user go back, forward, and reload via the menu?
- How does back on menu compare to back on controller?
- How often do users use the menu to go home, or to settings?

- Establish baselines about how users interact with websites via back/forward/refresh navigation
- Establish baselines about how often users use the menu and home screens

None. Could do UR but it's less telling than an aggregated data set.

No.

5) List all proposed measurements and indicate the category of data collection for each measurement, using the Firefox [data c](https://wiki.mozilla.org/Firefox/Data_Collection)[ollection ](https://wiki.mozilla.org/Firefox/Data_Collection)[categories](https://wiki.mozilla.org/Firefox/Data_Collection) on the found on the Mozilla wiki.   

<table>
  <tr>
    <td>Tracking Bug #</td>
  </tr>
  <tr>
    <td>Click on menu nav button (back, forward, reload)</td>
    <td>Interaction data</td>
    <td>#325</td>
  </tr>
  <tr>
    <td>Click on controller back button</td>
    <td>Interaction data</td>
    <td>#325</td>
  </tr>
  <tr>
    <td>Click on menu app nav button (home, settings)</td>
    <td>Interaction data</td>
    <td>#325</td>
  </tr>
</table>

I want to permanently monitor this data (@bbinto).

* Which release channels?

All.

* Which countries?

All

* Which locales?

All

* Any other filters?  Please describe in detail below.

No.

8) Please provide a general description of how you will analyze this data.

- See if users use back/forward/reload in unexpected ways for the product (e.g. they reload pages frequently)
- See if users access back/forward/reload at all (e.g. maybe to annoying to access in menu)
- See how often users use home screen (e.g. because entering urls via controller is difficult)
- See how often users change settings (e.g. especially useful after a new feature was added with settings)

9) Where do you intend to share the results of your analysis?

sql.tmo dashboards

### DIFF
--- a/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
+++ b/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
@@ -422,12 +422,12 @@ public class BrowserFragment extends WebFragment implements View.OnClickListener
         if (canGoBack()) {
             // Go back in web history
             goBack();
+            TelemetryWrapper.browserBackControllerEvent();
         } else {
             getFragmentManager().popBackStack();
             SessionManager.getInstance().removeCurrentSession();
         }
 
-        TelemetryWrapper.browserBackControllerEvent();
         return true;
     }
 
--- a/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
+++ b/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
@@ -36,6 +36,7 @@ import org.mozilla.focus.session.NullSession;
 import org.mozilla.focus.session.Session;
 import org.mozilla.focus.session.SessionCallbackProxy;
 import org.mozilla.focus.session.SessionManager;
+import org.mozilla.focus.telemetry.ControllerBackLocation;
 import org.mozilla.focus.telemetry.TelemetryWrapper;
 import org.mozilla.focus.utils.Browsers;
 import org.mozilla.focus.utils.Direction;
@@ -427,6 +428,7 @@ public class BrowserFragment extends WebFragment implements View.OnClickListener
             SessionManager.getInstance().removeCurrentSession();
         }
 
+        TelemetryWrapper.controllerBackEvent(ControllerBackLocation.BROWSER);
         return true;
     }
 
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
```diff
@@ -427,6 +428,7 @@ public boolean onBackPressed() {
             SessionManager.getInstance().removeCurrentSession();
         }
 
+        TelemetryWrapper.controllerBackEvent(ControllerBackLocation.BROWSER);
```
liuche: How does this interact with the "esc" case in Youtube? How does it count "closing the sidebar"? Do we actually want to count the *combined* number of times people go back in web history + navigate through the app? I think that this should either be two events (one for pop, one for browser back) or just put it in 'canGoBack'.

My concern is that this will not be a useful probe because it aggregates data from too many actions. Better to be specific.

app/src/main/java/org/mozilla/focus/telemetry/TelemetryWrapper.kt
```diff
@@ -129,6 +129,8 @@ object TelemetryWrapper {
         val RELOAD = "refresh"
         val BACK = "back"
         val FORWARD = "forward"
+        val HOME = "home"
```
liuche: Do you listen for the physical menu button press?

mcomella: Not at present: I filed #354 for the menu.

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
mcomella: I changed this probe – please see the new commit and the commit summary for an explanation.

Barbara said I should ignore youtube ESC for now.

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
```diff
@@ -428,7 +427,7 @@ public boolean onBackPressed() {
             SessionManager.getInstance().removeCurrentSession();
         }
 
-        TelemetryWrapper.controllerBackEvent(ControllerBackLocation.BROWSER);
+        TelemetryWrapper.browserBackControllerEvent();
```
liuche: I think we don't care about navigating "back" out of the browser app so I would put this in the "canGoBack" fork. But, also doesn't matter TOO much because worst case this will throw off telemetry by 1.

PR #342, opened 2018-01-19
Add tracking protection toggle to sidepane

liuche:
Adding a WIP PR, first parts are ready for review, but have a couple of more things to fix:
1) Update toggle colors when I get them from @lime124 
2) Focus always assumes that the webview is alive when tracking protection is toggled. This is not the case in FTV. My hunch right now is that Session creation is where we should check for Tracking Protection state, but I will finish this up tomorrow.


liuche:
Closing this, see #357 instead.

PR #357, opened 2018-01-20
Enable Tracking Protection through toggle

liuche:
This is a simpler version of #342 that doesn't mess around with Session.


app/src/main/res/values/styles.xml
```diff
@@ -93,4 +93,8 @@
         <item name="android:textStyle">normal</item>
         <item name="android:textColor">#FFFFFF</item>
     </style>
+
+    <style name="SwitchStyle" parent="AppTheme">
```
mcomella: Should we inherit from `Widget.AppCompat.CompoundButton.Switch`?

liuche: No, this is to override a theme, where the colorAccent is defined, without updating the whole app's theme. I originally tried setting the thumb/track through a selector, but 1) this way is much simpler, and 2) that way doesn't work for r22 (and only works on 23+).

PR #360, opened 2018-01-20
Onboarding

boek:


app/src/main/java/org/mozilla/focus/fragment/FirstrunFragment.kt
```diff
@@ -0,0 +1,42 @@
...
+    override fun onAttach(context: Context?) {
+        super.onAttach(context)
+    }
+
+    override fun onDetach() {
+        super.onDetach()
+    }
+
+    companion object {
+        const val FIRSTRUN_PREF = "firstrun_shown"
```
liuche: Nit: rename this FIRSTRUN_PREF to FIRSTRUN_SHOWN_PREF.

app/src/main/java/org/mozilla/focus/activity/MainActivity.java
```diff
@@ -283,6 +284,10 @@ public void onValueChanged(@NonNull List<Session> sessions) {
                     // We have at least one session. Show a fragment for the current session.
                     showBrowserScreenForCurrentSession();
                 }
+
+                if (Settings.getInstance(MainActivity.this).shouldShowFirstRun()) {
```
liuche: Hm, should this need to be in the session manager callback? I think you could just put it at the end of onCreate (assuming all the views are created by then).

app/src/main/java/org/mozilla/focus/activity/OnboardingActivity.kt
```diff
@@ -0,0 +1,17 @@
+package org.mozilla.focus.activity
+
+import android.os.Bundle
+import android.support.design.widget.Snackbar
```
liuche: Don't need snackbar import?

app/src/main/java/org/mozilla/focus/activity/OnboardingActivity.kt
```diff
@@ -0,0 +1,17 @@
+package org.mozilla.focus.activity
```
liuche: Nit: Add license

app/src/main/res/layout/activity_onboarding.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="utf-8"?>
```
liuche: Nit: license

app/src/main/res/values/styles.xml
```diff
@@ -97,4 +97,7 @@
     <style name="SwitchStyle" parent="AppTheme">
        <item name="colorAccent">@android:color/white</item>
     </style>
+
+    <style name="AppTheme.AppBarOverlay" parent="ThemeOverlay.AppCompat.Dark.ActionBar" />
```
liuche: These are newly added, not just ws changes, I'd take them out.

app/src/main/res/layout/activity_onboarding.xml
```diff
@@ -0,0 +1,14 @@
...
+<android.support.design.widget.CoordinatorLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
liuche: Okay a little complicated (nesting a ConstraintLayout inside a CoordinatorLayout) but if it works that's fine :)

app/src/main/java/org/mozilla/focus/activity/OnboardingActivity.kt
```diff
@@ -0,0 +1,48 @@
...
+
+class OnboardingActivity : AppCompatActivity() {
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        setContentView(R.layout.activity_onboarding)
+
+        enable_turbo_mode.setOnClickListener { _ ->
+            setTurboMode(true)
+            setOnboardShown()
```
liuche: I think we can move `setOnboardShown` to the onCreate actually. They don't need to interact with it to not have to see it again.

app/src/main/java/org/mozilla/focus/activity/OnboardingActivity.kt
```diff
@@ -0,0 +1,48 @@
...
+
+    private fun setOnboardShown() {
+        PreferenceManager.getDefaultSharedPreferences(this)
+                .edit()
+                .putBoolean(ONBOARD_PREF, true)
+                .apply();
+    }
+
+    companion object {
+        const val ONBOARD_PREF = "onboard_shown"
```
liuche: Nit: rename to ONBOARD_SHOWN_PREF to be clear.

app/src/main/java/org/mozilla/focus/utils/Settings.kt
```diff
@@ -42,6 +43,9 @@ class Settings private constructor(context: Context) {
                     false); */
             false
 
+    fun shouldShowOnboarding(): Boolean =
+            !preferences.getBoolean(OnboardingActivity.ONBOARD_PREF, false)
```
liuche: Let's set this default value to true - worst case they never see it, rather than worst case they see it *every* time.

app/src/main/java/org/mozilla/focus/activity/OnboardingActivity.kt
```diff
@@ -0,0 +1,48 @@
...
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        setContentView(R.layout.activity_onboarding)
+
+        enable_turbo_mode.setOnClickListener { _ ->
+            setTurboMode(true)
+            setOnboardShown()
+            finish()
+        }
+
```
liuche: This needs one more line to set the default state:
 `enable_turbo_mode.isPressed = true`

app/src/main/java/org/mozilla/focus/activity/MainActivity.java
mcomella: Yeah – we initialize the sessions and set the screen according to the observer, e.g. `showHomeScreen` and `showBrowserScreenForCurrentSession` are also in here.

app/src/main/java/org/mozilla/focus/utils/Settings.kt
mcomella: We can't do this – we rely on the default being false in order to show it the first time.

app/src/main/java/org/mozilla/focus/activity/OnboardingActivity.kt
mcomella: This seems to work fine without the suggested change, both my device and liuche's.

mcomella:
Close in favor of #362.

PR #361, opened 2018-01-20
Move isBlockingEnabled to global pref

mcomella:
Two major changes:
- Centralize access to TP in a property
- Cache global access directly in TrackingProtectionViewWebViewClient.blockingEnabled with shared prefs getter

I decide to add this second cache step to potentially improve perf

### DIFF
--- a/app/src/webview/java/org/mozilla/focus/webview/TrackingProtectionWebViewClient.java
+++ b/app/src/webview/java/org/mozilla/focus/webview/TrackingProtectionWebViewClient.java
@@ -5,6 +5,7 @@
 package org.mozilla.focus.webview;
 
 import android.content.Context;
+import android.content.SharedPreferences;
 import android.graphics.Bitmap;
 import android.net.Uri;
 import android.os.AsyncTask;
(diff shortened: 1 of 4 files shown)


app/src/webview/java/org/mozilla/focus/webview/TrackingProtectionWebViewClient.java
```diff
@@ -56,15 +66,19 @@ protected Void doInBackground(Void... voids) {
         // background loading of the lists as early as possible.
         triggerPreload(context);
 
-        this.blockingEnabled = true;
+        final Settings settings = Settings.getInstance(context);
+        settings.getPreferences().registerOnSharedPreferenceChangeListener(this);
```
liuche: Does this listener go away if the webview is destroyed? (Could this cause a crash?) Do we need to unregister it?

liuche: Actually, looks like it's a weak reference. Can't hurt to unregister the listener though.
https://developer.android.com/reference/android/content/SharedPreferences.html#registerOnSharedPreferenceChangeListener(android.content.SharedPreferences.OnSharedPreferenceChangeListener)

PR #363, opened 2018-01-20
Update TP toggle if user turns it off during onboarding.

liuche:
This updates the TP toggle to be off in the sidebar if the user turns it off during onboarding.


app/src/main/java/org/mozilla/focus/activity/MainActivity.java
```diff
@@ -505,7 +506,17 @@ private void showSettingsScreen() {
 
     private void showOnboardingScreen() {
         final Intent intent = new Intent(this, OnboardingActivity.class);
-        startActivity(intent);
+        startActivityForResult(intent, ONBOARDING_REQ_CODE);
+    }
+
+    @Override
+    protected void onActivityResult(int requestCode, int resultCode, Intent data) {
+        if (requestCode == ONBOARDING_REQ_CODE) {
```
mcomella: nit: `if (... && ...)`

PR #365, opened 2018-01-20
Workaround to make TP enabled state obvious.

liuche:
Change the alpha.


app/src/main/java/org/mozilla/focus/activity/MainActivity.java
```diff
@@ -302,6 +303,14 @@ public void onValueChanged(@NonNull List<Session> sessions) {
         WebViewProvider.preload(this);
     }
 
+    private void setTrackingProtectionSwitchOn(boolean isEnabled) {
```
mcomella: nit: naming isn't clear what it's actually doing but whatever

PR #374, opened 2018-01-23
Closes #373 - Update onboarding button highlight color.

### LINKED ISSUE #373: Update Onboarding "Turbo Mode" selected color
The highlighted state is hard to see.

liuche:
Did some back and forth with @lime124 and ended up with this.

Probably should be tested on a real tv rather than just my external monitor screen.

80% opacity on photonBlue60.

screenshot_1516665101
screenshot_1516665096


app/src/main/res/drawable/nav_button_selector.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
+   - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
+
+<selector xmlns:android="http://schemas.android.com/apk/res/android">
+    <item android:state_focused="true" android:drawable="@color/onboarding_button_highlight"/>
+    <item android:state_pressed="true" android:drawable="@color/onboarding_button_highlight"/>
```
mcomella: Traditionally we'd darken the button when it's pressed and modern Android often does the ripple effect when held (e.g. click the turbo mode checkbox) – should we be doing something like that? I think it's there in case the UI lags so you know your action registered.

This seems acceptable because it's not a touch screen device and the screen disappears after you interact but it could be good to be consistent.

fwiw, typically you set the background with the ripple effect (`?attr:selectableItemBackground` or something) and it'll darken the item so you don't actually need to do that much.

mcomella:
Going to land as is so we can get this in 1.1.

PR #386, opened 2018-01-23
Closes #383: simplify settings, particularly a11y code

### LINKED ISSUE #383: [a11y] Simplify settings a11y swapping
See: https://github.com/mozilla-mobile/firefox-tv/issues/293#issuecomment-359905812

mcomella:


app/src/main/java/org/mozilla/focus/fragment/NewSettingsFragment.kt
```diff
@@ -90,24 +90,9 @@ class NewSettingsFragment : Fragment() {
...
-        } else {
-            viewToFocus = telemetryView
-            viewToDisableFocus = telemetryButton
-        }
-
-        val isFocused = telemetryButton.isFocused || telemetryView.isFocused
-        viewToFocus.isFocusable = true
-        viewToDisableFocus.isFocusable = false
-        if (isFocused) {
-            viewToFocus.requestFocus()
```
liuche: Do you need to request focus if shouldFocus is true? Or does setting isFocusable automatically focus?

mcomella: We don't want to request focus because the child view may not be focused when a11y is turned off.

Current behavior: if the child view is focused and we turn a11y off, it clears focus. This is fine because the user can press left-right to regain focus on elements nearby – I don't think it's worth the extra code to fix it. I'll add a comment.

If a11y is disabled -> enabled, it works as expected (child view is focused).

PR #409, opened 2018-01-25
DO NOT MERGE

liuche:
Testing out disabling UITests to see if it fixes BB.


liuche:
Closing, BB master is green from "rebuild without cache", has nothing to do with disabling unit tests through Gradle.

Issue #403, opened 2018-01-24
Make BrowserFragment modular


mcomella:

mcomella:
This is now more: clean up remaining BrowserFragment code. It's reliant on the navigation overlay landing.

mcomella:
Not actively working on it.

PR #423, opened 2018-01-29
Issue #403: modularize BrowserFragment some more

mcomella:
#403: In order to make more progress here, I think I need to tackle how other classes interface with this one: WebFragment, the Cursor, MainActivity, and whoever is calling `canGoBack` & friends.

### DIFF
--- a/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
+++ b/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
@@ -11,7 +11,6 @@ import android.view.KeyEvent
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
-import android.view.WindowManager
 import android.widget.FrameLayout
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
```diff
@@ -236,3 +186,50 @@ class BrowserFragment : WebFragment(), CursorEvent {
...
+    override fun onBlockingStateChanged(isBlockingEnabled: Boolean) {}
+
+    override fun onLongPress(hitTarget: IWebView.HitTarget) {}
+
+    override fun onEnterFullScreen(callback: IWebView.FullscreenCallback, view: View?) {
+        fullscreenCallback = callback
+        if (view == null) return
+
+        with (browserFragment) {
+            browserContainer.visibility = View.INVISIBLE
```
liuche: I liked the comment about why we're hiding browserContainer - for the browser UI and web content.

PR #424, opened 2018-01-30
Closes #396: [Refactor] Fix issues with static analysis tools (linter, style guide)

### LINKED ISSUE #396: [Refactor] Fix issues with static analysis tools (linter, style guide)
Run the static analysis tools and fix any problems so we can add those tools to the tree.

`.gradlew app:checkstyle app:testFocusWebvieDebugUnitTest`

and do the same for ~~app:findbugs~~, app:lint, ~~app:detektCheck~~, app:ktlint, app:pmd. (Though if there are a lot of errors for these, we should figure out if we want to use them, or not.)

(You can see the tasks available with `.gradlew tasks`)

bkmunar:
WIP


app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
```diff
@@ -7,11 +7,7 @@ package org.mozilla.focus.fragment
 import android.arch.lifecycle.Observer
 import android.os.Bundle
 import android.text.TextUtils
-import android.view.KeyEvent
-import android.view.LayoutInflater
-import android.view.View
-import android.view.ViewGroup
-import android.view.WindowManager
+import android.view.*
```
mcomella: This is weird – we should use not `*` imports. I wonder if the IDE just did this (because that's how it's configured by default) and checkstyle didn't complain because it wasn't configured to...

Anyway, I'm not that opinionated about not using * imports (I guess it's so you don't accidentally clobber the namespace?).

bkmunar: yah the IDE just did this

PR #428, opened 2018-01-30
Closes #414: rm WebFragment in favor of composability

### LINKED ISSUE #414: [refactor] Modularize/kotlinify WebFragment
Interop is tricky and BrowserFragment relies on WebFragment so it'd be great to kotlinify WebFragment.

Consider removing the inheritance too – I'm not sure how much value it provides.

mcomella:


mcomella:
Spoke IRL: composability is better for re-use and more flexible, but is hard to implement. One problem I have with WebFragment is that it's coupling with BrowserFragment is unclear (i.e. how `abstract inflateLayout` is called from `onCreateView` so BrowserFragment has to use a non-conventional lifecycle method to inflate its view).

We could create composability without the problems liuche describes (too many entry points to use the class correctly) by:
- Creating a `LifecycleListenerFragment : LocaleAwareFragment()` which calls the appropriate classes and BrowserFragment extends from. We have the same problems with inheritance as the existing implementation has but we create composability of the WebViewLifecycleManager, reduce the BrowserFragment and WebFragment coupling, and clarify the WebFragments role.
- We can move the fragment lifecycle listener to an annotation processor to generate code, which is how I think the framework does it. This is a lot of unknown territory on my part, however.

I'm undecided if we should pursue the first approach ^ or just rename WebFragment -> WebviewLifecycleFragment to better clarify what it's doing (since it's hard to figure out the difference between `Browser` and `Web` fragment).

mcomella:
Closing in favor of less error prone to use through inheritance: #462.

PR #453, opened 2018-02-02
Refactor navigation to overlay

liuche:
Move navigation code into overlay. This doesn't remove any of the drawer code yet, but this is becoming monolithic so going to merge into the tree frequently.

### DIFF
--- a/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
+++ b/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
@@ -23,10 +23,13 @@ import org.mozilla.focus.session.Session
 import org.mozilla.focus.session.SessionCallbackProxy
 import org.mozilla.focus.session.SessionManager
 import org.mozilla.focus.telemetry.TelemetryWrapper
+import org.mozilla.focus.telemetry.UrlTextInputLocation
 import org.mozilla.focus.utils.Direction
 import org.mozilla.focus.utils.Edge
 import org.mozilla.focus.web.IWebView
+import org.mozilla.focus.widget.BrowserNavigationOverlay
 import org.mozilla.focus.widget.CursorEvent
+import org.mozilla.focus.widget.InlineAutocompleteEditText
 
 private const val ARGUMENT_SESSION_UUID = "sessionUUID"
 private const val SCROLL_MULTIPLIER = 45
@@ -34,7 +37,7 @@ private const val SCROLL_MULTIPLIER = 45
 /**
  * Fragment for displaying the browser UI.
  */
-class BrowserFragment : WebFragment(), CursorEvent {
+class BrowserFragment : WebFragment(), CursorEvent, BrowserNavigationOverlay.NavigationEventHandler {
     companion object {
         const val FRAGMENT_TAG = "browser"
 
@@ -64,6 +67,12 @@ class BrowserFragment : WebFragment(), CursorEvent {
         initSession()
     }
 
+    override fun onCreateView(inflater: LayoutInflater?, container: ViewGroup?, savedInstanceState: Bundle?): View? {
+        val view = super.onCreateView(inflater, container, savedInstanceState)
+        view?.browserOverlay?.setNavigationEventHandler(this)
+        return view
+    }
+
     private fun initSession() {
         val sessionUUID = arguments.getString(ARGUMENT_SESSION_UUID)
                 ?: throw IllegalAccessError("No session exists")
@@ -91,6 +100,16 @@ class BrowserFragment : WebFragment(), CursorEvent {
         })
     }
 
+    override fun onEvent(event: BrowserNavigationOverlay.NavigationEvent, value: String?, autocompleteResult: InlineAutocompleteEditText.AutocompleteResult?) {
+        when (event) {
+            BrowserNavigationOverlay.NavigationEvent.BACK -> if (canGoBack()) goBack()
+            BrowserNavigationOverlay.NavigationEvent.FORWARD -> if (canGoForward()) goForward()
+            BrowserNavigationOverlay.NavigationEvent.RELOAD -> reload()
+            BrowserNavigationOverlay.NavigationEvent.HOME -> (activity as MainActivity).showHomeScreen()
+            BrowserNavigationOverlay.NavigationEvent.SETTINGS -> (activity as MainActivity).showSettingsScreen()
+            BrowserNavigationOverlay.NavigationEvent.LOAD -> (activity as MainActivity).onTextInputUrlEntered(value!!, autocompleteResult!!, UrlTextInputLocation.MENU)
+        }
+    }
     // TODO: if we convert WebFragment to kotlin, these can become abstract properties
     override fun getSession(): Session {
         return session
--- a/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
+++ b/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
@@ -30,6 +30,7 @@ import org.mozilla.focus.web.IWebView
 import org.mozilla.focus.widget.BrowserNavigationOverlay
 import org.mozilla.focus.widget.CursorEvent
 import org.mozilla.focus.widget.InlineAutocompleteEditText
+import org.mozilla.focus.widget.NavigationEvent
 
 private const val ARGUMENT_SESSION_UUID = "sessionUUID"
 private const val SCROLL_MULTIPLIER = 45
@@ -67,10 +68,9 @@ class BrowserFragment : WebFragment(), CursorEvent, BrowserNavigationOverlay.Nav
         initSession()
     }
 
-    override fun onCreateView(inflater: LayoutInflater?, container: ViewGroup?, savedInstanceState: Bundle?): View? {
-        val view = super.onCreateView(inflater, container, savedInstanceState)
-        view?.browserOverlay?.setNavigationEventHandler(this)
-        return view
+    override fun onViewCreated(view: View?, savedInstanceState: Bundle?) {
+        view?.browserOverlay!!.setNavigationEventHandler(this)
+        super.onViewCreated(view, savedInstanceState)
     }
 
     private fun initSession() {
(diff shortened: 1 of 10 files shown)


mcomella:
@liuche Build failed: ktlint. Now could be a good time to set up a pre-push hook! @bkmunar what tools have you landed so far? checkstyle and ktlint?

bkmunar:
yes!

liuche:
Screenshot with opacity, basic icons.
screenshot_1517521855

app/src/main/java/org/mozilla/focus/widget/BrowserNavigationOverlay.kt
```diff
@@ -5,32 +5,77 @@
...
 
     init {
         LayoutInflater.from(context)
                 .inflate(R.layout.browser_overlay, this, true)
+        listOf(navButtonBack, navButtonForward, navButtonReload, navButtonHome, navButtonSettings)
+                .forEach { setOnClickListener(this) }
+        setupUrlInput()
+    }
+
+    private fun setupUrlInput() {
```
mcomella: nit: you could make this slightly less verbose by making `navUrlInput == this`:
```kotlin
private fun setupUrlInput() = with (navUrlInput) {
    imeOptions = ...
```

app/src/main/java/org/mozilla/focus/widget/BrowserNavigationOverlay.kt
```diff
@@ -5,32 +5,77 @@
...
+        navUrlInput.imeOptions = navUrlInput.imeOptions or ViewUtils.IME_FLAG_NO_PERSONALIZED_LEARNING
```
mcomella: I think we can allow personalized learning - I think we just don't do it for Focus

app/src/main/java/org/mozilla/focus/widget/BrowserNavigationOverlay.kt
```diff
@@ -5,32 +5,77 @@
...
+        navUrlInput.setOnCommitListener {
+            val userInput = navUrlInput.text.toString()
+            if (!TextUtils.isEmpty(userInput)) {
```
mcomella: nit: kotlin has a fn which is more readable: `userInput.isNotEmpty()`

app/src/main/java/org/mozilla/focus/widget/BrowserNavigationOverlay.kt
```diff
@@ -5,32 +5,77 @@
...
+            }
+        }
+        val autocompleteFilter = UrlAutoCompleteFilter()
+        autocompleteFilter.load(context.applicationContext)
+        navUrlInput.setOnFilterListener{ searchText, view -> autocompleteFilter.onFilter(searchText, view) }
+
+        navUrlInput.setOnBackPressedListener{
+            if (visibility == View.VISIBLE) {
+                requestFocus()
+                navUrlInput.requestFocus()
```
mcomella: I don't understand what's intended here: we should only have one requestFocus. I'm guessing if back is pressed, we want to leave the navUrlInput? Should it be `clearFocus`?

app/src/main/java/org/mozilla/focus/widget/BrowserNavigationOverlay.kt
```diff
@@ -5,32 +5,77 @@
...
+    }
+
+    override fun onClick(view: View?) {
+        val event = when (view?.id) {
+            R.id.navButtonBack -> NavigationEvent.BACK
+            R.id.navButtonForward -> NavigationEvent.FORWARD
+            R.id.navButtonReload -> NavigationEvent.RELOAD
+            R.id.navButtonHome -> NavigationEvent.HOME
+            R.id.navButtonSettings -> NavigationEvent.SETTINGS
+            else -> null
```
mcomella: nit: pretty sure you could do `else -> return`, negating the need for a null check below

If not, you could also try:
```kotlin
val event = when (...) {
    ....
} ?: return

eventHandler?.onEvent(event)
```

But I'm not so sure about the readability there. Or even:
```kotlin
when (...) {
   ...
}?.let { event ->
    eventHandler?.onEvent(event)
}
```

*edit*: but I think my suggestions above might be better, to move complexity into the enum. Leaving this here for edification.

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
```diff
@@ -64,6 +67,12 @@ class BrowserFragment : WebFragment(), CursorEvent {
         initSession()
     }
 
+    override fun onCreateView(inflater: LayoutInflater?, container: ViewGroup?, savedInstanceState: Bundle?): View? {
```
mcomella: nit: I think it's more appropriate to override `onViewCreated`, which is called after `onCreateView` when the view hierarchy is guaranteed to be created.

tbh, I don't know it's separated (fuck fragments) – maybe so sub-class can inflate a layout, declaring final, but you can still get a callback? Or maybe it has to do with killing the Fragment's view hierarchy and reviving it without killing the whole Fragment?

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
```diff
@@ -64,6 +67,12 @@ class BrowserFragment : WebFragment(), CursorEvent {
...
+        val view = super.onCreateView(inflater, container, savedInstanceState)
+        view?.browserOverlay?.setNavigationEventHandler(this)
```
mcomella: nit: I think you should be more aggressive about casting to non-null types – if view is null, the overlay won't be set and we won't crash but we won't know about the problem and we may get a lot of unhappy users not using the app because these buttons just suddenly stop working in some circumstances

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
```diff
@@ -91,6 +100,16 @@ class BrowserFragment : WebFragment(), CursorEvent {
...
+    override fun onEvent(event: BrowserNavigationOverlay.NavigationEvent, value: String?, autocompleteResult: InlineAutocompleteEditText.AutocompleteResult?) {
+        when (event) {
+            BrowserNavigationOverlay.NavigationEvent.BACK -> if (canGoBack()) goBack()
+            BrowserNavigationOverlay.NavigationEvent.FORWARD -> if (canGoForward()) goForward()
+            BrowserNavigationOverlay.NavigationEvent.RELOAD -> reload()
+            BrowserNavigationOverlay.NavigationEvent.HOME -> (activity as MainActivity).showHomeScreen()
+            BrowserNavigationOverlay.NavigationEvent.SETTINGS -> (activity as MainActivity).showSettingsScreen()
+            BrowserNavigationOverlay.NavigationEvent.LOAD -> (activity as MainActivity).onTextInputUrlEntered(value!!, autocompleteResult!!, UrlTextInputLocation.MENU)
+        }
+    }
```
mcomella: nit: whitespace below

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
```diff
@@ -91,6 +100,16 @@ class BrowserFragment : WebFragment(), CursorEvent {
         })
     }
 
+    override fun onEvent(event: BrowserNavigationOverlay.NavigationEvent, value: String?, autocompleteResult: InlineAutocompleteEditText.AutocompleteResult?) {
+        when (event) {
+            BrowserNavigationOverlay.NavigationEvent.BACK -> if (canGoBack()) goBack()
+            BrowserNavigationOverlay.NavigationEvent.FORWARD -> if (canGoForward()) goForward()
+            BrowserNavigationOverlay.NavigationEvent.RELOAD -> reload()
+            BrowserNavigationOverlay.NavigationEvent.HOME -> (activity as MainActivity).showHomeScreen()
```
mcomella: BrowserFragment should not know about MainActivity - MainActivity should set a listener on BrowserFragment for the necessary events.

If we do it only for small cases, it may not be the end of the world (especially because we're not really testing BF) but it's a nice to have

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
```diff
@@ -91,6 +100,16 @@ class BrowserFragment : WebFragment(), CursorEvent {
         })
     }
 
+    override fun onEvent(event: BrowserNavigationOverlay.NavigationEvent, value: String?, autocompleteResult: InlineAutocompleteEditText.AutocompleteResult?) {
```
mcomella: nit: one of our linters may eventually get mad at the line lengths in this function. if you move NavigationEvent to the top-level in BrowserNavigationOverlay, these imports would be shorter

mcomella: I would prefer if this was set on a separate class with a BrowserFragment reference in order to separate responsibilities. fwiw, this is how swift conventionally does it (with the delegates as an extension at the bottom of the file). e.g.:
```kotlin
class BrowserNavigationOverlay() {
    ...
}

class FragmentEventHandler(private val browserFragment: BrowserFragment) {
    override onEvent ...
}
```

Then again, maybe it's awkward to keep a reference to BrowserFragment if you pass in a lambda like I suggested earlier... I'm not sure! Sorry I don't have all the best answers, just a lot of possibilities... :\

mcomella: I also think `onEvent` should be more descriptive (if you keep it) – if I was reading the code, all I would know about this fn is it's `BrowserFragment.onEvent` – what does that do? What kind of events? What happens when there are `onEvent` name collisions?

app/src/main/java/org/mozilla/focus/widget/BrowserNavigationOverlay.kt
```diff
@@ -0,0 +1,81 @@
...
+            if (visibility == View.VISIBLE) {
+                requestFocus()
+                navUrlInput.requestFocus()
+            }
+        }
+    }
+
+    override fun onClick(view: View?) {
+        val event = when (view?.id) {
+            R.id.navButtonBack -> NavigationEvent.BACK
```
mcomella: You could simplify this by moving the complexity into the enum class:
```kotlin
enum class NavigationEvent { ...
    companion object {
        fun fromViewClick(viewID: Int) = when (viewID) {
            ...
```

And the calling code becomes:
```kotlin
val event = NavigationEvent.fromViewClick(view?.id) ?: return
eventHandler?.onEvent(event)
```

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
liuche: Agreed - this is just a stop-gap for now, I'll update that in a later PR.

mcomella: fwiw, you can add a comment saying it's temporary so I don't nit you about it. :)

liuche:
Apparently running `./gradlew lint` doesn't run `ktlint`.

PR #456, opened 2018-02-02
NO MERGE - UX overlay

liuche:
@lime124 here's the basic overlay, the two colors that you will want to modify are:

browser_overlay_background for the color/opacity of the overlay
homeUrlBarBackground for the color of the search entry bar

The xml file itself is browser_overlay (which you can jump to directly with shift-shift + typing)


liuche:
screenshot_1517521855

liuche:
Closing this bc UX has a pull of this branch now.

PR #457, opened 2018-02-02
Closes #455: German translation of "Keep Turbo Mode Enabled" doesnt m…

### LINKED ISSUE #455: German translation of "Keep Turbo Mode Enabled" doesn't mean anything in German
This was reported by a German user here: https://bugzilla.mozilla.org/show_bug.cgi?id=1435201
Apparently, what is currently stated means "exit the turbo mode activated" in German, which doesn't mean anything.

The correct translation would be "Turbo Modus aktiviert lassen" (English meaning: "leave/keep the turbo mode activated")".

String can be found here: https://github.com/mozilla-mobile/firefox-tv/blob/master/app/src/main/res/values-de/strings.xml#L191

bkmunar:
…ean anything in German


Sdaswani:
Agree with @liuche 👍

liuche:
Filed #458 for the releng - @bkmunar since these are small changes you can walk the whole process to release :)

bkmunar:
@liuche

PR #462, opened 2018-02-03
Closes #414: Simplify WebFragment, keeping inheritance

### LINKED ISSUE #414: [refactor] Modularize/kotlinify WebFragment

mcomella:
We decided in #428 that my composability approach was too error prone.


app/src/main/java/org/mozilla/focus/fragment/WebFragment.kt
```diff
@@ -49,17 +49,13 @@ abstract class WebFragment : LocaleAwareFragment() {
...
-        if (session != null) {
-            webView!!.setBlockingEnabled(session.isBlockingEnabled)
+        webView = (view.findViewById<View>(R.id.webview) as IWebView).apply {
+            callback = createCallback()
+            setBlockingEnabled(session.isBlockingEnabled)
         }
+        isWebViewAvailable = true
 
-        if (session == null || !session.hasWebViewState()) {
+        if (!session.hasWebViewState()) {
```
liuche: Okay, so we don't need this null check anymore? It shouldn't be a `!session?.hasWebViewState()` ? I'm curious why it was there in the first place then...

EDIT: Looked in the history, this was added to handle multiple tabs, so this should be fine.

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
```diff
@@ -44,6 +44,12 @@ class BrowserFragment : WebFragment(), CursorEvent {
         }
     }
 
+    // WebFragment expects a value for these properties before onViewCreated. We use a getter
+    // for the properties that reference session because it is lateinit.
+    override lateinit var session: Session
```
liuche: When is the session set? I remember looking through the Session code and seeing it was pretty complicated - is SessionManager in charge of returning the saved session or creating a new session?

mcomella: In onCreate, I called `initSession`:

```kotlin
        val sessionUUID = arguments.getString(ARGUMENT_SESSION_UUID)
                ?: throw IllegalAccessError("No session exists")
        session = if (sessionManager.hasSessionWithUUID(sessionUUID))
            sessionManager.getSessionByUUID(sessionUUID)
        else
            NullSession()

        setBlockingEnabled(session.isBlockingEnabled)
        session.url.observe(this, Observer { url -> this@BrowserFragment.url = url })
        session.loading.observe(this, object : NonNullObserver<Boolean>() {
                val activity = activity as MainActivity
                updateCursorState()
                if (!loading && activity.isReloadingForYoutubeDrawerClosed) {
                    activity.isReloadingForYoutubeDrawerClosed = false

                    // We send a play event which:
                    // - If we're on the video selection page, does nothing.
                    // - If we're in a fullscreen video, will show the play/pause controls on the screen so
                    // we don't just see a black screen.
                    activity.dispatchKeyEvent(KeyEvent(KeyEvent.ACTION_UP, KeyEvent.KEYCODE_MEDIA_PLAY_PAUSE))
                }
            }
        })
```

so SessionManager returns an instance in `onCreate` – it could be immutable if fragments didn't have a lifecycle.

app/src/main/java/org/mozilla/focus/fragment/WebFragment.kt
mcomella: I think session should never be null but because we were using Java, it was hard to guarantee that and we had inconsistent code where session was sometimes checked and sometimes not. Now we can guarantee that. :)

I also think there was a separate BrowserFragment instance for each session, hence the final nature of it.

PR #464, opened 2018-02-05
Fixes #460 - Dismisses video player onPause

### LINKED ISSUE #460: Fix App Exit
Issue 1               : App exits after performing mic search during the video playback  and pressing back button
Issue reproducible    : All compatible Fire TV devices.            
Reproducibility Rate  :  5/5
Issue reproducible in live: NO
 
Steps to reproduce
1. Install and launch the app
2. Select YouTube
3. Scroll Down and select any of the videos
4. While the video is playing performing Mic search and invoke any of the apps like Calm radio
5. Press back button using Remote/controller
 
Actual Result: App exits to device home screen
Expected Result: App should resume the video
 
Attachment:FireFox_Youtube_EXIT_2015.txt
Video Link:https://www.amazon.com/clouddrive/share/JkDBex6SLuHlSYnwziyGWodydZVhpX77DdJ...

boek:
https://github.com/mozilla-mobile/firefox-tv/issues/460#issuecomment-363221205

### DIFF
--- a/app/src/webview/java/org/mozilla/focus/webview/SystemWebView.java
+++ b/app/src/webview/java/org/mozilla/focus/webview/SystemWebView.java
@@ -39,7 +39,6 @@ public class SystemWebView extends NestedWebView implements IWebView, SharedPref
     private Callback callback;
     private FocusWebViewClient client;
     private final LinkHandler linkHandler;
-    
 
     public SystemWebView(Context context, AttributeSet attrs, AmazonWebKitFactory factory) {
         super(context, attrs);
@@ -175,6 +174,14 @@ public class SystemWebView extends NestedWebView implements IWebView, SharedPref
         }
 
         client.notifyCurrentURL(url);
+    }
+
+    @Override
+    public void onPause() {
+        // #460 - Ensure any videos that are playing get dismissed before we background the application.
+        callback.onExitFullScreen();
+
+        super.onPause();
     }
 
     @Override


app/src/webview/java/org/mozilla/focus/webview/SystemWebView.java
```diff
@@ -177,6 +176,15 @@ public void loadUrl(String url) {
         client.notifyCurrentURL(url);
     }
 
+    @Override
+    public void onPause() {
+        // Ensure any videos that are playing get dismissed before we background the application.
+        // https://github.com/mozilla-mobile/firefox-tv/issues/460
```
mcomella: nit; Just say #460

app/src/webview/java/org/mozilla/focus/webview/SystemWebView.java
```diff
@@ -177,6 +176,15 @@ public void loadUrl(String url) {
...
+        callback.onExitFullScreen();
```
liuche: Is there something that needs to be called in onResume, to undo this if we end up coming back to the video without doing something else?

boek: @liuche The crash happens if the app gets backgrounded in any way (Just bringing up the voice control and then pressing back will cause it). I'm not sure of other cases `onPause` will get called that we will need to account for?

liuche: Okay, so bringing up the voice control and dismissing it will return the video to an expected state? If so, that seems fine for handling the onResume bc it seems like it just works.

PR #469, opened 2018-02-07
[UX] Navigation overlay

liuche:
DO NOT MERGE

@lime124 here's another PR if you want to see more updates - the cursor nav is a bit broken (I think) but menu button opens the screen now. Doesn't work on Youtube.


liuche:
Closing in lieu of #479

Issue #443, opened 2018-01-31
Allow removing of tiles


liuche:
Allow deleting existing home tiles, and new ones. All tiles will be considered "pinned" (even the defaults).

SVG for menu icon: https://gist.github.com/liuche/5ca87ac2c9925cdb0e81094fa803fdfa
(this is the same as the one used in #513 so don't add this svg twice)

String for option: "Remove"

liuche:
This is blacklisting tiles - keep in mind that there will be two cases: removing existing tiles, and then removing "pinned" tiles the user chooses.

liuche:
Assigning to @lime124 for some mocks on what this "deletion" interaction should be. Easiest thing would be long-press to open up a context menu, but maybe this is not too discoverable. Adding new UI to the tiles could be okay, if it's simple, and easy to select with the remote (e.g., doesn't get confusing with moving left/right between tiles).

lime124:
@liuche this is my line of thinking. when a tile is selected, the little circle with a menu icon appears. pressing menu opens up a content menu for that tile with the delete option. for now it's only delete but in the future it can hold other things.

unfortunately my sketch expired today and i'm still waiting on a license from IT so i'm totally locked out of my files. this was the best i could capture. and the send one is new tab in firefox show the context menu idea.

<img width="374" alt="screen shot 2018-02-16 at 8 52 01 pm" src="https://user-images.githubusercontent.com/287322/36338098-461168d8-135b-11e8-85f3-5ac517e591c5.png">

<img width="394" alt="screen shot 2018-02-16 at 8 51 48 pm" src="https://user-images.githubusercontent.com/287322/36338097-45fa86d6-135b-11e8-83c5-18b6d2f3a8fa.png">

lime124:
based off of @mcomella 's suggestion, eitan agreed we should use " Fire TV has orientation text we can use: it tells the user the layout of the page and what to press to interact" to let the person know they can press the menu button to get that mini context menu.

liuche:
@mcomella I'm not sure what "orientation text" is - does @bkmunar need to do anything else to make that work, or do we get it for free?

mcomella:
@liuche It's kinda free: https://developer.amazon.com/docs/fire-tv/implement-voiceview-accessibility-features-fire-os.html#implementing-orientation-text-for-voiceview

liuche:
[...]

PR #471, opened 2018-02-07
Closes #440: Load default tiles from config

### LINKED ISSUE #440: Load default tiles from config
Instead of hard-coding the tiles, they should be loaded from a config. This may eventually need to support locale-specific default tiles.

bkmunar:


app/src/main/java/org/mozilla/focus/fragment/HomeFragment.kt
```diff
@@ -55,7 +59,12 @@ class HomeFragment : Fragment() {
     }
 
     private fun initTiles() = with (tileContainer) {
-        adapter = HomeTileAdapter(onUrlEnteredListener)
+        var homeTiles = mutableListOf<JSONObject>()
+        val jsonArray = HomeTileUtils.loadDefaultTiles(context).getJSONArray("default_tiles")
```
liuche: nit: pull this out into a constant

In general, you don't want to hard-code strings into the file because 1) it makes it more error prone if you use it elsewhere, and 2) people have to know what string to search for, and where it's used, rather than just being able to see it at the top of the file as a constant.

app/src/main/java/org/mozilla/focus/fragment/HomeFragment.kt
```diff
@@ -18,11 +17,16 @@ import android.widget.ImageView
...
 import kotlinx.android.synthetic.main.fragment_home.*
+import org.json.JSONObject
 import org.mozilla.focus.R
 import org.mozilla.focus.autocomplete.UrlAutoCompleteFilter
 import org.mozilla.focus.telemetry.TelemetryWrapper
 import org.mozilla.focus.telemetry.UrlTextInputLocation
+import org.mozilla.focus.utils.HomeTileUtils
 import org.mozilla.focus.utils.OnUrlEnteredListener
+import android.graphics.BitmapFactory
+
```
liuche: nit: extraneous spaces

app/src/main/java/org/mozilla/focus/fragment/HomeFragment.kt
```diff
@@ -55,7 +59,12 @@ class HomeFragment : Fragment() {
     }
 
     private fun initTiles() = with (tileContainer) {
-        adapter = HomeTileAdapter(onUrlEnteredListener)
+        var homeTiles = mutableListOf<JSONObject>()
+        val jsonArray = HomeTileUtils.loadDefaultTiles(context).getJSONArray("default_tiles")
+        for (i in 0..(jsonArray.length() - 1)) {
```
liuche: A forEach here would be better - maybe @mcomella has an even better kotlin suggestion!

app/src/main/java/org/mozilla/focus/fragment/HomeFragment.kt
```diff
@@ -75,29 +84,30 @@ class HomeFragment : Fragment() {
     }
 }
 
-private class HomeTileAdapter(val onUrlEnteredListener: OnUrlEnteredListener) :
+private class HomeTileAdapter(val onUrlEnteredListener: OnUrlEnteredListener, homeTiles: MutableList<JSONObject>, context: Context) :
         RecyclerView.Adapter<TileViewHolder>() {
-
+    val context = context
     val tiles = listOf(
```
liuche: You're processing each tile again, after pulling it from disk - I think you should just do everything at once, and pass in the list, rather than forming it again.

This code was originally here because we were hard-coding the list here, but we can move this to the init.

liuche: Optional: also think a little about how you would incorporate #443 - where would you want to process that? When would you need to handle the blacklist?

app/src/main/java/org/mozilla/focus/fragment/HomeFragment.kt
```diff
@@ -75,29 +84,30 @@ class HomeFragment : Fragment() {
...
+            HomeTile(homeTiles[8].getJSONObject("tile_9").getString("url"), homeTiles[8].getJSONObject("tile_9").getString("title"), homeTiles[8].getJSONObject("tile_9").getString("img")), // sign in required
+            HomeTile(homeTiles[9].getJSONObject("tile_10").getString("url"), homeTiles[9].getJSONObject("tile_10").getString("title"), homeTiles[9].getJSONObject("tile_10").getString("img")) // sign in required
     )
 
     override fun onBindViewHolder(holder: TileViewHolder, position: Int) = with (holder) {
         val item = tiles[position]
-        titleView.setText(item.titleRes)
-        iconView.setImageResource(item.imageRes)
+        titleView.setText(item.title)
+        val bmImg = BitmapFactory.decodeStream(context.assets.open("defaults/" + item.imagePath))
```
liuche: Again, don't hard-code this string.

app/src/main/java/org/mozilla/focus/utils/HomeTileUtils.java
```diff
@@ -0,0 +1,28 @@
...
+
+import android.content.Context;
+import org.json.JSONObject;
+import java.io.InputStream;
+
+public class HomeTileUtils {
+    public static JSONObject loadDefaultTiles(Context context) {
+        String json;
+        try {
+            InputStream is = context.getAssets().open("defaults/DefaultTiles.json");
```
liuche: Same: hard-coded string

But actually, since this is a util, you probably don't want to put the string into the util at all (though you're welcome to make the argument for it!). I would say the caller should be in charge of doing this (because you utils to be generic), and because this will eventually used by different locales.

app/src/main/java/org/mozilla/focus/utils/HomeTileUtils.java
```diff
@@ -0,0 +1,28 @@
...
+            int size = is.available();
+            byte[] buffer = new byte[size];
+            is.read(buffer);
+            is.close();
```
liuche: Should you move this close into a `finally` block? What if there's an error in the read?

app/src/main/java/org/mozilla/focus/fragment/HomeFragment.kt
mcomella: You could create an extension function for JSONArray like, `JSONArray.forEach` so no one else is tortured by this in the future, but otherwise I think this is the only way to iterate with this API.

For completeness, another suggestion would be using JSON APIs that support iterators (like Klaxon) but we've already discussed this and you've already made that decision. :)

app/src/main/java/org/mozilla/focus/utils/HomeTileUtils.java
mcomella: If you do this in kotlin, you can make this read simpler ([according to SO](https://stackoverflow.com/a/39500046)):
```kotlin
val inputAsString = inputStream.bufferedReader().use { it.readText() }  // defaults to UTF-8
```

tbh, I think we should avoid adding any more Java files because it makes interop more difficult.

mcomella: And that means no need to handle the `try/catch/finally` cases

bkmunar:
@liuche

app/src/main/assets/defaults/DefaultTiles.json
```diff
@@ -0,0 +1,74 @@
+{
+  "default_tiles": [
+    {
+      "tile_0": {
+        "url": "https://youtube.com/tv",
+        "title": "YouTube",
```
liuche: Thinking a little more about the structure of this json object, we should have an "identifier" field, with a string that describes the title but without spaces (like "youtube" or "hollywoodReporter").

app/src/main/java/org/mozilla/focus/fragment/HomeFragment.kt
```diff
@@ -31,6 +32,9 @@ class HomeFragment : Fragment() {
     lateinit var urlBar: LinearLayout
     var onUrlEnteredListener = object : OnUrlEnteredListener {} // default impl does nothing.
     val urlAutoCompleteFilter = UrlAutoCompleteFilter()
+    val HOME_TILES_PREFIX = "defaults/"
```
liuche: This isn't technically a prefix, this is the directory, so HOME_TILES_DIR.

app/src/main/java/org/mozilla/focus/fragment/HomeFragment.kt
```diff
@@ -55,7 +59,16 @@ class HomeFragment : Fragment() {
     }
 
     private fun initTiles() = with (tileContainer) {
-        adapter = HomeTileAdapter(onUrlEnteredListener)
+        var homeTiles = mutableListOf<HomeTile>()
+        val inputAsString = context.assets.open(HOME_TILES_JSON_PATH).bufferedReader().use { it.readText() }  // defaults to UTF-8
+        val jsonArray = JSONObject(inputAsString).getJSONArray(HOME_TILES_JSON_KEY)
+        for (i in 0..(jsonArray.length() - 1)) {
+            val url = jsonArray.getJSONObject(i).getJSONObject("tile_" + i.toString()).getString("url")
```
liuche: Let's get rid of this "tile_" string concatenation, I think it's unnecessary and adds complexity - json arrays should be ordered, right?

app/src/main/java/org/mozilla/focus/fragment/HomeFragment.kt
```diff
@@ -55,7 +59,16 @@ class HomeFragment : Fragment() {
...
+            val title = jsonArray.getJSONObject(i).getJSONObject("tile_" + i.toString()).getString("title")
+            val imgPath = HOME_TILES_PREFIX + jsonArray.getJSONObject(i).getJSONObject("tile_" + i.toString()).getString("img")
+            homeTiles.add(HomeTile(url, title, imgPath))
```
liuche: Add the "identifier" field to the HomeTile object and add a comment about it.

Basically, having an identifier will make it easier for people to access and check the object - could also a) make this optional, or just use the url, up to you.

bkmunar:
@liuche

app/src/main/assets/defaults/default_tiles.json
```diff
@@ -0,0 +1,64 @@
+{
+  "default_tiles": [
+    {
+        "identifier": "youtube",
+        "url": "https://youtube.com/tv",
+        "title": "YouTube",
+        "img": "tile_youtube.png"
+    },
+    {
+        "identifier": "google",
```
liuche: nit: googleVideo

app/src/main/java/org/mozilla/focus/fragment/HomeFragment.kt
```diff
@@ -55,7 +59,17 @@ class HomeFragment : Fragment() {
     }
 
     private fun initTiles() = with (tileContainer) {
-        adapter = HomeTileAdapter(onUrlEnteredListener)
+        val homeTiles = mutableListOf<HomeTile>()
+        val inputAsString = context.assets.open(HOME_TILES_JSON_PATH).bufferedReader().use { it.readText() }
+        val jsonArray = JSONObject(inputAsString).getJSONArray(HOME_TILES_JSON_KEY)
+        for (i in 0..(jsonArray.length() - 1)) {
+            val url = jsonArray.getJSONObject(i).getString("url")
```
liuche: Just get this JSONObject once, and call the getString, etc methods on it.

bkmunar:
@liuche done

Issue #477, opened 2018-02-08
Add colors for Overlay button states (selected, on, off, disabled, etc)


liuche:
A WIP idea for what this might look like.

<img width="454" alt="screen shot 2018-02-06 at 2 11 56 pm" src="https://user-images.githubusercontent.com/287322/35926838-060d9a2e-0bde-11e8-9bfc-bc5996c8b17b.png">

liuche:
Not sure if "Focused and Disabled" is a relevant state - when a button is disabled it can't gain focus (e.g., if you have no history to go back to you will not be able to select the back button)

Issue #415, opened 2018-01-26
Re-eval SessionManager/Session/LiveData; consider kotlinify


mcomella:
I always found LiveData a little weird and hard to follow - should we replace it? Perhaps the structure of SessionManager and SessionCallbackProxy is the root of the problems?

liuche:
Cleaning up titles so [meta] tag is clearer.

mcomella:
We said less important than 2.0 work.

mcomella:
We don't use tabs and it hasn't gotte in the way yet - let's just close it.

PR #479, opened 2018-02-10
Nav overlay

liuche:
Adds the overlay, rips out the drawer.


liuche:
@lime124 this is the next working version, feel free to take a look if you want - there's some weird behavior on YT that I need to push a commit for, but everything else structural should work (minus polish, button states)

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
```diff
@@ -132,6 +134,10 @@ class BrowserFragment : IWebViewLifecycleFragment(), BrowserNavigationOverlay.Na
         cursor = null
     }
 
+    override fun isBackEnabled() = canGoBack()
```
mcomella: The reason I have a remove todo on these is because they should be called directly on the webView, not the browserFragment. This:
1) Limits BrowserFragment's reponsibility
2) Better encapsulates WebView functionality
3) Prevents BrowserFragment from getting bloated from all of its contained functionality (Home screen, WebView, Cursor, full screen videos, etc.)

So I'd prefer this to be set on the WebView instead.

That being said, there are some reasons for having some functionality in BrowserFragment:
1) The functionality exceeds WebViews responsibilities, e.g. pressing back may go back in the WebView, or it might pop the fragment stack to show home
2) It allows easier access through property observers (e.g. we want to update multiple components when a value changes)

app/src/main/java/org/mozilla/focus/widget/BrowserNavigationOverlay.kt
```diff
@@ -42,7 +43,14 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
                               autocompleteResult: InlineAutocompleteEditText.AutocompleteResult? = null)
     }
 
+    interface BrowserNavigationStateProvider {
+        fun isBackEnabled(): Boolean
+        fun isForwardEnabled(): Boolean
+        fun getCurrentUrl(): String?
```
mcomella: nit: if you made this `var url: String?` right in the interface, you wouldn't need an extra getter function in BrowserFragment.

app/src/main/java/org/mozilla/focus/widget/BrowserNavigationOverlay.kt
```diff
@@ -94,14 +102,31 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
         eventHandler = handler
     }
 
+    fun setBrowserNavigationStateProvider(provider: BrowserNavigationStateProvider) {
```
mcomella: nit: just add a setter to the property

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
```diff
@@ -202,7 +202,10 @@ private class BrowserIWebViewCallback(
 
     private var fullscreenCallback: IWebView.FullscreenCallback? = null
 
-    override fun onPageStarted(url: String) {}
+    override fun onPageStarted(url: String) {
+        browserFragment.browserOverlay.updateNavigationButtons()
```
mcomella: In focus, sebastian intended to remove these callbacks and use SessionObservers instead – I'm not sure this is desirable #415 but that's the way the code is now. So this translates to:
```kotlin
session.loading.observe(... {
    // do the thing when true
```

I think we already have such an observer in this file.

app/src/main/java/org/mozilla/focus/session/SessionCallbackProxy.java
```diff
@@ -32,6 +32,7 @@ public void onPageStarted(@NonNull String url) {
         session.setProgress(MINIMUM_PROGRESS);
 
         session.setUrl(url);
+        delegate.onPageStarted(url);
```
mcomella: nit: If you use the Observers, remove this (that's why the delegates weren't attached in the first place).

But your confusion is a good example of why I don't love these observers. :)

app/src/main/java/org/mozilla/focus/widget/BrowserNavigationOverlay.kt
```diff
@@ -127,6 +128,17 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
         navButtonForward.isFocusable = canGoForward
 
         navUrlInput.setText(navigationProvider?.getCurrentUrl())
+
+        updateFocusOnNavigation()
+    }
+
+    private fun updateFocusOnNavigation() {
```
mcomella: I think requesting focus in the same method where we set focusability is confusing, e.g. if I'm writing new code and I want to update the focused state but not change focus, I'd probably call `updateNavigationButtons` without realizing it might request focus.

For example, one bug I think might exist in your patch is that `onPageStarted` will request focus every time, which may not be desirable, e.g. we have the overlay open, the page moves loads a new URL after some timeout, and then the focus changes inside the overlay.

We should request focus specifically when it's needed, which is only when the drawer is initially shown, if i'm not mistaken.

app/src/main/java/org/mozilla/focus/widget/BrowserNavigationOverlay.kt
```diff
@@ -127,6 +128,17 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
...
+        when {
+            findFocus() != null -> return
+            navButtonBack.isEnabled -> navButtonBack.requestFocus()
```
mcomella: I think we should requestFocus on the parent of these views, which can use `android:descendantFocusability` to ensure it's children are focused first, rather than writing out the focus patterns explicitly in code. So basically, "if visibility changes so view is shown, requestFocus on parent, which delegates to first focusable child".

app/src/main/java/org/mozilla/focus/activity/MainActivity.java
```diff
@@ -114,148 +83,14 @@ protected void onCreate(Bundle savedInstanceState) {
...
-                findViewById(R.id.urlView).requestFocus();
-                isDrawerOpen = true;
-
-                // Stop cursor movement upon drawer opening
-                // Need to fix follow-up issue https://github.com/mozilla-mobile/focus-video/issues/219
-                final BrowserFragment browserFragment = (BrowserFragment) getSupportFragmentManager().findFragmentByTag(BrowserFragment.FRAGMENT_TAG);
-                if (browserFragment != null &&
-                        browserFragment.getCursor() != null) {
-                    // For all intents and purposes, covering the UI with a menu is the same as onPause.
-                    browserFragment.getCursor().onPause();
```
mcomella: Make sure this gets revived.

app/src/main/java/org/mozilla/focus/activity/MainActivity.java
```diff
@@ -114,148 +83,14 @@ protected void onCreate(Bundle savedInstanceState) {
...
-                // One workaround is to reload the page, which we do here.
-                final BrowserFragment browserFragment = (BrowserFragment) getSupportFragmentManager().findFragmentByTag(BrowserFragment.FRAGMENT_TAG);
-                if (browserFragment != null && browserFragment.isVisible() && browserFragment.getUrl().contains("youtube.com/tv")) {
-                    browserFragment.reload();
-                    isReloadingForYoutubeDrawerClosed = true;
-                }
-
-                if (browserFragment != null && browserFragment.getCursor() != null) {
-                    // For all intents and purposes, covering the UI with a menu is the same as onPause.
-                    browserFragment.getCursor().onResume();
```
mcomella: Revive this too

app/src/main/res/layout/activity_main.xml
```diff
@@ -2,73 +2,11 @@
...
-    android:fitsSystemWindows="true"
-    tools:openDrawer="start">
+    android:focusable="true">
+</LinearLayout>
 
-    <RelativeLayout
-        android:layout_width="match_parent"
-        android:layout_height="match_parent">
-
-        <!-- This view must be specified first so that it's focused first when other views (e.g. menu drawer)
```
mcomella: Which reminds me, we should test a11y functionality with the overlay too (maybe file a follow-up to verify good behavior?)

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
```diff
@@ -96,8 +99,8 @@ class BrowserFragment : IWebViewLifecycleFragment(), BrowserNavigationOverlay.Na
         session.loading.observe(this, object : NonNullObserver<Boolean>() {
             public override fun onValueChanged(loading: Boolean) {
                 val activity = activity as MainActivity
```
mcomella: nit: no longer necessary

app/src/main/java/org/mozilla/focus/widget/BrowserNavigationOverlay.kt
```diff
@@ -115,6 +115,11 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
         if (toShow) {
             navUrlInput.requestFocus()
             updateNavigationButtons()
+        } else {
+            // #393: Youtube doesn't refocus properly, so refresh
+            if (navUrlInput.text.contains("youtube.com/tv")) {
```
mcomella: This might be wrong - what if the user types in "youtube.com/tv" and closes the drawer without loading? The UI state doesn't always match the model state. For this reason, I think we should delegate to BrowserFragment's – it knows the true url.

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
```diff
@@ -191,6 +191,10 @@ class BrowserFragment : IWebViewLifecycleFragment(), BrowserNavigationOverlay.Na
             val toShow = !browserOverlay.isVisible()
             browserOverlay.setOverlayVisibleByUser(toShow)
             cursor?.setEnabledForCurrentState()
+            // Fix this youtube focus hack in #393
```
mcomella: nit: ws above

app/src/main/java/org/mozilla/focus/widget/BrowserNavigationOverlay.kt
```diff
@@ -33,18 +38,33 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
...
+    }
+
+    interface BrowserNavigationStateProvider {
+        fun isBackEnabled(): Boolean
+        fun isForwardEnabled(): Boolean
+        fun getCurrentUrl(): String?
     }
 
     private var eventHandler: NavigationEventHandler? = null
+    private var navigationProvider: BrowserNavigationStateProvider? = null
```
mcomella: nit: `navigationStateProvider`

app/src/main/java/org/mozilla/focus/widget/BrowserNavigationOverlay.kt
```diff
@@ -33,18 +38,33 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
...
+                              autocompleteResult: InlineAutocompleteEditText.AutocompleteResult? = null)
+    }
+
+    interface BrowserNavigationStateProvider {
+        fun isBackEnabled(): Boolean
+        fun isForwardEnabled(): Boolean
+        fun getCurrentUrl(): String?
     }
 
     private var eventHandler: NavigationEventHandler? = null
```
mcomella: nit: `navigationHandler`

app/src/main/java/org/mozilla/focus/widget/BrowserNavigationOverlay.kt
```diff
@@ -68,11 +89,66 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
...
 
     fun setNavigationEventHandler(handler: NavigationEventHandler) {
         eventHandler = handler
     }
+
+    fun setBrowserNavigationStateProvider(provider: BrowserNavigationStateProvider) {
+        navigationProvider = provider
+    }
+
+    fun isVisible(): Boolean {
```
mcomella: nit: can use in `setOnBackpressedListener`

app/src/main/java/org/mozilla/focus/widget/BrowserNavigationOverlay.kt
```diff
@@ -94,6 +94,11 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
         eventHandler = handler
     }
 
+    fun setOverlayVisible(toShow: Boolean) {
+        visibility = if (toShow) VISIBLE else GONE
+        if (toShow) navButtonBack.requestFocus()
```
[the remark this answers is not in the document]
liuche: Otherwise nothing would be focused. I think this would be fixed with the changes you suggested on using descendantFocusability, I'll give that a try.

app/src/main/java/org/mozilla/focus/widget/BrowserNavigationOverlay.kt
liuche: interface setters need to be public, so this didn't work.

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
```diff
@@ -177,6 +177,7 @@ class BrowserFragment : IWebViewLifecycleFragment(), BrowserNavigationOverlay.Na
                 && event.repeatCount == 0) {
             val toShow = !browserOverlay.isVisible()
             browserOverlay.setOverlayVisibleByUser(toShow)
+            cursor?.setEnabledForCurrentState()
```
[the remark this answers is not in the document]
liuche: Grouping them into a private function so one doesn't get called without the other. Doesn't make sense to pass the cursor reference into the overlay for encapsulation reasons.

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
liuche: Okay, @mcomella but in this case, if we can just make `canGoBack` and `goBack` private, then we can use them right? I agree with you that they don't need to be exposed but I've been using them as convenience functions, and I think that comment is confusing.

app/src/main/java/org/mozilla/focus/widget/BrowserNavigationOverlay.kt
liuche: Focus needs to be set every time the drawer is shown, because just making a view VISIBLE does not give it focus.

app/src/main/java/org/mozilla/focus/widget/BrowserNavigationOverlay.kt
```diff
@@ -42,9 +47,17 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
...
                 .inflate(R.layout.browser_overlay, this, true)
-        listOf(navButtonBack, navButtonForward, navButtonReload, navButtonHome, navButtonSettings)
-                .forEach { it.setOnClickListener(this) }
+        listOf(navButtonBack, navButtonForward, navButtonReload, navButtonHome, navButtonSettings,
+                turboButton)
+                .forEach {
+                    it.setOnClickListener(this)
+                    DrawableCompat.setTintList(it.drawable.mutate(),
+                            ContextCompat.getColorStateList(context, R.color.overlay_button_selector))
+                    // Inactive state is used for Turbo mode
```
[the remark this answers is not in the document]
liuche: I'll deal with this in #477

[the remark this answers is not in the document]
liuche: Wait I lied, switch to isChecked in #516 .

PR #484, opened 2018-02-13
I410: clean up tracker related code in WebViewClients

mcomella:


liuche:
@mcomella why should this be moved to FocusWebViewClient rather than TrackingProtectionClient?

mcomella:
The hierarchy is `FocusWebViewClient : TrackingProtectionClient`. Formerly, the callback was used in TPClient but I removed that code so the callback can move to FWVClient, from a functional perspective. Conceptually, TPClient does not need to own an IWebViewCallback in order to function so it's unnecessary to put there. This also simplifies the responsibility of TPClient – it now only blocks requests, rather than managing a callback reference.

PR #485, opened 2018-02-13
Updates l10n tooling

boek:
* Points to new Firefox TV l10n repos
* Renames Focus to Firefox for Fire TV in translation files

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -556,17 +556,26 @@ be temporary, and you can try again later.</li>
     <string name="menu_drawer_home">Home</string>
 
     <!-- Home tile strings -->
-    <string name="tile_youtube_tv">YouTube</string>
-    <string name="tile_google_video_search">Video Search</string>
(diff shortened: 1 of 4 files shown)


boek:
@liuche They seem to work just fine now. I have a branch which has some changes to the tool to create new `.po` files based on the existing `.xml` files. It might be worth turning that into a script, but it seemed to be such a one off problem since we forked focus AND manually entered strings which put the project into a weird state.

boek:
Responded to one last l10n nit and added a description to all the home tile strings.

app/src/main/res/values/strings.xml
```diff
@@ -556,17 +556,26 @@ be temporary, and you can try again later.</li>
     <string name="menu_drawer_home">Home</string>
 
     <!-- Home tile strings -->
+
+    <!-- Trademarked brand name: Does not need to be translated -->
```
liuche: AH! These should get removed by #440, so up to you if we should wait on that or if you want to land this now.

PR #496, opened 2018-02-16
 Closes #287: Loading indicator 

### LINKED ISSUE #287: Loading/Progress indicators
We need some kind of progress indicators to give the user feedback that their input has been processed. Some sites (like YouTube) handle this themselves, but most »normal« websites don't. After clicking a link or entering an address, nothing happens for a while or the screen just turns blank without any indication as to whether the browser is still working.

bkmunar:


liuche:
@bkmunar btw all your checks are failing

liuche:
@bkmunar is this for UX review, or for code review? (bc all the checks are also broken)

If UX only, please put UX NO MERGE somewhere in the title, or somewhere.

lime124:
@bkmunar based off what i saw on the emulator. we might have to iterate a bit more once it gets on the tv...

- [ ] 8dp padding between the animation and the left side, between the animation and the text, and between the text and the right side
- [ ] let's make the height of the bar itself 32dp
- [ ] Font color: 2a2a2e, font size 14sp
- [ ] font should be regular. is the app using sans-serif everywhere? or is it roboto?
- [ ] epang suggested a quick fade out when it disappears
- [ ] can we add a drop shadow? i wasn't able to play with that in the emulator

thanks!

lime124:
@bkmunar a couple more comments from @e-pang
- [ ] update the gif with a new one from epang (he's going to speed up the animation and make it smaller)
- [ ] are we taking advantage of the grey (connecting) and blue (loading) states?

After this it should be in great shape to push to master for people to test on their tv's via the live app testing. We might have to throw in other resolution if it doens't look good. Thanks!!

bkmunar:
@lime124 haven't delved into the grey one yet since I'd need to investigate more into what APIs detect loading vs. connecting in android

app/src/main/java/org/mozilla/focus/utils/GifImageView.java
```diff
@@ -0,0 +1,100 @@
+/* -*- Mode: Java; c-basic-offset: 4; tab-width: 4; indent-tabs-mode: nil; -*-
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+// Source taken from https://stackoverflow.com/questions/42898968/adding-gif-in-android-studio
```
mcomella: nit: move to javadoc comment at top of class, i.e.:
```java
/**
 * Source taken from ...
 */
public class GifImageView ...
```

app/src/main/java/org/mozilla/focus/utils/GifImageView.java
```diff
@@ -0,0 +1,100 @@
...
+            invalidate();
+        }
+    }
+
+    public void setGifImageResource(int id) {
+        mInputStream = mContext.getResources().openRawResource(id);
+        init();
+    }
+
+    public void setGifImageUri(Uri uri) {
```
mcomella: nit: unused.

app/src/main/java/org/mozilla/focus/utils/GifImageView.java
```diff
@@ -0,0 +1,100 @@
...
+import java.io.FileNotFoundException;
+import android.util.Log;
+
+public class GifImageView extends View {
+
+    private InputStream mInputStream;
+    private Movie mMovie;
+    private int mWidth, mHeight;
+    private long mStart;
+    private Context mContext;
```
mcomella: nit: you can use `getContext()` so `mContext` is not necessary.

app/src/main/java/org/mozilla/focus/utils/GifImageView.java
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    @Override
+    protected void onMeasure(int widthMeasureSpec, int heightMeasureSpec) {
+        setMeasuredDimension(mWidth, mHeight);
+    }
+
+    @Override
+    protected void onDraw(Canvas canvas) {
+
```
mcomella: nit: rm whitespace

app/src/main/java/org/mozilla/focus/utils/GifImageView.java
```diff
@@ -0,0 +1,100 @@
...
+        long now = SystemClock.uptimeMillis();
+
+        if (mStart == 0) {
+            mStart = now;
+        }
+
+        if (mMovie != null) {
+
```
mcomella: nit: rm ws

app/src/main/java/org/mozilla/focus/utils/GifImageView.java
```diff
@@ -0,0 +1,100 @@
...
+        if (attrs.getAttributeName(1).equals("background")) {
+            int id = Integer.parseInt(attrs.getAttributeValue(1).substring(1));
+            setGifImageResource(id);
+        }
+    }
+
+
+    private void init() {
+        setFocusable(true);
+        mMovie = Movie.decodeStream(mInputStream);
```
mcomella: nit: this inputStream needs to be closed once it's used:
```java
try {
    mMovie = Movie.decodeStream(mInputStream);
} finally {
    mInputStream.close();
}
```

In this case, mInputStream doesn't need to be a member variable (which would make it easy to re-use it when it's already closed) – it you can pass it as an arg to init.

app/src/main/java/org/mozilla/focus/utils/GifImageView.java
```diff
@@ -0,0 +1,100 @@
...
+import java.io.InputStream;
+import android.graphics.Movie;
+import android.graphics.Canvas;
+import android.util.AttributeSet;
+import android.os.SystemClock;
+import android.net.Uri;
+import java.io.FileNotFoundException;
+import android.util.Log;
+
+public class GifImageView extends View {
```
mcomella: Did you test this on all the Fire devices in the office? Movie doesn't seem well supported according to other SO users: https://stackoverflow.com/questions/3660209/display-animated-gif#comment11621337_7772170 but I'm glad we're trying a built-in solution.

app/src/main/java/org/mozilla/focus/widget/ProgressBar.kt
```diff
@@ -0,0 +1,53 @@
...
+                    progressBar.visibility = View.INVISIBLE
+                }
+            }
+        })
+    }
+
+    init {
+        LayoutInflater.from(context)
+                .inflate(R.layout.progress_bar, this, true)
+        val gifImageView = findViewById<View>(R.id.gifImageView) as GifImageView
```
mcomella: nit: If you import the Kotlin android extensions, it has references to all the views so you don't need to do `findViewById` (i.e. this line can just be removed)

app/src/main/java/org/mozilla/focus/widget/ProgressBar.kt
```diff
@@ -0,0 +1,53 @@
...
+
+class ProgressBar @JvmOverloads constructor(
+        context: Context, attrs: AttributeSet? = null, defStyle: Int = 0)
+    : LinearLayout(context, attrs, defStyle) {
+
+    private var browserFrag: BrowserFragment? = null
+
+    fun setBrowserFragment(browserFragment: BrowserFragment) {
+        this.browserFrag = browserFragment
+        val textView = findViewById<TextView>(R.id.url)
```
mcomella: nit: unnecessary with kotlin android extensions (see below)

app/src/main/res/drawable/progress_bar_background.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
```
mcomella: nit: license header

app/src/main/java/org/mozilla/focus/widget/ProgressBar.kt
```diff
@@ -0,0 +1,90 @@
...
+import org.mozilla.focus.fragment.BrowserFragment
+import org.mozilla.focus.utils.GifImageView
+import java.lang.ref.WeakReference
+import java.util.concurrent.TimeUnit
+
+private const val HIDE_MESSAGE_ID = 0
+private const val HIDE_ANIMATION_DURATION_MILLIS = 250L
+private val HIDE_AFTER_MILLIS = TimeUnit.SECONDS.toMillis(3)
+
+class ProgressBar @JvmOverloads constructor(
```
mcomella: nit: There's a built-in Android class called ProgressBar - can we call this `FirefoxProgressBar`, or similar?

app/src/main/java/org/mozilla/focus/widget/ProgressBar.kt
```diff
@@ -0,0 +1,90 @@
...
+        context: Context, attrs: AttributeSet? = null, defStyle: Int = 0)
+    : LinearLayout(context, attrs, defStyle) {
+
+    private val hideHandler = ProgressBarHideHandler(this)
+    private var browserFrag: BrowserFragment? = null
+
+    fun setBrowserFragment(browserFragment: BrowserFragment) {
+        this.browserFrag = browserFragment
```
mcomella: nit: we don't need to keep a reference to browserFragment. As such, we should remove this assignment and change the fn name to `initialize` or similar.

app/src/main/java/org/mozilla/focus/widget/ProgressBar.kt
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    private fun setMaxVisibility() {
+        animate().cancel()
+        alpha = 1f
+    }
+
+    private fun resetCountdown() {
+        hideHandler.removeMessages(HIDE_MESSAGE_ID)
+        hideHandler.sendEmptyMessageDelayed(HIDE_MESSAGE_ID, HIDE_AFTER_MILLIS)
```
mcomella: We want to hide the progress bar after the page is done loading, not after HIDE_AFTER_MILLIS, right?

app/src/main/res/layout/progress_bar.xml
```diff
@@ -0,0 +1,36 @@
...
+    android:layout_marginStart="15dp"
+    android:layout_marginTop="15dp"
+    android:layout_marginBottom="15dp"
+    android:layout_marginEnd="15dp">
+
+    <org.mozilla.focus.utils.GifImageView
+        android:id="@+id/gifImageView"
+        android:layout_height="wrap_content"
+        android:layout_gravity="center"
+        android:layout_weight="1"
```
mcomella: nit: you can remove weight here and below – both weights default to 0 (i.e. the same as explicitly setting them to 1).

app/src/main/res/layout/progress_bar.xml
```diff
@@ -0,0 +1,36 @@
...
+        android:layout_width="wrap_content"
```
mcomella: nit: width and height should always be placed together

app/src/main/res/layout/progress_bar.xml
```diff
@@ -0,0 +1,36 @@
...
+    android:background="@drawable/progress_bar_background"
+    android:layout_marginStart="15dp"
+    android:layout_marginTop="15dp"
+    android:layout_marginBottom="15dp"
+    android:layout_marginEnd="15dp">
+
+    <org.mozilla.focus.utils.GifImageView
+        android:id="@+id/gifImageView"
+        android:layout_height="wrap_content"
+        android:layout_gravity="center"
```
mcomella: nit: this has no effect inside a linearLayout, afaik. Rm here and below

app/src/main/java/org/mozilla/focus/widget/ProgressBar.kt
bkmunar: The reason why we used the Cursor implementation of disappearing was because the listener to the loading isn't reliable as it loads and "finishes loading" every time the url changes (and the url changes quite frequently depending on the entered url, e.g. www.youtube.com --> https://www.youtube.com --> https://www.youtube.com/#tv or smth)

When I just relied on the listener, it faded in and out every time the url changed, so chenxia led me to the cursor implementation on how to handle this case. Thoughts?

liuche: @bkmunar The countdown suggestion is for solving the problem of the loads/redirects. However, I don't think starting this countdown every time you start the load makes sense (what if the network is really slow?). You should still hook this into the "done loading" cycle.

app/src/main/res/layout/progress_bar.xml
bkmunar: This won't be centered if it doesn't have at least "center_vertical"

app/src/main/res/layout/progress_bar.xml
```diff
@@ -0,0 +1,36 @@
...
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
+   - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    android:orientation="horizontal"
+    android:layout_width="wrap_content"
+    android:layout_height="32dp"
+    android:elevation="1dp"
+    android:background="@drawable/progress_bar_background"
+    android:layout_marginStart="15dp"
```
liuche: Can this just be layout_margin? You don't need to enumerate each of these.

bkmunar: did not know that was a thing lolol i was wondering what it was

app/src/main/AndroidManifest.xml
```diff
@@ -30,7 +30,7 @@
                 android:name="android.support.FILE_PROVIDER_PATHS"
                 android:resource="@xml/provider_paths"/>
         </provider>
-
+
```
mcomella: nit: unnecessary ws

app/src/main/java/org/mozilla/focus/utils/GifImageView.java
```diff
@@ -0,0 +1,92 @@
+/* -*- Mode: Java; c-basic-offset: 4; tab-width: 4; indent-tabs-mode: nil; -*-
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ *
+ * Source taken from https://stackoverflow.com/questions/42898968/adding-gif-in-android-studio */
```
mcomella: nit: Sorry if I was unclear - this should not be in the license header – it should be in the class Javadoc:
```java
/*
 * License
 */

package ...

import ...

/**
 * Source taken from...
 */
public class GifImageView ... {
    ...
}
```

app/src/main/java/org/mozilla/focus/utils/GifImageView.java
```diff
@@ -0,0 +1,92 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import android.graphics.Movie;
+import android.graphics.Canvas;
+import android.util.AttributeSet;
+import android.os.SystemClock;
+
+public class GifImageView extends View {
+
+    private InputStream mInputStream;
```
mcomella: nit: this is unnecessary – you can just pass it as an arg to `init()`

app/src/main/java/org/mozilla/focus/widget/FirefoxProgressBar.kt
```diff
@@ -0,0 +1,90 @@
...
+        gifImageView.setGifImageResource(R.drawable.progress_loading_indicator)
+
+        // hardware acceleration needs to be disabled for gif support in FirefoxProgressBar
+        // http://android-er.blogspot.it/2014/03/play-animated-gif-with.html
+        setLayerType(View.LAYER_TYPE_SOFTWARE, null)
+    }
+
+    fun startUpdates() {
+        setMaxVisibility()
+        resetCountdown()
```
mcomella: I don't think we'd want to start the hide timer when we start loading, only when it finishes: remove this

app/src/main/java/org/mozilla/focus/widget/FirefoxProgressBar.kt
```diff
@@ -0,0 +1,90 @@
...
+        LayoutInflater.from(context)
+                .inflate(R.layout.firefox_progress_bar, this, true)
+        gifImageView.setGifImageResource(R.drawable.progress_loading_indicator)
+
+        // hardware acceleration needs to be disabled for gif support in FirefoxProgressBar
+        // http://android-er.blogspot.it/2014/03/play-animated-gif-with.html
+        setLayerType(View.LAYER_TYPE_SOFTWARE, null)
+    }
+
+    fun startUpdates() {
```
mcomella: nit: it's not really starting updates – it's setting the visibility (maybe just remove this method since it'll only call one method after the call below)

app/src/main/java/org/mozilla/focus/widget/FirefoxProgressBar.kt
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    init {
+        LayoutInflater.from(context)
+                .inflate(R.layout.firefox_progress_bar, this, true)
+        gifImageView.setGifImageResource(R.drawable.progress_loading_indicator)
+
+        // hardware acceleration needs to be disabled for gif support in FirefoxProgressBar
+        // http://android-er.blogspot.it/2014/03/play-animated-gif-with.html
+        setLayerType(View.LAYER_TYPE_SOFTWARE, null)
```
mcomella: I'm uncomfortable with the potential side effects of this:
- Does drawing in software over a hardware layer force that hardware layer to draw in software too? We need to test this
- Using the CPU to draw animations will slow down pageload performance (which is using the CPU), especially on these slower devices

This person wrote a script to convert GIFs to AnimationDrawables: https://medium.com/@tushar.acharya/converting-animated-gifs-to-animationdrawables-c6594084d7ad Perhaps we should verify the scripts are sound and use that instead. So we can land this on live app testing, file a follow-up? @liuche What do you think?

app/src/main/java/org/mozilla/focus/widget/FirefoxProgressBar.kt
```diff
@@ -0,0 +1,90 @@
...
+        resetCountdown()
+    }
+
+    private fun setMaxVisibility() {
+        animate().cancel()
+        alpha = 1f
+    }
+
+    private fun resetCountdown() {
+        hideHandler.removeMessages(HIDE_MESSAGE_ID)
```
mcomella: nit: I'm not sure this hide handler is necessary – you might be able to call `animate().setStartDelay(...)` directly. The reason I do it for the cursor is because every time the user presses the directional buttons, we'll allocate and throw away some objects, which can be really bad depending on how much they tap it. In this case, I don't think we loading=false will get called often but I could be wrong (e.g. page loads so =false, then javascript runs, starts page loads again, etc.). I'm not sure how WebView will handle these cases.

It seems like the safer approach is to leave it as is though. :)

mcomella:
btw, please add me for review again when patches are ready for re-review, otherwise it doesn't show up in my dashboard – thanks!

Issue #446, opened 2018-01-31
Add "Settings" icon to home screen


liuche:
Since the overlay will only be on the browser, we need to add a button to open the settings screen to the home page.

[tif says:] we're going to put it between the search bar and the logo. we may have to shrink the width of the search bar to make everything look nice. the settings icon should be the same color as the search bar background which i think is 50% white.

![home](https://user-images.githubusercontent.com/287322/36115490-16b31d00-0fe8-11e8-8565-d9054bd6c578.png)

PR #498, opened 2018-02-16
#446 - Adds settings button to home screen

boek:
image

I did notice a new icon in #446. Wasn't sure if I should add a new one or replace the existing one.


lime124:
@boek that's my fault, i used the wrong icon in my sketch file. please use the settings icon you have.

liuche:
@boek please fix lint errors! Possibly consider adding a git hook to run lint and ktlint before pushing.

lime124:
Also, the icon is 50% white - same as the text in the search box. Which it looks like you did.

We'll have to add a focus state and a press state in a circle. same size as nav overaly. LMK if you want me to review more.

app/src/main/java/org/mozilla/focus/fragment/HomeFragment.kt
```diff
@@ -24,13 +25,19 @@ import org.mozilla.focus.telemetry.TelemetryWrapper
...
 
+interface HomeDelegate {
+    fun didPressSettings()
+}
+
 /** The home fragment which displays the navigation tiles of the app. */
 class HomeFragment : Fragment() {
     lateinit var urlBar: LinearLayout
     var onUrlEnteredListener = object : OnUrlEnteredListener {} // default impl does nothing.
+    var delegate: WeakReference<HomeDelegate>? = null
```
mcomella: This weak reference is unnecessary: Java is not reference-counted. Objects are GC'd when they're no longer connected to the main object tree by a strong reference, no matter how strong references they have pointing at them (i.e. so MainActivity <-strong-> HomeFragment can both be GC'd if no one in the main object tree has a reference to either of them).

The case for weak references in Android relates to the Context object: an Activity is a context and it's really big so you don't want to keep around any Contexts that don't need to be there or else OOM looms near. When you do certain actions, like rotating the device, the Activity is recreated. Because it's so easy to recreate the Activity, if you keep strong references to the Activity/Context in objects that outlive the Activity/Context (e.g. an async task), you can end up with multiple Contexts (the new one and the invalid ones in async tasks) taking up a lot of space. It's well documented if you want to do more reading: https://android-developers.googleblog.com/2009/01/avoiding-memory-leaks.html

In this particular case, the Activity lifecycle is a superset of the Fragment lifecycle so we don't need to do weak references.

---

nit: In any case, creating a weak reference is really the responsibility of the class receiving the object, so I'd actually change this var to a `fun setDelegate(arg: ...) { delegate = new WeakReference<...>(arg) }`, even if it's more verbose.

---

You mentioned on Vidyo about trying to make this a function type – it's probably hard to use function types in Java so this seems fine. If you really wanted to try:
- I think you could actually find the function types Kotlin is compiling down to ([like this?](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.reflect/-k-function/index.html)) and instantiate one of those, but it's probably really verbose and not worth the effort.
- You could create a kotlin file that MainActivity calls into which will set the listener. Maybe an extension function?

mcomella: nit: This delegation pattern is the one you want to use, however, the Java naming is usually `Listener` and more generic, e.g. this might be a, `NavigationListener` with a method like, `onSettingsSelected` (note also: `on...` vs. `did...`). Perhaps the interface should be defined inside the outer class so that it's namespaced (because the naming is generic): HomeFragment.NavigationListener

Then again, Kotlin is new and cool and not crusty-old Java so maybe this should be a `HomeActionListener` or similar and live at the top-level. 🤔 I guess the downside is it's harder to rename with the IDE later.

app/src/main/java/org/mozilla/focus/fragment/HomeFragment.kt
```diff
@@ -46,6 +53,15 @@ class HomeFragment : Fragment() {
         // todo: saved instance state?
         initTiles()
         initUrlInputView()
+
+        settingsButton.alpha = 0.4f
```
mcomella: nit: you can block related code operating on one object with `with()`:
```kotlin
with (settingsButton) {
    alpha = 0.4f
    setOnFocusChangeListener { ...
    ...
}
```

It's not always more readable but could be helpful. It's especially awesome for things like:
```kotlin
override fun onViewCreated...
    ...
    initSettingsButton()
}

fun initSettingsButton() = with (settingsButton) {
    alpha = 0.4f
    ...
}
```

Which I might recommend here. I find it's these initializers that create the broken windows `[1]` since they quickly sprawl out of control because everyone adds a few lines at a time and suddenly it's multiple pages long and impossible to understand.

`[1]:` Reference to broken window theory

app/src/main/java/org/mozilla/focus/fragment/HomeFragment.kt
```diff
@@ -46,6 +53,15 @@ class HomeFragment : Fragment() {
...
+        settingsButton.setOnFocusChangeListener { v, hasFocus ->
+            v.alpha = if (hasFocus) 1.0f else 0.4f
```
mcomella: fwiw, this can also be done with StateListDrawables (which can reference `#hex` or `@color/...`, since they're inflated into ColorDrawables, which contain alpha values).

In this case, I don't think it's worth it because of the added indirection/complexity/duplication of colors (since the color needs to be written twice, with different alpha values) but just wanted to keep you aware.

boek:
Blocked by #509

mcomella:
@boek ktlint failure – you could try adding a push hook so this doesn't happen often! https://mail.mozilla.org/pipermail/firefox-focus-public/2018-February/000095.html

app/src/main/res/layout/fragment_home.xml
```diff
@@ -44,13 +46,22 @@
                 android:importantForAutofill="no"
                 tools:targetApi="o" /><!-- importantForAutofill is only supported on Android O+ -->
 
+        <ImageButton
+            android:id="@+id/settingsButton"
+            android:layout_width="40dp"
+            android:layout_height="40dp"
+            android:layout_marginEnd="50dp"
+            android:background="@android:color/transparent"
+            android:backgroundTint="@android:color/holo_red_dark"
```
mcomella: nit: I think bgTint is unused.

Why is background transparent necessary?

app/src/main/res/layout/fragment_home.xml
```diff
@@ -44,13 +46,22 @@
...
+            android:src="@drawable/ic_settings" />
+
         <ImageView
                 android:layout_width="wrap_content"
                 android:layout_height="36dp"
                 android:layout_gravity="center_vertical"
-                android:scaleType="centerInside"
-                android:layout_marginStart="50dp"
                 android:layout_marginEnd="10dp"
+                android:scaleType="centerInside"
```
mcomella: nit: don't move unrelated line.

PR #508, opened 2018-02-20
Clean up main

liuche:
Pull things into another file so MainActivity is smaller, and convert MainActivity to kotlin.


app/src/main/java/org/mozilla/focus/fragment/FragmentDispatcher.kt
```diff
@@ -0,0 +1,134 @@
...
+import org.mozilla.focus.session.Source
+import org.mozilla.focus.telemetry.TelemetryWrapper
+import org.mozilla.focus.telemetry.UrlTextInputLocation
+import org.mozilla.focus.utils.OnUrlEnteredListener
+import org.mozilla.focus.utils.UrlUtils
+import org.mozilla.focus.widget.InlineAutocompleteEditText
+
+class FragmentDispatcher {
+
+    companion object {
```
mcomella: nit: if a class contains no member var/fn, you can actually make it top-level object and make "static" calls without a companion object:
```kotlin
object FragmentDispatcher {
    ...
```

app/src/main/java/org/mozilla/focus/fragment/FragmentDispatcher.kt
```diff
@@ -0,0 +1,134 @@
...
+import org.mozilla.focus.R
+import org.mozilla.focus.session.SessionManager
+import org.mozilla.focus.session.Source
+import org.mozilla.focus.telemetry.TelemetryWrapper
+import org.mozilla.focus.telemetry.UrlTextInputLocation
+import org.mozilla.focus.utils.OnUrlEnteredListener
+import org.mozilla.focus.utils.UrlUtils
+import org.mozilla.focus.widget.InlineAutocompleteEditText
+
+class FragmentDispatcher {
```
mcomella: nit: I wonder if it'd be more appropriate to name this after the functionality it serves (`ScreenController` or `ScreenSwitcher`), though perhaps that doesn't make it clear that these transitions are getting added to the fragment backstack.

app/src/main/java/org/mozilla/focus/fragment/FragmentDispatcher.kt
```diff
@@ -0,0 +1,134 @@
...
+    companion object {
+        /**
+         * Loads the given url. If isTextInput is true, there should be no null parameters.
+         */
+        @JvmStatic
+        fun onUrlEnteredInner(urlStr: String, isTextInput: Boolean,
+                              autocompleteResult: InlineAutocompleteEditText.AutocompleteResult?,
+                              inputLocation: UrlTextInputLocation?,
+                              fragmentManager: FragmentManager, sessionManager: SessionManager,
+                              context: Context) {
```
mcomella: nit: I think context should always be the first arg, if it's present. You can fix this quickly in the IDE with the `Change signature` action – it'll update the declaration and the call sites (though maybe it'll mangle line lengths).

app/src/main/java/org/mozilla/focus/activity/InfoActivity.java
```diff
@@ -104,7 +104,7 @@ public void onClick(View v) {
     @Override
     public View onCreateView(String name, Context context, AttributeSet attrs) {
         if (name.equals(IWebView.class.getName())) {
-            final View view = WebViewProvider.create(this, attrs, MainActivity.factory);
+            final View view = WebViewProvider.create(this, attrs, MainActivity.Companion.getFactory());
```
mcomella: nit: Add `@JvmStatic` to this method declaration and remove the additional `.Companion...`

app/src/main/java/org/mozilla/focus/activity/MainActivity.kt
```diff
@@ -0,0 +1,179 @@
...
+            window.addFlags(WindowManager.LayoutParams.FLAG_SECURE)
+        }
+
+        window.decorView.systemUiVisibility = View.SYSTEM_UI_FLAG_LAYOUT_STABLE or View.SYSTEM_UI_FLAG_LAYOUT_FULLSCREEN
+
+        setContentView(R.layout.activity_main)
+
+        fragmentContainer = findViewById(R.id.container)
+
+        val intent = SafeIntent(intent)
```
mcomella: nit: I don't love how this section of code is basically double-spaced.

app/src/main/java/org/mozilla/focus/activity/MainActivity.kt
```diff
@@ -0,0 +1,179 @@
...
+import org.mozilla.focus.utils.ViewUtils
+import org.mozilla.focus.web.IWebView
+import org.mozilla.focus.web.WebViewProvider
+import org.mozilla.focus.widget.InlineAutocompleteEditText
+
+class MainActivity : LocaleAwareAppCompatActivity(), OnUrlEnteredListener {
+
+    private val sessionManager = SessionManager.getInstance()
+
+    private var fragmentContainer: View? = null
```
mcomella: nit: could use kotlin android extensions here (and remove findById)

PR #509, opened 2018-02-20
#477 - Changes the button styles

boek:
image

Colors aren't final, but we can tweak them once they are 🙌

### DIFF
--- a/app/src/main/res/values/styles.xml
+++ b/app/src/main/res/values/styles.xml
@@ -99,7 +99,10 @@
     </style>
 
     <style name="NavigationButton" parent="Widget.AppCompat.ImageButton">
+        <item name="android:layout_width">@dimen/button_size</item>
+        <item name="android:layout_height">@dimen/button_size</item>
+        <item name="android:layout_marginEnd">@dimen/button_right_margin</item>
         <item name="android:padding">12dp</item>
-        <item name="android:background">?android:attr/selectableItemBackgroundBorderless</item>
+        <item name="android:background">@drawable/circle_button_background</item>
     </style>
 </resources>
--- a/app/src/main/res/values/styles.xml
+++ b/app/src/main/res/values/styles.xml
@@ -101,8 +101,9 @@
     <style name="NavigationButton" parent="Widget.AppCompat.ImageButton">
         <item name="android:layout_width">@dimen/button_size</item>
         <item name="android:layout_height">@dimen/button_size</item>
-        <item name="android:layout_marginEnd">@dimen/button_right_margin</item>
+        <item name="android:layout_marginEnd">24dp</item>
         <item name="android:padding">12dp</item>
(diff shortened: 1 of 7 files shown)


boek:
One question I have is about: https://github.com/mozilla-mobile/firefox-tv/pull/509/files#diff-e4590692aa6da9dc0da05cbe72efa571R51 — I've noticed we've intermixed hard coded size values and the constants in this file. Do we want to move back to using constants now that we have more time?

liuche:
Yeah - I think if it's a value that will reasonably be reused in places (like button size, or whatever), we should pull it out so it's easy for someone to use it if they add another button. But things like margins or padding of individual items will probably be too much trouble, if they pretty much only apply to one place.

When to pull out dimens is a hard question on Android ;)

boek:
Fixed the CI errors

mcomella:
My opinion: values should stay hard-coded unless they're intended to be re-used or overridden in some way. In theory, you can pull out everything from the layout files in Android (i.e. make them constants) but then 1) it makes it hard to see a layout file and understand what the layout should look like and 2) makes it hard to update by jumping around to many files (since there's one for every type of resource, nevermind overriding).

fwiw, not required reading, but I first got these ideas about how to write layout/styles from this article: http://blog.danlew.net/2014/11/19/styles-on-android/

app/src/main/res/values/styles.xml
```diff
@@ -99,7 +99,10 @@
     </style>
 
     <style name="NavigationButton" parent="Widget.AppCompat.ImageButton">
+        <item name="android:layout_width">@dimen/button_size</item>
```
mcomella: In general, be very careful about putting layout_width and height in a style – it makes it impossible to look at a layout and figure out how what the views should look like without jumping into another file. If multiple views in the file put their layout_width/height in styles, the problem only escalates.

This looks like an acceptable use case because there is so much duplication and it's pretty obvious what these buttons look like.

app/src/main/res/values/styles.xml
```diff
@@ -99,7 +99,10 @@
...
+        <item name="android:layout_height">@dimen/button_size</item>
+        <item name="android:layout_marginEnd">@dimen/button_right_margin</item>
```
mcomella: Now that width/height/marginEnd are only used in one place, I don't think we need them to read from dimens anymore.

app/src/main/res/values/styles.xml
```diff
@@ -99,7 +99,10 @@
...
         <item name="android:padding">12dp</item>
-        <item name="android:background">?android:attr/selectableItemBackgroundBorderless</item>
+        <item name="android:background">@drawable/circle_button_background</item>
```
mcomella: nit: ask Tif if we want to keep the ripple effect when the button is pressed

PR #518, opened 2018-02-22
Closes #516 - Make Checkable ImageButton for turbo overlay button.

### LINKED ISSUE #516: Use "isChecked" instead of "isActivated" in Turbo overlay button
Follow-up to #444 - in order to handle a11y correctly, we should not use isActivated.

liuche:


app/src/main/java/org/mozilla/focus/widget/BrowserNavigationOverlay.kt
```diff
@@ -107,7 +104,8 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
     override fun onClick(view: View?) {
         var event = NavigationEvent.fromViewClick(view?.id) ?: return
         if (event == NavigationEvent.TURBO) {
-            isTurboEnabled = !turboButton.isActivated
+            turboButton.toggle()
```
liuche: I'd like to put this into the view's onClick, but I'm not sure how to call the super of the view, either here or earlier when I set the onClick listeners.

liuche:
TC failing because testCustomAutocomplete is failing, even though it passes for me locally. Can repush with the test disabled, or see if it just clears up naturally.

app/src/main/java/org/mozilla/focus/widget/CheckableImageButton.kt
```diff
@@ -0,0 +1,58 @@
...
+import android.view.View
+import android.widget.Checkable
+import android.widget.ImageButton
+
+/**
+ * ImageButton that implements Checkable.
+ *
+ * NB: We need a checkable button with a centered drawable for the overlay.
+ * CheckBox uses a non-centered drawable, and we can't use ToggleButton
+ * because it uses android:background, which we need for button state,
```
mcomella: nit: how does ToggleButton use android:background such that we can't override it and put our background there? Add a comment

app/src/main/java/org/mozilla/focus/widget/CheckableImageButton.kt
```diff
@@ -0,0 +1,58 @@
...
+ * so we make a custom button.
+ *
+ * We need to override onCreateDrawableState (see CompoundButton source) to add
+ * the checked state to the attrs.
+ */
+class CheckableImageButton @JvmOverloads constructor(
+        context: Context, attrs: AttributeSet? = null, defStyle: Int = 0 )
+    : ImageButton(context, attrs, defStyle), Checkable {
```
mcomella: nit: style is:
```kotlin
class Clazz(
    properties: Array<Int>,
) : ImageButton...

File: app/src/main/java/org/mozilla/focus/widget/CheckableImageButton.kt [hunk 4]
```diff
@@ -0,0 +1,58 @@
...
+
+    private val _checkedStateSet = intArrayOf(android.R.attr.state_checked)
+
+    private var internalIsChecked = false
+
+    override fun isChecked(): Boolean {
```
mcomella: nit: `override fun isChecked() = internalIsChecked`

app/src/main/java/org/mozilla/focus/widget/CheckableImageButton.kt
```diff
@@ -0,0 +1,58 @@
...
+        return internalIsChecked
+    }
+
+    override fun toggle() {
+        isChecked = !internalIsChecked
+    }
+
+    override fun setChecked(checked: Boolean) {
+        internalIsChecked = checked
+        refreshDrawableState()
```
mcomella: nit: Don't we need to call `refreshDrawableState` when `toggle` is called too? If so, I think you should actually put this is in the property observer for `internalIsChecked` so it ensures our drawable state refreshes when our single source of truth changes (i.e. the internal checked state):
```kotlin
private var internalIsChecked: Boolean
    set(value) {
        field = value
        refreshDrawableState()
    }
```

app/src/main/java/org/mozilla/focus/widget/CheckableImageButton.kt
```diff
@@ -0,0 +1,58 @@
...
+ * so we make a custom button.
+ *
+ * We need to override onCreateDrawableState (see CompoundButton source) to add
+ * the checked state to the attrs.
+ */
+class CheckableImageButton @JvmOverloads constructor(
+        context: Context, attrs: AttributeSet? = null, defStyle: Int = 0 )
+    : ImageButton(context, attrs, defStyle), Checkable {
+
+    private val _checkedStateSet = intArrayOf(android.R.attr.state_checked)
```
mcomella: nit: move outside of class so it's static (not companion obj, that's just boilerplate here). I also don't think `_...` is necessary in the name – what are you trying to convey? `_variable` is only conventionally used when shadowing properties with a field, I think.

liuche:
Going to merge because saw some intermittents before, and this passes locally for me.

PR #524, opened 2018-02-23
Closes #519: FirefoxProgressBar constantly redraws itself, even when …

### LINKED ISSUE #519: FirefoxProgressBar constantly redraws itself, even when hidden
It works on `87a3260c` but stops working on `a3ce1676` (from #287), where the loading progress indicator was added. It hangs on `assertBodyText` when we stop to make sure the page was loaded. I'm not sure why this would happen... :\

bkmunar:
…hidden

### DIFF
--- a/app/src/main/java/org/mozilla/focus/widget/FirefoxProgressBar.kt
+++ b/app/src/main/java/org/mozilla/focus/widget/FirefoxProgressBar.kt
@@ -12,6 +12,7 @@ import android.os.Handler
 import android.os.Looper
 import android.os.Message
 import android.view.View
+import android.view.View.GONE
 import android.widget.LinearLayout
 import org.mozilla.focus.R
 import org.mozilla.focus.architecture.NonNullObserver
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mozilla/focus/widget/FirefoxProgressBar.kt
```diff
@@ -35,9 +35,11 @@ class FirefoxProgressBar @JvmOverloads constructor(
         browserFrag.session.loading.observe(browserFrag, object : NonNullObserver<Boolean>() {
             public override fun onValueChanged(loading: Boolean) {
                 if (loading) {
+                    visibility = View.VISIBLE
                     setMaxVisibility()
                 } else {
                     resetCountdown()
+                    visibility = View.GONE
```
mcomella: This is going to hide the view before the animation runs, right? Do we still want the animation?

mcomella: To follow up, this should be using an animation listener.

app/src/main/java/org/mozilla/focus/widget/FirefoxProgressBar.kt
```diff
@@ -35,9 +35,11 @@ class FirefoxProgressBar @JvmOverloads constructor(
         browserFrag.session.loading.observe(browserFrag, object : NonNullObserver<Boolean>() {
             public override fun onValueChanged(loading: Boolean) {
                 if (loading) {
```
mcomella: Another bug I noticed, we need to cancel any pending animation messages here as well.

PR #527, opened 2018-02-23
[2.1] Fixes #489 - Updates the visual style of the cursor

### LINKED ISSUE #489: update cursor/pointer
update from the blue circle to an svg.

boek:
image


app/src/main/java/org/mozilla/focus/browser/cursor/CursorView.kt
```diff
@@ -33,23 +29,13 @@ private val HIDE_AFTER_MILLIS = TimeUnit.SECONDS.toMillis(3)
...
-    }
+    private val CROSSHAIR_OFFSET = 6
 
     @UiThread
     fun updatePosition(x: Float, y: Float) {
-        // In onDraw, we offset the initial position from the origin: we undo that offset here.
-        translationX = x - radius
-        translationY = y - radius
+        // Offset the position to better align with the crosshairs.
+        translationX = x - CROSSHAIR_OFFSET
```
boek: I feel like there is probably a better way to do this. But when I tried to move the cursor it always got clipped by the view

app/src/main/java/org/mozilla/focus/browser/cursor/CursorView.kt
```diff
@@ -58,11 +44,9 @@ class CursorView(context: Context, attrs: AttributeSet) : View(context, attrs) {
     override fun onDraw(canvas: Canvas) {
         super.onDraw(canvas)
 
-        // We draw the cursor in the top-left corner, offset so the circle isn't clipped by its
-        // parent. The translationX/Y properties will draw it into the desired position on the
-        // screen. We do it this way to avoid allocating new RadialGradients, which are
-        // constructed in relation to the current position of the circle.
-        canvas.drawCircle(radius, radius, radius, paint)
+        val cursor = context.resources.getDrawable(R.drawable.pointer, context.theme)
```
mcomella: nit: onDraw gets called at 60FPS and getDrawable, called here, can read from disk – you should cache the result of this call and re-use it. I think `setBounds` could also be moved out of this function, provided you update the bounds `onLayout`

app/src/main/java/org/mozilla/focus/browser/cursor/CursorView.kt
mcomella: The position coming from `CursorViewModel` has no knowledge of the view hierarchy (beyond the window bounds) so it's position will be the center of the view while `translation` affects the top/left bounds.

I think you'll actually want to subtract `height / 2` and `width / 2` from translation (which was `radius` in the existing implementation).

app/src/main/res/drawable/pointer.xml
```diff
@@ -0,0 +1,10 @@
+<vector android:height="24dp" android:viewportHeight="72.0"
```
mcomella: nit: license.

fwiw, you can set up AS to automatically insert file headers (Settings -> Editor -> "File and Code Templates") but I've had bad experiences where it unsets itself (maybe on IDE updates?).

app/src/main/res/values/dimens.xml
```diff
@@ -46,8 +46,8 @@
 
     <dimen name="fab_margin">16dp</dimen>
 
-    <dimen name="remote_cursor_size">45dp</dimen>
-    <dimen name="remote_cursor_size2x">90dp</dimen>
+    <dimen name="remote_cursor_size">24dp</dimen>
+    <dimen name="remote_cursor_size2x">48dp</dimen>
```
mcomella: nit: this actually seems unused (I wonder why we needed it?)

app/src/main/java/org/mozilla/focus/browser/cursor/CursorView.kt
mcomella: Relatedly, now that the click point is *not* the center of the image, is clicking accurate in relation to the crosshairs on the image?

app/src/main/res/drawable/pointer.xml
mcomella: For correctness, this also needs xml version=...

Sdaswani:
@mcomella please bring up your concerns in triage tomorrow if there isn't discourse on it in the meantime. cc @bbinto @lime124

app/src/main/res/values/dimens.xml
boek: Not sure! I just figured I would update it for correctness

app/src/main/java/org/mozilla/focus/browser/cursor/CursorView.kt
boek: _Yes_. See further explanation in the other comment. I might be able to get it more accurate if I spend some more time on it

app/src/main/res/drawable/pointer.xml
boek: I tried adding `version="..."` but it threw an error. Do you have an example of what it should be?

app/src/main/java/org/mozilla/focus/browser/cursor/CursorView.kt
boek: This is true, but the click point is no longer the center, its only offset by 7 points.  Because the cursor can be positioned by a fraction of a point it looks like the click can get rounded when going into the webview. I spent some time trying to fix but couldn't find the best spot to put it in.

img_1196

app/src/main/res/drawable/pointer.xml
mcomella: `<?xml version="1.0" encoding="utf-8"?>`

https://github.com/mozilla-mobile/focus-android/blob/master/app/src/main/res/values/styles.xml#L1

Just open other files and copy paste. :)

app/src/main/java/org/mozilla/focus/browser/cursor/CursorView.kt
mcomella: tbh, this is probably good enough - no one is going to be this accurate when using the remote on their TV.

That being said, if we want to do better, we can put it in a place that's "good enough" and fat finger it, letting WebView handle disambiguation: we're sending MotionEvents, which are usually triggered by touch events, and you can actually specify the size of the press with this obtain method: https://developer.android.com/reference/android/view/MotionEvent.html#obtain(long,%20long,%20int,%20float,%20float,%20float,%20float,%20int,%20float,%20float,%20int,%20int)

app/src/main/res/drawable/pointer.xml
mcomella: Then again, liuche has a PR where there's no version in the vector file – maybe VectorDrawables just don't use it: https://github.com/mozilla-mobile/firefox-tv/pull/534/commits/d4b0a984497e17a7eb7bd8915df30049c67f302c#diff-4ba463c20b00ee02a671dbbf7412435bR1

mcomella: Since this is auto-generated, it's probably fine to roll without it.

boek: Ah, I added `<?xml version="1.0" encoding="utf-8"?>` to the top. Thought you meant to add it directly to that line.

boek:
Moving the discussion here: @lime124 

I spoke with @mcomella briefly today and he brought up some good points

>- I think the Windows 95 feel comes from: 1) how slanted the mouse cursor is and 2) the crosshair at >the end so I think we should try to fix those (unslant and remove?)
>- I think we can fix the black/white by just using some colors: here are some mock-ups: >https://www.figma.com/file/RJ6sBNEf3PUoTTZc7s7Q1V1T/Fire-TV-cursor Note sure how it’d actually >look on device though
>- Generally though, I have a little discomfort about using the mouse cursor on the TV – it feels almost >too pointed for something that can’t be that accurate. I’m not sure what a better solution is without >spending more time on it (that I’m not sure I have). Silk seems fine (a crosshair)

One more issue with the current crosshair is that it is pretty faint when sitting 8-10ft away from the screen.

lime124:
UX notes - totally ignore:
- at times the cursor can be hard to distinguish from the background. even the white stroke doesn't help. 
- i think the cursor might be too small
- the crosshair feels silly
- b/c the touchpoint is in the upper left, the cursor can go off screen at the bottom and right and become totally invisible. this is required though b/c you need to be able to go off screen a bit to scroll (pro for center point)

boek:
Closing this in favor of #674

PR #529, opened 2018-02-23
Fixes #513 - Addes wordmark and close hint to the overlay

### LINKED ISSUE #513: Add "Close" hint and logo to overlay
See #433 for mocks.

Neither part should be focusable.

Menu SVG: https://gist.github.com/liuche/5ca87ac2c9925cdb0e81094fa803fdfa
(this is the same as the one used in #443 so don't add this svg twice)

~~@mcomella do you know how a11y works for things that aren't selectable?~~
No a11y text needed

boek:
image

### DIFF
--- a/app/src/main/res/drawable/ic_wide_menu.xml
+++ b/app/src/main/res/drawable/ic_wide_menu.xml
@@ -1,6 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+
 <!-- This Source Code Form is subject to the terms of the Mozilla Public
    - License, v. 2.0. If a copy of the MPL was not distributed with this
    - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
+
 <vector xmlns:android="http://schemas.android.com/apk/res/android"
         android:width="24dp"
         android:height="24dp"
--- a/app/src/main/res/drawable/ic_wide_menu.xml
+++ b/app/src/main/res/drawable/ic_wide_menu.xml
@@ -1,0 +1,18 @@
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
+   - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+        android:width="24dp"
+        android:height="24dp"
+        android:viewportWidth="24.0"
+        android:viewportHeight="24.0">
+    <path
+        android:fillColor="#FF000000"
+        android:pathData="M18,7H6A1,1 0,0 1,6 5H18a1,1 0,0 1,0 2Z"/>
+    <path
+        android:fillColor="#FF000000"
+        android:pathData="M18,13H6a1,1 0,0 1,0 -2H18a1,1 0,0 1,0 2Z"/>
+    <path
+        android:fillColor="#FF000000"
+        android:pathData="M18,19H6a1,1 0,0 1,0 -2H18a1,1 0,0 1,0 2Z"/>
+</vector>
--- a/app/src/main/res/layout/browser_overlay.xml
+++ b/app/src/main/res/layout/browser_overlay.xml
@@ -41,6 +41,18 @@
             android:src="@drawable/turbo_selector"
             android:contentDescription="@string/turbo_mode"/>
 
+        <View
+            android:layout_width="0dp"
+            android:layout_height="0dp"
+            android:layout_weight="1" />
+
+        <ImageView
+            android:id="@+id/firefoxWordmark"
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:src="@drawable/ic_firefox_and_workmark" />
+
+
     </LinearLayout>
     <LinearLayout
         android:layout_width="match_parent"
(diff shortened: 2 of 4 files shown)


app/src/main/res/drawable/ic_wide_menu.xml
```diff
@@ -0,0 +1,18 @@
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
```
mcomella: nit: This should have xml version

app/src/main/res/layout/browser_overlay.xml
```diff
@@ -41,6 +41,18 @@
             android:src="@drawable/turbo_selector"
             android:contentDescription="@string/turbo_mode"/>
 
+        <View
+            android:layout_width="0dp"
+            android:layout_height="0dp"
+            android:layout_weight="1" />
```
mcomella: perf nit: there's an IDE warning for nested LinearLayouts where both use `layout_weight`s:

screen shot 2018-02-26 at 14 49 03

That being said, I feel like layout perf is trivial for self-containing views like this (unless it gets in the way of animations) so it's probably good enough and I wouldn't bother to change it. fwiw, I might design this as a root FrameLayout with four sub-containers for each corner. ConstraintLayout is also an option. I'm not sure how this designs would affect keyboard navigation, however.

app/src/main/res/layout/browser_overlay.xml
```diff
@@ -72,6 +84,25 @@
...
+
+        <ImageView
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:layout_weight="0"
+            android:tint="@android:color/white"
+            android:src="@drawable/ic_wide_menu" />
+
+        <TextView
```
mcomella: perf nit: you can remove the ImageView by adding a compound drawable directly into the TextView with `android:drawableStart` (or `compoundDrawable` whatever in the code). This is slightly better for performance, though it's harder to style the image as you want. If changing this starts giving you trouble, it's really not worth the time.

app/src/main/res/layout/browser_overlay.xml
```diff
@@ -72,6 +84,25 @@
...
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:layout_weight="0"
+            android:paddingLeft="5dp"
```
mcomella: nit: paddingStart, for easier RTL

app/src/main/res/values/strings.xml
```diff
@@ -565,4 +565,7 @@ be temporary, and you can try again later.</li>
     <string name="onboarding_turbo_mode_body">Turbo Mode automatically blocks trackers and ads to help get you where you\'re going faster. If there\'s a speed bump – like a site that isn\'t behaving the way you expect – just turn Turbo Mode off.</string>
     <string name="onboarding_turbo_mode_button_on">Keep Turbo Mode Enabled</string>
     <string name="onboarding_turbo_mode_button_off">Turn Turbo Mode Off</string>
+
+    <!-- Label for the hint to close the overlay-->
+    <string name="close_hint">Close</string>
```
mcomella: nit: this should be more specific to the UI element for which it's used, considering how hard it is to re-use strings: maybe `nav_overlay_close_hint`?

app/src/main/res/values/strings.xml
```diff
@@ -565,4 +565,7 @@ be temporary, and you can try again later.</li>
     <string name="onboarding_turbo_mode_body">Turbo Mode automatically blocks trackers and ads to help get you where you\'re going faster. If there\'s a speed bump – like a site that isn\'t behaving the way you expect – just turn Turbo Mode off.</string>
     <string name="onboarding_turbo_mode_button_on">Keep Turbo Mode Enabled</string>
     <string name="onboarding_turbo_mode_button_off">Turn Turbo Mode Off</string>
+
+    <!-- Label for the hint to close the overlay-->
```
mcomella: nit: since we don't extract these hints for l10n afaik, I think the comment is kind of redundant.

app/src/main/res/layout/browser_overlay.xml
```diff
@@ -72,6 +84,25 @@
...
+        <View
+            android:layout_width="0dp"
+            android:layout_height="0dp"
+            android:layout_weight="1" />
+
+        <ImageView
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:layout_weight="0"
+            android:tint="@android:color/white"
```
mcomella: This is white while the text looks to be some sort of shade of gray and it looks inconsistent: does the design dictate these should be the same color?

app/src/main/res/layout/browser_overlay.xml
boek: @mcomella I tried to go down the route of converting it into a ConstraintLayout but it started to increase the scope of this PR a bit. I definitely think we should do it as a follow up item though. What do you think?

boek: Tracking #540

mcomella:
I think yes, in an ideal world, but ultimately we can make this code
more maintainable or add some more features/polish and I think I'd chose
the latter until it's necessary not to (e.g. not doing so will waste
more time later or that it starts to affect performance).
Good call stopping yourself from scope creeping. :)


On Tue, Feb 27, 2018, at 17:39, Jeff Boek wrote:
> *@boek* commented on this pull request.


> 
> In app/src/main/res/layout/browser_overlay.xml[1]:


> > @@ -41,6 +41,18 @@ android:src="@drawable/turbo_selector"
> > android:contentDescription="@string/turbo_mode"/>  +        <View +
> > android:layout_width="0dp" +            android:layout_height="0dp"
> > +            android:layout_weight="1" />
>> @mcomella[2] I tried to go down the route of converting it into a
> ConstraintLayout but it started to increase the scope of this PR a
> bit. I definitely think we should do it as a follow up item though.
> What do you think?> — You are receiving this because you were mentioned. Reply to this
> email directly, view it on GitHub[3], or mute the thread[4].> 


Links:

  1. https://github.com/mozilla-mobile/firefox-tv/pull/529#discussion_r171091110
  2. https://github.com/mcomella
  3. https://github.com/mozilla-mobile/firefox-tv/pull/529#discussion_r171091110

PR #531, opened 2018-02-24
Added deletion logic

bkmunar:


bkmunar:
First part of https://github.com/mozilla-mobile/firefox-tv/issues/443

app/src/main/java/org/mozilla/focus/fragment/HomeFragment.kt
```diff
@@ -24,8 +24,11 @@ import org.mozilla.focus.telemetry.TelemetryWrapper
 import org.mozilla.focus.telemetry.UrlTextInputLocation
 import org.mozilla.focus.utils.OnUrlEnteredListener
 import android.graphics.BitmapFactory
+import android.content.Context.MODE_PRIVATE
 
 private const val COL_COUNT = 5
+val BLACKLIST = "blacklist"
```
liuche: Probably rename this to bundled_blacklist, to be explicit that this is for the bundled tiles.

app/src/main/java/org/mozilla/focus/fragment/HomeFragment.kt
```diff
@@ -75,13 +78,17 @@ class HomeFragment : Fragment() {
         val homeTiles = mutableListOf<HomeTile>()
         val inputAsString = context.assets.open(HOME_TILES_JSON_PATH).bufferedReader().use { it.readText() }
         val jsonArray = JSONObject(inputAsString).getJSONArray(HOME_TILES_JSON_KEY)
+        val userDetails = context.getSharedPreferences(HOME_TILES_PREFS, MODE_PRIVATE)
```
liuche: nit: rename homeTilesPrefs

liuche: Side note: Another way to do shared prefs without using a separate file is just to call `PreferenceManager.getDefaultSharedPreferences()`. That way you don't need to make a new "shared preferences" file. However, the tradeoff is that when you load SharedPreferences, you need to load the whole file - so if it's really big, this could slow down your app.

In the case of home tiles, where it's possible that people will want to have a ton of custom tiles, it's might be better to use a separate file - but at the same time, that's the first screen that gets loaded, so it might not matter. But better safe than sorry! so let's keep this separate pref file.

app/src/main/java/org/mozilla/focus/fragment/HomeFragment.kt
```diff
@@ -75,13 +78,17 @@ class HomeFragment : Fragment() {
...
+        val blacklist = userDetails.getStringSet(BLACKLIST, null)
```
liuche: nit: I would pull this whole part of making a list of HomeTiles into a function so it doesn't clutter up the init (and make it hard to read), something that takes the sharedprefs and home tiles jsonArray, and returns a list of HomeTiles, with the blacklist filtered out.

app/src/main/java/org/mozilla/focus/fragment/HomeFragment.kt
```diff
@@ -75,13 +78,17 @@ class HomeFragment : Fragment() {
         val homeTiles = mutableListOf<HomeTile>()
         val inputAsString = context.assets.open(HOME_TILES_JSON_PATH).bufferedReader().use { it.readText() }
         val jsonArray = JSONObject(inputAsString).getJSONArray(HOME_TILES_JSON_KEY)
```
liuche: Nit: rename jsonArray as bundledTiles

app/src/main/java/org/mozilla/focus/fragment/HomeFragment.kt
```diff
@@ -117,6 +124,13 @@ private class HomeTileAdapter(val onUrlEnteredListener: OnUrlEnteredListener, ho
             onUrlEnteredListener.onNonTextInputUrlEntered(item.url)
             TelemetryWrapper.homeTileClickEvent()
         }
+        itemView.setOnLongClickListener {
```
liuche: Add a TODO comment about replacing this with UI.

app/src/main/java/org/mozilla/focus/fragment/HomeFragment.kt
```diff
@@ -117,6 +124,13 @@ private class HomeTileAdapter(val onUrlEnteredListener: OnUrlEnteredListener, ho
...
+            deleteTile(item.id)
+            tiles.removeAt(position)
+            notifyItemRemoved(position)
+            notifyItemRangeChanged(position, tiles.size)
```
liuche: Do you need this notifyItemRangeChanged?

app/src/main/java/org/mozilla/focus/fragment/HomeFragment.kt
```diff
@@ -134,6 +148,19 @@ private class HomeTileAdapter(val onUrlEnteredListener: OnUrlEnteredListener, ho
...
 
+    fun deleteTile(id: String) {
+        var blacklist = context.getSharedPreferences(HOME_TILES_PREFS, MODE_PRIVATE).getStringSet(BLACKLIST, null)
+        if (blacklist != null) {
+            blacklist.add(id)
+        } else {
+            blacklist = mutableSetOf(id)
+        }
+        val edit = context.getSharedPreferences(HOME_TILES_PREFS, MODE_PRIVATE).edit()
+        edit.clear() // maybe copy array??
```
liuche: Yeah, fix this - I think the fact that this works is an unintended side effect (and it's not mentioned in the docs as a way to fix the "don't modify the string set" problem), so we shouldn't rely on it.

app/src/main/java/org/mozilla/focus/fragment/HomeFragment.kt
```diff
@@ -117,6 +124,13 @@ private class HomeTileAdapter(val onUrlEnteredListener: OnUrlEnteredListener, ho
             onUrlEnteredListener.onNonTextInputUrlEntered(item.url)
             TelemetryWrapper.homeTileClickEvent()
         }
+        itemView.setOnLongClickListener {
+            deleteTile(item.id)
```
liuche: Nit: be more specific in this method name - probably `addBundledTileToBlacklist` rather than `deleteTile`. It's not clear what "delete" means, does it mean that you're removing the tile from UI too?

Write maintainable code! That means naming methods to what they actually do - in the future, someone might see `deleteTile` and add some code there that belongs somewhere else, but if you name it `addBundledTileToBlacklist` no one will be confused.  ("Bundled" is optional, up to you)

app/src/main/java/org/mozilla/focus/fragment/HomeFragment.kt
bkmunar: Yes or else it crashes when I remove something from the [0,n-1) range and then n-1. Unless there's another way to notify the elements that their index has changed??

liuche: Mmmmm I'm pretty sure you don't need to call that, because in the documentation you should just be able to call "notifyItemRemoved". I'm currently trying to finish up pinning so I don't have a ton of time to try to debug this with you, but maybe try the following:

- Make sure the data source is updating correctly (is getItemCount returning the correct length?)
- Look at other existing code in Fennec ( https://dxr.mozilla.org/mozilla-central/source/mobile/android/base/java/org/mozilla/gecko/tabs/TabStripAdapter.java#74 or https://dxr.mozilla.org/mozilla-central/search?q=notifyitemremoved&redirect=true )
- Take a look at the docs ( https://developer.android.com/reference/android/support/v7/widget/RecyclerView.Adapter.html )
- See if other people have this problem on stackoverflow?

If you still can't figure this out, we can take a look at this together.

liuche:
I'm going to close this for now because this will need to be rebased on top of the pinned sites code, and doesn't have any UI or interaction.

PR #532, opened 2018-02-24
Closes #530: Close asset stream for home tile icons.

### LINKED ISSUE #530: InputStreams for home tile icons are not being closed
`BitmapFactory.decodeStream(context.assets.open(item.imagePath))`

mcomella:


mcomella:
@bkmunar I additionally added a commit that removes the reference to Context in the Adapter.

Context is a really large object so holding a reference longer than it's lifecycle can quickly create memory leaks that throw OOM errors, for example: https://android-developers.googleblog.com/2009/01/avoiding-memory-leaks.html

A general rule is if something doesn't contain UI, it probably shouldn't have a reference to a Context. In this case, an Adapter contains a reference to the data and generally outlives the UI, swapping the data into new UI as necessary. Since it outlives the UI, and the Context, it shouldn't hold a reference to it. That being said, it can get an always valid Context reference from the views it's holding, which I put in my commit.

This is a common Android problem so let me know if you have questions!

app/src/main/java/org/mozilla/focus/fragment/HomeFragment.kt
```diff
@@ -90,15 +89,14 @@ class HomeFragment : Fragment() {
...
+private class HomeTileAdapter(val onUrlEnteredListener: OnUrlEnteredListener, homeTiles: MutableList<HomeTile>) :
         RecyclerView.Adapter<TileViewHolder>() {
-    val context = context
     val tiles = homeTiles
 
     override fun onBindViewHolder(holder: TileViewHolder, position: Int) = with (holder) {
         val item = tiles[position]
         titleView.setText(item.title)
-        val bmImg = BitmapFactory.decodeStream(context.assets.open(item.imagePath))
+        val bmImg = itemView.context.assets.open(item.imagePath).use { BitmapFactory.decodeStream(it) }
```
liuche: Oh I see, `use` closes the stream after using it.

...ugh kotlin.

PR #534, opened 2018-02-27
Add pinning of custom sites to Overlay

liuche:
Added pinning for custom sites to overlay.

screenshot_1519703636

screenshot_1519691249


liuche:
@mcomella is there a better way to do null checks? I feel like I'm doing a lot of that because I'm checking if JSON strings are empty.

liuche:
Adding @lime124 to take a look at these screenshots.

lime124:
@liuche are we able to impact the placement of the toast?

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
```diff
@@ -135,7 +136,9 @@ class BrowserFragment : IWebViewLifecycleFragment(),
                 reload()
             }
             NavigationEvent.PIN_SITE -> {
-                // TODO: Add to pinned sites
+                if (webview.getUrl() != null) {
```
mcomella: You should use `BrowserFragment.url` – we set it outside of the webView so we have more control over it. THis is a hold-over from Focus that I don't entirely understand why.

mcomella: If you want to avoid the `!!` after the null check, you need to smart-cast. Smart-cast can only happen if the value is guaranteed not to change, which happens if you assign it to an immutable value:
```kotlin
val url = webview.getUrl()
if (url != null) {
    HomeTiles...
}
```

Or:

```kotlin
webview.getUrl()?.let { url -> // or use `it` directly
    HomeTiles...
}
```

Note that the `?` is really important: iirc, you can call `let` without it (which is really weird and not a good thing so I generally don't use let).

app/src/main/java/org/mozilla/focus/tiles/HomeTilesManager.kt
```diff
@@ -0,0 +1,50 @@
...
+
+package org.mozilla.focus.tiles
+
+import android.content.Context
+import android.content.Context.MODE_PRIVATE
+import android.content.SharedPreferences
+import org.json.JSONArray
+import org.json.JSONObject
+
+const val HOME_TILES_PREFS = "hometilesPrefs"
```
mcomella: nit: make these private if they're not used outside of this file.

mcomella: nit: camel-casing. Also, this is a shared preferences branch so maybe including "Prefs" in the string is not necessary.

app/src/main/java/org/mozilla/focus/tiles/HomeTilesManager.kt
```diff
@@ -0,0 +1,50 @@
...
+                .apply()
+    }
+
+    private fun getCustomSitesJSONArray(sharedPreferences: SharedPreferences): JSONArray {
+        val sitesListString = sharedPreferences.getString(CUSTOM_SITES_LIST, null)
+        return if (sitesListString != null) JSONArray(sitesListString) else JSONArray()
+    }
+
+    private fun makeSiteJSON(url: String): JSONObject {
+        val siteJSON = JSONObject()
```
mcomella: nit: you may be able to make this more clear with kotlin (at a small cost of an object allocation though I'd expect we'll be converting this from a `HomeTile` object in the future anyway):
```kotlin
private fun makeSiteJSON(url: String) = JSONObject(mapOf<String, String?>(
    KEY_URL to url,
    KEY_TITLE to url,
    KEY_IMG to null,
    KEY_ID to url
))
```

app/src/main/java/org/mozilla/focus/tiles/HomeTilesManager.kt
```diff
@@ -0,0 +1,50 @@
...
+
+        sitesJSONArray.put(makeSiteJSON(url))
+
+        sharedPreferences.edit()
+                .putString(CUSTOM_SITES_LIST, sitesJSONArray.toString())
+                .apply()
+    }
+
+    private fun getCustomSitesJSONArray(sharedPreferences: SharedPreferences): JSONArray {
+        val sitesListString = sharedPreferences.getString(CUSTOM_SITES_LIST, null)
```
mcomella: nit: you could simplify this with (or as one line):
```kotlin
val sitesListString = sp.getString(CUSTOM_SITES_LIST, "[]")
return JSONArray(sitesListString)
```

app/src/main/java/org/mozilla/focus/tiles/HomeTilesManager.kt
```diff
@@ -0,0 +1,50 @@
...
+const val HOME_TILES_PREFS = "hometilesPrefs"
+const val CUSTOM_SITES_LIST = "customSitesList"
+
+const val KEY_URL = "url"
+const val KEY_TITLE = "title"
+const val KEY_IMG = "img"
+const val KEY_ID = "identifier"
+
+object HomeTilesManager {
+    fun pinSite(context: Context, url: String) {
```
mcomella: nit: Can this be called by multiple threads? If not, annotate it. If so, make sure the get-and-set is atomic (either with locking or a data structure that makes it possible). Also if so, please reflag me for review.

app/src/main/java/org/mozilla/focus/tiles/HomeTilesManager.kt
```diff
@@ -0,0 +1,50 @@
...
+import org.json.JSONArray
+import org.json.JSONObject
+
+const val HOME_TILES_PREFS = "hometilesPrefs"
+const val CUSTOM_SITES_LIST = "customSitesList"
+
+const val KEY_URL = "url"
+const val KEY_TITLE = "title"
+const val KEY_IMG = "img"
+const val KEY_ID = "identifier"
```
mcomella: nit: I think `id` is fine - lets save a few characters on disk.

Why is the `id` necessary anyway? Can URL be used as the ID instead? URLs are guaranteed to be unique so provided we never add the same URL twice, it should be sufficient though if the code is wrong and we get into a state where there are duplicated URLs, it could be problem...

app/src/main/java/org/mozilla/focus/fragment/HomeFragment.kt
```diff
@@ -72,6 +74,7 @@ class HomeFragment : Fragment() {
 
     private fun initTiles() = with (tileContainer) {
         val homeTiles = mutableListOf<HomeTile>()
+        homeTiles.addAll(HomeTilesManager.getCustomHomeTilesList(context))
```
mcomella: nit: if you don't mind that the `getCustomHomeTilesList` gets mutated, you can do
```kotlin
val homeTiles = Home...getCustomHomeTilesList(context).toMutableList()
```

i.e. this will mutate the existing list, rather than copying it into a new one.

app/src/main/java/org/mozilla/focus/tiles/HomeTilesManager.kt
```diff
@@ -47,4 +60,14 @@ object HomeTilesManager {
     private fun getHomeTilesPreferences(context: Context): SharedPreferences {
         return context.getSharedPreferences(HOME_TILES_PREFS, MODE_PRIVATE)
     }
+
+    fun getCustomHomeTilesList(context: Context): List<HomeTile> {
+        val sharedPreferences = context.getSharedPreferences(HOME_TILES_PREFS, MODE_PRIVATE)
```
mcomella: nit: use getHomeTilesPreferences

app/src/main/java/org/mozilla/focus/tiles/HomeTilesManager.kt
```diff
@@ -47,4 +60,14 @@ object HomeTilesManager {
...
+        val sitesJSONArray = getCustomSitesJSONArray(sharedPreferences)
+        val homeTiles = mutableListOf<HomeTile>()
+        for (i in sitesJSONArray.length() - 1 downTo 0) {
```
mcomella: nit: fwiw, you could add a `JSONArray.map` extension fn, which would make this more readable, though this is pretty obnoxious since you have to deal with generics. I actually wrote a flatMapObj fn in another WIP: https://github.com/mcomella/firefox-tv/commit/72a4db66ef68d59afb6a4744ba6f02bfe689c713

app/src/main/java/org/mozilla/focus/tiles/HomeTilesManager.kt
```diff
@@ -18,6 +18,19 @@ const val KEY_TITLE = "title"
...
 const val KEY_ID = "identifier"
 
+data class HomeTile (
+        val url: String,
+        val title: String,
+        val imagePath: String?,
+        // unique id used to identify specific home tiles, e.g. for deletion, etc.
+        val id: String
+) {
+    constructor(jsonObject: JSONObject): this(jsonObject.getString(KEY_URL),
```
mcomella: nit: Since this is just a data container and its constructor is naturally used to fill its data, I think a more natural way of constructing this is statically:
```kotlin
data class HomeTile(
    ...
) {
    companion object {
        fun fromJSONObject(jsonObj: JSONObject): HomeTile? {
            return HomeTile(...)
        }
    }
}
```

Benefits:
- Can return null (e.g. if the JSONObject doesn't have the fields you expect/require)
- Defining multiple static constructors is more natural than having many constructors
- iirc, `this` constructor must be called before you can execute code (e.g. store local variables) which are helpful to make the code more readable.

app/src/main/java/org/mozilla/focus/fragment/HomeFragment.kt
mcomella: nit: this file is hard to read because there isn't much WS between lines. I think it'd help if you added a whitespace below this `getCustomTiles` event, to separate it from constructing the default tiles.

app/src/main/java/org/mozilla/focus/fragment/HomeFragment.kt
```diff
@@ -109,8 +112,10 @@ private class HomeTileAdapter(val onUrlEnteredListener: OnUrlEnteredListener, ho
     override fun onBindViewHolder(holder: TileViewHolder, position: Int) = with (holder) {
         val item = tiles[position]
         titleView.setText(item.title)
-        val bmImg = itemView.context.assets.open(item.imagePath).use { BitmapFactory.decodeStream(it) }
-        iconView.setImageBitmap(bmImg)
+        if (!item.imagePath.isNullOrEmpty()) {
+            val bmImg = itemView.context.assets.open(item.imagePath).use { BitmapFactory.decodeStream(it) }
+            iconView.setImageBitmap(bmImg)
+        }
```
mcomella: nit: should we have a placeholder if it's null/empty?

app/src/main/java/org/mozilla/focus/fragment/HomeFragment.kt
```diff
@@ -109,8 +112,10 @@ private class HomeTileAdapter(val onUrlEnteredListener: OnUrlEnteredListener, ho
     override fun onBindViewHolder(holder: TileViewHolder, position: Int) = with (holder) {
         val item = tiles[position]
         titleView.setText(item.title)
-        val bmImg = itemView.context.assets.open(item.imagePath).use { BitmapFactory.decodeStream(it) }
-        iconView.setImageBitmap(bmImg)
+        if (!item.imagePath.isNullOrEmpty()) {
```
mcomella: nit: `isNullOrBlank` is probably safer

liuche:
@lime124 I can center it, and then add x/y offsets - I can't anchor it to the pinned button, so here it is centered:

screenshot_1519772034

I can also try to eyeball what those offsets are, but I'm not sure what would happen on different sized screens, and we would have to update the offsets manually every time we move/add buttons.

What do you think?

lime124:
@liuche that's a really good point about different TVs and it potentially being oddly placed. I think this placement will work a lot better as it's closer to where the action was initiated and will (hopefully) be more visible/noticeable to people. Let's see how this works on larger TV sizes since at the moment I'm only on a monitor 😬

app/src/main/java/org/mozilla/focus/tiles/CustomTilesAccessor.kt
```diff
@@ -31,15 +31,50 @@ data class HomeTile (
...
 
-object HomeTilesManager {
+/**
+ * Static accessor of custom home tiles, that is backed by SharedPreferences.
+ *
+ * New sites are appended to the end of the list.
+ *
+ * This keeps a cached version of the custom home tiles that have been pinned,
+ * in order to be more performant when checking whether sites are pinned or not.
+ * In order to keep the cache consistent, should only be called from the UIThread.
```
mcomella: nit: Annotate the methods with `@UiThread`

app/src/main/java/org/mozilla/focus/tiles/CustomTilesAccessor.kt
```diff
@@ -31,15 +31,50 @@ data class HomeTile (
...
+        val tilesJSONArray = getCustomSitesJSONArray(getHomeTilesPreferences(context))
+        val lhm = LinkedHashMap<String, JSONObject>()
+        for (i in 0 until tilesJSONArray.length()) {
+            val tile = tilesJSONArray.getJSONObject(i)
+            lhm.put(tile.getString(KEY_URL), tile)
+        }
+        return lhm
+    }
+
+    fun isURLPinned(url: String): Boolean {
```
mcomella: nit: fun ... =

app/src/main/java/org/mozilla/focus/tiles/CustomTilesAccessor.kt
```diff
@@ -31,15 +31,50 @@ data class HomeTile (
...
+ *
+ * This keeps a cached version of the custom home tiles that have been pinned,
+ * in order to be more performant when checking whether sites are pinned or not.
+ * In order to keep the cache consistent, should only be called from the UIThread.
+ */
+object CustomTilesAccessor {
+    private lateinit var customTilesCache: LinkedHashMap<String, JSONObject>
+
+    // Cache pinned sites for perf beacues we need to check pinned state for every page load
+    fun initWithCache(customTiles: LinkedHashMap<String, JSONObject>) {
```
mcomella: nit: I think this should take a context and set the cache internally. The caller has no need to have access to the cache (and they could even mutate it if they wanted to).

Don't forget to privatize `getCustomTilesCache`.

app/src/main/java/org/mozilla/focus/tiles/CustomTilesAccessor.kt
```diff
@@ -31,15 +31,50 @@ data class HomeTile (
...
+ * Static accessor of custom home tiles, that is backed by SharedPreferences.
+ *
+ * New sites are appended to the end of the list.
+ *
+ * This keeps a cached version of the custom home tiles that have been pinned,
+ * in order to be more performant when checking whether sites are pinned or not.
+ * In order to keep the cache consistent, should only be called from the UIThread.
+ */
+object CustomTilesAccessor {
+    private lateinit var customTilesCache: LinkedHashMap<String, JSONObject>
```
mcomella: nit: it feels weird that this cache isn't `LinkedHashMap<String, HomeTile>`. Trade-offs:
- As HomeTile, it's type-safe and the data-type we want to manipulate ("more correct")
- Converting to JSONObjects can be expensive, but we don't do that much anyway (and we can cache those values in HomeTile if we wanted)
- Currently, we're allocating a bunch of HomeTile objects each time the home screen is shown, which can be expensive if it forces GC

mcomella: nit: -> customTileURLsToTile or similar

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
```diff
@@ -164,6 +164,7 @@ class BrowserFragment : IWebViewLifecycleFragment(),
     override fun isBackEnabled() = canGoBack()
     override fun isForwardEnabled() = canGoForward()
     override fun getCurrentUrl() = url
+    override fun isURLPinned() = if (url != null) CustomTilesAccessor.isURLPinned(url!!) else false
```
mcomella: nit: url can be mutated while this method is called so it'd be better to do (for a one-liner):
```kotlin
url?.let { CustomTilesAccessor.isUrlPinned(it) } ?: false
```

But maybe it's best not written as one line

app/src/main/java/org/mozilla/focus/activity/MainActivity.kt
```diff
@@ -41,6 +42,7 @@ class MainActivity : LocaleAwareAppCompatActivity(), OnUrlEnteredListener {
         super.onCreate(savedInstanceState)
 
         initAmazonFactory()
+        CustomTilesAccessor.initWithCache(CustomTilesAccessor.getCustomTilesCache(this))
```
mcomella: If we ever recreate MainActivity, this will be called again and it will break. Either:
1) Make sure init is only called once
2) Call this from FocusApplication.onCreate (seems like a bad idea from a perf perspective).

Also, calling this here will break if we ever refactor the code to enter from another Activity – if you made the cache lazily-initted in each of the public accessor methods, it'd avoid this problem (and the class would have fewer dependencies). If you decide to leave it as is, please document that this method needs to be called in the class javadoc.

app/src/main/java/org/mozilla/focus/fragment/HomeFragment.kt
```diff
@@ -27,14 +27,14 @@ import org.mozilla.focus.tiles.HomeTile
 import org.mozilla.focus.tiles.CustomTilesAccessor
 
 private const val COL_COUNT = 5
+const val HOME_TILES_DIR = "defaults/"
```
mcomella: nit: private? Can we move the other constants out too? (this file is really hard to read)

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
```diff
@@ -138,11 +139,13 @@ class BrowserFragment : IWebViewLifecycleFragment(),
             NavigationEvent.PIN_SITE -> {
                 if (webview.getUrl() != null) {
                     CustomTilesAccessor.pinSite(context, webview.getUrl()!!)
+                    Toast.makeText(context, R.string.notification_pinned_site, Toast.LENGTH_SHORT).show()
```
mcomella: Design nit: is Toast the appropriate API on TVs? Is Snackbar better?

app/src/main/res/values/strings.xml
```diff
@@ -565,4 +565,8 @@ be temporary, and you can try again later.</li>
     <string name="onboarding_turbo_mode_body">Turbo Mode automatically blocks trackers and ads to help get you where you\'re going faster. If there\'s a speed bump – like a site that isn\'t behaving the way you expect – just turn Turbo Mode off.</string>
     <string name="onboarding_turbo_mode_button_on">Keep Turbo Mode Enabled</string>
     <string name="onboarding_turbo_mode_button_off">Turn Turbo Mode Off</string>
+
+    <!-- Notification shown when saving/unsaving sites, which will be displayed on the homescreen as tiles -->
+    <string name="notification_pinned_site">Pinned to Firefox TV homescreen</string>
```
mcomella: nit: I'm concerned ppl will confuse this with, "Fire TV homescreen"

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
```diff
@@ -134,6 +134,12 @@ class BrowserFragment : IWebViewLifecycleFragment(),
                 isReloadingForYoutubeDrawerClosed = true
                 reload()
             }
+            NavigationEvent.PIN_SITE -> {
```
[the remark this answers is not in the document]
liuche: Not doing a sealed class, but removing the PIN_SITE, UNPIN_SITE enum cases, and fixing the TURBO mode one too.

mcomella:
> @mcomella is there a better way to do null checks? I feel like I'm doing a lot of that because I'm checking if JSON strings are empty when pulling them from SharedPreferences.

I think generally, your ways are:
- Chaining, e.g. `id?.toString()?.substring...`
- If != null with smart-casting (make sure they the values cannot be changed, e.g. generally immutable)
- Use default values instead, e.g. instead of returning null, return an empty string and pass that into a method that takes a string
- `if (value?.let { ...`

Specifically with pulling out JSON values, I convert them into non-null types, e.g. I'll convert them to immutable HomeTiles immediately and only manipulate them as HomeTiles, not accessing the JSON more than once.

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
liuche: thanks! that's exactly what I was looking for :)

app/src/main/java/org/mozilla/focus/tiles/HomeTilesManager.kt
liuche: This is something that will be used by the custom tiles code too, and if those change, we don't want to be tied to a url.

app/src/main/java/org/mozilla/focus/fragment/HomeFragment.kt
liuche: Not going to do this because CustomTileAccessor uses that list locally, without calling it, so it shouldn't be mutated.

app/src/main/java/org/mozilla/focus/fragment/HomeFragment.kt
liuche: I'd assume that the "get screenshot" code will provide something if it's empty? but if it doesn't we can add it later.

app/src/main/java/org/mozilla/focus/tiles/HomeTilesManager.kt
liuche: Should only be called from a user action (e.g., toggling the pin button, or maybe in the future deleting from the home screen, which will be UI thread)

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
liuche: No, toast is better.

app/src/main/res/values/strings.xml
liuche: Not sure what we can do about that...? @BrianNJones gave me the string and I don't know how else we could be more explicit without being incredibly wordy.

app/src/main/java/org/mozilla/focus/tiles/CustomTilesAccessor.kt
liuche: I don't think that's necessary, it's typed and you can find the definition.

app/src/main/java/org/mozilla/focus/activity/MainActivity.kt
liuche: I don't think this will break, because it just refreshes the cache.

But I'll switch this to getInstance anyway - probably good to try it out and if it doesn't work out I'll remember :P

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
```diff
@@ -134,10 +138,27 @@ class BrowserFragment : IWebViewLifecycleFragment(),
                 isReloadingForYoutubeDrawerClosed = true
                 reload()
             }
+            NavigationEvent.PIN_ACTION -> {
+                this@BrowserFragment.url?.let { url ->
+                    if (VAL_CHECKED == value) {
+                        CustomTilesAccessor.getInstance(context).pinSite(context, url)
+                        showCenteredToast(R.string.notification_pinned_site)
+                    } else {
```
mcomella: nit: I'd explicitly check `VAL_UNCHECKED == ...` and throw in the else case because we are comparing against a String

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
```diff
@@ -134,10 +138,27 @@ class BrowserFragment : IWebViewLifecycleFragment(),
...
+                        CustomTilesAccessor.getInstance(context).unpinSite(context, url)
+                        showCenteredToast(R.string.notification_unpinned_site)
+                    }
+                }
+            }
         }
         Unit
     }
 
+    private fun showCenteredToast(resId: Int) {
```
mcomella: nit: I think this should move to a helper to help keep BrowserFrag clean: `ToastUtils.showCentenedToast(resId: Int)`, for example.

mcomella:
> @lime124 I can center it, and then add x/y offsets - I can't anchor it to the pinned button, so here it is centered:

You could calculate the offsets necessary to anchor it on a button based on the button position, but it's error-prone (e.g. might appear offscreen) and probably not worth it.

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
```diff
@@ -155,6 +176,7 @@ class BrowserFragment : IWebViewLifecycleFragment(),
     override fun isBackEnabled() = canGoBack()
     override fun isForwardEnabled() = canGoForward()
     override fun getCurrentUrl() = url
+    override fun isURLPinned() = url?.let { CustomTilesAccessor.getInstance(context).isURLPinned(url!!) } ?: false
```
mcomella: Use `it`, not `url` (or bind to `url` with `?.let { url -> ... }`. If the compiler forces you to use `!!`, it means that some other thread could potentially change this value from the time you null-checked it to when you use it, which can cause bugs, so do your best not to use !! unless you're *sure* it won't change

app/src/main/java/org/mozilla/focus/fragment/HomeFragment.kt
```diff
@@ -23,16 +23,18 @@ import org.mozilla.focus.autocomplete.UrlAutoCompleteFilter
 import org.mozilla.focus.telemetry.TelemetryWrapper
 import org.mozilla.focus.telemetry.UrlTextInputLocation
 import org.mozilla.focus.utils.OnUrlEnteredListener
+import org.mozilla.focus.tiles.HomeTile
+import org.mozilla.focus.tiles.CustomTilesAccessor
 
 private const val COL_COUNT = 5
+const val HOME_TILES_DIR = "defaults/"
```
mcomella: nit: private? What about moving the other keys here too?

nit: This is `DEFAULT_HOME_TILES_DIR`. I feel picky about this because we're reading from `assets`, which we can't write to.

app/src/main/java/org/mozilla/focus/tiles/HomeTilesManager.kt
```diff
@@ -0,0 +1,116 @@
...
+package org.mozilla.focus.tiles
+
+import android.content.Context
+import android.content.Context.MODE_PRIVATE
+import android.content.SharedPreferences
+import android.support.annotation.UiThread
+import org.json.JSONArray
+import org.json.JSONObject
+
+private const val HOME_TILES = "homeTiles"
```
mcomella: nit: Sorry I was unclear: I would have left the var name PREF (to indicate where it's used) but removed Pref from the value (because it's redundant with the usage). Also, I'd probably prefix with PREF since it makes it more obvious when multiple of these are used together that they're related.

app/src/main/java/org/mozilla/focus/tiles/HomeTilesManager.kt
```diff
@@ -0,0 +1,116 @@
...
+private const val KEY_URL = "url"
+private const val KEY_TITLE = "title"
+private const val KEY_IMG = "img"
+private const val KEY_ID = "identifier"
+
+data class HomeTile(
+        val url: String,
+        val title: String,
+        val imagePath: String?,
+        // unique id used to identify specific home tiles, e.g. for deletion, etc.
```
mcomella: nit: since this is a property, you actually want to use javadoc syntax.

app/src/main/java/org/mozilla/focus/tiles/HomeTilesManager.kt
```diff
@@ -0,0 +1,116 @@
...
+                    jsonObject.getString(KEY_TITLE),
+                    jsonObject.optString(KEY_IMG),
+                    jsonObject.getString(KEY_ID))
+        }
+    }
+}
+
+/**
+ * Static accessor of custom home tiles, that is backed by SharedPreferences.
+ * This *MUST* be initialized before being used.
```
mcomella: nit: outdated.

app/src/main/java/org/mozilla/focus/tiles/HomeTilesManager.kt
```diff
@@ -0,0 +1,116 @@
...
+ *
+ * New sites are appended to the end of the list.
+ *
+ * This keeps a cached version of the custom home tiles that have been pinned,
+ * in order to be more performant when checking whether sites are pinned or not.
+ * In order to keep the cache consistent, should only be called from the UIThread.
+ */
+class CustomTilesAccessor(context: Context) {
```
mcomella: nit: `class CustomTilesAccessor private constructor(context: Context) {` ...

app/src/main/java/org/mozilla/focus/tiles/HomeTilesManager.kt
```diff
@@ -0,0 +1,116 @@
...
+    // Cache pinned sites for perf beacues we need to check pinned state for every page load
+    private var customTilesCache = getCustomTilesCache(context)
```
mcomella: nit: below the companion obj

app/src/main/java/org/mozilla/focus/tiles/HomeTilesManager.kt
```diff
@@ -0,0 +1,116 @@
...
+        val tilesJSONArray = getCustomSitesJSONArray(getHomeTilesPreferences(context))
+        val lhm = LinkedHashMap<String, HomeTile>()
+        for (i in 0 until tilesJSONArray.length()) {
+            val tile = tilesJSONArray.getJSONObject(i)
+            lhm.put(tile.getString(KEY_URL), HomeTile.fromJSONObject(tile))
+        }
+        return lhm
+    }
+
+    fun isURLPinned(url: String) = customTilesCache.containsKey(url)
```
mcomella: nit: these two need to be `@UiThread` too

app/src/main/java/org/mozilla/focus/widget/BrowserNavigationOverlay.kt
```diff
@@ -27,11 +27,15 @@ enum class NavigationEvent {
             R.id.navButtonHome -> HOME
             R.id.navButtonSettings -> SETTINGS
             R.id.turboButton -> TURBO
+            R.id.pinButton -> PIN_ACTION
             else -> null
         }
     }
 }
 
+const val VAL_CHECKED = "checked"
```
mcomella: nit: add these add to the companion object of NavigationEvent, otherwise they'll end up in the autocomplete for all files (which isn't terrible, but not ideal).

app/src/main/java/org/mozilla/focus/tiles/HomeTilesManager.kt
mcomella: nit: I prefer `CustomTilesStore` or similar. `Accessor` feels ambiguous: what is it accessing? Is this a wrapper with getters/setters around some private APIs? Is this persistent storage?

To me, `Store` indicates persistent storage and tells me where we're reading and storing it to.

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
liuche: Good call :)

app/src/main/java/org/mozilla/focus/tiles/HomeTilesManager.kt
liuche: Going to rename (back) to CustomTilesManager (lol) because this isn't really a Store either.

PR #539, opened 2018-02-27
Closes #535: Use separate Docker image on taskcluster.

### LINKED ISSUE #535: Taskcluster: Use seprate Docker image
Currently "mozillamobile/focus-android" is used on taskcluster. I'm going to update/change this image. To avoid breaking firefox-tv builds on taskcluster it makes sense to build a separate image and push it to docker hub so that it can be used by taskcluster.

pocmo:
To avoid builds breaking on taskcluster whenever the Focus image changes:
Let's use a separate image.

As a side effect this fixes archiving test artifacts too.

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -26,7 +26,7 @@ tasks:
     payload:
       maxRunTime: 3600
       deadline: "{{ '2 hours' | $fromNow }}"
-      image: 'mozillamobile/focus-android'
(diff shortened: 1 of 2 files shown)


pocmo:
@liuche You can use any Docker image you want (or create/build your own). I can add anyone interested to the 'mozillamobile' org on Docker Hub if you want to push your image to it (That's where taskcluster is going to download it from). Other than that no additional permission should be needed.

PR #546, opened 2018-02-28
No issue: update readme to say how to toggle menu

mcomella:
This will let emulators open the menu. The solution is a bit hacky but
it's quick and works. If we run into problems, we can rewrite it.

I canfirmed that this opens the menu in Debug configs but it does not on
Beta configs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,8 @@ can be connected to a Fire TV at a time. Note that while you can install on an
 Android TV emulator, the behavior is different from Fire TV's and should not be
 relied upon.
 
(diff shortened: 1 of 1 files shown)


liuche:
@mcomella maybe we should add something for non-macs? but maybe that's okay.

mcomella:
> @mcomella maybe we should add something for non-macs? but maybe that's okay.

I don't think we have the resources to support non-macs so until we actually know someone is contributing without a mac, I think we'll just be mac-centric.

PR #552, opened 2018-03-01
 Allow removing of tiles #443 

bkmunar:
#443

### DIFF
--- a/app/src/main/java/org/mozilla/focus/activity/MainActivity.kt
+++ b/app/src/main/java/org/mozilla/focus/activity/MainActivity.kt
@@ -19,6 +19,7 @@ import kotlinx.android.synthetic.main.activity_main.container
 import org.mozilla.focus.R
 import org.mozilla.focus.architecture.NonNullObserver
 import org.mozilla.focus.fragment.BrowserFragment
+import org.mozilla.focus.fragment.HomeFragment
 import org.mozilla.focus.fragment.ScreenController
 import org.mozilla.focus.locale.LocaleAwareAppCompatActivity
 import org.mozilla.focus.session.Session
@@ -158,11 +159,14 @@ class MainActivity : LocaleAwareAppCompatActivity(), OnUrlEnteredListener {
     override fun dispatchKeyEvent(event: KeyEvent): Boolean {
         val fragmentManager = supportFragmentManager
         val browserFragment = fragmentManager.findFragmentByTag(BrowserFragment.FRAGMENT_TAG) as BrowserFragment?
+        val homeFragment = fragmentManager.findFragmentByTag(HomeFragment.FRAGMENT_TAG) as HomeFragment?
 
-        return if (browserFragment == null || !browserFragment.isVisible) {
-            super.dispatchKeyEvent(event)
-        } else {
+        return if (browserFragment != null && browserFragment.isVisible) {
             browserFragment.dispatchKeyEvent(event) || super.dispatchKeyEvent(event)
+        } else if (homeFragment != null && homeFragment.isVisible) {
+            homeFragment.dispatchKeyEvent(event) || super.dispatchKeyEvent(event)
+        } else {
+            super.dispatchKeyEvent(event)
         }
     }
 
--- a/app/src/main/java/org/mozilla/focus/fragment/HomeFragment.kt
+++ b/app/src/main/java/org/mozilla/focus/fragment/HomeFragment.kt
@@ -14,9 +14,7 @@ import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
 import android.view.animation.DecelerateInterpolator
-import android.widget.ImageView
 import android.widget.LinearLayout
-import android.widget.TextView
 import kotlinx.android.synthetic.main.fragment_home.*
 import kotlinx.coroutines.experimental.CancellationException
 import kotlinx.coroutines.experimental.CommonPool
@@ -35,6 +33,11 @@ import org.mozilla.focus.tiles.CustomHomeTile
 import org.mozilla.focus.tiles.CustomTilesManager
 import org.mozilla.focus.tiles.HomeTile
 import org.mozilla.focus.utils.OnUrlEnteredListener
+import android.view.ContextMenu
+import android.view.MenuItem
+import android.view.KeyEvent
+import kotlinx.android.synthetic.main.home_tile.view.*
+import org.mozilla.focus.ext.toUri
 
 private const val COL_COUNT = 5
 private const val SETTINGS_ICON_IDLE_ALPHA = 0.4f
@@ -84,6 +87,39 @@ class HomeFragment : Fragment() {
         settingsButton.setOnClickListener { v ->
             onSettingsPressed?.invoke()
         }
+
+        registerForContextMenu(view)
+    }
+
+    override fun onCreateContextMenu(menu: ContextMenu?, v: View?, menuInfo: ContextMenu.ContextMenuInfo?) {
+        super.onCreateContextMenu(menu, v, menuInfo)
+        activity.menuInflater.inflate(R.menu.menu_context_hometile, menu)
+    }
+
+    override fun onContextItemSelected(item: MenuItem): Boolean {
+        when (item.itemId) {
+            R.id.remove -> {
+                val homeTileAdapter = tileContainer.adapter as HomeTileAdapter
+                val tileToRemove = homeTileAdapter.getItemAtPosition(getFocusedTilePosition())
+                // This assumes that since we're deleting from a Home Tile object that we created
+                // that the Uri is valid, so we do not do error handling here.
+                when (tileToRemove) {
+                    is BundledHomeTile -> {
+                        val tileUri = tileToRemove.url.toUri()
+                        if (tileUri != null) {
+                            BundledTilesManager.getInstance(context).unpinSite(context, tileUri)
+                            homeTileAdapter.removeItemAtPosition(getFocusedTilePosition())
+                        }
+                    }
+                    is CustomHomeTile -> {
+                        CustomTilesManager.getInstance(context).unpinSite(context, tileToRemove.url)
+                        homeTileAdapter.removeItemAtPosition(getFocusedTilePosition())
+                    }
+                }
+                return true
+            }
+        }
+        return false
     }
 
     override fun onDestroyView() {
@@ -125,11 +161,23 @@ class HomeFragment : Fragment() {
         @JvmStatic
         fun create() = HomeFragment()
     }
+
+    fun getFocusedTilePosition(): Int {
+        return (activity.currentFocus.parent as? RecyclerView)?.getChildAdapterPosition(activity.currentFocus) ?: RecyclerView.NO_POSITION
+    }
+
+    fun dispatchKeyEvent(event: KeyEvent): Boolean {
+        if (event.keyCode == KeyEvent.KEYCODE_MENU && getFocusedTilePosition() != RecyclerView.NO_POSITION) {
+            activity.openContextMenu(view)
+            return true
+        }
+        return false
+    }
 }
 
 private class HomeTileAdapter(
         private val uiLifecycleCancelJob: Job,
-        private val tiles: List<HomeTile>,
+        private val tiles: MutableList<HomeTile>,
         private val onUrlEnteredListener: OnUrlEnteredListener
 ) : RecyclerView.Adapter<TileViewHolder>() {
 
@@ -153,12 +201,28 @@ private class HomeTileAdapter(
             if (hasFocus) {
                 backgroundResource = R.drawable.home_tile_title_focused_background
                 textColor = Color.WHITE
+                menuButton.visibility = View.VISIBLE
             } else {
                 backgroundResource = 0
                 textColor = Color.BLACK
+                menuButton.visibility = View.GONE
             }
             titleView.setBackgroundResource(backgroundResource)
             titleView.setTextColor(textColor)
+        }
+    }
+
+    fun getItemAtPosition(position: Int): HomeTile? {
+        if (position > -1 && position < itemCount) {
+            return tiles[position]
+        }
+        return null
+    }
+
+    fun removeItemAtPosition(position: Int) {
+        if (position > -1 && position < itemCount) {
+            tiles.removeAt(position)
+            notifyItemRemoved(position)
         }
     }
 
@@ -191,6 +255,7 @@ private fun onBindCustomHomeTile(uiLifecycleCancelJob: Job, holder: TileViewHold
 private class TileViewHolder(
         itemView: View
 ) : RecyclerView.ViewHolder(itemView) {
-    val iconView = itemView.findViewById<ImageView>(R.id.tile_icon)
-    val titleView = itemView.findViewById<TextView>(R.id.tile_title)
+    val iconView = itemView.tile_icon
+    val titleView = itemView.tile_title
+    val menuButton = itemView.home_tile_menu_icon
 }
--- a/app/src/main/java/org/mozilla/focus/tiles/HomeTilesManager.kt
+++ b/app/src/main/java/org/mozilla/focus/tiles/HomeTilesManager.kt
@@ -19,6 +19,7 @@ import org.mozilla.focus.utils.UrlUtils
 import java.util.UUID
 
 private const val PREF_HOME_TILES = "homeTiles"
+private const val BUNDLED_SITES_ID_BLACKLIST = "blacklist"
 private const val CUSTOM_SITES_LIST = "customSitesList"
 
 private const val BUNDLED_HOME_TILES_DIR = "bundled"
@@ -49,10 +50,12 @@ class BundledTilesManager private constructor(context: Context) {
         val tilesJSONString = context.assets.open(HOME_TILES_JSON_PATH).bufferedReader().use { it.readText() }
         val tilesJSONArray = JSONArray(tilesJSONString)
         val lhm = LinkedHashMap<Uri, BundledHomeTile>(tilesJSONArray.length())
+        val blacklist = loadBlacklist(context)
         for (i in 0 until tilesJSONArray.length()) {
             val tile = BundledHomeTile.fromJSONObject(tilesJSONArray.getJSONObject(i))
-            // TODO: Check for blacklisted sites and don't add them
-            lhm.put(tile.url.toUri()!!, tile)
+            if (!blacklist.contains(tile.id)) {
+                lhm.put(tile.url.toUri()!!, tile)
+            }
         }
         return lhm
     }
@@ -76,9 +79,14 @@ class BundledTilesManager private constructor(context: Context) {
 
     @UiThread
     fun unpinSite(context: Context, uri: Uri): Boolean {
+        val blacklist = loadBlacklist(context)
+        val newBlacklist = blacklist.toMutableSet()
         for (pair in bundledTilesCache) {
             if (compareUri(uri, pair.key)) {
-                // TODO: Remove by pair.value.id (BEFORE removing from cache)
+                newBlacklist.add(pair.value.id)
+                context.getSharedPreferences(PREF_HOME_TILES, MODE_PRIVATE).edit()
+                        .putStringSet(BUNDLED_SITES_ID_BLACKLIST, newBlacklist)
+                        .apply()
                 bundledTilesCache.remove(pair.key)
                 return true
             }
@@ -90,6 +98,10 @@ class BundledTilesManager private constructor(context: Context) {
     fun loadImageFromPath(context: Context, path: String) = context.assets.open(
             "$BUNDLED_HOME_TILES_DIR/$path").use {
         BitmapFactory.decodeStream(it)
+    }
+
+    private fun loadBlacklist(context: Context): MutableSet<String> {
+        return context.getSharedPreferences(PREF_HOME_TILES, MODE_PRIVATE).getStringSet(BUNDLED_SITES_ID_BLACKLIST, mutableSetOf())
     }
 
     @UiThread
--- a/app/src/main/res/drawable/context_menu_button_background.xml
+++ b/app/src/main/res/drawable/context_menu_button_background.xml
@@ -1,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
+   - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
+<shape
+    xmlns:android="http://schemas.android.com/apk/res/android"
+    android:shape="oval">
+    <solid android:color="@color/default_background_gray" />
+    <stroke android:color="@color/overlay_button_border"
+        android:width="0.5dp" />
+</shape>
--- a/app/src/main/res/layout/fragment_home.xml
+++ b/app/src/main/res/layout/fragment_home.xml
@@ -5,6 +5,7 @@
 <!-- TODO: scrolll view -->
 <LinearLayout xmlns:android="http://schemas.android.com/apk/res/android" xmlns:tools="http://schemas.android.com/tools"
               android:orientation="vertical"
+              android:clipChildren="false"
               android:layout_width="match_parent"
               android:layout_height="match_parent"
               android:padding="50dp"
@@ -68,6 +69,7 @@
 
     <android.support.v7.widget.RecyclerView
             android:id="@+id/tileContainer"
+            android:clipChildren="false"
             android:layout_width="match_parent"
             android:layout_height="wrap_content"
             android:background="@android:color/transparent"/>
--- a/app/src/main/res/layout/home_tile.xml
+++ b/app/src/main/res/layout/home_tile.xml
@@ -2,7 +2,7 @@
 <!-- This Source Code Form is subject to the terms of the Mozilla Public
    - License, v. 2.0. If a copy of the MPL was not distributed with this
    - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
-<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+<FrameLayout xmlns:android="http://schemas.android.com/apk/res/android"
               xmlns:tools="http://schemas.android.com/tools"
               android:orientation="vertical"
               android:layout_width="140dp"
@@ -11,30 +11,49 @@
               android:foreground="@drawable/home_tile_selector"
               android:focusable="true"
               android:clickable="true"
+              android:clipChildren="false"
               android:layout_marginStart="12dp"
               android:layout_marginEnd="12dp"
               android:layout_marginBottom="18dp">
 
     <ImageView
+        android:id="@+id/home_tile_menu_icon"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:layout_gravity="end|top"
+        android:layout_marginEnd="-10dp"
+        android:layout_marginTop="-10dp"
+        android:background="@drawable/context_menu_button_background"
+        android:elevation="4dp"
+        android:padding="4dp"
+        android:src="@drawable/ic_wide_menu_home"
+        android:visibility="gone" />
+
+    <LinearLayout
+        android:layout_width="match_parent"
+        android:layout_height="match_parent"
+        android:orientation="vertical">
+
+        <ImageView
             android:id="@+id/tile_icon"
             android:layout_width="match_parent"
             android:layout_height="0dp"
-            android:scaleType="centerInside"
-            android:layout_weight="1"
             android:layout_margin="8dp"
+            android:layout_weight="1"
+            android:scaleType="centerInside"
             tools:ignore="ContentDescription"
-            tools:src="@drawable/tile_youtube"/>
+            tools:src="@drawable/tile_youtube" />
 
-    <TextView
-        android:id="@+id/tile_title"
-        android:layout_width="match_parent"
-        android:layout_height="wrap_content"
-        android:paddingTop="6dp"
-        android:paddingBottom="6dp"
-        android:gravity="center_horizontal"
-        android:textColor="@android:color/black"
-        android:fontFamily="sans-serif-medium"
-        android:maxLines="1"
-        tools:text="YouTube"/>
-
-</LinearLayout>
+        <TextView
+            android:id="@+id/tile_title"
+            android:layout_width="match_parent"
+            android:layout_height="wrap_content"
+            android:fontFamily="sans-serif-medium"
+            android:gravity="center_horizontal"
+            android:maxLines="1"
+            android:paddingBottom="6dp"
+            android:paddingTop="6dp"
+            android:textColor="@android:color/black"
+            tools:text="YouTube" />
+    </LinearLayout>
+</FrameLayout>
--- a/app/src/main/res/menu/menu_context_hometile.xml
+++ b/app/src/main/res/menu/menu_context_hometile.xml
@@ -1,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
+   - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
+<menu xmlns:android="http://schemas.android.com/apk/res/android">
+    <item android:id="@+id/remove"
+        android:title="@string/homescreen_tile_remove" />
+</menu>
--- a/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
+++ b/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
@@ -15,6 +15,7 @@ import android.view.ViewGroup
 import android.widget.FrameLayout
 import kotlinx.android.synthetic.main.fragment_browser.*
 import kotlinx.android.synthetic.main.fragment_browser.view.*
+import kotlinx.android.synthetic.main.home_tile.*
 import org.mozilla.focus.R
 import org.mozilla.focus.activity.MainActivity
 import org.mozilla.focus.architecture.NonNullObserver
--- a/app/src/main/res/drawable/ic_wide_menu_home.xml
+++ b/app/src/main/res/drawable/ic_wide_menu_home.xml
@@ -1,0 +1,21 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
+   - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
+
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:width="24dp"
+    android:height="24dp"
+    android:viewportWidth="24.0"
+    android:viewportHeight="24.0">
+    <path
+        android:fillColor="#2a2a2e"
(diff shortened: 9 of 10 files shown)


app/src/main/res/layout/home_tile.xml
```diff
@@ -1,40 +1,58 @@
...
     <ImageView
+        android:id="@+id/home_tile_menu_icon"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:layout_gravity="end|top"
+        android:layout_marginEnd="-15dp"
+        android:layout_marginTop="-15dp"
+        android:background="@drawable/circle_button_background"
+        android:elevation="4dp"
+        android:src="@drawable/ic_home_menu"
```
lime124: can we use the same icon Jeff has for the overlay - ic_wide_menu
and then tint the icon used for our purpose on the home screen as #2a2a2e

app/src/main/res/layout/home_tile.xml
```diff
@@ -1,40 +1,58 @@
...
+    android:focusable="true"
+    android:foreground="@drawable/home_tile_selector"
+    android:orientation="vertical">
 
     <ImageView
+        android:id="@+id/home_tile_menu_icon"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:layout_gravity="end|top"
+        android:layout_marginEnd="-15dp"
```
lime124: change both marginTop and marginEnd to -10dp

app/src/main/res/layout/home_tile.xml
```diff
@@ -1,40 +1,58 @@
...
+        android:layout_marginTop="-15dp"
+        android:background="@drawable/circle_button_background"
+        android:elevation="4dp"
+        android:src="@drawable/ic_home_menu"
+        android:visibility="gone" />
```
lime124: add padding of 4dp

app/src/main/res/layout/home_tile.xml
```diff
@@ -1,40 +1,58 @@
...
+    android:orientation="vertical">
 
     <ImageView
+        android:id="@+id/home_tile_menu_icon"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:layout_gravity="end|top"
+        android:layout_marginEnd="-15dp"
+        android:layout_marginTop="-15dp"
+        android:background="@drawable/circle_button_background"
```
lime124: change size of this to 24dp
color to e3e3e5 (same as progress/loading bar background) - it's a very similar grey so it won't look like it changed a whole lot

app/src/main/java/org/mozilla/focus/fragment/HomeFragment.kt
```diff
@@ -63,6 +74,22 @@ class HomeFragment : Fragment() {
...
             onSettingsPressed?.invoke()
         }
+
+        registerForContextMenu(view)
+    }
+
+    override fun onCreateContextMenu(menu: ContextMenu?, v: View?, menuInfo: ContextMenu.ContextMenuInfo?) {
+        super.onCreateContextMenu(menu, v, menuInfo)
+//        menu?.setHeaderTitle("Select Option")
+//        menu?.add(Menu.NONE, v!!.getId(), 0, "Remove tile")
```
lime124: we don't need the word 'tile'

lime124: also just want to double check that the string is from the string file so it gets translated.

lime124:
I couldn't quickly find where you reference the button_pressed color for the pressed state to comment on it. The press state should match the tile press state instead.

mcomella:
@bkmunar TaskCluster failures from lint.

Please set up [the pre-push hook](https://github.com/mozilla-mobile/firefox-tv/wiki/Recommended-pre-push-hook) so this doesn't happen!

app/src/main/java/org/mozilla/focus/activity/MainActivity.kt
```diff
@@ -158,11 +159,14 @@ class MainActivity : LocaleAwareAppCompatActivity(), OnUrlEnteredListener {
...
+        val homeFragment = fragmentManager.findFragmentByTag(HomeFragment.FRAGMENT_TAG) as HomeFragment?
 
-        return if (browserFragment == null || !browserFragment.isVisible) {
+        return if ((browserFragment == null || !browserFragment.isVisible) && (homeFragment == null || !homeFragment.isVisible)) {
             super.dispatchKeyEvent(event)
+        } else if (homeFragment!!.isVisible) {
+            homeFragment.dispatchKeyEvent(event) || super.dispatchKeyEvent(event)
         } else {
-            browserFragment.dispatchKeyEvent(event) || super.dispatchKeyEvent(event)
+            browserFragment!!.dispatchKeyEvent(event) || super.dispatchKeyEvent(event)
```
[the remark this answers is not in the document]
liuche: Why don't you check each one in order:
`browserFragment != null && browserFragment.isVisible`
and then 
`homeFragment != null && homeFragment.isVisible`
and then have an else?

app/src/main/java/org/mozilla/focus/fragment/HomeFragment.kt
```diff
@@ -28,20 +25,32 @@ import org.mozilla.focus.tiles.CustomTilesManager
 import org.mozilla.focus.tiles.BundledHomeTile
 import org.mozilla.focus.tiles.HomeTile
 import org.mozilla.focus.utils.OnUrlEnteredListener
+import android.content.Context.MODE_PRIVATE
```
liuche: I'm not sure why these imports moved from above to here, please fix your Android Studio to not autoformat these imports for no reason.

app/src/main/java/org/mozilla/focus/fragment/HomeFragment.kt
```diff
@@ -28,20 +25,32 @@ import org.mozilla.focus.tiles.CustomTilesManager
...
+import android.view.MenuItem
+import org.json.JSONArray
 
 private const val COL_COUNT = 5
 private const val BUNDLED_HOME_TILES_DIR = "bundled/"
 private const val HOME_TILES_JSON_PATH = BUNDLED_HOME_TILES_DIR + "bundled_tiles.json"
 private const val HOME_TILES_JSON_KEY = "bundled_tiles"
 private const val SETTINGS_ICON_IDLE_ALPHA = 0.4f
 private const val SETTINGS_ICON_ACTIVE_ALPHA = 1.0f
+private const val bundled_blacklist = "blacklist"
```
liuche: Will review again after these move to the HomeTileManager.

app/src/main/java/org/mozilla/focus/fragment/HomeFragment.kt
```diff
@@ -63,6 +74,22 @@ class HomeFragment : Fragment() {
...
         }
+
+        registerForContextMenu(view)
+    }
+
+    override fun onCreateContextMenu(menu: ContextMenu?, v: View?, menuInfo: ContextMenu.ContextMenuInfo?) {
+        super.onCreateContextMenu(menu, v, menuInfo)
+//        menu?.setHeaderTitle("Select Option")
+//        menu?.add(Menu.NONE, v!!.getId(), 0, "Remove tile")
+        activity.menuInflater.inflate(R.menu.menu_context_home, menu)
```
liuche: Nit: rename this because it's not just a home context menu, it's menu_context_hometile

app/src/main/java/org/mozilla/focus/fragment/HomeFragment.kt
```diff
@@ -63,6 +74,22 @@ class HomeFragment : Fragment() {
...
+    }
+
+    override fun onContextItemSelected(item: MenuItem?): Boolean {
```
liuche: Why is this an optional? Can you remove the `?` from MenuItem.

app/src/main/java/org/mozilla/focus/fragment/HomeFragment.kt
```diff
@@ -82,33 +109,76 @@ class HomeFragment : Fragment() {
 
         val inputAsString = context.assets.open(HOME_TILES_JSON_PATH).bufferedReader().use { it.readText() }
         val jsonArray = JSONObject(inputAsString).getJSONArray(HOME_TILES_JSON_KEY)
-        for (i in 0..(jsonArray.length() - 1)) {
-            val jsonObject = jsonArray.getJSONObject(i)
-            homeTiles.add(BundledHomeTile.fromJSONObject(jsonObject))
-        }
-
-        adapter = HomeTileAdapter(onUrlEnteredListener, homeTiles)
+        unpackJsonHomeTiles(homeTiles, jsonArray)
```
liuche: Shouldn't be necessary any more, see HomeTilesManager.

app/src/main/java/org/mozilla/focus/fragment/HomeFragment.kt
```diff
@@ -82,33 +109,76 @@ class HomeFragment : Fragment() {
...
+    }
+
     private fun initUrlInputView() = with (urlInputView) {
         setOnCommitListener {
             onUrlEnteredListener.onTextInputUrlEntered(text.toString(), urlInputView.lastAutocompleteResult, UrlTextInputLocation.HOME)
         }
         setOnFilterListener { searchText, view -> urlAutoCompleteFilter.onFilter(searchText, view) }
     }
 
+    fun addBundledTileToBlacklist(id: String) {
```
liuche: This should be moved to the BundledHomeTilesManager, with the exception of `removeItemAtPosition`

app/src/main/java/org/mozilla/focus/fragment/HomeFragment.kt
```diff
@@ -82,33 +109,76 @@ class HomeFragment : Fragment() {
...
 
         @JvmStatic
         fun create() = HomeFragment()
     }
+
+    fun getFocusedTile(): Int {
+        return (activity.currentFocus.parent as? RecyclerView)?.getChildAdapterPosition(activity.currentFocus) ?: RecyclerView.NO_POSITION
+    }
+
+    fun dispatchKeyEvent(event: KeyEvent): Boolean {
```
liuche: Just combine these, and update the non-relevant comment about overlay. Be more aware of updating things like comments if you're copypastaing stuff.

app/src/main/java/org/mozilla/focus/fragment/HomeFragment.kt
```diff
@@ -82,33 +109,76 @@ class HomeFragment : Fragment() {
...
+            return true
+        }
+        return false
+    }
 }
 
-private class HomeTileAdapter(val onUrlEnteredListener: OnUrlEnteredListener, val tiles: List<HomeTile>) :
+private class HomeTileAdapter(val onUrlEnteredListener: OnUrlEnteredListener, val tiles: MutableList<HomeTile>) :
         RecyclerView.Adapter<TileViewHolder>() {
+    var currentFocusedIndex: Int = -1
```
liuche: We talked about moving this out before - if it doesn't happen for this PR, file an issue with everything that needs to be cleaned up.

app/src/main/java/org/mozilla/focus/fragment/HomeFragment.kt
```diff
@@ -130,15 +200,25 @@ private class HomeTileAdapter(val onUrlEnteredListener: OnUrlEnteredListener, va
...
             }
             titleView.setBackgroundResource(backgroundResource)
             titleView.setTextColor(textColor)
         }
     }
 
+    fun removeItemAtPosition(position: Int) {
+        tiles.removeAt(position)
+        notifyItemRemoved(position)
+        notifyItemRangeChanged(position, itemCount)
```
liuche: I think this can be removed now that you've removed the long-click listener...maybe?f

app/src/main/java/org/mozilla/focus/fragment/HomeFragment.kt
```diff
@@ -158,4 +238,5 @@ private class TileViewHolder(
 ) : RecyclerView.ViewHolder(itemView) {
     val iconView = itemView.findViewById<ImageView>(R.id.tile_icon)
     val titleView = itemView.findViewById<TextView>(R.id.tile_title)
+    val menuButton = itemView.findViewById<ImageView>(R.id.home_tile_menu_icon)
```
liuche: These can just be accessed directly by id name with Kotlin, you don't need to do findViewById I think.

app/src/main/res/drawable/ic_home_menu.xml
```diff
@@ -0,0 +1,15 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
```
liuche: Add license.

app/src/main/res/layout/fragment_home.xml
```diff
@@ -1,50 +1,52 @@
...
-   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
-<!-- TODO: scrolll view -->
-<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android" xmlns:tools="http://schemas.android.com/tools"
-              android:orientation="vertical"
-              android:layout_width="match_parent"
-              android:layout_height="match_parent"
-              android:padding="50dp"
-              android:background="@drawable/background_home">
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. --><!-- TODO: scrolll view -->
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
liuche: (Make sure you fix your Android Studio so it doesn't do this anymore, or at least look at your commits with `git diff` so you don't commit formatting changes that obscure commit history.)

app/src/main/res/layout/fragment_home.xml
```diff
@@ -1,50 +1,52 @@
...
-                android:requiresFadingEdge="horizontal"
-                android:selectAllOnFocus="true"
-                android:textColor="#ffffff"
-                android:textColorHighlight="@color/colorAccent"
-                android:textColorHint="@color/colorTextInactive"
-                android:textSize="15sp"
-                android:textCursorDrawable="@drawable/cursor"
-                android:importantForAutofill="no"
-                tools:targetApi="o" /><!-- importantForAutofill is only supported on Android O+ -->
+        x
```
liuche: Remove this floating x!

app/src/main/java/org/mozilla/focus/activity/MainActivity.kt
```diff
@@ -158,11 +159,14 @@ class MainActivity : LocaleAwareAppCompatActivity(), OnUrlEnteredListener {
     override fun dispatchKeyEvent(event: KeyEvent): Boolean {
         val fragmentManager = supportFragmentManager
         val browserFragment = fragmentManager.findFragmentByTag(BrowserFragment.FRAGMENT_TAG) as BrowserFragment?
+        val homeFragment = fragmentManager.findFragmentByTag(HomeFragment.FRAGMENT_TAG) as HomeFragment?
 
-        return if (browserFragment == null || !browserFragment.isVisible) {
-            super.dispatchKeyEvent(event)
+        return if (browserFragment != null && browserFragment.isVisible) {
+            browserFragment.dispatchKeyEvent(event)
```
liuche: This still needs the `super.dispatchKey` in an || statement - browserFragment doesn't handle every case, and we always want it to fall to the super. See the code that was deleted, in comment below.

app/src/main/java/org/mozilla/focus/activity/MainActivity.kt
```diff
@@ -158,11 +159,14 @@ class MainActivity : LocaleAwareAppCompatActivity(), OnUrlEnteredListener {
...
+        } else if (homeFragment != null && homeFragment.isVisible) {
+            homeFragment.dispatchKeyEvent(event)
         } else {
-            browserFragment.dispatchKeyEvent(event) || super.dispatchKeyEvent(event)
```
liuche: See deleted code here: if browserfragment doesn't handle the code, we want this to be handled by the super.

Try to be more careful when moving behavior around - removing the `super` call would have introduced regressions.

app/src/main/java/org/mozilla/focus/fragment/HomeFragment.kt
```diff
@@ -35,10 +33,17 @@ import org.mozilla.focus.tiles.CustomHomeTile
...
+import android.view.MenuItem
+import android.view.KeyEvent
+import kotlinx.android.synthetic.main.home_tile.view.*
+import org.mozilla.focus.ext.toUri
 
 private const val COL_COUNT = 5
 private const val SETTINGS_ICON_IDLE_ALPHA = 0.4f
 private const val SETTINGS_ICON_ACTIVE_ALPHA = 1.0f
+private const val bundled_blacklist = "blacklist"
+private const val HOME_TILES_PREFS = "homeTilesPrefs"
```
liuche: This pref value has already been added to HomeTileManager - do not duplicate it here.

Please take a look at the code in HomeTileManager, and integrate your changes there. To keep the code clean (such as HomeFragment) we don't want to do logic in here that should be controlled by something like HomeTileManager. It makes the overall logic of HomeFragment harder to read, because 1) there is more code so you have to know what is important/not, and 2) it makes HomeTileManager also confusing because it's no longer the place to look for all the Home Tile code, some of it's hanging out here.

app/src/main/java/org/mozilla/focus/fragment/HomeFragment.kt
```diff
@@ -84,6 +90,28 @@ class HomeFragment : Fragment() {
...
+        super.onCreateContextMenu(menu, v, menuInfo)
+        activity.menuInflater.inflate(R.menu.menu_context_home, menu)
+    }
+
+    override fun onContextItemSelected(item: MenuItem?): Boolean {
+        when (item!!.itemId) {
+            R.id.remove -> {
+                val tileToRemove = homeTileAdapter.homeTiles[getFocusedTilePosition()]
+                if (tileToRemove is BundledHomeTile) {
+                    BundledTilesManager.getInstance(context).unpinSite(context, tileToRemove.url.toUri()!!)
```
liuche: You can't pass in a null uri here - if you look at the code in unpinSite, that's not handled. One thing you can do is check to make sure the url isn't null before passing it around.

Before you !! things (which you should do ONLY when you are absolutely sure that it's not null). I don't see a comment for why you're sure this isn't null, or (even better and preferred) a null check, so this seems like an unjustified !!.

Always be suspicious of your inputs, and don't assume that they aren't null because that causes crashes.

app/src/main/java/org/mozilla/focus/fragment/HomeFragment.kt
```diff
@@ -107,7 +135,9 @@ class HomeFragment : Fragment() {
             addAll(BundledTilesManager.getInstance(context).getBundledHomeTilesList())
         }
 
-        adapter = HomeTileAdapter(uiLifecycleCancelJob, homeTiles, onUrlEnteredListener)
+        val hta = HomeTileAdapter(uiLifecycleCancelJob, homeTiles, onUrlEnteredListener)
+        adapter = hta
+        homeTileAdapter = hta
```
liuche: lol what is this? why do you have so many adapters? This looks wrong. Do you need all these adapters? Since both of them are pointing to the same reference (hta) probably one of them is unnecessary. If you do need them both, add a comment as to why.

In general, this is bad code smell, because you have two similarly named variables, both being set to the same thnig. Before submitting something like this for review, do your own check to make sure you know why it's happening, and then add a comment explaining why it's necessary (because you *know* it's going to get flagged in review bc it looks so weird).

app/src/main/java/org/mozilla/focus/fragment/HomeFragment.kt
```diff
@@ -84,6 +90,28 @@ class HomeFragment : Fragment() {
...
+        super.onCreateContextMenu(menu, v, menuInfo)
+        activity.menuInflater.inflate(R.menu.menu_context_hometile, menu)
+    }
+
+    override fun onContextItemSelected(item: MenuItem): Boolean {
+        when (item.itemId) {
+            R.id.remove -> {
+                val tileToRemove = homeTileAdapter.homeTiles[getFocusedTilePosition()]
+                if (tileToRemove is BundledHomeTile) {
+                    BundledTilesManager.getInstance(context).unpinSite(context, tileToRemove.url.toUri()!!)
```
liuche: You can't pass in a null uri here - if you look at the code in unpinSite, that's not handled. One thing you can do is check to make sure the url isn't null before passing it around.

Before you !! things (which you should do ONLY when you are absolutely sure that it's not null). I don't see a comment for why you're sure this isn't null, or (even better and preferred) a null check, so this seems like an unjustified !!.

Always be suspicious of your inputs, and don't assume that they aren't null because that causes crashes.

app/src/main/java/org/mozilla/focus/fragment/HomeFragment.kt
```diff
@@ -125,16 +155,29 @@ class HomeFragment : Fragment() {
         @JvmStatic
         fun create() = HomeFragment()
     }
+
+    fun getFocusedTilePosition(): Int {
+        return (activity.currentFocus.parent as? RecyclerView)?.getChildAdapterPosition(activity.currentFocus) ?: RecyclerView.NO_POSITION
+    }
+
+    fun dispatchKeyEvent(event: KeyEvent): Boolean {
+        if (event.keyCode == KeyEvent.KEYCODE_DEL && getFocusedTilePosition() != RecyclerView.NO_POSITION) {
```
liuche: Why delete, is this for the emulator? If so, add a comment. Also reference the menu button behavior here.

app/src/main/java/org/mozilla/focus/fragment/HomeFragment.kt
```diff
@@ -125,16 +155,29 @@ class HomeFragment : Fragment() {
...
+    }
 }
 
 private class HomeTileAdapter(
         private val uiLifecycleCancelJob: Job,
-        private val tiles: List<HomeTile>,
+        tiles: MutableList<HomeTile>,
         private val onUrlEnteredListener: OnUrlEnteredListener
 ) : RecyclerView.Adapter<TileViewHolder>() {
+    var homeTiles = tiles
```
liuche: No need to reassign this, I think - can you just keep the change from above (`private val tiles`) and either rename to `homeTiles` if you want it to be named that way, and change it to `var` if you want to be able to mutate it?

By putting `private val` in the arugment list, you can automatically assign it to a variable and just use that.

app/src/main/java/org/mozilla/focus/fragment/HomeFragment.kt
```diff
@@ -153,16 +196,24 @@ private class HomeTileAdapter(
...
             titleView.setBackgroundResource(backgroundResource)
             titleView.setTextColor(textColor)
         }
     }
 
-    override fun getItemCount() = tiles.size
+    fun removeItemAtPosition(position: Int) {
+        homeTiles.removeAt(position)
+        notifyItemRemoved(position)
+        notifyItemRangeChanged(position, itemCount)
```
liuche: Mmmm since we removed the long-click listener, could you remove this?

In the future, if you choose not to address a comment, please comment in response so I know why.

app/src/main/java/org/mozilla/focus/tiles/HomeTilesManager.kt
```diff
@@ -19,6 +19,7 @@ import org.mozilla.focus.utils.UrlUtils
 import java.util.UUID
 
 private const val PREF_HOME_TILES = "homeTiles"
+private const val BUNDLED_SITES_BLACKLIST = "blacklist"
```
liuche: Oh, since you have this here, you can delete the earlier unnecessary references.

app/src/main/java/org/mozilla/focus/tiles/HomeTilesManager.kt
```diff
@@ -78,7 +81,12 @@ class BundledTilesManager private constructor(context: Context) {
     fun unpinSite(context: Context, uri: Uri): Boolean {
         for (pair in bundledTilesCache) {
             if (compareUri(uri, pair.key)) {
-                // TODO: Remove by pair.value.id (BEFORE removing from cache)
+                val blacklist = context.getSharedPreferences(PREF_HOME_TILES, MODE_PRIVATE).getStringSet(BUNDLED_SITES_BLACKLIST, mutableSetOf())
```
liuche: Since you're using this in multiple places, just make a private fun to load the blacklist.

app/src/main/java/org/mozilla/focus/tiles/HomeTilesManager.kt
```diff
@@ -78,7 +81,12 @@ class BundledTilesManager private constructor(context: Context) {
...
+                val edit = context.getSharedPreferences(PREF_HOME_TILES, MODE_PRIVATE).edit()
+                val newBlacklist = blacklist.toMutableSet()
+                newBlacklist.add(pair.value.id)
+                edit.putStringSet(BUNDLED_SITES_BLACKLIST, newBlacklist)
+                edit.apply()
```
liuche: You can just put the apply at the end of the previous edit line - see documentation, all of these put commands return an instance of the editor.

https://developer.android.com/reference/android/content/SharedPreferences.Editor.html

app/src/main/res/drawable/context_menu_button_background.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
+   - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
+<shape
+    xmlns:android="http://schemas.android.com/apk/res/android"
+    android:shape="oval">
+    <size
+        android:height="24dp"
```
liuche: Do you still need this size if the size is already defined in the vector as 24? If you do, then please pull them out into a dimens, something like home_tile_badge_size

app/src/main/res/drawable/context_menu_button_background.xml
```diff
@@ -0,0 +1,14 @@
...
+        android:width="24dp" />
+    <solid android:color="#e3e3e5" />
```
liuche: Is this color something that will probably be reused? If so, pull it out into colors.xml.

app/src/main/java/org/mozilla/focus/tiles/HomeTilesManager.kt
liuche: nit, be explicit that this is of ids.

BUNDLED_SITES_IDS_BLACKLIST
and
blacklist_ids

That way, anyone who is looking at the blacklist variable names knows what the list is composed of (ids), rather than having to guess, or find other code that actually adds items to the blacklist to figure that out.

app/src/main/res/values/strings.xml
```diff
@@ -559,6 +559,9 @@ be temporary, and you can try again later.</li>
     <string name="settings_cookies_dialog_title">Clear all cookies and site data</string>
     <string name="settings_cookies_dialog_content">Selecting ‘Clear Now’ will clear all cookies and site data stored by Firefox for Fire TV. This may sign you out of websites and remove offline web content.</string>
 
+    <!-- Home Tile context menu strings -->
+    <string name="home_tile_context_menu_remove">Remove</string>
```
liuche: Is this rebased off of master? I think this string is already in this string file, so you don't need to add a new one - just use the one in the file. IIRC you reviewed the code to add this string for me.

app/src/main/java/org/mozilla/focus/fragment/HomeFragment.kt
bkmunar: ^ just forgot to remove it, woopsss!!

app/src/main/java/org/mozilla/focus/fragment/HomeFragment.kt
```diff
@@ -153,15 +196,26 @@ private class HomeTileAdapter(
...
                 textColor = Color.BLACK
+                menuButton.visibility = View.GONE
             }
             titleView.setBackgroundResource(backgroundResource)
             titleView.setTextColor(textColor)
         }
     }
 
+    fun getItemAtPosition(position: Int): HomeTile {
+        return tiles[position]
```
liuche: Very dangerous! Since this is a public function, what if the position passed in is out of the range? You should check the index and return null if it's out of range. Also document that behavior in a javadoc style comment like /**

app/src/main/java/org/mozilla/focus/fragment/HomeFragment.kt
```diff
@@ -153,15 +196,26 @@ private class HomeTileAdapter(
...
+    }
+
+    fun removeItemAtPosition(position: Int) {
+        tiles.removeAt(position)
+        notifyItemRemoved(position)
```
liuche: Again, should probably check if this position is allowed to be removed, if this is a public function. If you call notifyItemRemoved and the position is out of range, you'll get a crash.

app/src/main/res/layout/fragment_home.xml
```diff
@@ -1,50 +1,51 @@
...
-                android:requiresFadingEdge="horizontal"
-                android:selectAllOnFocus="true"
-                android:textColor="#ffffff"
-                android:textColorHighlight="@color/colorAccent"
-                android:textColorHint="@color/colorTextInactive"
-                android:textSize="15sp"
-                android:textCursorDrawable="@drawable/cursor"
-                android:importantForAutofill="no"
-                tools:targetApi="o" /><!-- importantForAutofill is only supported on Android O+ -->
+
```
liuche: Nit: unnecessary newline

app/src/main/res/layout/fragment_home.xml
```diff
@@ -1,50 +1,51 @@
...
+            android:paddingStart="8dp"
+            android:paddingTop="8dp"
+            android:requiresFadingEdge="horizontal"
+            android:selectAllOnFocus="true"
+            android:textColor="#ffffff"
+            android:textColorHighlight="@color/colorAccent"
+            android:textColorHint="@color/colorTextInactive"
+            android:textCursorDrawable="@drawable/cursor"
+            android:textSize="15sp"
+            tools:targetApi="o" /><!-- importantForAutofill is only supported on Android O+ -->
```
liuche: I'm going to assume you didn't change this InlineAutocompleteEditText bc this is too many items to compare by hand.

app/src/main/res/layout/fragment_home.xml
```diff
@@ -1,50 +1,51 @@
-<?xml version="1.0" encoding="utf-8"?>
-<!-- This Source Code Form is subject to the terms of the Mozilla Public
+<?xml version="1.0" encoding="utf-8"?><!-- This Source Code Form is subject to the terms of the Mozilla Public
```
liuche: Did you change this file at all? It actually looks like ALL you did was add the `clipChildren=false` in the RecyclerView at the end. Can you revert this whole file and then just include that one change?

`git rebase -i` this commit, and then `edit` it
`git checkout app/src/.../fragment_home.xml`

app/src/main/res/layout/home_tile.xml
```diff
@@ -1,40 +1,59 @@
...
+<FrameLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:tools="http://schemas.android.com/tools"
+    android:layout_width="140dp"
+    android:layout_height="100dp"
+    android:layout_marginBottom="18dp"
+    android:layout_marginEnd="12dp"
+    android:layout_marginStart="12dp"
+    android:background="@drawable/home_tile_background"
+    android:clickable="true"
+    android:clipChildren="false"
```
liuche: Is this the only thing you added? What's that for?

app/src/main/res/menu/menu_context_hometile.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
+   - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
+<menu xmlns:android="http://schemas.android.com/apk/res/android"
+    >
```
liuche: This `>` shouldn't be on a separate line

app/src/main/java/org/mozilla/focus/fragment/HomeFragment.kt
liuche: @bkmunar you didn't address this comment.

app/src/main/java/org/mozilla/focus/tiles/HomeTilesManager.kt
```diff
@@ -51,8 +52,10 @@ class BundledTilesManager private constructor(context: Context) {
         val lhm = LinkedHashMap<Uri, BundledHomeTile>(tilesJSONArray.length())
         for (i in 0 until tilesJSONArray.length()) {
             val tile = BundledHomeTile.fromJSONObject(tilesJSONArray.getJSONObject(i))
-            // TODO: Check for blacklisted sites and don't add them
-            lhm.put(tile.url.toUri()!!, tile)
+            val blacklist = loadBlacklist(context)
```
liuche: Load this blacklist outside of the loop - this will load the blacklist for every item in the bundled tiles cache!

app/src/main/java/org/mozilla/focus/tiles/HomeTilesManager.kt
```diff
@@ -78,7 +81,12 @@ class BundledTilesManager private constructor(context: Context) {
     fun unpinSite(context: Context, uri: Uri): Boolean {
         for (pair in bundledTilesCache) {
             if (compareUri(uri, pair.key)) {
-                // TODO: Remove by pair.value.id (BEFORE removing from cache)
+                val blacklist = loadBlacklist(context)
```
liuche: Same, don't load the blacklist every time.

liuche:
@bkmunar I'm also getting a compile error when trying to build this.

app/src/main/java/org/mozilla/focus/fragment/HomeFragment.kt
```diff
@@ -84,6 +87,36 @@ class HomeFragment : Fragment() {
...
+    }
+
+    override fun onContextItemSelected(item: MenuItem): Boolean {
+        when (item.itemId) {
+            R.id.remove -> {
+                val homeTileAdapter = tileContainer.adapter as HomeTileAdapter
+                val tileToRemove = homeTileAdapter.getItemAtPosition(getFocusedTilePosition())
+                // This assumes that since we're deleting from a Home Tile object that we created
+                // that the Uri is valid, so we do not do error handling here.
+                if (tileToRemove is BundledHomeTile) {
```
liuche: I'm seeing a compile error here for some reason, it looks like maybe because this is in an `else` the tileToRemove.url isn't guaranteed to not be null?

You can make this a `when` case, since HomeTile is a sealed class (there are a fixed number of cases), so like:

```
when (tileToRemove) {
    is BundledHomeTile -> {...}
    is CustomHomeTile -> {...}
}

File: app/src/main/java/org/mozilla/focus/fragment/HomeFragment.kt [hunk 45]
```diff
@@ -84,6 +87,39 @@ class HomeFragment : Fragment() {
...
+    }
+
+    override fun onContextItemSelected(item: MenuItem): Boolean {
+        when (item.itemId) {
+            R.id.remove -> {
+                val homeTileAdapter = tileContainer.adapter as HomeTileAdapter
+                val tileToRemove = homeTileAdapter.getItemAtPosition(getFocusedTilePosition())
+                // This assumes that since we're deleting from a Home Tile object that we created
+                // that the Uri is valid, so we do not do error handling here.
+                when (tileToRemove) {
```
liuche: Actually I wonder what happens if this is null, as returned by the the getItemAtPosition....but I think it should be fine.

PR #553, opened 2018-03-01
No issue: move to sealed classes on HomeTiles

mcomella:


app/src/main/java/org/mozilla/focus/fragment/HomeFragment.kt
```diff
@@ -104,18 +106,23 @@ class HomeFragment : Fragment() {
...
-    val tiles = homeTiles
 
     override fun onBindViewHolder(holder: TileViewHolder, position: Int) = with (holder) {
         val item = tiles[position]
-        titleView.setText(item.title)
-        if (!item.imagePath.isNullOrBlank()) {
-            val bmImg = itemView.context.assets.open(DEFAULT_HOME_TILES_DIR + item.imagePath).use { BitmapFactory.decodeStream(it) }
-            iconView.setImageBitmap(bmImg)
+        when (item) {
+            is DefaultHomeTile -> {
```
liuche: Why do you need the `is` again?

app/src/main/java/org/mozilla/focus/tiles/HomeTile.kt
```diff
@@ -0,0 +1,68 @@
...
+    protected fun appendToJSONObject(obj: JSONObject) {
+        obj.put(KEY_URL, url)
+        obj.put(KEY_TITLE, title)
+    }
+}
+
+class DefaultHomeTile(
+        url: String,
+        title: String,
+        val imagePath: String?,
```
liuche: nit: this shouldn't be optional anymore for default home tiles - you can also remove the `isNullOrBlank` from HomeFragment too.

app/src/main/java/org/mozilla/focus/tiles/HomeTile.kt
```diff
@@ -0,0 +1,68 @@
...
+import java.util.UUID
+
+private const val KEY_URL = "url"
+private const val KEY_TITLE = "title"
+
+private const val KEY_IMG = "img"
+private const val KEY_ID = "id"
+
+sealed class HomeTile(val url: String, val title: String) {
+    protected fun appendToJSONObject(obj: JSONObject) {
```
liuche: I don't know if I like this `appendTo` - maybe this can create the base JSONObject with these shared key/value pairs? like super.createBaseJSON()

app/src/main/java/org/mozilla/focus/tiles/HomeTile.kt
```diff
@@ -0,0 +1,68 @@
...
+                    jsonObject.optString(KEY_IMG),
+                    jsonObject.getString(KEY_ID))
+        }
+    }
+}
+
+class CustomHomeTile(
+        url: String,
+        title: String,
+        val id: UUID
```
liuche: nit: Comment that UUID is used for screenshots, unless you add that later in your commit.

app/src/main/java/org/mozilla/focus/tiles/HomeTilesManager.kt
```diff
@@ -85,7 +56,7 @@ class CustomTilesManager private constructor(context: Context) {
 
     @UiThread
     fun pinSite(context: Context, url: String) {
-        customTilesCache.put(url, HomeTile(url, url, "", url))
+        customTilesCache.put(url, CustomHomeTile(url, "custom", UUID.randomUUID()))
```
liuche: Nit: comment TODO

app/src/main/java/org/mozilla/focus/ext/When.kt
```diff
@@ -0,0 +1,16 @@
...
+/**
+ * `when` is only exhaustive when its expression returns a value. Appending this
+ * value gives it a return value, forcing the compiler to check `when` for
+ * exhaustiveness.
+ *
+ * One downside is that it pollutes autocomplete, but let's try it out.
+ *
+ * Idea from SO: https://stackoverflow.com/a/44383076
+ */
+val Any?.forceExhaustive get() = Unit
```
liuche: Okay, so for void return types for when, it doesn't need to be exhaustive? That seems okay. How do you enforce this?

app/src/main/java/org/mozilla/focus/fragment/HomeFragment.kt
mcomella: `if (item is DefaultHomeTile)` – it's the operator to check the type of the variable. In Java, `is` would be `instanceof`

app/src/main/java/org/mozilla/focus/tiles/HomeTile.kt
mcomella: Got it: I'll do `super.toJSONObject` and enter my items into that.

app/src/main/java/org/mozilla/focus/ext/When.kt
mcomella: Yes, for `Unit` return types (void doesn't exist in Kotlin).

You can't enforce this, but if someone remembers to use it, the compiler won't let you compile without it. The alternatives weren't as easy to understand so I went with this one but I'm seeing more problems:
- It is literally first in autocomplete always
- Compiler warns about the return types of my functions (in my next PR)

I thought of one more alternative not addressed on SO: have a global `whenExhaustive` function. We can rethink this if it becomes a problem.

PR #561, opened 2018-03-02
Closes #442: add screenshots to custom tiles

### LINKED ISSUE #442: Save/load high-quality favicons for custom home tiles
This needs more spec (letter? screenshot?) and fallback (what if none of these options look good)?

mcomella:
This patch series is still missing:
- Placeholders when screenshots could not be fetched
- Polished animation for async screenshot fetching
- UI polish

I was thinking we could do those in follow-ups (if Tif doesn't get back to me in time).


mcomella:
> Anything else specific I should test for?

TBH, not sure – I think I was perhaps overly thorough here.

Issue #528, opened 2018-02-23
Add accessibility to CheckableImageButton


Sdaswani:
This already has one commit, which will make this feature barely accessible, but still needs:
- [x] Announce checked state on change
- [x] Announce checked state on focus (VoiceView handles order differently from TalkBack/developer doc recommendations)

This will involve either experimenting with the docs more for Amazon's VoiceView (see [wiki page](https://github.com/mozilla-mobile/firefox-tv/wiki/Accessibility-Caveats) for very basic docs) or figuring out some other way how Amazon is calling through the accessibility functions.

lime124:
this is for the accessibility comment from eitan? i don't think he meant a literal checkbox. we discussed over slack and he didn't mention needing to change the UI. @Sdaswani

Sdaswani:
cool @lime124 @eeejay so is there any work to do here?

mcomella:
The CheckBox UI widget gives us these a11y wins automatically but afaik, there is some restriction that prevented us from using it – @liuche? 

We could potentially try to extract the a11y code from the CheckBox widget, so getting it's a11y functionality without using the widget directly

liuche:
Shouldn't be hard, I thought we got those for free with my implementation but I think we just need to override some of these functions: https://developer.android.com/guide/topics/ui/accessibility/custom-views.html#populate-events

Bonus is that we could add our own custom accessibility text for turbo mode/pinning if we'd like.

eeejay:
Yeah, didn't look at the code, but if you have your own subclassed button
its easy to override the right methods. I would override
onInitializeAccessibilityNodeInfo() as well.

mcomella:
Best to do 2.0 but if it drops to 2.1, that might be okay too.

liuche:
Pushed a PR to this, which will make this feature barely accessible, but still needs:
- [ ] Announce checked state on focus

This will involve either experimenting with the docs more for Amazon's VoiceView (in [wiki page](https://github.com/mozilla-mobile/firefox-tv/wiki/Accessibility-Caveats) or decompiling the VoiceView code to see what order it calls things in.

liuche:
Unassigning now that first part is landed.

PR #570, opened 2018-03-03
Closes #551 - Add pin icon accessibility string

### LINKED ISSUE #551: Add accessibility string to "pin" behavior
Quoting from #441
liuche:
> Okay, so the string can be "Pin"? Just wanted to explicitly ask what string I should add as the label for the thumbtack icon. (or "Pin to top sites" or "Pin to home screen" or whatever was suggested earlier in this thread.)

BrianNJones:
> @liuche @lime124 I think we need a slightly more expansive accessibility string than "Pin." Recommend "Pin to homescreen"...is there concern that there may be confusion re: whether reference is to Firefox TV vs. device homescreen?

liuche:


liuche:
Ya a11y bug is still open, this is just so there is actually a string - #528

Issue #438, opened 2018-01-31
Add load page UI test


mcomella:
We should write the test to work on our code before the refactor (e.g. use v1.1 tag) and ensure it still works after the refactor. We should probably click a home tile because it'll be the least likely thing to break when loading a page.

PR #574, opened 2018-03-05
Adds espresso-contrib library and basic UI page load test

AaronMT:
- Adds espresso-contrib library to Gradle in order to extend Espresso to work with RecyclerViewAction interactions for the home-screen tiles

- Adds a basic page-load UI test


npark-mozilla:
This addresses #438

Issue #523, opened 2018-02-23
String translations/l10n import + export for v2.0


liuche:
Here are the outstanding strings that we need for translation:
- [x] Pinned to Firefox TV homescreen
- [x] Removed from Firefox TV homescreen
- [x] Pin (accessibility text for button)
- [x] Close (Close Hint in the overlay)
- [x] Remove (delete tile from home screen, but this can be reused if necessary)
- [ ] ? "is checked" for #528 (but we might get that for free)
- [x] String export
- [x] String import

Strings should be exported and passed off to @Delphine by 3/6, and deadline for re-import is 3/12.

Delphine:
Thanks!
We've got our l10n repos up and Pontoon hooked up for translation. Jeff did an export last week, so we can follow the same process from now on for strings :) thanks!

liuche:
@lime124 and @BrianNJones are there any other new strings that we need for 2.0? I think "delete" we can reuse.

lime124:
We don't have that many visible strings in general. For accessibility, I'm not sure where those strings are saved. Do you @liuche ? Ideally, I would like to just be able to look at the strings.xml file to give a better answer but right now there's too much extra stuff in there.

In the nav overlay there's:
- Search or enter address [in the text box]
- Close [isn't in the UI yet - @liuche you filed a ticket for that]
- the two strings above for the toast [pinned/removed from homescreen]

In the browser there aren't any strings that come to mind outside of the error strings we reused from Focus.

On the homsecreen:
- Search or enter address [which is the one as from the nav overlay]
- the tiles use URLs/page titles - things that don't need translation
- the context menu, we'll only have "Remove"

Settings, everything is the same. We _do_ need to update some of them but haven't done it. And we aren't doing parental controls.

And finally Turbo Mode onboarding, we didn't touch that screen at all either.

Surely I've forgotten something, that seems like very few strings...?

liuche:
Thanks for the follow-up, @lime124 ! Self-assigning to go through the strings Tif highlighted and get the strings in for exporting (and also go through any more of the mocks). @Delphine we'll plan to have an export to you on Thursday.

liuche:
@bbinto @Sdaswani what's the date for the last string import FROM l10n? 3/12 will give us 2 weekends for localization (but that's only one day before the release which is 3/13).

(3/9 was our original code freeze so I told Delphine that would be the earliest deadline for "last import".)

Sdaswani:
[...]

PR #579, opened 2018-03-05
Issue #523 - Remove l10n unnecessary comment.

liuche:
Removing this robotranslation stringfrom our string file entirely.


liuche:
Actually this is easier on top of the other l10n import commit - going to move it there.

PR #582, opened 2018-03-06
I511 Add "subdomain.domain" titles to custom tiles

mcomella:


app/src/main/java/org/mozilla/focus/utils/publicsuffix/PublicSuffixPatterns.java
```diff
@@ -41,7 +37,9 @@
         } catch (IOException e) {
             throw new IllegalStateException("resource publicsuffixlist could not be opened but is bundled with app", e);
         } finally {
-            IOUtils.safeStreamClose(reader);
+            try {
+                if (reader != null) { reader.close(); }
```
liuche: ugh java

I was going to say I remember this way to read streams and close them after using...

app/src/main/java/org/mozilla/focus/ext/String.kt
```diff
@@ -81,3 +83,13 @@ fun String?.toUri(): Uri? = if (this == null) {
 } else {
     Uri.parse(this)
 }
+
+fun String?.toJavaURI(): URI? = if (this == null) {
+    null
+} else {
+    try {
+        URI(this)
```
liuche: Should we use URI for the other stuff? I picked it arbitrarily, so there's no harm in changing from Uri to URI. 

lolwow

app/src/main/java/org/mozilla/focus/fragment/HomeFragment.kt
```diff
@@ -172,19 +175,35 @@ private class HomeTileAdapter(
...
-    launch(uiLifecycleCancelJob + CommonPool) {
-        val screenshot = HomeTileScreenshotStore.read(itemView.context, item.id) // TODO: if null, provide placeholder.
-        launch(UI) {
-            // Animate to avoid pop-in due to thread hand-offs. TODO: animation is janky.
-            ObjectAnimator.ofInt(iconView, "imageAlpha", 0, 255).apply {
-                interpolator = CUSTOM_TILE_ICON_INTERPOLATOR
-                duration = CUSTOM_TILE_ICON_TO_SHOW_MILLIS
-            }.start()
-            iconView.setImageBitmap(screenshot)
+    launch(uiLifecycleCancelJob + UI, CoroutineStart.UNDISPATCHED) {
```
liuche: Is `launch` okay here, rather than `runBlocking`? I'm not too familiar with this but I guess the UI thread (where `onBindCustomHomeTile` lives isn't going to die) Just reading through the docs.

liuche: Also why do you use a separate pool of threads?

app/src/main/java/org/mozilla/focus/fragment/HomeFragment.kt
```diff
@@ -172,19 +175,35 @@ private class HomeTileAdapter(
...
+        // Animate to avoid pop-in due to thread hand-offs. TODO: animation is janky.
+        AnimatorSet().apply {
+            interpolator = CUSTOM_TILE_ICON_INTERPOLATOR
+            duration = CUSTOM_TILE_TO_SHOW_MILLIS
+
+            val iconAnim = ObjectAnimator.ofInt(iconView, "imageAlpha", 0, 255)
+            val titleAnim = ObjectAnimator.ofFloat(titleView, "alpha", 0f, 1f)
+
+            playTogether(iconAnim, titleAnim)
+        }.start()
```
liuche: Do you need a timeout for this, or is it okay because either these files exist or they don't?

app/src/main/java/org/mozilla/focus/fragment/HomeFragment.kt
```diff
@@ -172,19 +175,35 @@ private class HomeTileAdapter(
...
+
+        // Animate to avoid pop-in due to thread hand-offs. TODO: animation is janky.
+        AnimatorSet().apply {
+            interpolator = CUSTOM_TILE_ICON_INTERPOLATOR
+            duration = CUSTOM_TILE_TO_SHOW_MILLIS
+
+            val iconAnim = ObjectAnimator.ofInt(iconView, "imageAlpha", 0, 255)
+            val titleAnim = ObjectAnimator.ofFloat(titleView, "alpha", 0f, 1f)
+
+            playTogether(iconAnim, titleAnim)
```
liuche: Since you're only awaiting on the two results, does it make sense to use an async rather than a launch, because you don't need the Job that's returned?

liuche:
Failing test on TC, but maybe that's okay.

mcomella:
Logs: https://taskcluster-artifacts.net/a7cbJJR9Rp24P_CnBPR_Gw/0/public/tests/testAmazonWebviewBetaUnitTest/classes/org.mozilla.focus.utils.publicsuffix.TestPublicSuffix.html

I haven't seen these failures before:
```
org.junit.ComparisonFailure: for input:a网络A.网络.cn|| expected:<[网络.]cn> but was:<[]cn>
	at org.junit.Assert.assertEquals(Assert.java:115)
	at org.mozilla.focus.utils.publicsuffix.TestPublicSuffix.testGetPublicSuffixZeroAdditionalParts(TestPublicSuffix.java:109)
```

```
org.junit.ComparisonFailure: expected:<a网络A[]> but was:<a网络A[.网络]>
	at org.junit.Assert.assertEquals(Assert.java:115)
	at org.junit.Assert.assertEquals(Assert.java:144)
	at org.mozilla.focus.utils.publicsuffix.TestPublicSuffix.testStripPublicSuffix(TestPublicSuffix.java:52)
```

So I don't think it's related to the assets problem.

app/src/main/java/org/mozilla/focus/utils/publicsuffix/PublicSuffixPatterns.java
mcomella: We could add a safe stream close ext fn but I figured when in kotlin, we'd just close streams and let the app crash if they can't be closed because something really funky is going on.

app/src/main/java/org/mozilla/focus/ext/String.kt
mcomella: I don't actually know the differences – I think we should decide on a standard though, but I just wanted to use this code as is because I chose `URI` for a reason when I originally wrote it (maybe that's the way it was before though x_x).

app/src/main/java/org/mozilla/focus/fragment/HomeFragment.kt
mcomella: `launch` and `async` are pretty similar: they start a coroutine context and do the work on the specified thread. Where I think they differ is their return value and thus their intent: if you want a return value, use async: e.g. `val screenshot = async.await()`. If you don't and you just want the Job reference to manage the thread, e.g. `job.join()` and `job.cancel()`, use launch.

To answer your question, `launch` doesn't return a result so it'd be tedious to try to get a return value from it.

app/src/main/java/org/mozilla/focus/fragment/HomeFragment.kt
mcomella: The two coroutines builders that are common are `launch` and `async`. At the top-level, they basically do the same thing, except for their return types (Job and Deferred respectively, the latter which can return a value).

`runBlocking` is idiomatically only used for tests – instead of handing the thread to someone else when the function is suspended, it blocks the thread, which ensures the outer function doesn't return (e.g. in the case of tests where the test completes when the function returns). 

The value in coroutines is of not blocking the thread you're running on but still being able to write intuitive concurrent code (i.e. it's written as if it's sequential) so it generally doesn't make sense to use `runBlocking` in production code (and the IDE will warn you too!).

```kotlin
launch(UI) {
    Log.d("lol", "a")
    delay(2000)
    Log.d("lol", "b")
}
```
Here the coroutine will suspend at the delay, allowing the UI thread to handle other tasks (like responding to touch events).

```kotlin
runBlocking(Ui) {
    Log.d("...
    delay(2000)
    Log.d("...
}
```

Here the coroutine will suspend at delay, *blocking* the UI thread from handling other tasks and probably locking up your UI thread. It's equivalent to not launching a coroutine and just calling `Thread.sleep(2000)`.

mcomella: wrt threading, when using coroutines, you can specify: 1) UI thread, 2) a specific executor, or 3) the common thread pool shared by the app.

If you're not running on the UI thread, it's generally best to put it on a common thread pool so that 1) we're not wasting resources by spawning and destroying new threads or 2) hoarding memory by keeping an executor just for this task.

In this case, we spawn two jobs without side effects (kinda – reading from disk might technically be a side effect but we lock and are not mutating state) so we can put them on two separate threads to take advantage of the concurrency, wait for them to complete, and perform the updated operations on the UI thread.

---

One problem comes from this though: putting tasks on a thread pool (as opposed to a single thread) creates concurrency – I'll spare you the details (and my wrists) on how to fix that until you need them. :)

app/src/main/java/org/mozilla/focus/fragment/HomeFragment.kt
mcomella: For screenshots, we return if the file doesn't exist so no timeout necessary.

For public suffix, we crash if we're missing the file so no timeout necessary.

In theory, if this did get blocked, it wouldn't take up CPU (since it's suspended) and it'd get cancelled when the parent lifecycle job is cancelled anyway so it'll clean itself up.

Good question though. :)

mcomella:
Same failure on re-run. I think Beta is the first job running (alphabetical!) so for some reason this failures on TC but not locally.

mcomella:
So I installed Docker and can reproduce in the container. I noticed that if I run the container with LC_ALL set to C.UTF-8, I can no longer reproduce. iiuc, it looks like the JVM is being forced to used non-UTF-8 encoding and that causes the tests to fail.

Here's the command where the tests pass:
```sh
docker run -e LC_ALL=C.UTF-8 -it mozillamobile/firefox-tv /bin/bash
```

And the command that TC is using just in case the artifacts of the task die:
```sh
git fetch https://github.com/mcomella/firefox-tv.git i511-titles && git config advice.detachedHead false && git checkout 3c78731d5422ed0b33e32b8c4de4fc9083fd907e && echo \"--\" > .adjust_token && ./gradlew clean assemble lint checkstyle ktlint pmd test
```

It seems like this wouldn't repro on device, assuming the JVM's aren't forced into using non-UTF-8 so I'm going to file a follow-up, hide the tests, and land.

mcomella:
@liuche Oops, I accidentally landed despite making a TC config change. Want to look? https://github.com/mozilla-mobile/firefox-tv/pull/582/commits/0f4a1660d4657ae670745a296ecbf422dd16e94f

PR #584, opened 2018-03-06
Closes #573: Move custom tiles to the end of the homescreen.

### LINKED ISSUE #573: Pinned sites should be added to the end of the list
I think it should be put at the bottom, or at least Youtube shouldn't be displaced. Not sure what @lime124 had in mind though.

mcomella:


app/src/main/java/org/mozilla/focus/fragment/HomeFragment.kt
```diff
@@ -103,8 +103,8 @@ class HomeFragment : Fragment() {
 
     private fun initTiles() = with (tileContainer) {
         val homeTiles = mutableListOf<HomeTile>().apply {
-            addAll(CustomTilesManager.getInstance(context).getCustomHomeTilesList())
             addAll(BundledTilesManager.getInstance(context).getBundledHomeTilesList())
+            addAll(CustomTilesManager.getInstance(context).getCustomHomeTilesList())
```
liuche: The list gets reversed in `HomeTilesManager` so you should remove that line too, and return a copy.

PR #591, opened 2018-03-06
Fixes #549 overlay focus woes

### LINKED ISSUE #549: Navigating below bottom row of buttons leads to no buttons being selected
A button (or the url text) should always be selected.

boek:


app/src/main/java/org/mozilla/focus/widget/InlineAutocompleteEditText.java
```diff
@@ -646,19 +646,11 @@ private void setIsKeyFromRemoteAppOrSoftKeyboardClear(final boolean isKeyFromRem
...
-                || keyCode != KeyEvent.KEYCODE_DPAD_RIGHT
-                || keyCode != KeyEvent.KEYCODE_DPAD_UP
-                || keyCode != KeyEvent.KEYCODE_DPAD_UP_LEFT
-                || keyCode != KeyEvent.KEYCODE_DPAD_UP_RIGHT;
-
-        if (shouldDispatch) {
-            return super.dispatchKeyEvent(event);
-        }
+        final boolean shouldDispatch = !(keyCode == KeyEvent.KEYCODE_DPAD_DOWN
+                || keyCode == KeyEvent.KEYCODE_DPAD_LEFT
```
liuche: Are we not worried about these other down-right, up-left keys anymore?

boek: @liuche I tried every way possible and noticed the combos never got hit. It always sent two events. Maybe there is a remote that has diagonal buttons? I removed it to simplify the code a bit

mcomella: The Android code is pretty generic and assumes you can have any device (e.g. a video game controller). I think ppl could pair bluetooth controllers to their devices and this combination could get hit – I think I'd prefer to leave it in. If we want to simplify the code, we could stick this in a helper function or use a range (since the DPAD_* are almost certainly ints that are all next to one another).

app/src/main/java/org/mozilla/focus/widget/BrowserNavigationOverlay.kt
```diff
@@ -147,6 +147,18 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
 
         pinButton.isChecked = navigationStateProvider?.isURLPinned() ?: false
 
+        // Prevent the focus from looping to the bottom row when reaching the last
+        // focusable element in the top row
+        navButtonReload.nextFocusLeftId = when {
```
mcomella: fwiw, this might fix itself if we used a different layout pattern (instead of the nested LinearLayouts) but we agreed it wasn't worth fixin atm.

Issue #490, opened 2018-02-15
Telemetry for custom tiles


liuche:
- [x] differentiate between clicking a custom tile vs default tiles
- [x] delete custom/default tile
- [x] pin/unpin custom tile
- [x] count # of custom tiles, and total tiles (to create a histogram)

We also need to restore the telemetry removed with the Overlay addition:
- [x] Back/forward/reload
- [x] home
- [x] settings
- [x] Turbo
- [x] URL bar
- [x] Overlay opened/closed

mcomella:
Waiting on deletion of tiles: #443.

PR #602, opened 2018-03-07
I490 telemetry

mcomella:
Working on data review form...

### DIFF
--- a/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
+++ b/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
@@ -15,7 +15,6 @@ import android.view.ViewGroup
 import android.widget.FrameLayout
 import kotlinx.android.synthetic.main.fragment_browser.*
 import kotlinx.android.synthetic.main.fragment_browser.view.*
-import kotlinx.android.synthetic.main.home_tile.*
 import org.mozilla.focus.R
 import org.mozilla.focus.activity.MainActivity
 import org.mozilla.focus.architecture.NonNullObserver
@@ -133,7 +132,7 @@ class BrowserFragment : IWebViewLifecycleFragment(),
             NavigationEvent.SETTINGS -> ScreenController.showSettingsScreen(fragmentManager)
             NavigationEvent.LOAD -> {
                 (activity as MainActivity).onTextInputUrlEntered(value!!, autocompleteResult!!, UrlTextInputLocation.MENU)
-                showOverlay(false)
+                setOverlayVisibileByUser(false)
             }
             NavigationEvent.RELOAD_YT -> {
                 isReloadingForYoutubeDrawerClosed = true
@@ -191,7 +190,7 @@ class BrowserFragment : IWebViewLifecycleFragment(),
                 goBack()
                 TelemetryWrapper.browserBackControllerEvent()
             }
-            browserOverlay.isVisible -> showOverlay(false)
+            browserOverlay.isVisible -> setOverlayVisibileByUser(false)
             else -> {
                 fragmentManager.popBackStack()
                 SessionManager.getInstance().removeCurrentSession()
@@ -232,7 +231,7 @@ class BrowserFragment : IWebViewLifecycleFragment(),
     private fun handleSpecialKeyEvent(event: KeyEvent): Boolean {
         if (event.keyCode == KeyEvent.KEYCODE_MENU && event.action == KeyEvent.ACTION_UP) {
             val toShow = !browserOverlay.isVisible
-            showOverlay(toShow)
+            setOverlayVisibileByUser(toShow)
             // Fix this youtube focus hack in #393
             if (!toShow && webView!!.isYoutubeTV) {
                 webView?.requestFocus()
@@ -249,10 +248,11 @@ class BrowserFragment : IWebViewLifecycleFragment(),
         return false
     }
 
-    private fun showOverlay(toShow: Boolean) {
-        browserOverlay.setOverlayVisibleByUser(toShow)
+    private fun setOverlayVisibileByUser(toShow: Boolean) {
+        browserOverlay.isVisible = toShow
         if (toShow) cursor?.onPause() else cursor?.onResume()
         cursor?.setEnabledForCurrentState()
+        TelemetryWrapper.drawerShowHideEvent(toShow)
     }
 }
 
--- a/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
+++ b/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
@@ -248,6 +248,13 @@ class BrowserFragment : IWebViewLifecycleFragment(),
         return false
     }
 
+    /**
+     * Changes the overlay visibility: this should be called instead of changing
+     * [BrowserNavigationOverlay.isVisible] directly.
+     *
+     * It's important this is only called for user actions because our Telemetry
+     * is dependent on it.
+     */
     private fun setOverlayVisibileByUser(toShow: Boolean) {
         browserOverlay.isVisible = toShow
         if (toShow) cursor?.onPause() else cursor?.onResume()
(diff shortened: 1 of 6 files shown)


mcomella:

- How do users interact with the new navigation overlay? (open/close, button clicks, enable/disable turbo, pin/unpin sites, use url bar) This mostly existed in the previous browselane
- How often do users click on home tiles? Custom vs. bundled?
- How often do users delete tiles? Custom vs. bundled?
- How many tiles do users have on their home screens? How many are custom?

- Establish baseline interactions in overlay
- Identify how useful custom tile feature is to users; this allows users to better navigate using the d-pad remote and we need to know if there are alternatives/additions we should consider.

None: we need aggregate info and UR isn't sufficient.

No.

5) List all proposed measurements and indicate the category of data collection for each measurement, using the Firefox [data c](https://wiki.mozilla.org/Firefox/Data_Collection)[ollection ](https://wiki.mozilla.org/Firefox/Data_Collection)[categories](https://wiki.mozilla.org/Firefox/Data_Collection) on the found on the Mozilla wiki.   

<table>
  <tr>
    <td>Tracking Bug #</td>
  </tr>
  <tr>
    <td>Overlay click: back/forward/reload/home/settings</td>
    <td>Interaction</td>
    <td>#490</td>
  </tr>
  <tr>
    <td>Overlay switch: turbo/(un)pin</td>
    <td>Interaction</td>
    <td>#490</td>
  </tr>
  <tr>
    <td>Custom and total home tile count</td>
    <td>Interaction</td>
    <td>#490</td>
  </tr>
  <tr>
    <td>Home tile deleted (custom? bundled?)</td>
    <td>Interaction</td>
    <td>#490</td>
  </tr>
  <tr>
    <td>Home tile clicked: custom? bundled?</td>
    <td>Interaction</td>
    <td>#490</td>
  </tr>
</table>


I want to permanently monitor this data. (@bbinto)

* Which release channels?

All.

* Which countries?

All.

* Which locales?

All.

* Any other filters?  Please describe in detail below.

Nah.

In settings screen, users can opt-out.

Barbara will monitor after release to see 1) how users interact with overlay and 2) how useful the home tile feature is, in addition to review store feedback.

sql.tmo, mailing lists as we discuss new features, team meetings

app/src/main/java/org/mozilla/focus/widget/BrowserNavigationOverlay.kt
```diff
@@ -72,7 +72,6 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
         get() = Settings.getInstance(context).isBlockingEnabled
         set(value) {
             Settings.getInstance(context).isBlockingEnabled = value
```
liuche: Since you're removing the other line you can do a `set(value) = Settings...` :P

app/src/main/java/org/mozilla/focus/telemetry/TelemetryWrapper.kt
```diff
@@ -266,9 +268,30 @@ object TelemetryWrapper {
         TelemetryEvent.create(Category.ACTION, Method.CLICK, Object.MENU, button.value).queue()
     }
 
-    @JvmStatic
-    fun menuAppNavEvent(button: MenuAppNavButton) {
-        TelemetryEvent.create(Category.ACTION, Method.CLICK, Object.MENU, button.value).queue()
+    fun overlayClickEvent(event: NavigationEvent, isTurboButtonChecked: Boolean, isPinButtonChecked: Boolean) {
```
liuche: Hmm, I wonder if we can bundle `is*Checked` into the `NavigationEvent`? Maybe more sealed classes, but maybe overkill.

app/src/main/java/org/mozilla/focus/widget/BrowserNavigationOverlay.kt
```diff
@@ -117,16 +116,20 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
     override fun onClick(view: View?) {
         val event = NavigationEvent.fromViewClick(view?.id) ?: return
         var value: String? = null
+
+        val isTurboButtonChecked = turboButton.isChecked
```
liuche: Do we need these values? Can you just use `isTurboEnabled` or `isPinButtonChecked` after it's set, without creating this additional value?

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
```diff
@@ -232,7 +231,7 @@ class BrowserFragment : IWebViewLifecycleFragment(),
     private fun handleSpecialKeyEvent(event: KeyEvent): Boolean {
         if (event.keyCode == KeyEvent.KEYCODE_MENU && event.action == KeyEvent.ACTION_UP) {
             val toShow = !browserOverlay.isVisible
-            showOverlay(toShow)
```
liuche: I didn't think there was a difference between these two so it would be fine to swap them, but what about https://github.com/mozilla-mobile/firefox-tv/commit/8747c4f4236c995e66348745d71e05acb9ce1691 ? Should that have fixed that bug, if they're the same, minus telemetry?

app/src/main/java/org/mozilla/focus/telemetry/TelemetrySettingsProvider.kt
```diff
@@ -0,0 +1,37 @@
...
+    override fun containsKey(key: String?): Boolean {
+        return key == PREF_CUSTOM_HOME_TILE_COUNT ||
+                key == PREF_TOTAL_HOME_TILE_COUNT ||
+                super.containsKey(key)
+    }
+
+    override fun getValue(key: String?): Any? = when (key) {
+        PREF_CUSTOM_HOME_TILE_COUNT -> CustomTilesManager.getInstance(appContext).tileCount
+        PREF_TOTAL_HOME_TILE_COUNT -> CustomTilesManager.getInstance(appContext).tileCount +
+                BundledTilesManager.getInstance(appContext).tileCount
```
liuche: I think here you might need to remove the home tiles blacklist.

app/src/main/java/org/mozilla/focus/widget/BrowserNavigationOverlay.kt
mcomella: I actually can't because it's not returning a value. Nice try though. ;)

app/src/main/java/org/mozilla/focus/widget/BrowserNavigationOverlay.kt
mcomella: You can do that but it's more error-prone – there are two steps to autocomplete that you can get wrong each time it's typed. Having it as one local variable means it's less likely to be mistyped. Also, (this doesn't matter since we're on the UI thread but if the code is refactored...) if the isChecked value is updated async, this ensures we're using the same value for both fields.

I'm going to leave it as is, but yeah, I'm conflicted about the best approach here.

app/src/main/java/org/mozilla/focus/telemetry/TelemetryWrapper.kt
mcomella: You can't assign values to enums (they're singletons) so my desired case is: `onNavigationEvent(event, /* any params necessary */)`. This doesn't work though: NavigationEvent is overloaded so that it gets called for any Overlay -> BrowserFragment callsbacks, so e.g. a `LOAD` or `RELOAD_YT` event needs to call onNavigationEvent but doesn't have the parameters to do so.

In theory, we could do `onNavigationEvent(event, someSealedClassArgs)` but that seems like an unnecessary abstraction.

Fun fact: Swift solves this sealed class/enum problem by letting you assign values to enums. What do you know? :)

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
mcomella: That bug is irrelevant – the problem in that bug is that you can call 1) BrowserFragment's hide or 2) BrowserOverlay's hide. Ideally, the code could be architected differently so you can't do that.

What I did in this bug wasn't swap these implementations, but rename BrowserFragment's to be the same as BrowserOverlay's so it's more apparent that maybe you'd want to use it instead of touching the overlay directly and to make it clear that this should only be called from a user action (i.e. so telemetry isn't messed up). I'll add a comment.

app/src/main/java/org/mozilla/focus/telemetry/TelemetrySettingsProvider.kt
mcomella: The bundled tiles cache removes any tiles on the blacklist so we can use this value directly.

PR #605, opened 2018-03-07
Closes #601: home tiles overlapping url bar

### LINKED ISSUE #601: Home Tiles overlapping urlbar
screenshot_1520385592

Should be a quick fix - `android:clipChildren` shouldn't be false for the home tile container, add padding to the top of the RecyclerView.

If that doesn't work, can also try using an `inset` to create padding, or a `layer-list`.

bkmunar:


app/src/main/res/layout/fragment_home.xml
```diff
@@ -69,6 +68,8 @@
 
     <android.support.v7.widget.RecyclerView
             android:id="@+id/tileContainer"
+            android:clipToPadding="false"
+            android:paddingTop="10dp"
```
boek: If we're adding 10dp of padding here, should we -10dp from the urlbar so the spacing stays the same?

bkmunar: o0o0o0oo0o0o0o0o0o0o0o0o0o0o0

boek:
LGTM

app/src/main/res/layout/fragment_home.xml
mcomella: fwiw, negative values are bad practice in Android (I think it's just because it's confusing) and the IDE should warn you as such.

bkmunar: So if we didn't have the extra padding already, how would you do a negative padding? Is there another attribute for it?

boek: @mcomella ah, I meant just take 10dp off of the other padding. Not use a negative value! My bad! I should have been more descriptive

PR #619, opened 2018-03-07
Closes #598: enlarge screenshots; round corners

### LINKED ISSUE #598: Enlarge the custom tile screenshots and round the corners
Break-out: #568, maybe like https://user-images.githubusercontent.com/759372/36880009-5f53a12e-1d7b-11e8-84df-03ba4cc8de24.png

bkmunar:
…t found, like focus


mcomella:
@bkmunar This is #598 - please update the title and commit message.

Also, please rebase. I moved some files around – git is usually smart about this – but if you run into trouble, let me know and we can back it out.

bkmunar:
Converted file to Kotlin!

However, it seems to be crashing whenever I try to assign `RectF(drawable.bounds)` outside of the onDraw function, even in a `println` statement in the init.

The only success I've had is when I tried to override the `setImageDrawable` function and try to assign the drawable.bounds there. But it just shows the image really quick and then fades out to blank, probably meaning that drawable isn't null, but it doesn't have the bounds we want at that point in time. 

I've also tried overriding `onAttachedToWindow` and `setFrame` but to no avail. Would love ideas on how to fix!! Or we can land and then file follow up fixes, I'm not sure what is important since this is a UI polish.

mcomella:
**Important**: I want you to allocate the RectF object in advance, but you can still assign the bounds in `onDraw`! You shouldn't need to find where the bounds are set in advance!

---

Just to address your debugging:

> it seems to be crashing whenever I try to assign RectF(drawable.bounds) outside of the onDraw function, even in a println statement in the init.

I would expect this: Android inflates the XML using the `(Context, Attribute...)` constructor so the drawable will be null at this point. If you try to access the bounds, you'll crash!

> The only success I've had is when I tried to override the setImageDrawable function and try to assign the drawable.bounds there.

Exactly – this is where the drawable gets set and is first non-null.

> But it just shows the image really quick and then fades out to blank, probably meaning that drawable isn't null, but it doesn't have the bounds we want at that point in time.

Seems reasonable, good debugging!

app/src/main/java/org/mozilla/focus/home/HomeFragment.kt
```diff
@@ -41,6 +41,8 @@ import org.mozilla.focus.utils.OnUrlEnteredListener
 private const val COL_COUNT = 5
 private const val SETTINGS_ICON_IDLE_ALPHA = 0.4f
 private const val SETTINGS_ICON_ACTIVE_ALPHA = 1.0f
+private const val BUNDLED_TILE_VIEW_TYPE = 0
```
mcomella: nit: Add a line above to separate these constants from the others.

You should group related ideas together so you can quickly scan between groups, rather than having to look at all the individual items in these groups.

app/src/main/java/org/mozilla/focus/home/HomeFragment.kt
```diff
@@ -212,6 +214,11 @@ private class HomeTileAdapter(
         }
     }
 
+    override fun getItemViewType(position: Int) = when (getItemAtPosition(position)) {
+        is BundledHomeTile -> BUNDLED_TILE_VIEW_TYPE
+        else -> CUSTOM_TILE_VIEW_TYPE
```
mcomella: You should explicitly use `is CustomHomeTile` and avoid using `else`. Else removes the purpose of making a list exhaustive because when you add new types, they'll all fall under the else case rather than forcing a compile error. Else also prevents you from considering edge cases you may not have thought about (e.g. if you're checking for 0 or 1, you could do `if (val > 0) thing(); else otherThing();` but you might forget the negative value case.

app/src/main/java/org/mozilla/focus/home/HomeFragment.kt
```diff
@@ -228,9 +235,13 @@ private class HomeTileAdapter(
 
     override fun getItemCount() = tiles.size
 
-    override fun onCreateViewHolder(parent: ViewGroup, viewType: Int) = TileViewHolder(
-            LayoutInflater.from(parent.context).inflate(R.layout.home_tile, parent, false)
-    )
+    override fun onCreateViewHolder(parent: ViewGroup, viewType: Int): TileViewHolder {
+        var layout = R.layout.bundled_home_tile
+        if (viewType == CUSTOM_TILE_VIEW_TYPE) {
```
mcomella: nit: similar idea to the above: you want to explicitly specify all the states you expect this to be in (and crash if you're not in an expected state!).

Also, you generally want to avoid mutability where possible because someone later can change a value you depend on.

In this case, you can:
```kotlin
val layout = when (viewType) {
```

app/src/main/java/org/mozilla/focus/widget/RoundedImageView.kt
```diff
@@ -0,0 +1,58 @@
...
+import android.graphics.Canvas
+import android.graphics.PorterDuff
+import android.graphics.PorterDuffXfermode
+import android.graphics.RectF
+import android.graphics.drawable.BitmapDrawable
+import android.support.v7.widget.AppCompatImageView
+import android.util.AttributeSet
+
+/*
+ * Code taken from https://stackoverflow.com/questions/20743859/imageview-rounded-corners/23713740#23713740
```
mcomella: This isn't the javadoc pattern: please look in other files to match them. Having me correct it in review is time-consuming for both of us.

app/src/main/java/org/mozilla/focus/widget/RoundedImageView.kt
```diff
@@ -0,0 +1,58 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
```
mcomella: This isn't the license pattern: please look in other files to match them.

app/src/main/java/org/mozilla/focus/widget/RoundedImageView.kt
```diff
@@ -0,0 +1,58 @@
...
+import android.support.v7.widget.AppCompatImageView
+import android.util.AttributeSet
+
+/*
+ * Code taken from https://stackoverflow.com/questions/20743859/imageview-rounded-corners/23713740#23713740
+ *
+ * */
+class RoundedImageView(context: Context, attributes: AttributeSet) : AppCompatImageView(context, attributes) {
+    private val mCornerRadius = 6.0f
+    private var rectF: RectF? = null
```
mcomella: nit: You can simplify the code by constructing it here.

app/src/main/java/org/mozilla/focus/widget/RoundedImageView.kt
```diff
@@ -0,0 +1,58 @@
...
+import android.graphics.drawable.BitmapDrawable
+import android.support.v7.widget.AppCompatImageView
+import android.util.AttributeSet
+
+/*
+ * Code taken from https://stackoverflow.com/questions/20743859/imageview-rounded-corners/23713740#23713740
+ *
+ * */
+class RoundedImageView(context: Context, attributes: AttributeSet) : AppCompatImageView(context, attributes) {
+    private val mCornerRadius = 6.0f
```
mcomella: nit: `private const val` and should be defined above this class

app/src/main/res/layout/custom_home_tile.xml
```diff
@@ -0,0 +1,59 @@
...
+        android:background="@drawable/context_menu_button_background"
+        android:elevation="4dp"
+        android:padding="4dp"
+        android:src="@drawable/ic_wide_menu_home"
+        android:visibility="gone" />
+
+    <LinearLayout
+        android:layout_width="match_parent"
+        android:layout_height="match_parent"
+        android:layout_marginTop="4dp"
```
mcomella: I didn't realize these files would be so similar so I mislead you again – sorry about that. :(

I think in general having so much code duplication is bad practice. I think there are two approaches:
- Can you we have one layout? i.e. if we used one layout file with no dynamic changes, would Tif mind if the bundled tiles grew in size too?
- If she'd mind, your initial approach of changing the margins (and cornerRadius!) programmatically would have been better. You can `getLayoutMargins()`, update the values, and `setLayoutMargins()` (you may need to cast to MarginLayoutParams).

Do you mind trying these approaches?

fwiw, I have a bad habit of giving really general advice instead of advice specific to the problem someone is trying to solve (which happened here - XML is generally preferred over code) so I'll try to be better about this in the future.

app/src/main/java/org/mozilla/focus/widget/RoundedImageView.kt
```diff
@@ -0,0 +1,58 @@
...
+    private var rectF: RectF? = null
+    private val porterDuffXferMode = PorterDuffXfermode(PorterDuff.Mode.SRC_IN)
+
+    override fun onDraw(canvas: Canvas) {
+        val myDrawable = drawable
+
+        if (myDrawable != null && myDrawable is BitmapDrawable && mCornerRadius > 0) {
+            val paint = myDrawable.paint
```
mcomella: As discussed on Slack, please figure out how this fn works so I can ask you questions about it! Thanks!

bkmunar:
Am doing the second approach you suggested... it takes out a WHOLE lotta code including the custom classes, etc. 

Still working on fixing the height and width programatically but it's almost there!! Commented on the code explaining what it's doing (or rather, what I think it's doing hehe). What a learning experience this is 😅

Sdaswani:
@bkmunar will this make 2.0? don't love landing a whole lot of new code past soft code freeze.

PR #628, opened 2018-03-08
Closes #587 - Clean up webview in onDestroy rather than onDestroyView

### LINKED ISSUE #587: App Crashing After Pinning Youtube Video
EDIT: As @bkmunar and I are investigating, we're back-tracking all the way to ~~pre-pin~~ cursor refactor code, and we're still see this crash. ~~It looks like it might be switching from the overlay to Home/Settings that is causing a crash.~~ Through bisecting, we believe it's related to when the webview is destroyed (but since the repro isn't consistent, this is a best guess).

STR:

1. Visit Youtube
2. Play a Video (note you may have to start playing it because of https://github.com/mozilla-mobile/firefox-tv/issues/585)
3. ~~Let it play for a bit~~ Fast-forward ~1/3, then open overlay
4. Pin the site
5. From overlay, go Home (or to Settings)

Observe.

Actual: App Crashes (_sometimes_, so try a couple of times)
Expected: App doesn't crash

liuche:


app/src/main/java/org/mozilla/focus/web/IWebViewLifecycleFragment.kt
```diff
@@ -77,13 +77,11 @@ abstract class IWebViewLifecycleFragment : LocaleAwareFragment() {
...
 
-    override fun onDestroyView() {
-        super.onDestroyView()
-        if (webView != null) {
-            webView!!.callback = null
-            webView!!.destroy()
-            webView = null
-        }
+    override fun onDestroy() {
+        super.onDestroy()
```
mcomella: Hm, I think we shouldn't be constructing new WebView's in onCreateView if we're destroying them only in onDestroy – it seems like we'll have a memory leak. I filed a follow-up: https://github.com/mozilla-mobile/firefox-tv/issues/652

PR #631, opened 2018-03-09
Issue #528 - Add accessibility for CheckableImageButton

liuche:
This is a partial fix - upon checking/unchecking an item, VoiceView will announce the checked state, but upon focusing, the checked state is not announced, and calling `onInitializeAccessibilityEvent` doesn't seem to be doing anything. Will keep testing other options and look at Amazon accessibility docs.

### DIFF
--- a/app/src/main/java/org/mozilla/focus/widget/CheckableImageButton.kt
+++ b/app/src/main/java/org/mozilla/focus/widget/CheckableImageButton.kt
@@ -8,6 +8,7 @@ package org.mozilla.focus.widget
 import android.content.Context
 import android.util.AttributeSet
 import android.view.View
+import android.view.accessibility.AccessibilityNodeInfo
 import android.widget.Checkable
 import android.widget.ImageButton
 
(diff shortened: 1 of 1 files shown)


liuche:
Okay, so even after patterning this off of [CompoundButton](http://androidxref.com/7.1.2_r36/xref/frameworks/base/core/java/android/widget/CompoundButton.java#352) from the Android source, this wasn't working. Checkbox state is still not announced on focus.

After looking at the [Amazon accessibility docs](https://developer.amazon.com/docs/fire-tv/implement-accessibility-fire-os.html#path-of-an-accessibility-event) and following them, it looks like 1) they are different from what the Android accessibility docs suggest, and 2) they don't really work as advertised.

This works to some degree, but isn't ideal - VoiceView will tell you the state of the Turbo Mode/Pin Site button (checked/unchecked) after you toggle it, but on selecting it, the button doesn't announce its state.

I've added a wiki page titled "Accessibility Caveats" so other people don't have to fall down this hole again, and can go directly to the source.

app/src/main/java/org/mozilla/focus/widget/CheckableImageButton.kt
```diff
@@ -56,4 +57,10 @@ class CheckableImageButton @JvmOverloads constructor(
         toggle()
         return super.performClick()
     }
+
+    override fun onInitializeAccessibilityNodeInfo(info: AccessibilityNodeInfo?) {
+        super.onInitializeAccessibilityNodeInfo(info)
+        info?.isCheckable = true
```
mcomella: nit: add a comment indicating what this is intended to do.

PR #646, opened 2018-03-09
Closes #598: enlarge screenshots; round corners

### LINKED ISSUE #598: Enlarge the custom tile screenshots and round the corners

bkmunar:


bkmunar:
screen shot 2018-03-09 at 12 03 37 pm

app/src/main/java/org/mozilla/focus/home/HomeFragment.kt
```diff
@@ -259,9 +266,27 @@ private fun onBindCustomHomeTile(uiLifecycleCancelJob: Job, holder: TileViewHold
         }
 
         // Wait for both to complete so we can animate them together.
-        iconView.setImageBitmap(screenshot.await()) // TODO: if null, provide placeholder.
+        val screenshotToRound = screenshot.await() // TODO: if null, provide placeholder.
+        val cornerRadius = 8.toFloat()
```
mcomella: Numbers defined in code are pixel values but we should be specifying them in DP so the changes look consistent across devices. The easiest way to do this is define the value in dimens.xml and access it with `context.getDimensionPixelSize`.

In this case, you can use the same corner radius we define for the home tiles themselves.

Please grab all the numbers from resources!

---

No longer necessary but you can create float literals with `8f`.

app/src/main/java/org/mozilla/focus/home/HomeFragment.kt
```diff
@@ -259,9 +266,27 @@ private fun onBindCustomHomeTile(uiLifecycleCancelJob: Job, holder: TileViewHold
...
+        // Need copy of screenshot bitmap because it is immutable
+        val copyOfScreenshot = Bitmap.createBitmap(screenshotToRound.width, screenshotToRound.height, screenshotToRound.config)
```
mcomella: Please make this a helper function – the code gets really hard to read when there are many low-level methods.

I think this should be an extension function: `Bitmap.withRoundedCorners(cornerRadius: Float)`

app/src/main/java/org/mozilla/focus/home/HomeFragment.kt
```diff
@@ -259,9 +266,27 @@ private fun onBindCustomHomeTile(uiLifecycleCancelJob: Job, holder: TileViewHold
...
+        val canvas = Canvas(copyOfScreenshot)
+        val paint = Paint()
+        // Need to set isAntiAlias to true because it smooths out the edges of what is being drawn
```
mcomella: nit: there should generally be a line of WS above `//` comments.

app/src/main/java/org/mozilla/focus/home/HomeFragment.kt
```diff
@@ -259,9 +266,27 @@ private fun onBindCustomHomeTile(uiLifecycleCancelJob: Job, holder: TileViewHold
...
+        paint.isAntiAlias = true
+        // Need the shader to replicate the edge color if the shader draws outside of its original bounds
+        paint.shader = BitmapShader(screenshotToRound, Shader.TileMode.CLAMP, Shader.TileMode.CLAMP)
```
mcomella: I don't understand - when will the shader draw outside the original bounds?

I think we just need the BitmapShader in order to use the bitmap in the paint, which is how it gets in the canvas.

app/src/main/java/org/mozilla/focus/home/HomeFragment.kt
```diff
@@ -259,9 +266,27 @@ private fun onBindCustomHomeTile(uiLifecycleCancelJob: Job, holder: TileViewHold
...
+        canvas.drawRoundRect(RectF(0.0f, 0.0f, screenshotToRound.width.toFloat(), screenshotToRound.height.toFloat()), cornerRadius, cornerRadius, paint)
+
+        iconView.setImageBitmap(copyOfScreenshot)
         titleView.text = title.await()
 
+        val layoutParams = LinearLayout.LayoutParams(LinearLayout.LayoutParams.MATCH_PARENT, 130)
```
mcomella: RecyclerView re-uses views so if you change these layout parameters, you also have to change them in `onBindBundled` (since the views could be re-used for bundle tiles.

You should remove these layout params from the XML, add a comment mentioning they're defined in code, and then define them in both `onBind*` methods (or perhaps define them outside, perhaps inside the when, so that we're less likely to forget to update both).

mcomella: nit: I think you should actually be able to get the params from the view, modify them, and then reset them on the view, rather than creating a new instance.

mcomella:
Another perf note: it's not great that we're copying bitmaps on the UI thread but they're small so it should be okay.

app/src/main/java/org/mozilla/focus/ext/Bitmap.kt
```diff
@@ -26,3 +31,18 @@ fun Bitmap.arePixelsAllTheSame(): Boolean {
 
     return true
 }
+
+fun Bitmap.withRoundedCorners(cornerRadius: Float): Bitmap? {
+    // Need copy of screenshot bitmap because it is immutable
```
mcomella: nit: rm comment. We probably don't want to overwrite the bitmap anyway.

app/src/main/java/org/mozilla/focus/ext/Bitmap.kt
```diff
@@ -26,3 +31,18 @@ fun Bitmap.arePixelsAllTheSame(): Boolean {
 
     return true
 }
+
+fun Bitmap.withRoundedCorners(cornerRadius: Float): Bitmap? {
```
mcomella: nit: add `@CHeckResult` annotation - it'll make it more clear this Bitmap is not being mutated.

app/src/main/java/org/mozilla/focus/ext/Bitmap.kt
```diff
@@ -26,3 +31,18 @@ fun Bitmap.arePixelsAllTheSame(): Boolean {
...
+    // Need copy of screenshot bitmap because it is immutable
+    val newBitmapWithScreenshotDims = Bitmap.createBitmap(width, height, config)
+    val canvas = Canvas(newBitmapWithScreenshotDims)
+    val paint = Paint()
+
+    // Need to set isAntiAlias to true because it smooths out the edges of what is being drawn
+    paint.isAntiAlias = true
+
+    // Need the shader to replicate the edge color if the shader draws outside of its original bounds
```
mcomella: nit: I don't think we use this functionality. I would remove this comment (unless you think it's helpful).

app/src/main/java/org/mozilla/focus/home/HomeFragment.kt
```diff
@@ -214,6 +222,17 @@ private class HomeTileAdapter(
         }
     }
 
+    fun setLayoutParams(iconView: View, tileMarginValue: Int, tileHeight: Int) {
+        val marginLayoutParams = iconView.layoutParams as ViewGroup.MarginLayoutParams
+        val marginValue = iconView.resources.getDimensionPixelSize(tileMarginValue)
+        marginLayoutParams.setMargins(marginValue, marginValue, marginValue, marginValue)
+        val layoutParams = iconView.layoutParams as LinearLayout.LayoutParams
```
mcomella: nit: you should be able to set the margins on this reference too (i.e. don't also get `marginLayoutParams`

app/src/main/java/org/mozilla/focus/home/HomeFragment.kt
```diff
@@ -260,15 +279,14 @@ private fun onBindCustomHomeTile(uiLifecycleCancelJob: Job, holder: TileViewHold
...
+        // Wait for both to complete so we can animate them together.
+        val screenshot = screenshotDeferred.await() // TODO: if null, provide placeholder.
 
         // NB: Don't suspend after this point (i.e. between view updates like setImage)
         // so we don't see intermediate view states.
         // TODO: It'd be less error-prone to launch { /* bg work */ launch(UI) { /* UI work */ } }
-        iconView.setImageBitmap(screenshot)
-        titleView.text = title
+        iconView.setImageBitmap(screenshot.withRoundedCorners(iconView.resources.getDimension(R.dimen.home_tile_corner_radius)))
+        titleView.text = titleDeferred.await()
```
mcomella: Don't move this: it suspends the UI thread and causes pop-in (see comment above). You should put it back where it was.

app/src/main/res/layout/home_tile.xml
```diff
@@ -29,17 +29,18 @@
         android:visibility="gone"
         android:duplicateParentState="true" />
 
+
```
mcomella: nit: excess ws

app/src/main/java/org/mozilla/focus/home/HomeFragment.kt
```diff
@@ -214,6 +222,17 @@ private class HomeTileAdapter(
         }
     }
 
+    fun setLayoutParams(iconView: View, tileMarginValue: Int, tileHeight: Int) {
+        val marginLayoutParams = iconView.layoutParams as ViewGroup.MarginLayoutParams
+        val marginValue = iconView.resources.getDimensionPixelSize(tileMarginValue)
+        marginLayoutParams.setMargins(marginValue, marginValue, marginValue, marginValue)
+        val layoutParams = iconView.layoutParams as LinearLayout.LayoutParams
+        layoutParams.weight = 1f
```
mcomella: This is a constant: we can define it in XML.

app/src/main/res/values/dimens.xml
```diff
@@ -51,6 +51,10 @@
     <dimen name="remote_cursor_size2x">90dp</dimen>
 
     <dimen name="button_size">48dp</dimen>
+    <dimen name="bundled_home_tile_height">0dp</dimen>
+    <dimen name="bundled_home_tile_margin_value">8dp</dimen>
+    <dimen name="custom_home_tile_height">65dp</dimen>
+    <dimen name="custom_home_tile_margin_value">5dp</dimen>
```
mcomella: You added a binary `dump.hprof` file: please remove it.

app/src/main/java/org/mozilla/focus/home/HomeFragment.kt
```diff
@@ -214,6 +222,17 @@ private class HomeTileAdapter(
...
+        layoutParams.width = LinearLayout.LayoutParams.MATCH_PARENT
+        layoutParams.height = iconView.resources.getDimensionPixelSize(tileHeight)
+        layoutParams.gravity = Gravity.CENTER
```
mcomella: Same: this is a constant: you can define it XML. THis probably didn't work before because you weren't re-using the original layout params.

mcomella: You also need to do `iconView.layoutParams = layoutParams`. This forces it to refresh (an Android gotcha).

app/src/main/java/org/mozilla/focus/home/HomeFragment.kt
```diff
@@ -214,6 +222,17 @@ private class HomeTileAdapter(
...
     }
 
+    fun setLayoutParams(iconView: View, tileMarginValue: Int, tileHeight: Int) {
+        val marginLayoutParams = iconView.layoutParams as ViewGroup.MarginLayoutParams
+        val marginValue = iconView.resources.getDimensionPixelSize(tileMarginValue)
+        marginLayoutParams.setMargins(marginValue, marginValue, marginValue, marginValue)
+        val layoutParams = iconView.layoutParams as LinearLayout.LayoutParams
+        layoutParams.weight = 1f
+        layoutParams.width = LinearLayout.LayoutParams.MATCH_PARENT
+        layoutParams.height = iconView.resources.getDimensionPixelSize(tileHeight)
```
mcomella: Setting an explicit height isn't necessary: weight is setting the height for us. We only need to update the margins

app/src/main/java/org/mozilla/focus/home/HomeFragment.kt
```diff
@@ -214,6 +222,17 @@ private class HomeTileAdapter(
         }
     }
 
+    fun setLayoutParams(iconView: View, tileMarginValue: Int, tileHeight: Int) {
```
mcomella: nit: private

app/src/main/java/org/mozilla/focus/home/HomeFragment.kt
```diff
@@ -260,14 +274,14 @@ private fun onBindCustomHomeTile(uiLifecycleCancelJob: Job, holder: TileViewHold
             }
         }
 
-        // We wait for both to complete so we can animate them together.
-        val screenshot = screenshotDeferred.await()
+        // Wait for both to complete so we can animate them together.
+        val screenshot = screenshotDeferred.await() // TODO: if null, provide placeholder.
```
mcomella: nit: rebase error: please don't change these lines

app/src/main/res/layout/home_tile.xml
```diff
@@ -34,12 +34,14 @@
         android:layout_height="match_parent"
         android:orientation="vertical">
 
+        <!-- margins are defined
```
mcomella: nit: one line

app/src/main/res/layout/home_tile.xml
```diff
@@ -34,12 +34,14 @@
...
+           - in HomeFragment.kt -->
         <ImageView
             android:id="@+id/tile_icon"
-            android:layout_width="match_parent"
             android:layout_height="0dp"
-            android:layout_margin="8dp"
+            android:layout_width="match_parent"
             android:layout_weight="1"
+            android:layout_gravity="center"
```
mcomella: nit: I think this line is unnecessary.

bkmunar:
@mcomella yup passed by @lime124

PR #651, opened 2018-03-09
Fixes #596 - Fixes the pressed state for the context menu button

### LINKED ISSUE #596: Context menu indicator doesn't change colors when home tile is pressed
screenshot_1520379035

boek:
Had to change the foreground to its own view so I could add the effect to the tile and the context menu button

### DIFF
--- a/app/src/main/res/drawable/context_menu_button_background.xml
+++ b/app/src/main/res/drawable/context_menu_button_background.xml
@@ -2,10 +2,16 @@
 <!-- This Source Code Form is subject to the terms of the Mozilla Public
    - License, v. 2.0. If a copy of the MPL was not distributed with this
    - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
-<shape
-    xmlns:android="http://schemas.android.com/apk/res/android"
-    android:shape="oval">
-    <solid android:color="@color/default_background_gray" />
-    <stroke android:color="@color/overlay_button_border"
-        android:width="0.5dp" />
-</shape>
+<selector xmlns:android="http://schemas.android.com/apk/res/android">
+    <item android:state_pressed="true">
+        <shape android:shape="oval">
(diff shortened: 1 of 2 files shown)


app/src/main/res/drawable/context_menu_button_background.xml
```diff
@@ -2,10 +2,16 @@
...
-    android:shape="oval">
-    <solid android:color="@color/default_background_gray" />
-    <stroke android:color="@color/overlay_button_border"
-        android:width="0.5dp" />
-</shape>
\ No newline at end of file
+<selector xmlns:android="http://schemas.android.com/apk/res/android">
+    <item android:state_pressed="true">
+        <shape android:shape="oval">
+            <solid android:color="#5b5b5c" />
```
mcomella: Ideally this was the same pressed color as the overlay view but I can see why it's not (because we use alpha for the overlay).

app/src/main/res/layout/home_tile.xml
```diff
@@ -59,4 +59,10 @@
             android:textColor="@android:color/black"
             tools:text="YouTube" />
     </LinearLayout>
+
+    <View
```
mcomella: nit: Add a comment explaining why this is necessary.

app/src/main/res/drawable/context_menu_button_background.xml
boek: Yeah, we wanted the same visual effect but that was hard to achieve with one overlay as this left the bounds of the home_tile.

app/src/main/res/layout/home_tile.xml
boek: https://github.com/boek/firefox-tv/blob/85acba3a056743da561046030fcae775597facfa/app/src/main/res/layout/home_tile.xml#L63

PR #653, opened 2018-03-10
Fixes #548 - Prevents the URL from being updated while the EditText has focus

### LINKED ISSUE #548: URL bar will overwrite user input when loading
STR:
1) Click IMDB

The next steps should be performed quickly, while the page loads

2) Open menu
3) Start typing

Expected: input stays
Actual: input is overwritten by redirects

boek:


app/src/main/java/org/mozilla/focus/widget/BrowserNavigationOverlay.kt
```diff
@@ -105,6 +105,13 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
         autocompleteFilter.load(context.applicationContext)
         setOnFilterListener { searchText, view -> autocompleteFilter.onFilter(searchText, view) }
 
+        // #548: Get the current URL when we leave resign focus
+        setOnFocusChangeListener { _, hasFocus ->
+            if (!hasFocus) {
+                setText(navigationStateProvider?.getCurrentUrl())
```
mcomella: Since the url bar is the first thing focused when this screen is shown, we don't set the text when the overlay is initially shown: we should do that in `onStart()`

mcomella: This unsets any input the user has entered when using the url bar – I don't like that (and desktop doesn't do that). I think we should just remove this OnFocusChangeListener entirely and just prevent the url bar from changing if the user already has focus.

mcomella:
@boek btw, not sure how many cycles you have left for Fire TV (or if you want to avoid context switching costs) – if you want, we can take this over. Just let us know!

boek:
@mcomella Thanks for the bump! For some reason I totally thought this had landed.. I'll address it today!

mcomella:
I'm thinking through this again: I think this could be used to mislead the user that they're on a secure property when they're not. STR:
- User goes to legitimate, secure website
- User opens overlay, highlighting the URL bar (by default) - this stops updates
- Website redirects user to another domain (no longer secure to enter data from first data)
- User closes overlay, thinking they're on the first domain and *may enter secure data into this unsecure second domain*

As such, I think we want to do something else – sorry I hadn't thought about this sooner. Here are the guidelines:
- If the URL updates, the user should know about it
- We should not disrupt the user's typing

It sounds like your original approach (don't update if the user has the soft keyboard up) was a good one but we struggled to implement it.

One proxy idea I came up with: allow the URL to change until the user enters some text. After this point, do not change the URL (until the overlay is hidden and shown again). Since the user has entered text, they now they've screwed with the URL and hopefully know it's no longer accurate.

We could also leave the implementation the way it is in this bug and consider showing the progress bar with the current URL when the overlay is closed, i.e. move the source of truth from the overlay to the progress bar (but overloading the progress bar could be confusing too).

Note: sometimes desktop will overwrite things the user is typing (like if you setTimeout 5s to change window.location and start typing in the URL bar). I'm not sure what rules they use, but they don't have a modal overlay either.

@boek What do you think? Once you've figured out a solution, perhaps we should rope in the security team too?

mcomella:
Closing in favor of #714.

PR #655, opened 2018-03-10
Closes #589 - Persist overlay state when going back from Settings.

### LINKED ISSUE #589: Hitting back from Settings shouldn't hide the overlay
Filing from LAT feedback.

liuche:
Doesn't persist webview state (we'd have to pull the reference out into BrowserFragment, which maybe isn't that hard hmmm....


liuche:
It's true, random boolean flags is kind of messy - well, we can leave this open bc it's pretty small, and then decide whether or not to merge it in.

@bbinto can you give some feedback on this based on the comments above^?

liuche:
Okay, I agree with @mcomella's points and concerns, but upon doing more research, the Fragment paradigm differentiates between "fragment state" and "UI state", and fragments don't provide a way to persist UI state that through `onSaveInstanceState`. From the docs, `onSaveInstanceState` can be called any time before `onDestroy`, but fragments aren't destroyed when being moved to the back stack.

PR #664, opened 2018-03-12
Closes #623: Polish: screenshot placeholder size

### LINKED ISSUE #623: Polish: screenshot placeholder size
Tif okayed #599 on Slack but we want to polish it before release.

Added complication: Bryan's patch could affect this code so we need to wait for it to land.

bkmunar:
UI approved by @lime124 
screen shot 2018-03-12 at 3 56 00 pm

### DIFF
--- a/app/src/main/java/org/mozilla/focus/home/HomeFragment.kt
+++ b/app/src/main/java/org/mozilla/focus/home/HomeFragment.kt
@@ -261,8 +261,10 @@ private fun onBindCustomHomeTile(uiLifecycleCancelJob: Job, holder: TileViewHold
         val validUri = item.url.toJavaURI()
 
         val screenshotDeferred = async {
-            val screenshot = HomeTileScreenshotStore.read(itemView.context, item.id)
-            screenshot ?: HomeTilePlaceholderGenerator.generate(itemView.context, item.url)
+            val homeTileCornerRadius = itemView.resources.getDimension(R.dimen.home_tile_corner_radius)
+            val homeTilePlaceholderCornerRadius = itemView.resources.getDimension(R.dimen.home_tile_placeholder_corner_radius)
+            val screenshot = HomeTileScreenshotStore.read(itemView.context, item.id)?.withRoundedCorners(homeTileCornerRadius)
+            screenshot ?: HomeTilePlaceholderGenerator.generate(itemView.context, item.url).withRoundedCorners(homeTilePlaceholderCornerRadius)
         }
 
         val titleDeferred = if (validUri == null) {
@@ -281,7 +283,7 @@ private fun onBindCustomHomeTile(uiLifecycleCancelJob: Job, holder: TileViewHold
         // NB: Don't suspend after this point (i.e. between view updates like setImage)
         // so we don't see intermediate view states.
         // TODO: It'd be less error-prone to launch { /* bg work */ launch(UI) { /* UI work */ } }
-        iconView.setImageBitmap(screenshot.withRoundedCorners(iconView.resources.getDimension(R.dimen.home_tile_corner_radius)))
+        iconView.setImageBitmap(screenshot)
         titleView.text = title
 
         // Animate to avoid pop-in due to thread hand-offs. TODO: animation is janky.
--- a/app/src/main/java/org/mozilla/focus/home/HomeTilePlaceholderGenerator.kt
+++ b/app/src/main/java/org/mozilla/focus/home/HomeTilePlaceholderGenerator.kt
@@ -17,18 +17,13 @@ import org.mozilla.focus.utils.UrlUtils
 class HomeTilePlaceholderGenerator {
 
     companion object {
-        private val TEXT_SIZE_DP = 52f
+        private val TEXT_SIZE_DP = 22f
         private val DEFAULT_ICON_CHAR = '?'
 
         fun generate(context: Context, url: String?): Bitmap {
             val startingChar = getRepresentativeCharacter(url)
-
-            // When this method is called, the view we're setting this bitmap on has no size (maybe
-            // because it doesn't contain an image?) so we use a hard-coded estimated size instead.
-            val width = context.resources.getDimensionPixelSize(R.dimen.home_tile_icon_estimated_width)
-            val height = context.resources.getDimensionPixelSize(R.dimen.home_tile_icon_estimated_height)
-
-            val bitmap = Bitmap.createBitmap(width, height, Bitmap.Config.ARGB_8888)
+            val dimen = context.resources.getDimensionPixelSize(R.dimen.home_tile_placeholder_icon_size)
+            val bitmap = Bitmap.createBitmap(dimen, dimen, Bitmap.Config.ARGB_8888)
             bitmap.eraseColor(ContextCompat.getColor(context, R.color.tv_ink))
             return drawCharacterOnBitmap(context, startingChar, bitmap)
         }
--- a/app/src/main/res/values/dimens.xml
+++ b/app/src/main/res/values/dimens.xml
@@ -17,6 +17,8 @@
     <!-- Specified in AdaptiveIconDrawable docs. -->
 
     <dimen name="home_tile_corner_radius">4dp</dimen>
+    <dimen name="home_tile_placeholder_corner_radius">12dp</dimen>
+    <dimen name="home_tile_placeholder_icon_size">50dp</dimen>
     <dimen name="urlbar_corner_radius">4dp</dimen>
 
     <dimen name="remote_cursor_size">45dp</dimen>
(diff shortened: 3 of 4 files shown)


app/src/main/java/org/mozilla/focus/home/HomeTilePlaceholderGenerator.kt
```diff
@@ -49,7 +51,7 @@ class HomeTilePlaceholderGenerator {
                     canvas.height / 2.0f - (paint.descent() + paint.ascent()) / 2.0f,
                     paint)
 
-            return bitmap
+            return bitmap.withRoundedCorners(DEFAULT_TILE_CORNER_RADIUS)
```
mcomella: We're doing this twice: the code in HomeFragment still rounds the corners.

app/src/main/res/values/dimens.xml
```diff
@@ -59,6 +59,6 @@
 
     <!-- In HomeTileScreenshotStore, we generate screenshots of this size and scale them:
          it should be a fine estimate for now. -->
-    <dimen name="home_tile_icon_estimated_width">480px</dimen>
-    <dimen name="home_tile_icon_estimated_height">270px</dimen>
+    <dimen name="home_tile_icon_estimated_width">50dp</dimen>
```
mcomella: I see that the placeholder *does not* fill the space we give it (i.e. the same space as a screenshot). Are these numbers being used as absolute values (i.e. it shouldn't be scaled up or down)? In that case, this is no longer an estimate: please rename these fields and update the comment to be more accurate to their current usage (and mention that they're meant to be used absolutely).

On that note, we probably only need one of these if it's a square.

app/src/main/java/org/mozilla/focus/home/HomeTilePlaceholderGenerator.kt
```diff
@@ -12,12 +12,14 @@ import android.net.Uri
...
 import org.mozilla.focus.R
+import org.mozilla.focus.ext.withRoundedCorners
 import org.mozilla.focus.utils.UrlUtils
 
 class HomeTilePlaceholderGenerator {
 
     companion object {
-        private val TEXT_SIZE_DP = 52f
+        private val TEXT_SIZE_DP = 22f
+        private val DEFAULT_TILE_CORNER_RADIUS = 25f
```
mcomella: While I'm not entirely opposed to leaving this in the file, we should be adjusting this to DP (like how textSize does it) otherwise the corner rounding amount could change per device.

bkmunar:
@mcomella made changes, although not sure about the `withRoundedCorners` call in the ScreenshotStore. What do u think?

app/src/main/java/org/mozilla/focus/home/HomeTilePlaceholderGenerator.kt
```diff
@@ -12,21 +12,23 @@ import android.net.Uri
...
 
         fun generate(context: Context, url: String?): Bitmap {
             val startingChar = getRepresentativeCharacter(url)
 
             // When this method is called, the view we're setting this bitmap on has no size (maybe
-            // because it doesn't contain an image?) so we use a hard-coded estimated size instead.
-            val width = context.resources.getDimensionPixelSize(R.dimen.home_tile_icon_estimated_width)
-            val height = context.resources.getDimensionPixelSize(R.dimen.home_tile_icon_estimated_height)
+            // because it doesn't contain an image?) so we use a hard-coded size instead.
+            val width = context.resources.getDimensionPixelSize(R.dimen.home_tile_placeholder_icon_size)
```
mcomella: nit: `val dimen = ...` and use this dimen var twice, rather than accessing the resource off disk twice.

app/src/main/java/org/mozilla/focus/home/HomeTilePlaceholderGenerator.kt
```diff
@@ -12,21 +12,23 @@ import android.net.Uri
...
 import org.mozilla.focus.R
+import org.mozilla.focus.ext.withRoundedCorners
 import org.mozilla.focus.utils.UrlUtils
 
 class HomeTilePlaceholderGenerator {
 
     companion object {
-        private val TEXT_SIZE_DP = 52f
+        private val TEXT_SIZE_DP = 22f
+        private val DEFAULT_TILE_CORNER_RADIUS = 12f
```
mcomella: nit: append with `_DP` to make it clear that this is using dp.

app/src/main/java/org/mozilla/focus/home/HomeTileScreenshotStore.kt
```diff
@@ -134,7 +136,7 @@ object HomeTileScreenshotStore {
             null
         } else {
             file.inputStream().use {
-                BitmapFactory.decodeStream(it, null, BITMAP_FACTORY_OPTIONS)
+                BitmapFactory.decodeStream(it, null, BITMAP_FACTORY_OPTIONS).withRoundedCorners(context.resources.getDimension(R.dimen.home_tile_corner_radius))
```
mcomella: Oh, I see what the problem is here: we need to round in two different places. I don't think we should be modifying the items in the store: let's modify them right before we set them (switching on the tile type) (i.e. don't round the corners inside the placeholderGenerator too). e.g.:
```kotlin
        val screenshotDeferred = async {
            val screenshot = HomeTileScreenshotStore.read(itemView.context, item.id)?.withRoundedCorners(...)
            screenshot ?: HomeTilePlaceholderGenerator.generate(itemView.context, item.url).withRoundedCorners(...)
        }
```

I prefer this approach because it keeps the related code together (and hopefully reduces duplication) and because the HomeTileScreenshotStore should have one responsibility: retrieving items from disk verbatim and that's it (e.g. if we wanted to create a cache, I'd wrap the Store we have rather than putting a cache inside of it).

app/src/main/res/values/dimens.xml
```diff
@@ -57,8 +57,6 @@
     <!-- We want to keep button_size and urlbar height the same to keep them aligned -->
     <dimen name="urlbar_height">@dimen/button_size</dimen>
 
-    <!-- In HomeTileScreenshotStore, we generate screenshots of this size and scale them:
-         it should be a fine estimate for now. -->
-    <dimen name="home_tile_icon_estimated_width">480px</dimen>
-    <dimen name="home_tile_icon_estimated_height">270px</dimen>
+    <!-- In HomeTileScreenshotStore, we generate placeholder screenshot icons of this size -->
```
mcomella: nit: I think this is redundant to the name: I think you can remove it.

I had the comment before to emphasize that we're making an estimate which is scaled to the right size, rather than giving an exact value (which is how I'd expect to use it).

app/src/main/java/org/mozilla/focus/home/HomeTilePlaceholderGenerator.kt
```diff
@@ -17,18 +17,17 @@ import org.mozilla.focus.utils.UrlUtils
...
         private val DEFAULT_ICON_CHAR = '?'
 
         fun generate(context: Context, url: String?): Bitmap {
             val startingChar = getRepresentativeCharacter(url)
 
             // When this method is called, the view we're setting this bitmap on has no size (maybe
-            // because it doesn't contain an image?) so we use a hard-coded estimated size instead.
-            val width = context.resources.getDimensionPixelSize(R.dimen.home_tile_icon_estimated_width)
-            val height = context.resources.getDimensionPixelSize(R.dimen.home_tile_icon_estimated_height)
+            // because it doesn't contain an image?) so we use a hard-coded size instead.
```
mcomella: nit: this comment isn't useful anymore: can you figure out why? Please respond here and remove it.

If you ever find comments/code that you think are unnecessary, feel free to remove it! Someone will complain if it shouldn't have been removed. :)

app/src/main/java/org/mozilla/focus/home/HomeFragment.kt
```diff
@@ -261,8 +261,8 @@ private fun onBindCustomHomeTile(uiLifecycleCancelJob: Job, holder: TileViewHold
         val validUri = item.url.toJavaURI()
 
         val screenshotDeferred = async {
-            val screenshot = HomeTileScreenshotStore.read(itemView.context, item.id)
-            screenshot ?: HomeTilePlaceholderGenerator.generate(itemView.context, item.url)
+            val screenshot = HomeTileScreenshotStore.read(itemView.context, item.id)?.withRoundedCorners(itemView.resources.getDimension(R.dimen.home_tile_corner_radius))
```
mcomella: nit: we shouldn't use such long line lengths because it makes code hard to read. That being said, this will get fixed when we set up detektCheck.

app/src/main/java/org/mozilla/focus/home/HomeTilePlaceholderGenerator.kt
bkmunar: I think it's because we're simply just giving it a hard-coded size at this point so no need to justify.

mcomella: Yep! 👍

PR #668, opened 2018-03-13
I662 rm unused code/resources

mcomella:


app/src/main/java/org/mozilla/focus/search/SearchEngineManager.java
```diff
@@ -63,9 +45,6 @@
     private static final String PREF_KEY_CUSTOM_SEARCH_VERSION = "pref_custom_search_version";
```
liuche: Probably don't need these two either.

app/src/main/java/org/mozilla/focus/search/SearchEngineParser.java
```diff
@@ -115,7 +114,6 @@ private static void readUrl(XmlPullParser parser, SearchEngine searchEngine) thr
                 searchEngine.resultsUris.add(uri);
             }
         } else if (type.equals(URLTYPE_SUGGEST_JSON)) {
-            searchEngine.suggestUri = uri;
```
liuche: do nothing?

app/src/main/java/org/mozilla/focus/utils/HtmlLoader.java
```diff
@@ -50,47 +47,4 @@ public static String loadResourceFile(@NonNull final Context context,
         }
     }
 
-    private final static byte[] pngHeader = new byte[] { -119, 80, 78, 71, 13, 10, 26, 10 };
-
-    public static String loadPngAsDataURI(@NonNull final Context context,
```
liuche: We don't need this? Was it used for error pages...?

app/src/main/java/org/mozilla/focus/utils/ThreadUtils.java
```diff
@@ -1,45 +0,0 @@
...
-
-import android.os.Handler;
-import android.os.Looper;
-
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
-
-import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
-
-public class ThreadUtils {
```
liuche: Do we just use kotlin coroutines instead, and no longer care about posting to UI/background threads? If we're going to use threads in the future is this worth keeping?

app/src/main/res/values/strings.xml
```diff
@@ -353,24 +68,6 @@
     <!-- This label is shown above a list of apps that can be used to open a given link -->
     <string name="external_multiple_apps_matched_exit">Exit Private Browsing?</string>
```
liuche: Are these two needed?

app/src/main/res/values-ca/strings.xml
```diff
@@ -3,80 +3,31 @@
   <string name="action_cancel">Cancel·la</string>
```
liuche: I don't know if this deletion is necessary - this should be done with a string export rather than a manual? deletion, because we should export so it doesn't get too out of date with the l10n repo (though actually, maybe it's fine because this is mainly a string deletion?). And we should do a string export/import without any new strings so Delphine doesn't need to look at a horrible diff with new strings and deleted strings.

app/src/main/res/xml/settings.xml
```diff
@@ -68,52 +54,6 @@
...
-        -->
-
-    </PreferenceCategory>
-
-    <PreferenceCategory
-        android:title="@string/preference_category_mozilla">
-
-        <org.mozilla.focus.widget.DefaultBrowserPreference
-            android:layout="@layout/focus_preference"
-            android:key="@string/pref_key_default_browser"
```
liuche: What about these `@string/pref`'s? the ones ahunt added (for string compare reasons, I don't remember the specifics).

app/src/main/java/org/mozilla/focus/search/SearchEngineManager.java
mcomella: Removed in a future commit, I think.

app/src/main/java/org/mozilla/focus/utils/HtmlLoader.java
mcomella: It's used in focus:about: we commented out the code to load the PNG in LocalizedContent.loadAbout. We can re-add it if we need pngs in the about pages.

app/src/main/java/org/mozilla/focus/utils/ThreadUtils.java
mcomella: My plan is to just use kotlin coroutines instead.

The value of this file is that it 1) creates a centralized background thread for the app and 2) makes the calling background methods simpler in Java.

For 1), coroutines creates a shared ThreadPool we can post to, we can use existing primitives that spawn threads (like AsyncTask), or we can create an app-wide Executor that we can post to. For 2) we're using kotlin and lambdas make this much simpler.

app/src/main/res/values-ca/strings.xml
mcomella: @liuche I don't understand the l10n system too well so I don't know. What do you propose?

I'd think it's better to give us a clean repository slate for Firefox TV, which is what I'm trying to do here. I figured the import/export tools would just delete all these already-translated strings. I agree with the idea that we should just export this so it's easier for the l10n team.

app/src/main/res/xml/settings.xml
mcomella: They're used in other places for whatever reason – the IDE has removed any that are actually unused.

I was going for "good enough" here rather than trying to get every last piece.

app/src/main/res/values/strings.xml
mcomella: Probably not: this is called by the `IntentUtils.handleExternalUri`, which we don't really support ATM. That being said, I haven't removed much code from existing methods that are called (e.g. how this method is probably called from MainActivity) – I just removed code that the IDE determined was unused. I figure we can clean these types of methods up incrementally, as we start changing the code that uses them.

PR #669, opened 2018-03-13
Issue #425: upgrade gradle plugin to 3.0.1

mcomella:

PR #672, opened 2018-03-13
Upgradle gradle plugin to 3.0.1 🔄🔄

mcomella:
Replaces #669.

TaskCluster failures are:
```
* What went wrong:
A problem occurred configuring project ':app'.
> You have not accepted the license agreements of the following SDK components:
  [Android SDK Build-Tools 26.0.2].
  Before building your project, you need to accept the license agreements and complete the installation of the missing components using the Android Studio SDK Manager.
  Alternatively, to learn how to transfer the license agreements from one workstation to another, go to http://d.android.com/r/studio-ui/export-licenses.html
* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output.
```

@pocmo Any idea how I can fix this?


liuche:
@mcomella when you get the answer to this, can you also document this on the wiki? Seems like something we'll need to fix any time we update SDK/gradle/plugins.

pocmo:
> @pocmo Any idea how I can fix this?

Yes!

Normally gradle and the Android plugin will just download all (new) dependencies and Android SDK bits if needed. However sometimes the Android SDK requires you to accept a (new/updated) license when downloading a new version of something. This is always a pain in automation. When creating the Docker image we accept all licenses automatically. So what I did was:

* Update the Docker image to explicitly download 26.0.2 (In addition to what is already there. To not break builds that happen before this PR gets merged.)

* Test running the gradle commands taskcluster would run for a PR locally in the image. Once for master (To verify everything still works) and then for your branch (to verify that everything will work after the merge).

* Now I am going to push the new image to Docker Cloud so that taskcluster will be able to use it. (This is currently happening. I'll comment again once it is done).

After that you should be able to re-trigger this PR (Pro-Tip: Changing the title of the PR will trigger a re-run).

From my local run I can tell you already that the new tools (or due to the new gradle version?) find more lint issues that need to be fixed to get a successful run.

pocmo:
I opened PR #676 for the Dockerfile changes.

pocmo:
> (This is currently happening. I'll comment again once it is done).

Push to Docker hub completed! I'll re-trigger this PR. As mentioned above this will fail with lint issues - but at least this should give us a nice lint report (and verify that what I did was actually correct).

pocmo:
Seems like the latest run didn't use the new image yet? I'll re-trigger once more just to be sure.

pocmo:
Ah, yes, now it's the failure I was expecting (*):

> Lint found errors in the project; aborting build.

Here's the report:
https://taskcluster-artifacts.net/S9QK_voBS8KP0Uy518EMcw/0/public/lint-results.html

The only error seems to be InvalidVectorPath and I had to fix some of them in the Focus repository too. Maybe it's easier to just re-import the ones where lint is complaining?

(*) I noticed that taskcluster keeps a virtual machine running a task alive for a brief moment of time. If a new task arrives (for the same image) then it will re-use the machine and not re-download the image. I assume this happened here. So far it was only a matter of minutes until taskcluster killed the machine. The right way to do this would be to version the image so that taskcluster picks the actual one you want it to pick.

mcomella:
I think the only `InvalidVectorPath` files we removed: I'm going to rebase and try pushing.

mcomella:
It passes, yay!

app/src/test/java/org/mozilla/focus/search/SearchEngineParserTest.java
```diff
@@ -84,6 +84,8 @@ private static void collectSearchPlugins(Collection<Object[]> searchPlugins, Fil
     }
 
     private static File getBasePath() {
+        // Ordinarily, we could use RuntimeEnvironment.application to get the resources but
```
liuche: Yes, comments! :D

PR #674, opened 2018-03-14
Fixes #489 - Updates the visual style of the cursor

### LINKED ISSUE #489: update cursor/pointer

boek:
image


app/src/main/java/org/mozilla/focus/browser/cursor/CursorView.kt
```diff
@@ -38,11 +36,18 @@ class CursorView(context: Context, attrs: AttributeSet) : View(context, attrs) {
...
+
+    private val cursorBg = BitmapFactory.decodeResource(resources, R.drawable.cursor_bg)
+    private val cursorBgBlendingMode = PorterDuffXfermode(PorterDuff.Mode.MULTIPLY)
+    private val cursorFg = BitmapFactory.decodeResource(resources, R.drawable.cursor_fg)
+    private val cursorFgBlendingMode = PorterDuffXfermode(PorterDuff.Mode.SCREEN)
+
+    init {
+        // In order to draw the bitmaps with the blending mode and
+        // alpha we have to set the LayerType to software
+        setLayerType(View.LAYER_TYPE_SOFTWARE, null)
```
mcomella: We probably don't want to do this: it's going to negatively affect page load performance if the Cursor needs to be redrawn (which I expect to be any time the content under the cursor is redrawn or the cursor is moved).

That is, unless you figured out that this is actually only drawn once and the GPU handles translations and redraws. In which case, please add a comment explaining this to be the case.

Do you disagree?

app/src/main/java/org/mozilla/focus/browser/cursor/CursorView.kt
```diff
@@ -58,11 +63,13 @@ class CursorView(context: Context, attrs: AttributeSet) : View(context, attrs) {
     override fun onDraw(canvas: Canvas) {
         super.onDraw(canvas)
 
-        // We draw the cursor in the top-left corner, offset so the circle isn't clipped by its
-        // parent. The translationX/Y properties will draw it into the desired position on the
-        // screen. We do it this way to avoid allocating new RadialGradients, which are
-        // constructed in relation to the current position of the circle.
-        canvas.drawCircle(radius, radius, radius, paint)
+        // We have to paint a background for the shadow to apepar
```
mcomella: nit: appear

app/src/main/java/org/mozilla/focus/browser/cursor/CursorView.kt
```diff
@@ -58,11 +63,13 @@ class CursorView(context: Context, attrs: AttributeSet) : View(context, attrs) {
...
+        canvas.drawRect(0f, 0f, width + 0f, height + 0f, paint)
```
mcomella: nit: let's rename paint to make it apparent how it's being used (with a transparent color).

app/src/main/java/org/mozilla/focus/browser/cursor/CursorView.kt
boek: @mcomella I agree, but I wasn't able to to keep the transparency without that. Any ideas for an alternative?

mcomella: @boek Not off the top of my head but if a UI effect hurts performance, especially pageload, I'd rather not do it.

Thoughts:
- The fennec platform team deals with low-level rendering: perhaps they have some ideas about how to do this? Talk to :snorp
- Have you asked UX if there is a similar, but simpler, implementation they could suggest?

boek:
Waiting on feedback from @lime124 re: The cursor without the blending modes.

lime124:
@boek just collecting feedback from some people!

lime124:
i didn't get any feedback other than love it! after using it, i do think we need a press state. should that be a new ticket? @boek ?

Sdaswani:
cool let's merge, and then since @boek is working on focus Android full time now, other engineers can iterate on it.

app/src/main/java/org/mozilla/focus/browser/cursor/CursorView.kt
```diff
@@ -38,13 +34,12 @@ class CursorView(context: Context, attrs: AttributeSet) : View(context, attrs) {
...
     private val paint = Paint().apply {
         style = Paint.Style.FILL
-        alpha = CURSOR_ALPHA
         isAntiAlias = true
-        shader = RadialGradient(radius, radius, /* radius */ radius,
-                ContextCompat.getColor(context, R.color.teal50), ContextCompat.getColor(context, R.color.photonBlue50),
-                Shader.TileMode.CLAMP)
     }
 
+    private val cursorBg = BitmapFactory.decodeResource(resources, R.drawable.cursor_bg)
```
mcomella: nit: can you optimize these assets if you haven't done so already? I use [ImageOptim](https://imageoptim.com/mac)

mcomella: These resources need to be in a -*hdpi folder, or else they won't get scaled (i.e. this will appear as different sizes on 720p and 1080p). Please put it in one of these folders (I'm not sure which one is appropriate to make this appear to be the same size – you'll have to calculate it from the multiplication scales: see the multipliers here: https://developer.android.com/training/multiscreen/screendensities.html#TaskProvideAltBmp)

mcomella:
Closing in favor of #711.

PR #679, opened 2018-03-14
Closes #663:  Move code out of activity/ and fragment/ folders

### LINKED ISSUE #663: Move code out of activity/ and fragment/ folders
These packages are not helpful and we should move the code into more appropriate directories like "home" and "browser".

bkmunar:


app/src/androidTest/java/org/mozilla/focus/activity/BrowserOverlayTest.java
```diff
@@ -39,7 +40,7 @@
 import static android.support.test.espresso.web.sugar.Web.onWebView;
 import static org.hamcrest.Matchers.allOf;
 import static org.hamcrest.Matchers.containsString;
-import static org.mozilla.focus.activity.OnboardingActivity.ONBOARD_SHOWN_PREF;
+import static org.mozilla.focus.home.OnboardingActivity.ONBOARD_SHOWN_PREF;
 
 @RunWith(AndroidJUnit4.class)
 public class BrowserOverlayTest {
```
mcomella: These tests also need to move to match the components they're testing.

app/src/main/java/org/mozilla/focus/browser/MainActivity.kt
```diff
@@ -3,7 +3,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-package org.mozilla.focus.activity
+package org.mozilla.focus.browser
```
mcomella: I think MainActivity and ScreenController should be at the top-level (`org.mozilla.focus`): they switch between all the other features, which are composed into different packages, such as `home/` for the home screen feature.

app/src/main/java/org/mozilla/focus/home/NewSettingsFragment.kt
```diff
@@ -2,7 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-package org.mozilla.focus.fragment
+package org.mozilla.focus.home
```
mcomella: NewSettings isn't a part of home – it can be accessed from the browserOverlay too. I think it should be at the top-level.

app/src/main/java/org/mozilla/focus/home/OnboardingActivity.kt
```diff
@@ -2,7 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-package org.mozilla.focus.activity
+package org.mozilla.focus.home
```
mcomella: Onboarding is also not a "home" feature.

PR #680, opened 2018-03-14
Closes #634: With > 1 page of home tiles, sometimes cursor disappears…

### LINKED ISSUE #634: With > 1 page of home tiles, sometimes cursor disappears when scrolling down
STR (on Fire Stick, not sure if matters)
- Add > 3 rows of home tiles so it exceeds 1 page
-  Open home
- Click down to enter tiles (from url bar)
- Click right
- Click down 3 times, until you scroll the screen

Expected: item is highlighted
Actual: item is not highlighted: url bar is highlighted

Note: this is not reproducible 100% of the time.

Maybe our recyclerView is creating new views when we scroll down so sometimes the view is laid out and focused, but other times the view isn't laid out yet and we select the url bar outside the container.

bkmunar:
… when scrolling down

### DIFF
--- a/app/src/main/java/org/mozilla/focus/home/HomeFragment.kt
+++ b/app/src/main/java/org/mozilla/focus/home/HomeFragment.kt
@@ -12,6 +12,7 @@ import android.support.v4.app.Fragment
 import android.support.v4.content.ContextCompat
 import android.support.v7.widget.GridLayoutManager
 import android.support.v7.widget.RecyclerView
+import android.util.TypedValue
 import android.view.ContextMenu
 import android.view.KeyEvent
 import android.view.LayoutInflater
@@ -21,6 +22,7 @@ import android.view.ViewGroup
 import android.view.animation.DecelerateInterpolator
 import android.widget.LinearLayout
 import kotlinx.android.synthetic.main.fragment_home.*
+import kotlinx.android.synthetic.main.home_tile.*
 import kotlinx.android.synthetic.main.home_tile.view.*
 import kotlinx.coroutines.experimental.CancellationException
 import kotlinx.coroutines.experimental.CompletableDeferred
@@ -90,6 +92,19 @@ class HomeFragment : Fragment() {
         }
 
         registerForContextMenu(view)
+
+        tileContainer.addOnScrollListener(object : RecyclerView.OnScrollListener() {
+            override fun onScrolled(recyclerView: RecyclerView?, dx: Int, dy: Int) {
+                val gridLayoutManager = recyclerView?.layoutManager as GridLayoutManager
+                val lastVisibleItem = gridLayoutManager.findLastCompletelyVisibleItemPosition()
+                // We add a scroll offset, revealing the next row to hint that there are more home tiles
+                if (dy > 0 && getFocusedTilePosition() > lastVisibleItem) {
+                    val scrollOffset = TypedValue.applyDimension(
+                            TypedValue.COMPLEX_UNIT_DIP, home_tile.height.toFloat() / 2, context.resources.displayMetrics)
+                    recyclerView.smoothScrollBy(0, scrollOffset.toInt())
+                }
+            }
+        })
     }
 
     override fun onCreateContextMenu(menu: ContextMenu?, v: View?, menuInfo: ContextMenu.ContextMenuInfo?) {
--- a/app/src/main/res/layout/home_tile.xml
+++ b/app/src/main/res/layout/home_tile.xml
@@ -4,6 +4,7 @@
    - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
 <FrameLayout xmlns:android="http://schemas.android.com/apk/res/android"
               xmlns:tools="http://schemas.android.com/tools"
+              android:id="@+id/home_tile"
               android:orientation="vertical"
               android:layout_width="140dp"
               android:layout_height="100dp"


app/src/main/java/org/mozilla/focus/home/HomeFragment.kt
```diff
@@ -90,6 +92,22 @@ class HomeFragment : Fragment() {
         }
 
         registerForContextMenu(view)
+
+        tileContainer.addOnScrollListener(object : RecyclerView.OnScrollListener() {
+            override fun onScrolled(recyclerView: RecyclerView?, dx: Int, dy: Int) {
+                if (dy > 0) {
+                    val layoutManager = recyclerView?.layoutManager as GridLayoutManager
+                    val visibleItemCount = layoutManager.childCount
```
liuche: Is this actually correct? Can you explain to me why this would work for what we want it to do, and does it work for all cases?

I'm not stoked about code from Stackoverflow, without comments or explanation...please add code comments for why this works and what problem this is fixing, or at least a comment in the github.

Ideally the goal for code (or hacks) is to make it easy to understand, optionally with comments detailing when/why this works (if the logic isn't simple/obvious). That way, if it introduces bugs, anyone could glance at the code and see "oh this is a wrong assumption" and have a fix.

app/src/main/java/org/mozilla/focus/home/HomeFragment.kt
```diff
@@ -90,6 +92,26 @@ class HomeFragment : Fragment() {
...
+                // Only add scroll offset when user is scrolling down
+                if (dy > 0) {
+                    val layoutManager = recyclerView?.layoutManager as GridLayoutManager
+                    val visibleItemCount = layoutManager.childCount
+                    val totalItemCount = layoutManager.itemCount
+                    val pastVisibleItems = layoutManager.findFirstVisibleItemPosition()
+                    // This checks to see if we have reached the last row of the RecyclerView after a scroll occurs.
+                    // If we have not reached that point, we additionally scroll with a small offset
+                    // to reveal the next row, thus preventing the refocusing of the URL bar.
+                    if (pastVisibleItems + visibleItemCount < totalItemCount) {
```
liuche: If we have more than three rows (say 4), won't this scroll extra when you move down to the second row, even if you haven't reached the "bottom" of the visible items? because
`pastVisibleItems` = 0
`visibleItemCount` = 3 x 5 = 15
`totalItemCount` = 4 x 5 = 20 > 15

app/src/main/java/org/mozilla/focus/home/HomeFragment.kt
```diff
@@ -90,6 +92,26 @@ class HomeFragment : Fragment() {
         }
 
         registerForContextMenu(view)
+
+        tileContainer.addOnScrollListener(object : RecyclerView.OnScrollListener() {
+            override fun onScrolled(recyclerView: RecyclerView?, dx: Int, dy: Int) {
+                // Only add scroll offset when user is scrolling down
+                if (dy > 0) {
+                    val layoutManager = recyclerView?.layoutManager as GridLayoutManager
```
liuche: You don't have to use GridLayoutManager, can use a parent because you don't use any GridLayoutManager-specific calls.

app/src/main/java/org/mozilla/focus/home/HomeFragment.kt
liuche: I think we want to try to scroll extra either when we reach the bottom "visible" row, or when we reach a certain margin from the bottom.

bkmunar: No it won't because onScrolled only gets called when you move down to the third row, the view has to physically scroll for the function to be called, not the focus to be changed.

bkmunar: So as long as there are still items left underneath what was just scrolled, an extra scroll is needed, that's how I thought about it

app/src/main/java/org/mozilla/focus/home/HomeFragment.kt
bkmunar: What do u mean by use a parent, I used GridLayoutManager because I needed access to those APIs for the children stuff

liuche: Here's the docs for GridLayoutManager - at the top is its inheritance chain. If you use GridLayoutManager, that's more constrained than it needs to be, its parents have have methods for `childCount`, etc as well. Unless there's one of these calls that is Grid-specific? In general, try to use the most general parent as possible.

https://developer.android.com/reference/android/support/v7/widget/GridLayoutManager.html

app/src/main/java/org/mozilla/focus/home/HomeFragment.kt
liuche: Okay, I see, that makes sense. So one problem I see is that a scroll action will trigger more nested scroll actions because this is called on every scroll, so the view will just autoscroll until it reaches the bottom. This isn't a problem if you don't have that many sites (or just one more row).

liuche: Since we're not in a rush to land, I think we should try to take care of edge cases. If you come up with a fix for this edge case, please push it as a *separate commit*.

app/src/main/java/org/mozilla/focus/home/HomeFragment.kt
```diff
@@ -90,6 +92,20 @@ class HomeFragment : Fragment() {
         }
 
         registerForContextMenu(view)
+
+        tileContainer.addOnScrollListener(object : RecyclerView.OnScrollListener() {
+            override fun onScrolled(recyclerView: RecyclerView?, dx: Int, dy: Int) {
+                val gridLayoutManager = recyclerView?.layoutManager as GridLayoutManager
+                val lastVisibleItem = gridLayoutManager.findLastCompletelyVisibleItemPosition()
+                // If the focused tile position is greater than the last fully visible item position,
+                // we must add a scroll offset to reveal the next row of tiles to prevent the URL bar from being focused.
```
liuche: Nit: I'd just say this is to hint that there are more items by revealing the next row - I know that "fixing jump of focus to urlbar" is what this bug is for, but that's more of a confusing side effect.

PR #684, opened 2018-03-15
I472 60fps cursor

mcomella:
This patch series:
- Smooths the cursor movement
- Smooths page scrolling
- Makes sure all user interactions are adjusted by time (which guarantees equal distance moved, no matter what the framerate is, e.g. we'll continue to scroll the same distance even during page load unlike the existing code)
- Otherwise preserves cursor/scrolling behavior. I think we can improve here by at least speeding up the scroll speed but maybe also applying slightly deceleration to scrolling (rather than immediate stops) and perhaps speeding up the cursor.

### DIFF
--- a/app/src/main/java/org/mozilla/focus/browser/cursor/CursorViewModel.kt
+++ b/app/src/main/java/org/mozilla/focus/browser/cursor/CursorViewModel.kt
@@ -19,11 +19,11 @@ import java.util.EnumSet
 import java.util.concurrent.TimeUnit
 import kotlin.properties.Delegates
 
-private const val UPDATE_DELAY_MILLIS = 20L
+private const val UPDATE_DELAY_MILLIS = 17L // ~60 FPS.
 private const val UPDATE_DELAY_MILLIS_F = UPDATE_DELAY_MILLIS.toFloat() // Avoid conversion in update loop.
 
 private const val ACCEL_MODIFIER = 0.98f
-private const val MAX_VELOCITY = 25f
+private const val MAX_VELOCITY = 21.25f
 
 private const val DOWN_TIME_OFFSET_MILLIS = 100
 
--- a/app/src/main/java/org/mozilla/focus/browser/cursor/CursorViewModel.kt
+++ b/app/src/main/java/org/mozilla/focus/browser/cursor/CursorViewModel.kt
@@ -20,9 +20,10 @@ import java.util.concurrent.TimeUnit
 import kotlin.properties.Delegates
 
 private const val UPDATE_DELAY_MILLIS = 20L
+private const val UPDATE_DELAY_MILLIS_F = UPDATE_DELAY_MILLIS.toFloat() // Avoid conversion in update loop.
 
-private const val FRICTION = 0.98f
-private const val MAX_SPEED = 25f
+private const val ACCEL_MODIFIER = 0.98f
+private const val MAX_VELOCITY = 25f
 
 private const val DOWN_TIME_OFFSET_MILLIS = 100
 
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/mozilla/focus/browser/cursor/CursorViewModel.kt
```diff
@@ -75,32 +76,35 @@ class CursorViewModel(
...
-        // increase smoothness when the system can't keep up with our desired framerate.
+        // opposed to the amount of time we expect to pass between frames): this guarantees equal
+        // distance travelled when the system can't keep up with our desired framerate but the
+        // cursor will noticeably skip ahead as frames are dropped.
         //
-        // This adjustment could be expressed more naturally if this algorithm was expressed
-        // as a series of kinematic equations, i.e. vnew = vold + accel * deltaTime.
-        val deltaVel = (vel + 1) * FRICTION - vel
-        val timeAdjustedDeltaVel = deltaVel * (deltaMillis / UPDATE_DELAY_MILLIS.toFloat())
-        vel = Math.min(MAX_SPEED, vel + timeAdjustedDeltaVel)
```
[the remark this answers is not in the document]
liuche: ...but I'm also having trouble figuring out what parts are actually correct, especially if we're fixing the calculations in this review.

app/src/main/java/org/mozilla/focus/browser/cursor/CursorViewModel.kt
```diff
@@ -75,32 +76,35 @@ class CursorViewModel(
...
+            updatePosForVel(dir, finalVel, framesPassed)
         }
 
         clampPos(pos, maxBounds)
         onUpdate(pos.x, pos.y, getScrollVel())
     }
 
-    private fun updatePosForVel(dir: Direction, vel: Float) {
+    private fun updatePosForVel(dir: Direction, vel: Float, framesPassed: Float) {
+        val timeAdjustedVel = vel * framesPassed
```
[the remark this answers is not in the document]
liuche: This should not be `vel` because you're adding it to distance - just `distance`?

app/src/main/java/org/mozilla/focus/browser/cursor/CursorViewModel.kt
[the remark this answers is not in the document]
liuche: We should pull this calculation into a different function like `getNewVelocity` and comment that the values and alg are experimentally obtained rather than adhering to any kinematic math.

app/src/main/java/org/mozilla/focus/browser/cursor/CursorViewModel.kt
```diff
@@ -43,7 +43,7 @@ private const val DOWN_TIME_OFFSET_MILLIS = 100
  * @param simulateTouchEvent Takes the given touch event and simulates a touch to the screen.
  */
 class CursorViewModel(
-        private val onUpdate: (x: Float, y: Float, scrollVel: PointF) -> Unit,
+        private val onUpdate: (x: Float, y: Float, percentMaxScrollVel: PointF, framesPassed: Float) -> Unit,
```
liuche: nit: rename `framesPassed` too to match multiplier from earlier.

liuche: What is the difference between `onUpdate` and `updatePosForVel`?

app/src/main/java/org/mozilla/focus/browser/cursor/CursorViewModel.kt
```diff
@@ -62,7 +62,7 @@ class CursorViewModel(
                 isInitialPosSet = true
                 pos.set(value.x / 2, value.y / 2) // Center.
             }
-            onUpdate(pos.x, pos.y, getScrollVel())
+            onUpdate(pos.x, pos.y, getPercentMaxScrollVel(), 1f)
```
liuche: Why is `onUpdate called from within `onUpdate`? or am I reading this wrong.

app/src/main/java/org/mozilla/focus/browser/cursor/CursorViewModel.kt
```diff
@@ -19,9 +19,10 @@ import java.util.EnumSet
 import java.util.concurrent.TimeUnit
 
 private const val UPDATE_DELAY_MILLIS = 20L
+private const val UPDATE_DELAY_MILLIS_F = UPDATE_DELAY_MILLIS.toFloat() // Avoid conversion in update loop.
 
-private const val FRICTION = 0.98f
-private const val MAX_SPEED = 25f
+private const val ACCEL_MODIFIER = 0.98f
```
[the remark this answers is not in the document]
mcomella: Renamed `DELTA_VEL_CHANGE_PER_FRAME` (since we decided this isn't really accel below).

app/src/main/java/org/mozilla/focus/browser/cursor/CursorViewModel.kt
mcomella: > nit: rename framesPassed too to match multiplier from earlier.

@liuche I don't understand, what do you mean?

mcomella: > What is the difference between onUpdate and updatePosForVel?

- `updatePosForVel` is a helper function mutates the current object to 
- `onUpdate` is a callback to inform listeners that the position has changed. `on*` is conventionally for callbacks like this in Java.

app/src/main/java/org/mozilla/focus/browser/cursor/CursorViewModel.kt
mcomella: `onUpdate` is called from the setter of `maxBounds`, not `onUpdate`.

mcomella:
@liuche Comments addressed and commit with fixes has been appended.

app/src/main/java/org/mozilla/focus/browser/cursor/CursorViewModel.kt
```diff
@@ -74,18 +74,25 @@ class CursorViewModel(
...
         // cursor will noticeably skip ahead as frames are dropped.
         //
-        // The way we express acceleration is not natural, but kept for legacy reasons.
+        // This adjustment could be expressed more naturally if this algorithm was expressed
+        // as a series of kinematic equations, i.e. vnew = vold + accel * deltaTime.
         val framesPassed = deltaMillis / UPDATE_DELAY_MILLIS_F
-        val accel = (vel + 1) * ACCEL_MODIFIER - vel
-        val timeAdjustedAccel = accel * framesPassed
+        val deltaVelPerFrame = getDeltaVelPerFrame()
+        val deltaVelForCurrentUpdate = deltaVelPerFrame * framesPassed
```
liuche: Inline.

app/src/main/java/org/mozilla/focus/browser/cursor/CursorViewModel.kt
```diff
@@ -21,7 +21,7 @@ import java.util.concurrent.TimeUnit
 private const val UPDATE_DELAY_MILLIS = 17L // ~60 FPS.
 private const val UPDATE_DELAY_MILLIS_F = UPDATE_DELAY_MILLIS.toFloat() // Avoid conversion in update loop.
 
-private const val ACCEL_MODIFIER = 0.98f
+private const val DELTA_VEL_CHANGE_PER_FRAME = 0.98f
```
liuche: This should still be accel, not vel - otherwise this would indicate that the cursor slows down over time.

app/src/main/java/org/mozilla/focus/browser/cursor/CursorViewModel.kt
```diff
@@ -74,18 +74,25 @@ class CursorViewModel(
     private var updateLoop: Deferred<Unit>? = null
 
     private fun update(deltaMillis: Long) {
+        /**
+         * Gets the amount velocity should change per frame. The algorithm was determined
+         * experimentally and does not adhere to kinematic math.
+         */
+        fun getDeltaVelPerFrame() = (vel + 1) * DELTA_VEL_CHANGE_PER_FRAME - vel
```
liuche: I still think it's a little weird that we have this constant for acceleration change over time, but don't use it properly...? anyway, it's fine, just weird.

app/src/main/java/org/mozilla/focus/browser/cursor/CursorViewModel.kt
mcomella: Going to switch it back to ACCEL_MODIFIER because `CHANGE_PER_FRAME` indicates this is the whole change per frame, as opposed to some constant that is used in an algorithm to change the velocity.

PR #686, opened 2018-03-15
Closes #671: Bundled home tile has no padding between icon and title …

### LINKED ISSUE #671: Bundled home tile has no padding between icon and title background
Looks like a regression from #598: the bundled home tiles got larger (I'm not sure if this was intentional).

device-2018-03-13-163038

@lime124 Can you confirm this is an issue?

bkmunar:
…background


bkmunar:
@mcomella I did show that to @lime124 and she said the size was right. I didn't show the highlighted state to her however but if I get rid of that margin the size will be smaller and won't be as big as what I originally copied from the spec.

bkmunar:
screen shot 2018-03-15 at 3 29 08 pm

lime124:
ya let's just keep it that size then so we don't have to futz with the highlight state. i don't see the point in putting a ton of effort if this may change in the upcoming releases. i don't think it makes it _that_ much easier to read. thanks!

PR #699, opened 2018-03-20
 Closes #537: Run detekt on taskcluster (and fix issues) 

### LINKED ISSUE #537: Run detekt on taskcluster (and fix issues)
`detekt` doesn't seem to run for pull requests or master merges. So far there are 35 smells reported.

bkmunar:
I deleted ComplexMethod, TooManyFunctions, ComplexCondition, ForbiddenComment, LongParameterList, ModifiedOrder, NestedBlockDepth from the detekt config because I think that these rules are not good standards/reqs to base our code on.


.taskcluster.yml
```diff
@@ -36,7 +36,7 @@ tasks:
           && git config advice.detachedHead false
           && git checkout {{event.head.sha}}
           && echo "--" > .adjust_token
-          && ./gradlew clean assemble lint checkstyle ktlint pmd test
+          && ./gradlew clean assemble lint checkstyle ktlint pmd test detektCheck
```
mcomella: nit: please run this before `test`: afaik, test uses the most CPU time so it'd be great if we can stop builds before it runs. Here and below.

mcomella: Is this using the config file?

bkmunar: yes, it is. detekt is configured to use the config file in gradle

app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt
```diff
@@ -38,6 +38,9 @@ enum class NavigationEvent {
     }
 }
 
+private const val NAVIGATION_BUTTON_ENABLED_ALPHA = 1.0f
```
mcomella: nit: constants should only be at the very top of the file, not between classes.

app/src/main/java/org/mozilla/focus/browser/BrowserFragment.kt
```diff
@@ -36,6 +36,7 @@ import org.mozilla.focus.iwebview.IWebViewLifecycleFragment
 import org.mozilla.focus.widget.InlineAutocompleteEditText
 
 private const val ARGUMENT_SESSION_UUID = "sessionUUID"
+private const val TOAST_Y_OFFSET = 200
```
mcomella: nit: nl above. These are unrelated values.

quality/detekt.yml
```diff
@@ -175,35 +176,16 @@ complexity:
   LongMethod:
     active: true
     threshold: 20
-  NestedBlockDepth:
-    active: true
-    threshold: 3
```
mcomella: I see what you're saying about the thresholds now. Can we make this larger so it isn't hard to fix? Maybe 4? It'd be nice if it automatically failed things we'd nit in review.

quality/detekt.yml
```diff
@@ -175,35 +176,16 @@ complexity:
...
-  LongParameterList:
-    active: true
-    threshold: 5
```
mcomella: We could also make this really high, like 10.

quality/detekt.yml
```diff
@@ -175,35 +176,16 @@ complexity:
...
   LargeClass:
     active: true
     threshold: 150
   ComplexInterface:
     active: false
     threshold: 10
     includeStaticDeclarations: false
-  ComplexMethod:
-    active: true
```
mcomella: nit: active: false instead, here and below? I think we should still keep all the checks in the config, just so it's easier to turn them on again

quality/detekt.yml
```diff
@@ -252,8 +228,6 @@ style:
   MethodNameEqualsClassName:
     active: false
     ignoreOverriddenFunction: true
-  ModifierOrder:
-    active: true
```
mcomella: I think we should leave this one on: while I don't care about all the modifier orders, I really want `private/public/etc.` on the outside.

quality/detekt.yml
```diff
@@ -271,7 +245,7 @@ style:
     active: true
   MaxLineLength:
     active: true
-    maxLineLength: 120
+    maxLineLength: 180
```
mcomella: nit: Would 140 be too much trouble? 180 seems pretty long.

PR #711, opened 2018-03-21
Issue #489: Update cursor/pointer

mcomella:
Waiting on NI from Tif in issue, #489.


mcomella:
@bkmunar FYI: this wasn't ready to land - I was waiting on UX review from Tif. In the future, I'll put the NI on the PR, rather than the issue, but if you have other suggestions for how we should fix this in the future, let me know.

bkmunar:
oh no sorry!!! Yeah I didn't know T_T

PR #715, opened 2018-03-22
Update search config

mkaply:


mkaply:
> I assume this was intentional?

Yes. Yahoo JP is a separate thing.

> fwiw, we're not currently using these files, except to load the default search engine – I wonder if we should remove/cleanup this code since it's not in use (and thus not maintained)...

So there's no search selection on FireTV? IF so, yeah, we should probably remove. How are you choosing the default engine?

Ideally, when were componentizing this stuff, we should build a search component that does the right thing and can be used/not used.

mcomella:
> So there's no search selection on FireTV? IF so, yeah, we should probably remove. How are you choosing the default engine?

We're loading the default engine through this code but there's no search selection screen at the moment, no.

Issue #706, opened 2018-03-21
Pressed state for cursor


mcomella:
*edit:* Updated assets with larger cursor: [Cursor2-assets.zip](https://github.com/mozilla-mobile/firefox-tv/files/1842947/Cursor2-assets.zip)

---

Follow-up to #489, which just changes the UI.

@lime124 I assume we want this in v2.1 if possible?

boek:
![screen recording 2018-03-19 at 03 34 pm](https://user-images.githubusercontent.com/1250545/37724647-7bb86886-2cee-11e8-84e4-4b2d9fae8203.gif)
![cursor-bg-active](https://user-images.githubusercontent.com/1250545/37724681-8ea4a202-2cee-11e8-8200-467ae888fde7.png)
![cursor-bg](https://user-images.githubusercontent.com/1250545/37724683-8ec1f0c8-2cee-11e8-8c4b-343549b860f5.png)
![cursor-fg-active](https://user-images.githubusercontent.com/1250545/37724684-8edfb982-2cee-11e8-8409-75af3736d51a.png)
![cursor-fg](https://user-images.githubusercontent.com/1250545/37724686-8f0b3af8-2cee-11e8-9b4f-f100c198d402.png)

lime124:
yes please!

PR #722, opened 2018-03-22
Closes #721: Use larger Cursor assets.

### LINKED ISSUE #721: UX review updated cursor: adjust the size
Assets from https://github.com/mozilla-mobile/firefox-tv/issues/721#issuecomment-375487382: 

[Cursor2-assets.zip](https://github.com/mozilla-mobile/firefox-tv/files/1839591/Cursor2-assets.zip)

---

It accidentally landed early in https://github.com/mozilla-mobile/firefox-tv/pull/711#pullrequestreview-105908701 and it's probably not worth backing out.

Here's the old conversation from the old issue, #489:

---

@lime124 The new asset has padding built into it, which means the cursor (constrained by the 45dp we specify) appears smaller:

device-2018-03-21-134601

device-2018-03-21-134503

Is that okay? If we want to make it bigger, we're going to need larger assets.

fwiw, I notice I tend to overshoot things more often with this cursor (though the hit area is smaller than the cursor so maybe that's fine).

---

From Tif:

One usability problem I'm seeing with the old, bigger cursor vs the new, smaller one is that on my TV (which has lots of overscan) I 'lose" the cursor when scroll up/down or side to side. The smaller one completely goes off screen, whereas some part of the cursor remains visible with the larger one.

Are other people experiencing that too?

To fix the overshot problem, can we make the target area bigger without changing the cursor size?

mcomella:
The previous new cursor (#489) was smaller than the existing cursor,
which was especially a problem for overscan, so this fixes that problem.


lime124:
Up close it felt way too big. But scooting my chair all the way back it feels appropriate . Looks good. Let's merge and get people to try it out in LAT. Does this cover press state too @mcomella ?

mcomella:
> Does this cover press state too @mcomella ?

No, that's #706, but it's straightforward so I think it can be done today.

Issue #712, opened 2018-03-21
Cursor can "over scroll" the view


mcomella:
STR:
- Open a page that has more than one page length (e.g. google video)
- Move to the top of the screen and continue moving up, as if scrolling
- Scroll down

Expected: scrolling immediately starts
Actual: there is a pause, then scrolling starts

I think this is a regression from #684, where we moved from `flingScroll` to `scrollBy`. I'm guessing `scrollBy` isn't constrained by the amount the WebView can scroll (i.e. 0 to page length).

mcomella:
I wrote up an analysis of what methods we can use to do this on StackOverflow: https://stackoverflow.com/a/49439945/2219998

mcomella:
I needed to verify my solution working with a shrinking page so I created a test page: http://mcomella.xyz/test/shrinking_page.html

PR #723, opened 2018-03-23
Issue #712: Clamp WebView scrolling so it doesn't "overscroll".

mcomella:


app/src/amazonWebview/java/org/mozilla/focus/webview/FirefoxAmazonWebView.kt
```diff
@@ -116,6 +116,21 @@ internal class FirefoxAmazonWebView(
         destroyDrawingCache()
         return outBitmap
     }
+
+    override fun scrollByClamped(vx: Int, vy: Int) {
+        // This is not a true clamp: it can only stop us from
+        // continuing to scroll if we've already overscrolled.
+        val scrollX = clampScroll(vx) { canScrollHorizontally(it) }
```
liuche: I'm a little confused by this kotlinism - what do the curly braces mean? Does this has to do with the function being inlined (?) so the `canScrollHorizontally` isn't passed in as an argument within the parens (like `scroll`), but is rather passed in through curly braces?

Is there another way to pass this into the function? e.g., what if the ordering of `scroll` and `canScroll` were reversed, how would this be written?

mcomella:
> I feel like this 0/scroll could also just be a ternary in scrollByClamped rather than calling clampScroll and adding another level of function, but that's fine with me.

My goal was to reduce code duplication: we do the same bit of logic twice and there a few different steps so it seemed simpler to call out to a function.

What is your concern with pushing it out to a function?

If your concern is performance, inlined functions can written right into the code – i.e. from a perf perspective, it's as if I'd just written the code directly there.

If your concern is readability/jumping, yeah, maybe it's less readable but I think it's arguable either way whether the code safety (via deduplication) or readability is more important.

app/src/amazonWebview/java/org/mozilla/focus/webview/FirefoxAmazonWebView.kt
mcomella: In Kotlin, if the last argument to a function is an anonymous function, you can move the anonymous function outside of the parenthesis. For example, these two invocations are equivalent:
```kotlin
val scrollX = clampScroll(vx) { canScrollHorizontally(it) }
val scrollX = clampScroll(vx, { canScrollHorizontally(it) })
```

---

> what if the ordering of scroll and canScroll were reversed, how would this be written?

If the anonymous function is not last, it must be passed inside the parenthesis:
```kotlin
val scrollX = clampScroll({canScrollHorizontally()it) }, vx)
```

> Does this has to do with the function being inlined (?)

No, inlining does not affect argument order. It primarily affects what your generated code looks like and [how `return` works](https://kotlinlang.org/docs/reference/inline-functions.html#non-local-returns).

---

The reason why Kotlin lets you move the anonymous function outside the parenthesis is to create Domain Specific Languages (or DSLs). For example, there could be an API to generate HTML with:
```kotlin
html {
    body {
        p { "Hey liuche!" }
    }
}
```

Instead of:
```kotlin
html.create(
    body.create(
        p.create("Hey liuche!")
    )
)
```

For sufficiently complex structures, it really helps to omit the redundancy.

In theory, we could also...
```kotlin
linearLayout {
    textView {
        withAttrs(
            layoutWidth = MATCH_PARENT,
            layoutHeight = MATCH_PARENT,
            text = "Hey liuche!"
        )
    }
}
```

PR #724, opened 2018-03-23
Closes #706: Pressed state for cursor

### LINKED ISSUE #706: Pressed state for cursor

bkmunar:


app/src/main/res/drawable/cursor_full_active.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
+   - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
+<layer-list xmlns:android="http://schemas.android.com/apk/res/android">
```
mcomella: nit: it'd be easier to review this patch if you didn't add these files in this commit. I had to skim ahead to ensure they're used appropriately, unchanged, etc. (or removed in the case of that last file).

But generally, good job splitting this up into separate commits! :) Do you know about interactive rebase? `git rebase -i <ancestor-commit>`. This is how you can move commits, squash them (e.g. like moving the unnecessary commit file in here), etc.

app/src/main/java/org/mozilla/focus/browser/cursor/CursorKeyDispatcher.kt
```diff
@@ -30,7 +30,7 @@ class CursorKeyDispatcher(
         val remoteKey = RemoteKey.fromKeyEvent(event)
         if (remoteKey == RemoteKey.CENTER ||
                 event.keyCode == KeyEvent.KEYCODE_ENTER) { // For keyboard and emulator use.
-            dispatchTouchEventOnCurrentPosition(event.action)
+            dispatchTouchEventOnCurrentPosition(event)
```
mcomella: Similarly, so I don't have to "review" these changes twice, it'd be ideal if they were in the first commit.

app/src/main/java/org/mozilla/focus/browser/cursor/CursorView.kt
```diff
@@ -36,6 +39,16 @@ class CursorView(context: Context, attrs: AttributeSet) : ImageView(context, att
         resetCountdown()
     }
 
+    fun updateCursorPressedState(event: KeyEvent) {
+        val remoteKey = RemoteKey.fromKeyEvent(event)
+        if (event.action == KeyEvent.ACTION_DOWN && (remoteKey == RemoteKey.CENTER
```
mcomella: nit: boolean operators should always be at the end of the line and you should generally group related statements. In this case:
```kotlin
        // Enter is for keyboard and emulator use.
        if (event.action == KeyEvent.ACTION_DOWN &&
                (remoteKey == RemoteKey.CENTER || event.keyCode == KeyEvent.KEYCODE_ENTER)) {
```
Assuming line length isn't too long.

app/src/main/java/org/mozilla/focus/browser/cursor/CursorController.kt
```diff
@@ -54,7 +54,10 @@ class CursorController(
             KeyEvent.ACTION_UP -> viewModel.onDirectionKeyUp(dir)
             else -> Unit
         }
-    }, dispatchTouchEventOnCurrentPosition = { action -> viewModel.onSelectKeyEvent(action) })
+    }, dispatchTouchEventOnCurrentPosition = { event ->
+        viewModel.onSelectKeyEvent(event.action)
+        view.updateCursorPressedState(event)
```
mcomella: nit: There's discrepancy here: one method says what has happened (onSelectKeyEvent) while the other says what it's going to do (updateCursorPressedState).

I'm not sure which is better – probably the latter, to help with code readability. I think we can leave it the way it is for now, but for future reference.

app/src/main/java/org/mozilla/focus/browser/cursor/CursorView.kt
```diff
@@ -36,6 +39,16 @@ class CursorView(context: Context, attrs: AttributeSet) : ImageView(context, att
...
+                        || event.keyCode == KeyEvent.KEYCODE_ENTER)) { // For keyboard and emulator use.
+            setImageResource(R.drawable.cursor_full_active)
+        } else {
```
mcomella: nit: You're assuming the argument passed in will either be ACTION_DOWN or ACTION_UP and no other key presses. However, if this code is refactored by another developer, they may not know about your assumption and the `else` case might start catching other actions!

You should use an explicit else case here, or restrict your arguments (either through types or assertion-like expressions) so that your assumptions are either removed or explicitly documented.

fwiw, if this method was called `onSelectKeyPressed`, it would be another way of documenting this assumption but I think maybe you're right that this is better called updateCursorPressedState. :)

app/src/main/java/org/mozilla/focus/browser/cursor/CursorView.kt
```diff
@@ -36,6 +39,18 @@ class CursorView(context: Context, attrs: AttributeSet) : ImageView(context, att
         resetCountdown()
     }
 
+    fun updateCursorPressedState(event: KeyEvent) {
+        val remoteKey = RemoteKey.fromKeyEvent(event)
+        // For keyboard and emulator use.
```
mcomella: nit: You need to explain that ENTER is for keyboard/emulator use, not CENTER.

mcomella: nit: Comments should generally always have whitespace above them. In this case, I would:
```kotlin
// For...
val remoteKey ==
if (remoteKey...
```

I personally rather group unrelated code under a comment (unless it's really confusing) than "double-space" the code, but I can see the argument for the other direction.

app/src/main/java/org/mozilla/focus/browser/cursor/CursorKeyDispatcher.kt
```diff
@@ -10,7 +10,11 @@ import org.mozilla.focus.utils.Direction
 import org.mozilla.focus.utils.RemoteKey
 
 /** Converts raw key events into high-level commands the view model can recognize. */
-class CursorKeyDispatcher(var isEnabled: Boolean, private val viewModel: CursorViewModel) {
+class CursorKeyDispatcher(
+        var isEnabled: Boolean,
+        private val onDirectionKey: (dir: Direction, action: Int) -> Unit,
+        private val dispatchTouchEventOnCurrentPosition: (event: KeyEvent) -> Unit
```
mcomella: nit: -> `onSelectKey`

PR #725, opened 2018-03-23
[NO MERGE] Double-tap to page up/down.

mcomella:
Pushing so people can pull down the code from the PR.


mcomella:
We don't need to keep this open.

PR #739, opened 2018-03-29
Fixes #713 - Handle back button once for overlay/keyboard.

### LINKED ISSUE #713: Pressing back on controller from overlay soft keyboard dismisses overlay
STR:
- go to page
- open overlay
- click url bar to enter text and raise soft keyboard
- hit back

Expected: soft keyboard dismissed
Actual: keyboard and overlay dismissed

I think the problem is that the soft keyboard is not intercepting the back pressed event, even though it handles it, so the back event also gets sent to the overlay, from which we dismiss the keyboard.

Unfortunately, it seems non-trivial to detect if a soft keyboard is raised, so this might be non-trivial to fix: Jeff tried https://github.com/mozilla-mobile/firefox-tv/issues/548#issuecomment-371977621 with little success and the internet seems to detect layout size changes, which doesn't work for us because the soft keyboard is an overlay.

One way to fix this might be to make the overlay part of the system back stack (e.g. launch it as fragment).

Workarounds:
- Close overlay with back and reopen with menu
- Click "previous" on the soft keyboard

liuche:
Handle menu-hiding only on keyDown of back button.

We have a lot of back handling code, and `onBackPressed` doesn't pass the keyUp/keyDown differentiation, so if we're using both `dispatchKeyEvent` to handle BACK_KEYPRESS and `onBackPressed`, they must both use keyDown.

### DIFF
--- a/app/src/main/java/org/mozilla/focus/browser/BrowserFragment.kt
+++ b/app/src/main/java/org/mozilla/focus/browser/BrowserFragment.kt
@@ -208,11 +208,11 @@ class BrowserFragment : IWebViewLifecycleFragment() {
     private fun handleSpecialKeyEvent(event: KeyEvent): Boolean {
         val keyCodeIsMenu = event.keyCode == KeyEvent.KEYCODE_MENU
         val keyCodeIsBack = event.keyCode == KeyEvent.KEYCODE_BACK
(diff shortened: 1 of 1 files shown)


liuche:
mcomella is busy with other things, so going to land with this r+.

PR #743, opened 2018-03-29
Revert "your rights" strings deletion

liuche:
Will need to do an export of strings again too after this lands.


mcomella:
screenshot_1522727661

I think this looks good, given what we had, but this page could use a redesign. :)

Issue #429, opened 2018-01-30
[April 30] Improve crash reporting by integrating Sentry


mcomella:
- [x] Add and init Sentry
- [x] Ensure crashes are caught (bg thread checks periodically, [according to docs](https://docs.sentry.io/clients/java/config/#buffering-events-to-disk))
- [x] Look at crash logs to make sure no sensitive data (urls, identifiers) are being collected
- [x] Decide how to include Sentry "secret" - manually on build? Integrate builds with TC (later)?
- [x] Handle ProGuard UUIDs
- [x] Include `release`? Or will this be handled automatically? *Happens automatically: uses versionName*
- [x] Distinguish between LAT and release? *This should be handled automatically with version number where LAT is `M.m.b.*`*
- [x] We need to decide if we need/want to include a toggle in settings to opt-in/out
- [x] Should we remove user UUID field?
- [x] Document crash reporting (a la [Firefox iOS Sentry wiki page](https://github.com/mozilla-mobile/firefox-ios/wiki/Crash-Reporting-with-Sentry)) and what types of data is being collected
- [x] Update external docs (SUMO)? *afaict, there are no specific docs for Firefox TV so there's nothing to update*
- [ ] Email a few notes on things we can do with Sentry (e.g. breadcrumbs)
- [ ] Releng checklist

---

Apparently we've used Sentry on iOS. Which ones respect user privacy?

Sdaswani:
@mcomella Sentry is approved by our Legal / Policy team so should be OK on the privacy front.

mcomella:
FYI – asking for @pocmo 's opinion via IRC.

mcomella:
A summary of Sebastian's opinion via email (I would say it verbatim but I'm not sure he wants me to):
- Sentry would be fastest to add without legal/privacy overhead: we already run our own instance, which doesn't send data to third parties
- It's used on iOS so Stefan may have an opinion
- Ideally, we should look at a data dump to ensure nothing critical like URLs are being collected
- After playing with a hosted version of Sentry, Sebastian felt it was okay – it only sends crashes and doesn't know session data so it can't give us the crash rates and in particular compare across channels (has Beta regressed from Release?)
- Sebastian was wondering if we could extract Fennec's crash reporter into a library and use our crash stats interface but hasn't looked into it yet

boek:
[...]

PR #745, opened 2018-03-30
I429 add sentry

mcomella:


mcomella:

How often do users crash? Why are they crashing? How can we fix it?

* Ensure product quality

Amazon store crash reporting which is built into the release process, but that's deprecated and will be removed in a few weeks. Otherwise, none are really available.

Somewhat, with Amazon crash reporting but again, this will be removed in a few weeks.

<table>
  <tr>
    <td>Tracking Bug #</td>
  </tr>
  <tr>
    <td>Crash reports</td>
    <td>Technical data, interaction data</td>
    <td>#429</td>
  </tr>
</table>

For a sample of what is collected, see [the crash on Sentry](https://sentry.prod.mozaws.net/operations/firefox-tv-debug/issues/3417804/) or the [raw JSON blob](https://gist.github.com/mcomella/50622aef817b40a20714b8550fb19991).


* I want to permanently monitor this data. (@bbinto)

All release users.

We have a "Send usage data" switch in preferences.

We will regularly look at the crash reports to make sure there are no disruptively high crashes and fix them if there are.

Internally, with the team, in order to fix crashes.

app/build.gradle
```diff
@@ -231,26 +231,39 @@ tasks.whenTaskAdded { task ->
...
+    def SENTRY_DEBUG_FILE_PATH = "${rootDir}/.sentry_dsn_debug"
+    def SENTRY_RELEASE_FILE_PATH = "${rootDir}/.sentry_dsn_release"
     def variantName = variant.getName()
 
-    print("Adjust token (" + variantName + "): ")
+    print("Sentry DSN (" + variantName + "): ")
+
+    def filePath
+    if (variantName.contains("Debug")) {
+        filePath = SENTRY_DEBUG_FILE_PATH
```
liuche: Why is there a Debug sentry path?

mcomella:
@liuche Data collection docs are up: https://github.com/mozilla-mobile/firefox-tv/wiki/Crash-reporting-with-Sentry

app/src/main/java/org/mozilla/focus/telemetry/SentryWrapper.kt
```diff
@@ -0,0 +1,41 @@
...
+ *
+ * With the current implementation, to enable Sentry on Beta/Release builds, add a
+ * <project-dir>/.sentry_dsn_release file with your key. To enable Sentry on Debug
+ * builds, add a .sentry_dsn_debug key and replace the [DataUploadPreference.isEnabled]
+ * value with true (upload is disabled by default in dev builds).
+ */
+object SentryWrapper {
+
+    fun init(context: Context) {
+        onIsEnabledChanged(context, DataUploadPreference.isEnabled(context))
```
liuche: What if this is changed later, this doesn't get re-inited right? I guess it's not a big deal if we miss a Sentry upload if someone decides to turn it on after the app starts, and this is simpler.

liuche:
1) Is there or will there be **documentation** that describes the schema for the ultimate data set available publicly, complete and accurate?

Yes, in Firefox TV Sentry wiki page: https://github.com/mozilla-mobile/firefox-tv/wiki/Crash-reporting-with-Sentry

2) Is there a control mechanism that allows the user to turn the data collection on and off?

Yes, part of Firefox TV data controls

3) If the request is for permanent data collection, is there someone who will monitor the data over time?**

Permanent

4) Using the **[category system of data types](https://wiki.mozilla.org/Firefox/Data_Collection)** on the Mozilla wiki, what collection type of data do the requested measurements fall under?  **

Type 1, Type 2

5) Is the data collection request for default-on or default-off?

default on

6) Does the instrumentation include the addition of **any *new* identifiers** (whether anonymous or otherwise; e.g., username, random IDs, etc.  See the appendix for more details)?

Yes, crash id but not linked to user

7) Is the data collection covered by the existing Firefox privacy notice? 

Yes, covered in Focus Privacy Policy

8) Does there need to be a check-in in the future to determine whether to renew the data? (Yes/No) (If yes, set a todo reminder or file a bug if appropriate)**
No, crash reporting core functionality

app/src/main/java/org/mozilla/focus/telemetry/SentryWrapper.kt
mcomella: We always call `Sentry.init` when the activity is created, even if it's disabled. We pass in an invalid client if it's disabled, which prevents upload (I need to verify this doesn't record events while it's disabled either).

Whenever the toggle changes, we call init with a new client, which should resume sentry again.

We don't need to call our own init more than once because it all ends up in the same place.

app/build.gradle
mcomella: There are two sentry endpoints/keys: one for debug and one for release. Debug, I assume, is for us to test to make sure Sentry is working correctly. Release is release.

I separated the file paths so that it's harder to ship the wrong key.

mcomella:
In my additional enabled/disabled testing, I did `Sentry.captureMessage(String)` and noticed it does not upload any Strings when Sentry disabled.

However, I tried throwing an exception. When:
- Sentry is just enabled and crashes, one crash report is left on disk.
- Sentry is disabled and crashes, no crash report is left on disk
- Sentry is disabled, crashes, is re-enabled, crashes, **two crash reports are left on disk**
- Sentry is disabled, crashes twice, is re-enabled, crashes, two crash reports are left on disk

However, I verified what is uploaded: it actually uploads the enabled crash report twice (i.e. they look identical). Maybe this is a side effect of calling `Sentry.init` twice, with different parameters? It seemed fine though. 🤔 In any case, I think it's unlikely someone will be enabling and disabling sentry many times in a single session, so I think this behavior is acceptable. I kind of wish we had some tests though so this doesn't regress in subsequent Sentry updates: I'll file a bug.

PR #786, opened 2018-04-07
Closes #785: Remove DOMFocusedElement globalOnFocusChangeListener in …

### LINKED ISSUE #785: Memory leak: fragments aren't being garbage-collected
STR:
- From home screen, open Settings (move to gear) and close settings (via back button) 3 times
- In Android profiler, force GC
- Get heap dump
- Search for "SettingsFragment"

Expected: 0 references to SettingsFragment
Actual: 3 references

mcomella:
…onStop.

ViewTreeObserver listeners outlive the view hierarchy so we need to be
sure to remove it when it's no longer needed.

This appears to fix the memory leak:
- In my STR, GC'ing once causes 2 instances to be retained instead of 3.
- If I GC again, there are 0 instances retained.
- LeakCanary doesn't warn me with this implementation.

### DIFF
--- a/app/src/gecko/java/org/mozilla/focus/iwebview/WebViewProvider.java
+++ b/app/src/gecko/java/org/mozilla/focus/iwebview/WebViewProvider.java
@@ -98,6 +98,14 @@ public class WebViewProvider {
         }
 
         @Override
+        public void onStart() {
+        }
+
+        @Override
+        public void onStop() {
+        }
+
+        @Override
         public void scrollByClamped(int vx, int vy) {
 
         }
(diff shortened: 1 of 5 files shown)


app/src/gecko/java/org/mozilla/focus/iwebview/WebViewProvider.java
```diff
@@ -97,6 +97,14 @@ public void onResume() {
 
         }
 
+        @Override
+        public void onStart() {
```
liuche: Why onStart/onStop, rather than onCreate/onDestroy? (is onDestroy not guaranteed to be called?)

mcomella: `onDestroy` isn't guaranteed to be called. Generally, it will be called when the app will continue (e.g. Activity or Fragment is being explicitly removed) but it's safer to call in onStop.

Then again, with my mindset, nothing would go in onDestroy so maybe I need to rethink that. 🤔

PR #792, opened 2018-04-10
Closes #507: don't create a new instance of WebView each time an intent is created

### LINKED ISSUE #507: New Instance of Webview Created Every Time?
This report came from Amazon:

[02/20/18 01:27 PM] Haldeman, Andy: Hi Susheel, this is a minor issue, but our team noticed that if you try to start Firefox via ADB, it eventually hangs.
[02/20/18 01:28 PM] Haldeman, Andy: Here is the command:
[02/20/18 01:28 PM] Haldeman, Andy: adb shell am start -a android.intent.action.VIEW -n 'org.mozilla.tv.firefox/org.mozilla.focus.activity.MainActivity' -d 'http://youtube.com/tv'
[02/20/18 01:28 PM] Haldeman, Andy: We discovered it is because a new instance of Web View is created each time.
[02/20/18 01:28 PM] Haldeman, Andy: Not a high priority fix, but wanted to pass it along to your engineers.

mcomella:
*edit:* I've tested it successfully: see below.

I think this is the fix ~~**but I need to do further testing**~~, tomorrow morning. Note:
- The extended commit summary on the final commit explains why this fixed the issue
- A lot of the commits are removing unused code that simplifies the code so that I can make the later refactors I do, i.e. not much actually changes in this patch
- This change uncovered a lot of cruft in the way we handle sessions (i.e. we can create more than one), the browser fragment and session (it was really hard to load a URL from outside the browserFragment), state restoration, and intent handling. I'll file some follow-up bugs.

---

Here's the behavior after the bug from the final commit summary:
    We used to create a new session every time we receive an intent, which
    would add the existing BrowserFragment/WebView to the back and stack
    causing redundant WebViews and resource usage.

    At the end of this patch series, when we receive an intent, the behavior is:
    - If BrowserFragment is visible, we re-use it
    - If BF is not visible, we'll create a new Session, which creates a new
    BrowserFragment. This could cause multiple WebViews if BF was in the
    back stack but this is unlikely with the current code (only when
    Settings is shown)
    - Since we create and destroy many BF/WebView instances (e.g. home is
    shown so we destroy and when a site is clicked, we recreate it), we're
    relying on GC to ensure there is only one WebView instance at a time.
    Curiously, force GC in the debugger does not seem to always remove inactive
    WebView instances: either there's a bug in force GC or someone is
    holding a reference to it and it only gets cleaned up after several GC.
    It would likely be most resource efficient (assuming WebView is resource
    intense even when it's inactive) to  re-use a single BrowserFragment/WebView.

    This change uncovered a lot of cruft in the way we handle sessions (i.e.
    we can create more than one), the browser fragment and session (it was
    really hard to load a URL from outside the browserFragment), state
    restoration, and intent handling.

Notably, this patch series won't always stop us from recreating the WebView, but it will ensure we should only have one (as long as GC behaves).

### DIFF
--- a/app/src/main/java/org/mozilla/focus/IntentValidator.kt
+++ b/app/src/main/java/org/mozilla/focus/IntentValidator.kt
@@ -1,0 +1,66 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.focus
+
+import android.content.Context
+import android.content.Intent
+import android.os.Bundle
+import android.text.TextUtils
+import org.mozilla.focus.session.Source
+import org.mozilla.focus.utils.SafeIntent
+import org.mozilla.focus.utils.UrlUtils
+
+typealias OnValidBrowserIntent = (url: String, source: Source) -> Unit
+
+/**
+ * A container for functions that parse Intents and notify the application of their validity.
+ *
+ * This class uses callbacks to notify the application in order to decouple the Intent parsing
+ * from the application logic, which also allows us to test the functionality.
+ *
+ * The functions in this class take [SafeIntent] in order to encourage the caller to use SafeIntents
+ * in their code.
+ */
+object IntentValidator {
+
+    fun validateOnCreate(context: Context, intent: SafeIntent, savedInstanceState: Bundle?, onValidBrowserIntent: OnValidBrowserIntent) {
+        if ((intent.getFlags() and Intent.FLAG_ACTIVITY_LAUNCHED_FROM_HISTORY) != 0) {
+            // This Intent was launched from history (recent apps). Android will redeliver the
+            // original Intent (which might be a VIEW intent). However if there's no active browsing
+            // session then we do not want to re-process the Intent and potentially re-open a website
+            // from a session that the user already "erased".
+            return
+        }
+
+        if (savedInstanceState != null) {
+            // We are restoring a previous session - No need to handle this Intent.
+            return
+        }
+
+        validate(context, intent, onValidBrowserIntent)
+    }
+
+    fun validate(context: Context, intent: SafeIntent, onValidBrowserIntent: OnValidBrowserIntent) {
+        val action = intent.action
+
+        if (Intent.ACTION_VIEW.equals(action)) {
+            val dataString = intent.getDataString()
+            if (TextUtils.isEmpty(dataString)) {
+                return // If there's no URL in the Intent then we can't create a session.
+            }
+
+            onValidBrowserIntent(dataString, Source.VIEW)
+        } else if (Intent.ACTION_SEND.equals(action)) {
+            val dataString = intent.getStringExtra(Intent.EXTRA_TEXT)
+            if (TextUtils.isEmpty(dataString)) {
+                return
+            }
+
+            val isSearch = !UrlUtils.isUrl(dataString)
+            val url = if (isSearch) UrlUtils.createSearchUrl(context, dataString) else dataString
+            onValidBrowserIntent(url, Source.SHARE)
+        }
+    }
+}
--- a/app/src/main/java/org/mozilla/focus/MainActivity.kt
+++ b/app/src/main/java/org/mozilla/focus/MainActivity.kt
@@ -76,7 +76,7 @@ class MainActivity : LocaleAwareAppCompatActivity(), OnUrlEnteredListener {
     }
 
     private fun onValidBrowserIntent(url: String, source: Source) {
-        sessionManager.createSession(source, url)
+        ScreenController.showBrowserScreenForUrl(supportFragmentManager, url, source)
     }
 
     override fun applyLocale() {
--- a/app/src/main/java/org/mozilla/focus/MainActivity.kt
+++ b/app/src/main/java/org/mozilla/focus/MainActivity.kt
@@ -134,8 +134,8 @@ class MainActivity : LocaleAwareAppCompatActivity(), OnUrlEnteredListener {
 
     override fun onNonTextInputUrlEntered(urlStr: String) {
         ViewUtils.hideKeyboard(container)
-        ScreenController.onUrlEnteredInner(this, supportFragmentManager, sessionManager, urlStr,
-                false, null, null)
+        ScreenController.onUrlEnteredInner(this, supportFragmentManager, urlStr, false,
+                null, null)
     }
 
     override fun onTextInputUrlEntered(urlStr: String,
@@ -143,8 +143,8 @@ class MainActivity : LocaleAwareAppCompatActivity(), OnUrlEnteredListener {
                                        inputLocation: UrlTextInputLocation?) {
         ViewUtils.hideKeyboard(container)
         // It'd be much cleaner/safer to do this with a kotlin callback.
-        ScreenController.onUrlEnteredInner(this, supportFragmentManager, sessionManager, urlStr,
-                true, autocompleteResult, inputLocation)
+        ScreenController.onUrlEnteredInner(this, supportFragmentManager, urlStr, true,
+                autocompleteResult, inputLocation)
     }
 
     override fun dispatchKeyEvent(event: KeyEvent): Boolean {
--- a/app/src/main/java/org/mozilla/focus/MainActivity.kt
+++ b/app/src/main/java/org/mozilla/focus/MainActivity.kt
@@ -16,12 +16,14 @@ import com.amazon.android.webkit.AmazonWebKitFactory
 import kotlinx.android.synthetic.main.activity_main.*
 import org.mozilla.focus.architecture.NonNullObserver
 import org.mozilla.focus.browser.BrowserFragment
+import org.mozilla.focus.ext.toSafeIntent
 import org.mozilla.focus.home.HomeFragment
 import org.mozilla.focus.iwebview.IWebView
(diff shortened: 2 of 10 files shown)


app/src/main/java/org/mozilla/focus/IntentValidator.kt
```diff
@@ -0,0 +1,66 @@
...
+ *
+ * This class uses callbacks to notify the application in order to decouple the Intent parsing
+ * from the application logic, which also allows us to test the functionality.
+ *
+ * The functions in this class take [SafeIntent] in order to encourage the caller to use SafeIntents
+ * in their code.
+ */
+object IntentValidator {
+
+    fun validateOnCreate(context: Context, intent: SafeIntent, savedInstanceState: Bundle?, onValidBrowserIntent: OnValidBrowserIntent) {
```
sputnikci: \[detekt\] INFO: \[style/MaxLineLength\] Line detected that is longer than the defined maximum line length in the code style\.

app/src/main/java/org/mozilla/focus/ScreenController.kt
```diff
@@ -23,10 +23,9 @@ object ScreenController {
      * Loads the given url. If isTextInput is true, there should be no null parameters.
      */
     fun onUrlEnteredInner(context: Context, fragmentManager: FragmentManager,
-                          sessionManager: SessionManager,
                           urlStr: String,
-                          isTextInput: Boolean, autocompleteResult: InlineAutocompleteEditText.AutocompleteResult?,
-                          inputLocation: UrlTextInputLocation?) {
+                          isTextInput: Boolean,
+                          autocompleteResult: InlineAutocompleteEditText.AutocompleteResult?, inputLocation: UrlTextInputLocation?) {
```
sputnikci: \[detekt\] INFO: \[style/MaxLineLength\] Line detected that is longer than the defined maximum line length in the code style\.

Sdaswani:
Thanks for the update @mcomella !

app/src/main/java/org/mozilla/focus/ScreenController.kt
```diff
@@ -57,11 +57,7 @@ object ScreenController {
                     .addToBackStack(null)
                     .commit()
         } else {
-            if (isSearch) {
-                SessionManager.getInstance().createSearchSession(Source.USER_ENTERED, updatedUrlStr, searchTerms)
-            } else {
-                SessionManager.getInstance().createSession(Source.USER_ENTERED, updatedUrlStr)
-            }
+            SessionManager.getInstance().createSession(Source.USER_ENTERED, updatedUrlStr)
```
liuche: Delete the variable `searchTerms` as well, from above. The `isUrl` check above could be made into a ternary.

app/src/main/java/org/mozilla/focus/ScreenController.kt
```diff
@@ -41,24 +41,9 @@ object ScreenController {
...
-            fragmentManager.beginTransaction()
-                    .replace(R.id.container, browserFragment)
-                    .addToBackStack(null)
-                    .commit()
-        } else {
-            SessionManager.getInstance().createSession(Source.USER_ENTERED, updatedUrlStr)
-        }
+        showBrowserScreenForUrl(fragmentManager, updatedUrlStr)
 
+        val isSearch = !TextUtils.isEmpty(searchTerms)
```
liuche: `isSearch` and `isUrl` are basically the same variable, maybe you can combine them.

app/src/main/java/org/mozilla/focus/ScreenController.kt
```diff
@@ -115,4 +100,24 @@ object ScreenController {
...
+        // - We only support one session but it creates a new session when there's no BrowserFragment
+        // such as each time we open a URL from the home screen.
+        // - It doesn't handle the case where the BrowserFragment is non-null but not
+        // visible: this can happen when a BrowserFragment is in the back stack, e.g. if this
+        // method is called from Settings.
+        //
+        // However, from a user perspective, the behavior is correct (e.g. back stack functions
+        // correctly with multiple sessions).
+        val browserFragment = fragmentManager.findFragmentByTag(BrowserFragment.FRAGMENT_TAG) as? BrowserFragment
+        if (browserFragment != null && browserFragment.isVisible) {
```
liuche: Hmm, okay - thanks for the detailed comment.

app/src/main/java/org/mozilla/focus/ext/Intent.kt
```diff
@@ -0,0 +1,10 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.focus.ext
+
+import android.content.Intent
+import org.mozilla.focus.utils.SafeIntent
+
+fun Intent.toSafeIntent() = SafeIntent(this)
```
liuche: Yum extensions.

app/src/main/java/org/mozilla/focus/ScreenController.kt
```diff
@@ -117,7 +117,7 @@ object ScreenController {
             // for visibility in addition to existence.
             browserFragment.loadUrl(url)
         } else {
-            SessionManager.getInstance().createSession(Source.USER_ENTERED, url)
+            SessionManager.getInstance().createSession(source, url)
```
liuche: ...what other sources are there?

app/src/main/java/org/mozilla/focus/MainActivity.kt
```diff
@@ -48,7 +50,7 @@ class MainActivity : LocaleAwareAppCompatActivity(), OnUrlEnteredListener {
 
         setContentView(R.layout.activity_main)
 
-        sessionManager.handleIntent(this, intent, savedInstanceState)
+        IntentValidator.validateOnCreate(this, intent, savedInstanceState, ::onValidBrowserIntent)
```
liuche: What is this :: notation?

app/src/main/java/org/mozilla/focus/IntentValidator.kt
```diff
@@ -0,0 +1,66 @@
...
+
+typealias OnValidBrowserIntent = (url: String, source: Source) -> Unit
+
+/**
+ * A container for functions that parse Intents and notify the application of their validity.
+ *
+ * This class uses callbacks to notify the application in order to decouple the Intent parsing
+ * from the application logic, which also allows us to test the functionality.
+ *
+ * The functions in this class take [SafeIntent] in order to encourage the caller to use SafeIntents
```
liuche: +1

app/src/main/java/org/mozilla/focus/MainActivity.kt
```diff
@@ -76,7 +76,7 @@ class MainActivity : LocaleAwareAppCompatActivity(), OnUrlEnteredListener {
     }
 
     private fun onValidBrowserIntent(url: String, source: Source) {
-        sessionManager.createSession(source, url)
+        ScreenController.showBrowserScreenForUrl(supportFragmentManager, url, source)
```
liuche: lol - I was expecting a super-long last commit, but this is pretty good (albeit anticlimactic).

app/src/main/java/org/mozilla/focus/ScreenController.kt
mcomella: We need `isUrl` to decide whether or not to send telemetry but I'll remove `searchTerms`.

app/src/main/java/org/mozilla/focus/ScreenController.kt
mcomella: VIEW, SHARE, from intents: it's in the next commit. I don't know what these are used for or if they're important, however...

app/src/main/java/org/mozilla/focus/MainActivity.kt
mcomella: `::` is slightly overloaded (sigh, I know, I know...) but generally indicates access to meta information like `MainActivity::class` instead of `MainActivity.class` in Java. `::` without a receiver indicates an implicit `this` receiver, i.e. `this::`.

In our particular case, `::` is used to provide a function reference, e.g. `this::aFunctionOfTheClass`. Taking in a function is the same if it's a lambda or if it's a named function, e.g. in this contrived example:
```kotlin
fun isLessThanTwo(value: Int) = value < 2

fun main() {
    val values = listOf(1, 2, 3, 4)

    // These are equivalent!
    val namedFunctionReference = values.filter(::isLessThanTwo)
    val anonymousFunctionReference = values.filter { it < 2 }
}
```

Named function references are useful for improving readability of complex expressions, e.g. `values.filter(::isPrime)`, or that you don't want to duplicate the code for (like `onValidBrowserIntent`, here).

app/src/main/java/org/mozilla/focus/IntentValidator.kt
mcomella: Sputnik isn't reading our detektConfig, unfortunately, but this doesn't fail locally.

app/src/main/java/org/mozilla/focus/ScreenController.kt
mcomella: Sputnik isn't reading our detektConfig, unfortunately, but this doesn't fail locally.

mcomella:
> Are these session/webview cruft/bugs relevant in Focus? I know they still use sessions, but do they also create many WebView instances?

With brief testing, it looks like they only have one WebView and one BrowserFragment: their behavior for switching between tabs (i.e. reloading the whole page) backs this up.

PR #796, opened 2018-04-10
Closes #768:  [Home] Combine layouts for home and overlay

### LINKED ISSUE #768: [Home] Combine layouts for home and overlay
Move the RecyclerView from HomeFragment to overlay to match mocks in #759

bkmunar:


app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt
```diff
@@ -4,23 +4,60 @@
...
+import org.mozilla.focus.home.HomeTile
+import org.mozilla.focus.home.HomeTilePlaceholderGenerator
+import org.mozilla.focus.home.HomeTileScreenshotStore
 import org.mozilla.focus.telemetry.TelemetryWrapper
+import org.mozilla.focus.utils.FormattedDomain
+import org.mozilla.focus.utils.OnUrlEnteredListener
 import org.mozilla.focus.utils.Settings
 import org.mozilla.focus.widget.InlineAutocompleteEditText
 
+/**
```
sputnikci: \[detekt\] INFO: \[comments/CommentOverPrivateProperty\] Private properties should be named such that they explain themselves even without a comment\.

app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt
```diff
@@ -52,6 +89,14 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
         fun isURLPinned(): Boolean
     }
 
+    private lateinit var onUrlEnteredListener: OnUrlEnteredListener // default impl does nothing.
+
+    /**
```
sputnikci: \[detekt\] INFO: \[comments/CommentOverPrivateProperty\] Private properties should be named such that they explain themselves even without a comment\.

app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt
```diff
@@ -4,23 +4,60 @@
...
+import android.support.v7.widget.RecyclerView
 import android.util.AttributeSet
+import android.util.TypedValue
 import android.view.LayoutInflater
 import android.view.View
+import android.view.ViewGroup
+import android.view.animation.DecelerateInterpolator
 import android.widget.LinearLayout
 import kotlinx.android.synthetic.main.browser_overlay.view.*
+import kotlinx.android.synthetic.main.home_tile.view.*
```
sputnikci: \[detekt\] INFO: \[style/WildcardImport\] Wildcard imports should be replaced with imports using fully qualified class names\. Wildcard imports can lead to naming conflicts\. A library update can introduce naming clashes with your classes which results in compilation errors\.

app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt
```diff
@@ -75,13 +120,41 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
...
         navCloseHint.compoundDrawablesRelative.forEach(tintDrawable)
         navUrlInput.compoundDrawablesRelative.forEach(tintDrawable)
+
+        tileContainer.addOnScrollListener(object : RecyclerView.OnScrollListener() {
+            override fun onScrolled(recyclerView: RecyclerView?, dx: Int, dy: Int) {
+                val gridLayoutManager = recyclerView?.layoutManager as GridLayoutManager
+                val lastVisibleItem = gridLayoutManager.findLastCompletelyVisibleItemPosition()
+                // We add a scroll offset, revealing the next row to hint that there are more home tiles
+                if (dy > 0 && getFocusedTilePosition() > lastVisibleItem) {
+                    val scrollOffset = TypedValue.applyDimension(
```
sputnikci: \[detekt\] INFO: \[style/MaxLineLength\] Line detected that is longer than the defined maximum line length in the code style\.

app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt
```diff
@@ -184,4 +261,166 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
...
+    }
+
+    override fun getItemCount() = tiles.size
+
+    override fun onCreateViewHolder(parent: ViewGroup, viewType: Int) = TileViewHolder(
+            LayoutInflater.from(parent.context).inflate(R.layout.home_tile, parent, false)
+    )
+}
+
+private fun onBindBundledHomeTile(holder: TileViewHolder, tile: BundledHomeTile) = with (holder) {
```
sputnikci: \[detekt\] INFO: \[style/MaxLineLength\] Line detected that is longer than the defined maximum line length in the code style\.

app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt
```diff
@@ -184,4 +261,166 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
...
+    val bitmap = BundledTilesManager.getInstance(itemView.context).loadImageFromPath(itemView.context, tile.imagePath)
+    iconView.setImageBitmap(bitmap)
+
+    titleView.text = tile.title
+}
+
+private fun onBindCustomHomeTile(uiLifecycleCancelJob: Job, holder: TileViewHolder, item: CustomHomeTile) = with (holder) {
+    launch(uiLifecycleCancelJob + UI, CoroutineStart.UNDISPATCHED) {
```
sputnikci: \[detekt\] INFO: \[style/MaxLineLength\] Line detected that is longer than the defined maximum line length in the code style\.

app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt
```diff
@@ -184,4 +261,166 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
...
+        val validUri = item.url.toJavaURI()
+
+        val screenshotDeferred = async {
```
sputnikci: \[detekt\] INFO: \[style/MaxLineLength\] Line detected that is longer than the defined maximum line length in the code style\.

app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt
```diff
@@ -184,4 +261,166 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
...
+            val homeTileCornerRadius = itemView.resources.getDimension(R.dimen.home_tile_corner_radius)
+            val homeTilePlaceholderCornerRadius = itemView.resources.getDimension(R.dimen.home_tile_placeholder_corner_radius)
```
sputnikci: \[detekt\] INFO: \[style/MaxLineLength\] Line detected that is longer than the defined maximum line length in the code style\.

app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt
```diff
@@ -184,4 +261,166 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
...
+            }
+        }
+
+        // We wait for both to complete so we can animate them together.
+        val screenshot = screenshotDeferred.await()
+        val title = titleDeferred.await()
+
+        // NB: Don't suspend after this point (i.e. between view updates like setImage)
+        // so we don't see intermediate view states.
+        // TODO: It'd be less error-prone to launch { /* bg work */ launch(UI) { /* UI work */ } }
```
sputnikci: \[detekt\] INFO: \[style/ForbiddenComment\] Flags a forbidden comment\. Defaults values are TODO:, FIXME: or STOPSHIP:

app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt
```diff
@@ -184,4 +261,166 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
...
+        iconView.setImageBitmap(screenshot)
+        titleView.text = title
+
+        // Animate to avoid pop-in due to thread hand-offs. TODO: animation is janky.
```
sputnikci: \[detekt\] INFO: \[style/ForbiddenComment\] Flags a forbidden comment\. Defaults values are TODO:, FIXME: or STOPSHIP:

app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt
```diff
@@ -184,4 +261,166 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
...
+        AnimatorSet().apply {
+            interpolator = CUSTOM_TILE_ICON_INTERPOLATOR
+            duration = CUSTOM_TILE_TO_SHOW_MILLIS
+
+            val iconAnim = ObjectAnimator.ofInt(iconView, "imageAlpha", 0, 255)
```
sputnikci: \[detekt\] INFO: \[style/MagicNumber\] Report magic numbers\. Magic number is a numeric literal that is not defined as a constant and hence it's unclear what the purpose of this number is\. It's better to declare such numbers as constants and give them a proper name\. By default, \-1, 0, 1, and 2 are not considered to be magic numbers\.

app/src/main/java/org/mozilla/focus/home/HomeFragment.kt
```diff
@@ -192,132 +166,4 @@ class HomeFragment : Fragment() {
...
-}
-
-private class TileViewHolder(
-        itemView: View
-) : RecyclerView.ViewHolder(itemView) {
-    val iconView = itemView.tile_icon
-    val titleView = itemView.tile_title
-    val menuButton = itemView.home_tile_menu_icon
-}
+}
```
sputnikci: \[detekt\] INFO: \[style/NewLineAtEndOfFile\] Checks whether files end with a line separator\.

app/src/main/java/org/mozilla/focus/browser/BrowserFragment.kt
```diff
@@ -121,16 +121,18 @@ class BrowserFragment : IWebViewLifecycleFragment() {
                 this@BrowserFragment.url?.let { url ->
                     when (value) {
                         NavigationEvent.VAL_CHECKED -> {
-                            CustomTilesManager.getInstance(context).pinSite(context, url,
-                                    webView?.takeScreenshot())
+                            browserOverlay.addPinnedSiteToTiles(CustomTilesManager.getInstance(context).pinSite(context, url,
```
sputnikci: \[detekt\] INFO: \[style/MaxLineLength\] Line detected that is longer than the defined maximum line length in the code style\.

app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt
```diff
@@ -75,13 +121,40 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
...
+        initTiles()
         setupUrlInput()
         turboButton.isChecked = Settings.getInstance(context).isBlockingEnabled
         navButtonSettings.setImageResource(R.drawable.ic_settings) // Must be set in code for SVG to work correctly.
 
         val tintDrawable: (Drawable?) -> Unit = { it?.setTint(ContextCompat.getColor(context, R.color.tv_white)) }
         navCloseHint.compoundDrawablesRelative.forEach(tintDrawable)
         navUrlInput.compoundDrawablesRelative.forEach(tintDrawable)
+
+        tileContainer.addOnScrollListener(object : RecyclerView.OnScrollListener() {
```
liuche: As mentioned earlier, move this whole recyclerview code into a different file - this is already nicely modular, no reason to glop it into a big file.

app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt
```diff
@@ -185,3 +263,131 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
...
+                setLayoutMarginParams(iconView, R.dimen.bundled_home_tile_margin_value)
+            }
+            is CustomHomeTile -> {
+                onBindCustomHomeTile(uiLifecycleCancelJob, holder, item)
+                setLayoutMarginParams(iconView, R.dimen.custom_home_tile_margin_value)
+            }
+        }.forceExhaustive
+
+        itemView.setOnClickListener {
+            onUrlEnteredListener.onNonTextInputUrlEntered(item.url)
```
liuche: Now that this is part of the overlay, you should use the overlay methods to "send" information to the browser, rather than doing it directly. So take a look at how the overlay does this.

app/src/main/java/org/mozilla/focus/browser/BrowserFragment.kt
```diff
@@ -127,10 +127,12 @@ class BrowserFragment : IWebViewLifecycleFragment() {
...
                             url.toUri()?.let {
-                                if (BundledTilesManager.getInstance(context).unpinSite(context, it)
-                                        || CustomTilesManager.getInstance(context).unpinSite(context, url)) {
-                                    showCenteredTopToast(context, R.string.notification_unpinned_site, 0, TOAST_Y_OFFSET)
+                                if (BundledTilesManager.getInstance(context).unpinSite(context, it)) {
+                                    browserOverlay.removePinnedSiteFromTiles(it.toString())
+                                } else if (CustomTilesManager.getInstance(context).unpinSite(context, url)) {
+                                    browserOverlay.removePinnedSiteFromTiles(url)
                                 }
+                                showCenteredTopToast(context, R.string.notification_unpinned_site, 0, TOAST_Y_OFFSET)
```
liuche: This isn't correct, the original code shows a toast if one of these cases is true.

app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt
```diff
@@ -236,7 +235,7 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
         }
     }
 
-    override fun onNonTextInputUrlEntered(urlStr: String) = with(navUrlInput) {
+    override fun onNonTextInputUrlEntered(urlStr: String) = with (navUrlInput) {
```
liuche: nit: extra space - does the linter catch this?

app/src/main/java/org/mozilla/focus/browser/BrowserFragment.kt
```diff
@@ -127,8 +127,9 @@ class BrowserFragment : IWebViewLifecycleFragment() {
                         }
                         NavigationEvent.VAL_UNCHECKED -> {
                             url.toUri()?.let {
-                                if (BundledTilesManager.getInstance(context).unpinSite(context, it)) {
-                                    browserOverlay.removePinnedSiteFromTiles(it.toString())
+                                val bundledTileId = BundledTilesManager.getInstance(context).unpinSite(context, it)
+                                if (bundledTileId != null) {
+                                    browserOverlay.removePinnedSiteFromTiles(bundledTileId)
```
liuche: This has some code smells here - it already more than doubles the previous code length, makes duplicate calls, and checks a bunch of cases. It also causes code complexity down the line (by making it handle two types of inputs) and also passes around `null` as an expected value that needs to be checked. (Using `null` to signal state is not a good idea). [You should read that Clean Code book :P ]

Can you think of a cleaner way to structure this code?

I'd encourage you not to "just write code that works" or "fix problems created in previous commits" - I think it's easy to do that, but if code looks messy, consider how you can clean it up.

app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt
liuche: Why do you need to have this in a separate function, rather than directly in the loadUrl interface? Can you explain why you did it this way?

app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt
```diff
@@ -329,6 +333,11 @@ class HomeTileAdapter(
         return null
     }
 
+    fun addItemToFrontOfAdapter(homeTile: HomeTile) {
+        tiles.add(0, homeTile)
```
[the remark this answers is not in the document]
bkmunar: I originally thought tiles were added to the front, but I guess they're not????

app/src/main/java/org/mozilla/focus/browser/HomeTileRecyclerView.kt
```diff
@@ -0,0 +1,184 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
liuche: omgomgty @bkmunar this is so much cleaner!!!!!! 😍

app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt
```diff
@@ -7,20 +7,31 @@ package org.mozilla.focus.browser
 import android.content.Context
```
liuche: Nice comment, and in the future, if you move a ton of code, that in and of itself should be a commit (even if it doesn't compile, but make a note in the commit message). That way, I don't need to re-review code that already is present and just being moved so I can hunt down the changes you made - I can just look at the subsequent commit.

app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt
```diff
@@ -75,13 +92,42 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
...
+    }
+
+    private fun initTiles() = with (tileContainer) {
+        val homeTiles = mutableListOf<HomeTile>().apply {
+            addAll(BundledTilesManager.getInstance(context).getBundledHomeTilesList())
+            addAll(CustomTilesManager.getInstance(context).getCustomHomeTilesList())
+        }
+
+        adapter = HomeTileAdapter(uiLifecycleCancelJob, homeTiles, loadUrl = { urlStr ->
+            onNonTextInputUrlEntered(urlStr)
```
liuche: I'm a little confused where onNonTextInputUrlEntered is, because I can't find it in the rest of the file.

app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt
```diff
@@ -54,7 +54,7 @@ enum class NavigationEvent {
 
 class BrowserNavigationOverlay @JvmOverloads constructor(
         context: Context, attrs: AttributeSet? = null, defStyle: Int = 0 )
-    : LinearLayout(context, attrs, defStyle), View.OnClickListener {
+    : LinearLayout(context, attrs, defStyle), View.OnClickListener, OnUrlEnteredListener {
```
liuche: Why do you need to implement OnUrlEnteredListener? This is just adding an unnecessary layer of interfaces.

app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt
liuche: ^why not directly into loadUrl interface?

app/src/main/java/org/mozilla/focus/browser/HomeTileRecyclerView.kt
```diff
@@ -0,0 +1,184 @@
...
+    fun getItemAtPosition(position: Int): HomeTile? {
+        if (position > -1 && position < itemCount) {
+            return tiles[position]
+        }
+        return null
+    }
+
+    fun addItemToAdapter(homeTile: HomeTile) {
+        tiles.add(homeTile)
+        notifyItemInserted(tiles.lastIndex)
```
liuche: Hmmm this still has the problem of hard-coding the insertion. What if UX changes how the insertion works? I'd rather have this be correct for everything, rather than just fixed for now.

Thinking about this from a MVC perspective (and happy to do this IRL if you want), there are two things we want:
* No assumptions about *where* the tile is being inserted
* Efficient insertion (e.g., not invalidating the whole list)

After we pin the site, we can get the new cached list of tiles, right? My thought would be, can we do something here like `updateAdapterSingleInsertion` where the argument is the new list of tiles? That way we can make this more general, and the method can handle any "single insertion" adapter changes without the assumption of "new tiles get added to the end" being hidden in the method call.

app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt
```diff
@@ -234,4 +234,23 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
             updateOverlayForCurrentState()
         }
     }
+
+    override fun onNonTextInputUrlEntered(urlStr: String) = with (navUrlInput) {
+        if (urlStr.isNotEmpty()) {
+            val cachedAutocompleteResult =
+                    lastAutocompleteResult // setText clears the reference so we cache it here.
+            setText(cachedAutocompleteResult.text)
+            onNavigationEvent?.invoke(NavigationEvent.LOAD, urlStr, cachedAutocompleteResult)
```
liuche: Looking through the NavigationEvent LOAD code here, it eventually calls into `MainActivity.onTextInputUrlEntered` which will send the wrong telemetry. (This code is kind of confusing, so I also filed #800 )

app/src/main/java/org/mozilla/focus/home/HomeTilesManager.kt
```diff
@@ -93,10 +93,10 @@ class BundledTilesManager private constructor(context: Context) {
                         .putStringSet(BUNDLED_SITES_ID_BLACKLIST, newBlacklist)
                         .apply()
                 bundledTilesCache.remove(pair.key)
-                return true
+                return pair.value.id
             }
         }
-        return false
+        return null
```
liuche: I still don't love passing null around as a type, but I guess this is a not-too-messy way to get the tile id on removal, and other Java `.remove` functions do it too.

app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt
```diff
@@ -117,14 +116,21 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
         })
     }
 
-    private fun initTiles() = with (tileContainer) {
-        val homeTiles = mutableListOf<HomeTile>().apply {
+    private fun initHomeTilesCache(): MutableList<HomeTile> {
```
liuche: Why do you need to return cache for initialization?

app/src/main/java/org/mozilla/focus/browser/BrowserFragment.kt
```diff
@@ -127,12 +127,14 @@ class BrowserFragment : IWebViewLifecycleFragment() {
...
                                     webView?.takeScreenshot())
+                            browserOverlay.addPinnedSiteToTiles()
                             showCenteredTopToast(context, R.string.notification_pinned_site, 0, TOAST_Y_OFFSET)
                         }
                         NavigationEvent.VAL_UNCHECKED -> {
                             url.toUri()?.let {
-                                if (BundledTilesManager.getInstance(context).unpinSite(context, it)
-                                        || CustomTilesManager.getInstance(context).unpinSite(context, url)) {
+                                val tileId = BundledTilesManager.getInstance(context).unpinSite(context, it) ?: CustomTilesManager.getInstance(context).unpinSite(context, url)
+                                if (!tileId.isNullOrEmpty()) {
```
liuche: Add a comment for why this tileId might be null or empty.

app/src/main/java/org/mozilla/focus/browser/BrowserFragment.kt
```diff
@@ -127,12 +127,14 @@ class BrowserFragment : IWebViewLifecycleFragment() {
...
+                                    browserOverlay.removePinnedSiteFromTiles(tileId!!)
```
liuche: Okay, red flag, `!!` is not okay. Try to use `!!` it as little as possible, basically, only when there isn't another way and you actually **want** it to crash if it's null. (e.g. you're in some really unexpected state, and you'd rather crash so you can get a crash report, or whatever)

In this case, you actually know it's not null, so (I learned by asking mcomella) if you do a null check, Kotlin will cast it for you. Since you want to also do the empty check, you should do something like
`if (tileId != null && !tileId.isEmpty())`

app/src/main/java/org/mozilla/focus/home/HomeTilesManager.kt
```diff
@@ -171,11 +171,11 @@ class CustomTilesManager private constructor(context: Context) {
     }
 
     @UiThread
-    fun unpinSite(context: Context, url: String): Boolean {
-        val tile = customTilesCache.remove(url) ?: return false
+    fun unpinSite(context: Context, url: String): String? {
```
liuche: Add a javadoc style comment here that says when you return a null, same as above.

app/src/main/java/org/mozilla/focus/browser/HomeTileRecyclerView.kt
```diff
@@ -97,9 +97,20 @@ class HomeTileAdapter(
         return null
     }
 
-    fun addItemToAdapter(homeTile: HomeTile) {
-        tiles.add(homeTile)
-        notifyItemInserted(tiles.lastIndex)
+    fun updateAdapterSingleInsertion(homeTiles: MutableList<HomeTile>) {
```
liuche: This is a lot of non-obvious code - add a javadoc function comment explaining what you're trying to do and **why** you're doing it.

E.g., think about the case where you're a new hire reading this code - it would be helpful to explain why we're doing this comparison, rather than having to read through the code and then realize that it's comparing item-by-item.

app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt
```diff
@@ -240,4 +240,14 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
             updateOverlayForCurrentState()
         }
     }
+
+    fun addPinnedSiteToTiles() {
+        val adapter = tileContainer.adapter as HomeTileAdapter
+        adapter.updateAdapterSingleInsertion(initHomeTilesCache())
```
liuche: Okay, red flag, you're calling `init` again - why?

We made a cache so that we don't have to load it from disk all the time - it's even called "cache".

app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt
```diff
@@ -240,4 +240,14 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
...
+    }
+
+    fun removePinnedSiteFromTiles(tileIdOrUrl: String) {
```
liuche: Why is this still `tileIdOrUrl`?

app/src/main/java/org/mozilla/focus/browser/HomeTileRecyclerView.kt
```diff
@@ -97,9 +97,20 @@ class HomeTileAdapter(
...
+        var i = 0
```
liuche: Hm, it looks like you're doing a lot of checks here rather than thinking about the problem, and this could be a little more efficient. (If this were an interview question, I'd 100% ask for this code to be more concise :P)

- why not use a `for` loop instead of `while`? that handles the incrementing for you
- how about returning if you find and notify the item inserted, rather than continuing to iterate through even after you've found the item?
- why do you need the last check on size?

app/src/main/java/org/mozilla/focus/browser/BrowserFragment.kt
```diff
@@ -127,12 +127,14 @@ class BrowserFragment : IWebViewLifecycleFragment() {
                         NavigationEvent.VAL_CHECKED -> {
                             CustomTilesManager.getInstance(context).pinSite(context, url,
                                     webView?.takeScreenshot())
+                            browserOverlay.addPinnedSiteToTiles()
```
liuche: Nit: I'd rename this to be `refreshTilesForInsertion`

liuche:
Going to wait for this to go green before landing.

PR #803, opened 2018-04-13
I764 pocket keys

mcomella:


mcomella:
Looks like I forgot to replace one `SENTRY_FIELD` with `fieldName`: I guess I never tested the case where I had no keys.

Issue #806, opened 2018-04-13
Upgrade gradle plugin to 3.1.1


mcomella:
Due to a bug in lint, this is needed to continue my work in #805.

PR #811, opened 2018-04-13
I805 lint baseline

mcomella:
Ignore #806 which must land first for this to work.


mcomella:
Broke on the errors in the gradle changes: I'm repushing.

PR #814, opened 2018-04-13
Closes #782:  [Home] Handle tile "removal" now that menu button is overloaded 

### LINKED ISSUE #782: [Home] Handle tile "removal" now that menu button is overloaded
With the overlay, the menu button will close the overlay, but that means we can't also use it for deleting tiles anymore.

Some options we discussed were:
* Remove "remove" - don't allow removing for this release
* Long-press "select" button to trigger remove context menu
* Add new "doorhanger" UI to add remove option

Whichever option we choose, we should also revisit it in a sebsequent release (maybe 3.1).

bkmunar:


liuche:
@bkmunar on a quick glance, this PR doesn't implement the toast hint that's part of the mocks, with the new string.

app/src/main/java/org/mozilla/focus/browser/BrowserFragment.kt
```diff
@@ -165,20 +171,54 @@ class BrowserFragment : IWebViewLifecycleFragment() {
...
             }
+            registerForContextMenu(browserOverlay.tileContainer)
         }
 
         layout.progressBar.initialize(this)
 
         return layout
     }
 
+    override fun onContextItemSelected(item: MenuItem): Boolean {
```
mcomella: nit: you can save an indent with:
```kotlin
override fun onContextItemSelected(item: MenuItem): Boolean = when (item.itemId) {
    R.id.remove -> ...
```

The return type isn't strictly necessary but it helps to be explicit (e.g. `when` might infer a `Any?` return type depending on the values you return) – I'm still going back and forth whether I prefer to declare the types or not

app/src/main/java/org/mozilla/focus/browser/BrowserFragment.kt
```diff
@@ -165,20 +171,54 @@ class BrowserFragment : IWebViewLifecycleFragment() {
...
+                    is CustomHomeTile -> {
+                        CustomTilesManager.getInstance(context).unpinSite(context, tileToRemove.url)
+                        homeTileAdapter.removeItemAtPosition(browserOverlay.getFocusedTilePosition())
+                    }
+                }
+                TelemetryWrapper.homeTileRemovedEvent(tileToRemove)
+                return true
+            }
+        }
+        return false
```
mcomella: You should try to make your `when` exhaustive, even if it has to have an else case (e.g. for ints like these):
```kotlin
return when (item.itemId) {
    R.id.remove -> {
        ...
        true
    }
    else -> false
}

File: app/src/main/java/org/mozilla/focus/browser/BrowserFragment.kt [hunk 3]
```diff
@@ -165,20 +171,54 @@ class BrowserFragment : IWebViewLifecycleFragment() {
...
+                val tileToRemove = homeTileAdapter.getItemAtPosition(browserOverlay.getFocusedTilePosition())
+                        ?: return false
+                // This assumes that since we're deleting from a Home Tile object that we created
+                // that the Uri is valid, so we do not do error handling here.
+                when (tileToRemove) {
+                    is BundledHomeTile -> {
+                        val tileUri = tileToRemove.url.toUri()
+                        if (tileUri != null) {
+                            BundledTilesManager.getInstance(context).unpinSite(context, tileUri)
+                            homeTileAdapter.removeItemAtPosition(browserOverlay.getFocusedTilePosition())
```
mcomella: Can we write a `HomeTilesManager.removeItem(focusTilePosition, tileUri)` method? It'd help simplify this method.

app/src/main/java/org/mozilla/focus/browser/BrowserFragment.kt
```diff
@@ -237,6 +277,13 @@ class BrowserFragment : IWebViewLifecycleFragment() {
             activity.dispatchKeyEvent(escKeyEvent)
             return true
         }
+
+        if (browserOverlay.isVisible && event.keyCode == KeyEvent.KEYCODE_DPAD_CENTER &&
+                event.action == KeyEvent.ACTION_DOWN
+                && event.repeatCount > 1) {
```
mcomella: Why can't we use `View.setOnLongClickListener` instead of this custom handling?

mcomella:
As Chenxia mentions, it'd be great to have the toast too.

app/src/main/java/org/mozilla/focus/browser/BrowserFragment.kt
bkmunar: We discussed that we should keep it this way since the inner nested stuff gets a bit complicated and we need to return early in some cases.

app/src/main/java/org/mozilla/focus/browser/BrowserFragment.kt
```diff
@@ -165,25 +170,51 @@ class BrowserFragment : IWebViewLifecycleFragment() {
...
             visibility = overlayVisibleCached ?: View.GONE
+            openHomeTileContextMenu = {
+                activity.openContextMenu(browserOverlay.tileContainer)
+            }
             onPreSetVisibilityListener = { isVisible ->
                 // The overlay can clear the DOM and a previous focused element cache (e.g. reload)
                 // so we need to do our own caching: see FocusedDOMElementCache for details.
                 if (!isVisible) { webView?.focusedDOMElement?.cache() }
             }
+            registerForContextMenu(browserOverlay.tileContainer)
```
mcomella: nit: group this together with `openHomeTileContextMenu`:
```kotlin
...
onPerSet...

openHomeTileContext = ...
registerForContext ...
```

The reason I think this is important is because this method is starting to get long. After 4 or 5 lines (since people can only keep [~4-7 things in their head at once](https://en.wikipedia.org/wiki/The_Magical_Number_Seven,_Plus_or_Minus_Two)), it's important to start grouping.

app/src/main/java/org/mozilla/focus/browser/BrowserFragment.kt
```diff
@@ -165,25 +170,51 @@ class BrowserFragment : IWebViewLifecycleFragment() {
...
         return layout
     }
 
+    override fun onContextItemSelected(item: MenuItem): Boolean {
+        when (item.itemId) {
+            R.id.remove -> {
+                val homeTileAdapter = tileContainer.adapter as HomeTileAdapter
+                val tileToRemove = homeTileAdapter.getItemAtPosition(browserOverlay.getFocusedTilePosition())
+                        ?: return false
+                // This assumes that since we're deleting from a Home Tile object that we created
```
mcomella: nit: nl above. You should generally have whitespace above `//` comments

app/src/main/java/org/mozilla/focus/browser/BrowserFragment.kt
```diff
@@ -238,6 +269,7 @@ class BrowserFragment : IWebViewLifecycleFragment() {
             activity.dispatchKeyEvent(escKeyEvent)
             return true
         }
+
```
mcomella: nit: unrelated change

app/src/main/java/org/mozilla/focus/browser/BrowserFragment.kt
```diff
@@ -165,25 +170,51 @@ class BrowserFragment : IWebViewLifecycleFragment() {
...
     override fun onDestroyView() {
         super.onDestroyView()
         lifecycle.removeObserver(cursor!!)
         cursor = null
         overlayVisibleCached = browserOverlay.visibility
     }
 
+    override fun onCreateContextMenu(menu: ContextMenu?, v: View?, menuInfo: ContextMenu.ContextMenuInfo?) {
+        super.onCreateContextMenu(menu, v, menuInfo)
```
mcomella: You don't want to call super: it will call `Activity.onCreateContextMenu`. Check the docs

app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt
```diff
@@ -68,6 +70,16 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
      */
     private var uiLifecycleCancelJob: Job
 
+    /**
+     * Setting the openContextMenu function in the HomeTileAdapter is fragile
+     * since we need to use Delegates.observable to update the function. However,
+     * the function will be updated by then because it's only used when a user
+     * needs to long press on a tile, after all the views have been created.
+     */
+    var openHomeTileContextMenu: (() -> Unit) by Delegates.observable(fun () { }) { _, _, newValue ->
```
mcomella: nit: type can be inferred:
```kotlin
var openHomeTileContextMenu by Delegates.observable({}) { _, _, newValue ->

File: app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt [hunk 10]
```diff
@@ -68,6 +70,16 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
      */
     private var uiLifecycleCancelJob: Job
 
+    /**
+     * Setting the openContextMenu function in the HomeTileAdapter is fragile
+     * since we need to use Delegates.observable to update the function. However,
```
mcomella: Why do we need to use Delegates.observable? Add to the comment.

nit: the comment's a little confusing too: "the function will be updated by then". When's then? Consider rewording it.

app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt
```diff
@@ -68,6 +70,16 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
      */
     private var uiLifecycleCancelJob: Job
 
+    /**
```
mcomella: nit: `/**` comments are used to describe to the caller what the function/variable does when they use it. `//` (repeated when multiple lines) is used to describe implementation details. In this case, we're describing the latter!

app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt
```diff
@@ -68,6 +70,16 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
...
+     * Setting the openContextMenu function in the HomeTileAdapter is fragile
+     * since we need to use Delegates.observable to update the function. However,
+     * the function will be updated by then because it's only used when a user
+     * needs to long press on a tile, after all the views have been created.
+     */
+    var openHomeTileContextMenu: (() -> Unit) by Delegates.observable(fun () { }) { _, _, newValue ->
+        setOpenContextMenuFunctionInAdapter(newValue)
```
mcomella: nit: call the code directly instead of creating a new function. Adding unnecessary functions pollutes auto-complete and makes it hard to figure out what a class does.

app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt
```diff
@@ -114,6 +126,8 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
                 }
             }
         })
+
+        Toast.makeText(context, R.string.remove_tile_toast_title, Toast.LENGTH_LONG).show()
```
mcomella: This is fine but just thoughts about the UX:
- Tif told me the spec was it shows up when we click tiles (but that seems wrong since it's too late)
- This also seems annoying to show every time, but better than ^

Let's land and see what people think.

app/src/main/java/org/mozilla/focus/browser/HomeTileRecyclerView.kt
```diff
@@ -44,7 +44,8 @@ private val CUSTOM_TILE_ICON_INTERPOLATOR = DecelerateInterpolator()
 class HomeTileAdapter(
```
mcomella: nit: File name does not represent what's in the file: can you fix the name (maybe in the last commit so it's easier to read the history)?

app/src/main/java/org/mozilla/focus/home/HomeTilesManager.kt
```diff
@@ -215,5 +215,19 @@ class HomeTilesManager {
...
+
+        fun removeHomeTile(homeTile: HomeTile, context: Context) {
+            when (homeTile) {
+                is BundledHomeTile -> {
+                    val tileUri = homeTile.url.toUri()
+                    if (tileUri != null) {
+                        BundledTilesManager.getInstance(context).unpinSite(context, tileUri)
+                    }
+                }
+                is CustomHomeTile -> {
```
mcomella: nit: can be one line without braces

app/src/main/res/values/strings.xml
```diff
@@ -211,4 +211,5 @@ be temporary, and you can try again later.</li>
     %1$s will be the title of the recommended video
     %2$s will be the source of the video (e.g. YouTube). -->
     <string name="pocket_video_feed_a11y_feed_item_focused">%1$s on %2$s</string>
+    <string name="remove_tile_toast_title">Press and hold SELECT to unpin from homescreen</string>
```
mcomella: nit: I prelanded landed this with our string export – can you use that string instead?

app/src/main/java/org/mozilla/focus/browser/HomeTileRecyclerView.kt
```diff
@@ -65,18 +66,21 @@ class HomeTileAdapter(
             TelemetryWrapper.homeTileClickEvent(it.context, item)
         }
 
+        itemView.setOnLongClickListener {
+            openContextMenu?.invoke()
```
mcomella: nit: `openContextMenu` -> `onTileLongClick`

To decouple a hierarchy of objects, the child shouldn't know what the parent is going to do in the callback so we name it generically. If we call it something like `openContextMenu`:
- The name needs to change if the behavior changes, or this child gets a new parent
- Having an `openContextMenu` may encourage someone to create another callback to have another action occur, e.g. `displayToast`, even if they both occur in the long click listener.

app/src/main/java/org/mozilla/focus/browser/HomeTileRecyclerView.kt
```diff
@@ -114,6 +118,13 @@ class HomeTileAdapter(
         notifyItemInserted(homeTiles.lastIndex)
     }
 
+    fun getItemAtPosition(position: Int): HomeTile? {
```
mcomella: nit: You can shorten this a tad and make it more readable:
```kotlin
fun getItemAtPosition(position: Int): HomeTile? = if (position > -1 && position < itemCount) {
    tiles[position]
} else {
    null
}
```

It may seem like a trivial difference but I find 1) keeping the file short can improve readability and 2) having the indented section display the return value makes it clear where the checks are and where the return values are.

Issue #766, opened 2018-04-05
[Home] Consolidate back button behavior and ensure back behavior is correct


liuche:
Now that we're handling showing the home/overlay in two places (as an overlay, and as a default home screen), we need to make sure the back button behavior isn't spread out everywhere (MainActivity, BrowserFragment, HomeFragment).

PR #815, opened 2018-04-14
Update MainActivity handling to not use the Homescreen

liuche:
This only changes the calling order from MainActivity, and removes HomeFragment - #768 does the UI changes.


liuche:
@mcomella I did not end up making any of the changes you were suggesting as follow-ups to #507. This ended up being cleaner than I expected, but let me know if you think it should integrate the follow-ups you suggested. We may also be able to do more work as part of #766 .

app/src/amazonWebview/java/org/mozilla/focus/webview/FocusWebViewClient.java
```diff
@@ -234,9 +234,10 @@ public boolean shouldOverrideUrlLoading(AmazonWebView view, String url) {
             return true;
         }
 
-        // Home screen loads a blank page and shows the overlay
+        // Home screen should show a blank webview. Since WebView doesn't (and shouldn't) have a
```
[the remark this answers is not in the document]
liuche: Homescreen isn't separate from BrowserFragment.

app/src/main/java/org/mozilla/focus/browser/BrowserFragment.kt
```diff
@@ -173,6 +173,13 @@ class BrowserFragment : IWebViewLifecycleFragment() {
         return layout
     }
 
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
```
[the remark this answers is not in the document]
liuche: I assume that Pocket Videos will just be loading a fragment anyway. This breaks history behavior if users expect "pocket to be a web page" and want to go "back" to it, but since back doesn't work on Youtube anyway (#733), I didn't handle that case here.

app/src/main/java/org/mozilla/focus/browser/BrowserFragment.kt
```diff
@@ -173,8 +173,12 @@ class BrowserFragment : IWebViewLifecycleFragment() {
         return layout
     }
 
+    private fun isHomepage(): Boolean {
```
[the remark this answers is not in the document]
liuche: Hmmm, I guess I'll address this "back behavior" in #766...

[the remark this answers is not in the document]
liuche: Well, I think this a being a little nitpicky - the current alternative is to spread out "url == HOME_URL" everywhere, which I think is worse. At least this is an easy single accessor, so if we want to change it, it's easy to find.

app/src/main/java/org/mozilla/focus/browser/BrowserFragment.kt
```diff
@@ -193,7 +197,6 @@ class BrowserFragment : IWebViewLifecycleFragment() {
                 webView?.goBack()
                 TelemetryWrapper.browserBackControllerEvent()
             }
-            browserOverlay.isVisible -> setOverlayVisibleByUser(false)
```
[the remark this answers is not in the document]
liuche: Hmmm yeah this is a little more complicated - short of hardcoding "if ishomepage && !hashistory", I'm not sure how else we should do this. Can handle this in #766 .

app/src/main/java/org/mozilla/focus/MainActivity.kt
```diff
@@ -54,12 +55,10 @@ class MainActivity : LocaleAwareAppCompatActivity(), OnUrlEnteredListener {
...
                 if (sessions.isEmpty()) {
-                    // There's no active session. Show the URL input screen so that the user can
-                    // start a new session.
-                    ScreenController.showHomeScreen(supportFragmentManager, this@MainActivity)
-                } else {
-                    ScreenController.showBrowserScreenForCurrentSession(supportFragmentManager, sessionManager)
+                    // There's no active session. Start a new session with "homepage".
+                    sessionManager.createSession(Source.USER_ENTERED, URL_HOME)
                 }
+                ScreenController.showBrowserScreenForCurrentSession(supportFragmentManager, sessionManager)
```
mcomella: nit: This is going to get called twice: once now and once when the session is created. I think you should do this instead:
```kotlin
if (sessions.isEmpty()) {
    ScreenController.showBrowserScreenForUrl(URL_HOME)
} else {
    ScreenController.showBrowserScreenForCurrentSession(...)
}

File: app/src/main/java/org/mozilla/focus/browser/BrowserFragment.kt [hunk 6]
```diff
@@ -45,7 +48,9 @@ private const val TOAST_Y_OFFSET = 200
 class BrowserFragment : IWebViewLifecycleFragment() {
     companion object {
         const val FRAGMENT_TAG = "browser"
-        const val URL_HOME = "firefox:home"
+        const val APP_URL_PREFIX = "firefox:"
+        const val HOME_SUFFIX = "home"
+        const val APP_URL_HOME = APP_URL_PREFIX + HOME_SUFFIX
```
mcomella: nit: "${APP_URL_PREFIX}home" and rm constant above (since it's only used once).

app/src/main/java/org/mozilla/focus/browser/BrowserFragment.kt
```diff
@@ -288,6 +287,13 @@ private class BrowserIWebViewCallback(
     override fun onBlockingStateChanged(isBlockingEnabled: Boolean) {}
 
     override fun onLongPress(hitTarget: IWebView.HitTarget) {}
+    override fun shouldInterceptRequest(url: String) {
+        launch(UI) {
```
mcomella: nit: make a comment that this may not be called from the UI thread and that's why we launch

app/src/amazonWebview/java/org/mozilla/focus/webview/FirefoxAmazonWebView.kt
```diff
@@ -112,6 +112,7 @@ internal class FirefoxAmazonWebView(
         // called by webview when clicking on a link, and not when opening a new page for the
         // first time using loadUrl().
         if (!client.shouldOverrideUrlLoading(this, url)) {
+            callback?.shouldInterceptRequest(url)
```
mcomella: This gets called twice: here and in shouldInterceptRequest. I don't think this call is necessary but if you don't, the screen flashes when `loadUrl` is initially called: I think this is because `shouldInterceptRequest` gets called from a background thread so the webView has time to blank (white) before it displays the overlay while calling `shouldInterceptRequest` from `loadUrl` is on the UI thread and happens ASAP. One hacky way to fix this is to set the default color of the WebView.

I think we should try to remove the redundant call but I'm not sure the best way to fix the flash-in issue.

app/src/amazonWebview/java/org/mozilla/focus/webview/FocusWebViewClient.java
```diff
@@ -111,9 +114,24 @@ public void onLoadResource(AmazonWebView view, String url) {
...
+                    callback.shouldInterceptRequest(request);
+                    final ByteArrayInputStream homeDataStream = makeEmptyDataStream();
+                    return new AmazonWebResourceResponse("text/html", "utf-8", homeDataStream);
+            }
+        }
+
         return super.shouldInterceptRequest(view, request);
     }
 
+    private ByteArrayInputStream makeEmptyDataStream() {
```
mcomella: nit: -> `getBlankPageStream` or `getBlankPageInputStream()`

app/src/amazonWebview/java/org/mozilla/focus/webview/FocusWebViewClient.java
```diff
@@ -111,9 +114,24 @@ public void onLoadResource(AmazonWebView view, String url) {
     // WebResourceRequest was added in API21 and there is no equivalent AmazonWebResourceRequest
     @Override
     public AmazonWebResourceResponse shouldInterceptRequest(AmazonWebView view, String request) {
+        if ((request != null && request.startsWith(APP_URL_PREFIX))) {
+            switch (request) {
+                case APP_URL_HOME:
+                    // Home screen should show a blank webview behind the overlay, but keep the url.
+                    callback.shouldInterceptRequest(request);
```
mcomella: nit: -> `onShouldInterceptRequest`. If you don't expect it to replace the method, you should make it clear this is a callback.

app/src/amazonWebview/java/org/mozilla/focus/webview/FocusWebViewClient.java
```diff
@@ -111,9 +114,24 @@ public void onLoadResource(AmazonWebView view, String url) {
     // WebResourceRequest was added in API21 and there is no equivalent AmazonWebResourceRequest
     @Override
     public AmazonWebResourceResponse shouldInterceptRequest(AmazonWebView view, String request) {
+        if ((request != null && request.startsWith(APP_URL_PREFIX))) {
```
mcomella: nit: This if has an extra set of parenthesis.

app/src/amazonWebview/java/org/mozilla/focus/webview/FocusWebViewClient.java
```diff
@@ -109,9 +114,24 @@ public void onLoadResource(AmazonWebView view, String url) {
...
+            switch (request) {
+                case APP_URL_HOME:
+                    // Home screen should show a blank webview behind the overlay, but keep the url.
+                    callback.shouldInterceptRequest(request);
+                    final ByteArrayInputStream homeDataStream = makeEmptyDataStream();
+                    return new AmazonWebResourceResponse("text/html", "utf-8", homeDataStream);
```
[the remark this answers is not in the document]
liuche: I assumed that this was sufficient:
// Home screen should show a blank webview behind the overlay, but keep the url.

app/src/amazonWebview/java/org/mozilla/focus/webview/FirefoxAmazonWebView.kt
liuche: Hmm, okay - I'll file a follow-up for this, but take it out for now.

liuche:
TC errors, but we need to do a string export to get rid of localized strings that were removed.

PR #832, opened 2018-04-18
Closes #831: Use new versionCodeOverride API to fix dynamic versionCo…

### LINKED ISSUE #831: versionCode override is broken with latest android gradle plugin
Updating the gradle plugin broke our versionCode override code: see https://github.com/mozilla-mobile/focus-android/pull/2371 for more details and a solution.

mcomella:
…des.

This broke when we updated the Android gradle plugin. This code is
inspired by Focus: https://github.com/mozilla-mobile/focus-android/pull/2371

I confirmed this worked by looking at the contents of the
app/build/outputs/apk/amazonWebview/release/output.json file to verify
the versionCode was no longer 11.

---

We'll need to land this on `releases/v2.1` too.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -206,7 +206,9 @@ android.applicationVariants.all { variant ->
     def isBetaOrRelease = buildType == "release" || buildType == "beta"
 
     if (isBetaOrRelease) {
-        variant.mergedFlavor.versionCode = generatedVersionCode
+        variant.outputs.all { output ->
+            setVersionCodeOverride(generatedVersionCode)
+        }
     }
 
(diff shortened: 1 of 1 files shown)


liuche:
lgtm esp since it works for Focus.

Found the docs, if you were interested in all the other things you can do with the new `variants`:
https://developer.android.com/studio/build/gradle-plugin-3-0-0-migration.html#manifestOutputFile

mcomella:
I'll wait on TC, just to be safe...

mcomella:
TC is r+ and travis is redundant: landing.

mcomella:
Now all green: I should just wait a minute, huh?

PR #843, opened 2018-04-23
Closes #770: [Home] Update overlay buttons to reflect current navigat…

### LINKED ISSUE #770: [Home] Update overlay buttons to reflect current navigation state
Reflect the "home" state without a browser stack in the overlay buttons.

Back, forward, refresh, pin should all be disabled and un-activated.

This should also fix the problems with the navigation buttons incorrectly gaining focus when moving through tiles (introduced in #768).

bkmunar:
…ion state

### DIFF
--- a/app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt
+++ b/app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt
@@ -13,6 +13,7 @@ import android.util.AttributeSet
 import android.util.TypedValue
 import android.view.LayoutInflater
 import android.view.View
+import android.widget.ImageButton
 import android.widget.LinearLayout
 import kotlinx.android.synthetic.main.browser_overlay.view.*
 import kotlinx.android.synthetic.main.home_tile.view.*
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt
```diff
@@ -179,7 +181,14 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
         navButtonForward.isFocusable = canGoForward
         navButtonForward.alpha = if (canGoForward) NAVIGATION_BUTTON_ENABLED_ALPHA else NAVIGATION_BUTTON_DISABLED_ALPHA
 
+        val isPinEnabled = navigationStateProvider?.isPinEnabled() ?: false
+        pinButton.isEnabled = isPinEnabled
```
mcomella: nit: since we repeat this code so much, can we add a helper function?

app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt
```diff
@@ -179,7 +181,14 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
...
         pinButton.isChecked = navigationStateProvider?.isURLPinned() ?: false
+        pinButton.alpha = if (isPinEnabled) NAVIGATION_BUTTON_ENABLED_ALPHA else NAVIGATION_BUTTON_DISABLED_ALPHA
```
mcomella: nit: Do we need `isFocusable` too, like the other buttons? Another reason we want helper functions. :)

PR #847, opened 2018-04-23
I845 fix lint

mcomella:


mcomella:
Lint passes locally, as well as pre-push hook, so I'm just going to land this.

PR #858, opened 2018-04-26
No issue: string import 2018/04/25

mcomella:
Dunno that this needs review but here you go!


liuche:
@mcomella maybe as part of this you can also remove the "unused strings" check.

mcomella:
> maybe as part of this you can also remove the "unused strings" check.

I already have, this looks like a different failure:
```
A problem occurred configuring root project 'firefox-tv'.
> Could not resolve all files for configuration ':classpath'.
   > Could not download bcpkix-jdk15on.jar (org.bouncycastle:bcpkix-jdk15on:1.56)
      > Could not get resource 'https://plugins.gradle.org/m2/org/bouncycastle/bcpkix-jdk15on/1.56/bcpkix-jdk15on-1.56.jar'.
         > Could not HEAD 'https://plugins.gradle.org/m2/org/bouncycastle/bcpkix-jdk15on/1.56/bcpkix-jdk15on-1.56.jar'. Received status code 504 from server: Gateway Time-out
```

Looks like an intermittent error with internet connection: I'll try re-running.

PR #860, opened 2018-04-26
Closes #825: Combined overlay doesnt cancel lifecycle jobs

### LINKED ISSUE #825: Combined overlay doesn't cancel lifecycle jobs
Regression from merging home screen and navigationOverlay: #768.

It creates a Job() in `init` but never cancels them. The instantiation of the job and calling `cancel` on it should be tied to the lifecycle of the views/owner: e.g. for Fragments, `onCreateView` and `onDestroyView`. I'm not sure what we should do about `uiLifecycleCancelJob` now being owned by a View.

If we don't do this, we make leak view hierarchies and contexts.

bkmunar:


app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt
```diff
@@ -66,7 +66,7 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
      * Used to cancel background->UI threads: we attach them as children to this job
      * and cancel this job at the end of the UI lifecycle, cancelling the children.
      */
-    private var uiLifecycleCancelJob: Job
+    var uiLifecycleCancelJob: Job
```
mcomella: You've got the code in place where it works – great! But is it maintainable?

In general, each abstraction layer should manage anything it allocates and shouldn't pass it to its children or parents.

In this specific case, should BrowserFragment be controlling the lifecycle of a member of the BrowserNavigationOverlay's? Instead, I think BrowserFragment should own the uiLifecycleCancelJob (I'll let you figure out what I mean here).

A few concrete examples why this is less maintainable:
- What if BrowserFragment has another child that requires a uiLifecycleCancelJob? What do we have to do to make that work?
- What if BrowserNavigationOverlay, a View, is re-used in another Fragment or Activity? Will we remember to reach inside it to cancel the lifecycle job?

mcomella:
Bryan and I spoke about this in person: since we start the async jobs in View.init and Android is inflating the view for us, there's no good way to pass in the uiLifecycleJob. We could consider other solutions (separating model from view seems the most probable) but it'll add complexity that I don't think is probably worth it. Let's roll with this.

mcomella:
@bkmunar Can you add a comment explaining this is not ideal (maybe where we cancel) but other solutions would add complexity?

Issue #765, opened 2018-04-05
Download and cache Pocket video data


mcomella:
Questions:
- How often should this list be refreshed? Determined by UX or from the API cache header?
- Will we actually cache or just download in real-time? How should we animate if we don't cache?

PR #863, opened 2018-04-26
Issue #765: Fetch Pocket data, cache it, and add to basic placeholder data

mcomella:
@bkmunar This review may introduce a lot of new concepts for you (e.g. coroutines) so let me know if you have questions! This PR essentially gets Pocket data from the network, caches it, and shoves it into the UI. The structure is relatively simple so I think you should be concerned about:
- Do we handle (network) errors appropriately?
- Does the cache work correctly? Are there any situations it may not get updated correctly?
- Are there any concurrency bugs?

I think this PR did a decent job of separating out commits: it adds one concept/file at a time and then builds upon it in the next commit. I recommend reviewing by reading the commits in order and seeing what you like (or not!) about this and adapt it into your own PRs!

Let me know if you have questions!


mcomella:
Tests pass locally.

app/src/main/java/org/mozilla/focus/MainActivity.kt
```diff
@@ -93,8 +94,14 @@ class MainActivity : LocaleAwareAppCompatActivity(), OnUrlEnteredListener {
...
     }
 
+    override fun onStart() {
+        super.onStart()
+        Pocket.startBackgroundUpdates()
+    }
+
     override fun onStop() {
         super.onStop()
+        Pocket.stopBackgroundUpdates() // Don't regularly hit the network in the background.
```
liuche: In general since this is a tv with dedicated network whose primary use case is watching video content, I think background network is probably okay.

app/src/main/java/org/mozilla/focus/home/pocket/Pocket.kt
```diff
@@ -0,0 +1,121 @@
...
+     * The videos are updated asynchronously so calling this twice in succession may return
+     * different lists.
+     *
+     * This method assumes [startBackgroundUpdates] has been called before being accessed.
+     */
+    @AnyThread // videosCache is synchronized.
+    fun getRecommendedVideos() = videosCache
+    private fun getRecommendedVideosNoCacheAsync() = async { PocketEndpoint.getRecommendedVideos() }
+
+    private fun startBackgroundUpdatesInner() = launch {
```
liuche: This seems fine, launching a coroutine and using `cancel` to stop it when we're backgrounded.

Do you need any error handling, exception catching?

mcomella: afaik, I caught all the exceptions and return null in those cases so no, I think we've handled all the exception handling. That being said, I could have been more thorough and wrote tests for these error cases to ensure we actually return null for all of them.

Issue #767, opened 2018-04-05
[Pocket] Integrate Pocket data with feed


mcomella:
Pocket data from #765 and feed from #761.

mcomella:
![screenshot_1524859552](https://user-images.githubusercontent.com/759372/39382851-d91cf1fa-4a1b-11e8-9f10-3615cf15e282.png)

@aminalhazwani Ta-dah! There are two stages of loading: the one you have in the mocks and the other when we have the text, but not the images yet. For this second, I use photonGrey50, which is the same color as the thumbnails in the first loading stage.

You should be able to try this out in my PR (you need to select the middle tile and go up due to the way I laid it out.

PR #867, opened 2018-04-27
Issue #767: Add Pocket data into Pocket feed

mcomella:
Ignore #765.


mcomella:
FYI: I'm going to get around the CI errors by not packaging the Picasso support library deps, which can sometimes cause runtime errors. Seems to work locally for now. We should package them again when we upgrade the SDK, #883?

Issue #774, opened 2018-04-05
[Pocket] Add Pocket video megatile


mcomella:
(recyclerView span, selection state, menu, visual design refinements?)

mcomella:
We could maybe use the [SpannedGridLayoutManager](https://github.com/Arasthel/SpannedGridLayoutManager), discovered in this SO post: https://stackoverflow.com/a/39484534

liuche:
We talked about making *not* part of the GridLayout and having a separate section so it's easier.

mcomella:
Here's my WIP so far:
![device-2018-04-27-174429](https://user-images.githubusercontent.com/759372/39390082-a6e76bec-4a44-11e8-8bdb-4ec5ed6f2635.png)

It's in my branch: https://github.com/mcomella/firefox-tv/tree/i774-megatile

mcomella:
After my first, PR here are the remaining issues:

- The spacing is mostly correct but it seems that there's too much vertical margin in the megatile (are images scaling correctly?), which might cause the cut-off on the bottom
- Round corners of thumbnails
- Focused state of megatile
- Colors of the rest of the overlay (#818).

wrt 1), we have to crop or skew the image to get it to match specs so I'm talking to Amin about resolving that. I took care of round corners and I'm looking at focused state now.

mcomella:
After my second PR:

After this PR series, we still need to resolve any UX nits Amin may have
and figure out how to size the images (they're not to spec because their
aspect ratios are different than the specs for the views: e.g. should we crop?)

mcomella:
Amin's last bit of feedback and the last thing to do for this bug:
>> In my opinion, this looks better uncropped but there's too much vertical padding. I think we should shrink the megatile and not crop the images. What do you think?
>
> I agree, thanks for the suggestion. Let's not crop the thumbs, and with this decision the pocket video thumb are 150dp x 84dp making the pocket megatile 870dp x 108dp height, these changes do not affect the padding of the megatile which remains 12dp. I am going to update and upload new mocks tomorrow!

PR #877, opened 2018-05-01
Issue #774: Polish the megatile

mcomella:
Builds on the previous commits for #774.

    After this PR series, we still need to resolve any UX nits Amin may have
    and figure out how to size the images (they're not to spec because their
    aspect ratios are different than the specs for the views: e.g. should we crop?)


app/src/main/java/org/mozilla/focus/home/pocket/PocketVideoMegaTile.kt
```diff
@@ -23,12 +24,17 @@ class PocketVideoMegaTile(
     var pocketVideos by Delegates.observable<List<PocketVideo>?>(null) { _, _, newVideos ->
         if (newVideos == null) return@observable
         thumbnailViews.forEachIndexed { i, thumbnailView ->
-            PicassoWrapper.client.load(newVideos[i].thumbnailURL).into(thumbnailView)
+            PicassoWrapper.client.load(newVideos[i].thumbnailURL)
+                    .transform(roundCornerTransformation)
```
bkmunar: got it, that's smart

PR #891, opened 2018-05-04
Closes #887: Scroll the whole overlay instead of RecyclerView

### LINKED ISSUE #887: Scroll whole overlay when scrolling tiles
Right now, we only scroll the tiles with the fixed elements (ignore WIP spacing):
device-2018-04-30-143125

Amin thinks we should scroll the whole home overlay:
> > How should the overlay scroll when selecting multiple tiles?
>
> If possible I would make everything scroll, no fixed elements. With the d-pad I don't see any real advantage of having fixed elements, or am I missing something @mcomella?

For this, you should:
- disable scrolling in the tile RecyclerView
- Enable scrolling on the whole overlay (with ScrollView?)
- Make sure the whole tiles are visible, like we had before the merge
- Remove any "whole-tile visible" code ^ we don't use now

mcomella:
screenshot_1525394224

*edit: I added 4dp of padding over this screenshot*

### DIFF
--- a/app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt
+++ b/app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt
@@ -11,7 +11,6 @@ import android.support.v4.content.ContextCompat
 import android.support.v7.widget.GridLayoutManager
 import android.support.v7.widget.RecyclerView
 import android.util.AttributeSet
-import android.util.TypedValue
 import android.view.LayoutInflater
 import android.view.View
 import android.widget.ImageButton
@@ -20,7 +19,6 @@ import android.widget.ScrollView
 import android.widget.Toast
 import kotlinx.android.synthetic.main.browser_overlay.view.*
 import kotlinx.android.synthetic.main.browser_overlay_top_nav.view.*
-import kotlinx.android.synthetic.main.home_tile.view.*
 import kotlinx.coroutines.experimental.Job
 import kotlinx.coroutines.experimental.android.UI
 import kotlinx.coroutines.experimental.launch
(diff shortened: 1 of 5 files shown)


app/src/main/res/layout/browser_overlay.xml
```diff
@@ -2,78 +2,90 @@
...
-    android:orientation="vertical"
-    android:descendantFocusability="afterDescendants"
-    android:paddingTop="27dp"
-    >
+    android:fillViewport="true"
+    android:focusable="false"
+    android:scrollbars="none">
+    <!-- Fill viewport? When don't have enough tiles. -->
 
     <LinearLayout
```
liuche: Nit: comment that scrollview can only have one child? Maybe this isn't necessary, but looked a little inefficient.

app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt
```diff
@@ -287,3 +289,27 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
...
+ * clicks up twice quickly), it will skip and not scroll smoothly. Since we don't scroll often,
+ * this seems fine.
+ */
+class BrowserNavigationOverlayScrollView(
+        context: Context, attrs: AttributeSet
+) : ScrollView(context, attrs) {
+
+    private val deltaScrollPadding = resources.getDimensionPixelSize(R.dimen.browser_overlay_delta_scroll_padding)
+
+    override fun computeScrollDeltaToGetChildRectOnScreen(rect: Rect?): Int {
```
liuche: There's already some scroll code for the RecyclerView, to autoscroll so the next row is showing - how does this interact with that? Should that be deleted?

app/src/main/res/layout/browser_overlay.xml
mcomella: I think it can have multiple children: it's designed this way because ScrollView is a FrameLayout and we wanted a LinearLayout.

mcomella:
@liuche Please rereview the last two commits.

Issue #795, opened 2018-04-10
Back stack code is hard to change for a variety of reasons: consider simplifying it


mcomella:
Sorry if this is difficult to understand: this code is really difficult to understand and I'm having difficulty summarizing concrete steps to break it down/change it.

## User value
From a user's perspective, this code mostly works. However, if we modify the back stack behavior or add an additional screen, the development time to make the change may be long and the change error-prone. In extreme cases, we may be forced to simplify this code then.

If an Intent opens the app while it's on the Settings screen, we'll create multiple WebView instances and take up resources. This will be cleared when home is selected.

## Details
This code is an intersection of several components and has several issues. Feel free to break these down if necessary but I'm having difficult doing it:
- The code to load a URL implicitly assumes that there will never be a BrowserFragment in the back stack so if there is, we'll create multiple WebView instances (e.g. when Settings is shown so BrowserFragment is in the back stack and we receive an Intent)
- We create and destroy BrowserFragments + WebViews as the browser screen visibility changes. In the current workflow, there should generally only be one pair with strong references. We know we'll need these: should we just keep them around long-term instead of churning? (#654)
- The Session abstraction creates confusion: it restores WebView state, to create a new BrowserFragment, you want to call, `Session.createSession`, etc. Should remove as part of #794?
- The code is difficult to read because many actions pass asynchronously through Observables, which isn't well served by IDE inspection (e.g. go-to-definition).

Other issues that came out of this research are #794, #793, and #654.

liuche:
> (e.g. when Settings is shown so BrowserFragment is in the back stack and we receive an Intent)

From Settings, I think we're opening content like About (and links therein). This is something that should be in a separate webview IMO, because it has nothing to do with the user's web history and in a mental model should be "native" - sharing the webviews/backstack here would lead to confusion.

mcomella:
> From Settings, we're opening content like About (and links therein). This is something that should be in a separate webview because it has nothing to do with the user's web history

I mean if the user had visited some sites, opens settings, leaves the app, clicks a link (or otherwise sends Firefox a browsing intent), and we launch a browserfragment on top of settings. In this case, there are two WebView/BF instances consuming resources.

mcomella:
Dupe #766.

mcomella:
As mentioned in https://github.com/mozilla-mobile/firefox-tv/pull/892#issuecomment-386762792, this isn't really a dupe.

PR #892, opened 2018-05-04
Clean up BrowserFragment back behavior

liuche:
We removed the HomeFragment, and now only BrowserFragment and MainActivity have non-trivial back behavior.

This doesn't fix multiple sessions, multiple webviews (as described in #795).


liuche:
@mcomella you might consider re-opening #795 because I didn't address a lot of those problems in this PR, but probably with a rename because it doesn't seem like a back stack problem.

PR #898, opened 2018-05-08
Closes #885: Follow-up: Toast for "Unpin" hint is not shown according…

### LINKED ISSUE #885: Follow-up: Toast for "Unpin" hint is not shown according to spec
Follow-up from #782 
### Steps to reproduce
* Open Firefox TV
* Go to home screen

### Expected behavior
* Toast hint shows up when tile is selected, max 3 times

### Actual behavior
* Toast hint shows up over Turbo mode screen
* Toast doesn't show up when navigating through tiles

### Device information

* Fire TV device: 25
* Firefox version: master

bkmunar:
… to spec

### DIFF
--- a/app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt
+++ b/app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt
@@ -7,6 +7,7 @@ package org.mozilla.focus.browser
 import android.content.Context
 import android.graphics.Rect
 import android.graphics.drawable.Drawable
+import android.preference.PreferenceManager
 import android.support.v4.content.ContextCompat
 import android.support.v7.widget.GridLayoutManager
 import android.support.v7.widget.RecyclerView
@@ -35,6 +36,9 @@ import kotlin.properties.Delegates
 
 private const val NAVIGATION_BUTTON_ENABLED_ALPHA = 1.0f
 private const val NAVIGATION_BUTTON_DISABLED_ALPHA = 0.3f
+
+private const val SHOW_UNPIN_TOAST_COUNTER_PREF = "show_upin_toast_counter"
+private const val MAX_UNPIN_TOAST_COUNT = 3
 
 private const val COL_COUNT = 5
 
@@ -76,6 +80,10 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
      * and cancel this job at the end of the UI lifecycle, cancelling the children.
      */
     var uiLifecycleCancelJob: Job
+
+    // We need this in order to show the unpin toast, at max, once per
+    // instantiation of the BrowserNavigationOverlay
+    var canShowUpinToast: Boolean = false
 
     // Setting the onTileLongClick function in the HomeTileAdapter is fragile
     // since we init the tiles in View.init and Android is inflating the view for us,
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt
```diff
@@ -135,7 +136,16 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
                     onNavigationEvent?.invoke(NavigationEvent.LOAD_TILE, urlStr, null)
                 }
             }
-        }, onTileLongClick = openHomeTileContextMenu)
+        }, onTileLongClick = openHomeTileContextMenu, onTileFocused = {
+            val prefInt = PreferenceManager.getDefaultSharedPreferences(context).getInt(SHOW_UNPIN_TOAST_COUNTER_PREF, 0)
```
liuche: Hm, I guess this shows the toast 3 times on a single launch. I don't think that really makes sense - probably my fault for not being entirely clear "when tile is first selected on a launch, max 3 times".

The approach to use sharedprefs seems fine, but let me clarify this behavior in the bug.

app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt
```diff
@@ -135,7 +136,16 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
...
+            if (prefInt < 2) {
```
liuche: This 2 should be a constant so that it's named, like MAX_UNPIN_TOAST_COUNT, rather than just a random 2 in the code.

app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt
liuche: Can you add a constant that gets set once per launch, maybe in `initTiles`, or in one of the `onCreate*`?

liuche:
@bkmunar how do you know this works? Can you describe why you think this change works? What situations did you test in, and does that cover all the expected cases?

bkmunar:
Well I have tested that the toast only shows up either once per launch or after clearing the history. This is because initTiles() (and the init function for the BrowserNavigationOverlay for that matter) is instantiated in those two cases. I think this is appropriate behavior for this feature (although there may be a discussion if it should happen after clearing history. I think it's ok, but I think making it happen once per launch would involve another pref).

liuche:
@bkmunar since the overlay shows up in two instances (launch, opening the overlay), did you test that behavior, and what did you see? Does it behave the way you'd expect it to?

(To be clear, I'm not asking these questions to be challenging - the goal is to make sure that the code is behaving correctly, and part of that is understanding how it behaves in all the cases where it's used and making sure it's what we expect. So in this case, I'd expect you to test all the cases, and raise any questions you have.)

bkmunar:
It seems to work fine in both those cases. After launch and user focuses on a tile for the first time, one toast shows up. After navigation to a site and overlay is pulled up, the toast shows up depending on if we've shown the toast already or not.

bkmunar:
I did notice that the BNO init() is called after viewing the Pocket feed thus making the toast triggerable again. Not sure if this is what we want, but if not, I think we should resort to a pref that is changed every time the app is created/destroyed if we really want the behavior to be specific.

app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt
```diff
@@ -77,6 +81,8 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
      */
     var uiLifecycleCancelJob: Job
 
+    var canShowUpinToast: Boolean = false
```
liuche: Nit: add a comment here about why we have this.

Issue #918, opened 2018-05-22
Failed to resolve: org.mozilla.telemetry:telemetry:1.2.0


SeburoSUMO:
### Steps to reproduce
Downloaded .zip from https://github.com/mozilla-mobile/firefox-tv and extracted.
Opened Android Studio 3.1.2 and imported the extracted file as a new project.
Let Android Studio build the app

### Expected behavior
The application should build without error

### Actual behavior
App does not build
Message of "Failed to resolve: org.mozilla.telemetry:telemetry:1.2.0" is displayed.  This references line 168 in app, that states "implementation 'org.mozilla.telemetry:telemetry:1.2.0'"

### Device information
Android Studio 3.1.2 on Ubuntu 18.04

mcomella:
We're blocked by components: https://github.com/mozilla-mobile/android-components/issues/208

pocmo:
The telemetry library is back on jcenter. Unfortunately only with the new group id. I'll try to prepare a patch.

mcomella:
This was fixed on master, https://github.com/mozilla-mobile/firefox-tv/commit/226585b84fb97b508e08806a2dca670d1a15f3e0, but older builds will continue to be broken until this is fixed: https://github.com/mozilla-mobile/android-components/issues/208

PR #920, opened 2018-05-28
Issue #918: Update outdated dependencies and use new artifact names (telemetry and kotlin stdlib).

pocmo:
I was only planning to update the telemetry library, but Gradle was warning about outdated dependencies and it couldn't resolve the runtime dependency of android.arch.*. 

So I updated some of the other dependencies too and did a quick run of the app to make sure it still works.


mcomella:
> Gradle was warning about outdated dependencies

@pocmo Like Gradle failed the build because the dependencies were outdated? Which part of gradle? Command line? Is this Android lint? I haven't seen this before.

mcomella:
These changes make sense to me.

PR #922, opened 2018-06-01
Pre-land strings.

liuche:
Wanted to generalize our strings, keeping some slight duplication in ("Remove from homescreen") but when those strings get switched, they can be removed.

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -171,20 +171,35 @@ be temporary, and you can try again later.</li>
     <string name="settings_cookies_dialog_content">Selecting ‘Clear Now’ will clear all cookies and site data stored by Firefox for Fire TV. This may sign you out of websites and remove offline web content.</string>
 
     <string name="turbo_mode">Turbo Mode</string>
+    <!-- Displayed as notification when Turbo mode is turned on -->
+    <string name="turbo_mode_enabled_toast">Turbo Mode enabled</string>
+    <!-- Displayed as notification when Turbo mode is turned off -->
+    <string name="turbo_mode_disabled_toast">Turbo Mode disabled</string>
     <string name="onboarding_turbo_mode_title">Turbo Mode helps you browse faster</string>
     <string name="onboarding_turbo_mode_body">Turbo Mode automatically blocks trackers and ads to help get you where you\'re going faster. If there\'s a speed bump – like a site that isn\'t behaving the way you expect – just turn Turbo Mode off.</string>
     <string name="onboarding_turbo_mode_button_on">Keep Turbo Mode Enabled</string>
     <string name="onboarding_turbo_mode_button_off">Turn Turbo Mode Off</string>
+    <!-- Spoken to screen reader users to explain "off" action of Turbo mode button -->
+    <string name="turbo_mode_disable_a11y">Turn Turbo Mode Off</string>
+    <!-- Spoken to screen reader users to explain "on" action of Turbo mode button -->
+    <string name="turbo_mode_enable_a11y">Turn Turbo Mode On</string>
 
     <!-- Text displayed in a toast to explain to the user how to unpin a tile from the homescreen.
          SELECT refers to a button on the remote: the button at the center of the 4-way directional navigation. -->
     <string name="homescreen_unpin_tutorial_toast">Press and hold SELECT to unpin from homescreen</string>
     <string name="homescreen_tile_remove">Remove</string>
+    <!-- Spoken to screen reader users to explain action of deactivating Pin button state -->
+    <string name="homescreen_unpin_a11y">Unpin</string>
 
+    <!-- Spoken to screen reader users to describe Firefox homescreen -->
+    <string name="homescreen_title">Homescreen</string>
     <string name="pin_label">Pin to homescreen</string>
+    <!-- Short version of notification displayed when a site is added to homescreen -->
+    <string name="notification_pinned_general">Pinned to homescreen</string>
+    <!-- Short version of notification displayed when a site is removed from homescreen -->
+    <string name="notification_unpinned_general">Removed from homescreen</string>
     <string name="notification_pinned_site">Pinned to Firefox TV homescreen</string>
     <string name="notification_unpinned_site">Removed from Firefox TV homescreen</string>
-
 
     <!-- The title of a tutorial shown to introduce Pocket video recommendations.
     %1$s will be replaced with the Pocket brand name. -->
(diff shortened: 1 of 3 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -171,21 +171,27 @@ be temporary, and you can try again later.</li>
...
     <!-- Text displayed in a toast to explain to the user how to unpin a tile from the homescreen.
          SELECT refers to a button on the remote: the button at the center of the 4-way directional navigation. -->
     <string name="homescreen_unpin_tutorial_toast">Press and hold SELECT to unpin from homescreen</string>
     <string name="homescreen_tile_remove">Remove</string>
+    <string name="homescreen_unpin">Unpin</string>
 
+    <string name="homescreen_title">Homescreen</string>
     <string name="pin_label">Pin to homescreen</string>
+    <string name="notification_pinned_general">Pinned to homescreen</string>
+    <string name="notification_unpinned_general">Unpinned from homescreen</string>
```
mcomella: Amin suggested "Removed from|Pinned to Firefox homescreen", so I guess we should do substitution there too.

app/src/main/res/values/strings.xml
```diff
@@ -171,21 +171,34 @@ be temporary, and you can try again later.</li>
...
+    <!-- Displayed as notification when Turbo mode is turned on -->
+    <string name="turbo_mode_enabled">Turbo Mode enabled</string>
+    <!-- Displayed as notification when Turbo mode is turned off -->
+    <string name="turbo_mode_disabled">Turbo Mode disabled</string>
     <string name="onboarding_turbo_mode_title">Turbo Mode helps you browse faster</string>
     <string name="onboarding_turbo_mode_body">Turbo Mode automatically blocks trackers and ads to help get you where you\'re going faster. If there\'s a speed bump – like a site that isn\'t behaving the way you expect – just turn Turbo Mode off.</string>
     <string name="onboarding_turbo_mode_button_on">Keep Turbo Mode Enabled</string>
     <string name="onboarding_turbo_mode_button_off">Turn Turbo Mode Off</string>
+    <!-- Spoken to screen reader users to explain action of Turbo mode button -->
+    <string name="turbo_mode_enable">Turn Turbo Mode On</string>
```
mcomella: nit: turbo_mode_enable_a11y or similar

app/src/main/res/values/strings.xml
```diff
@@ -171,21 +171,34 @@ be temporary, and you can try again later.</li>
...
 
     <string name="turbo_mode">Turbo Mode</string>
+    <!-- Displayed as notification when Turbo mode is turned on -->
+    <string name="turbo_mode_enabled">Turbo Mode enabled</string>
+    <!-- Displayed as notification when Turbo mode is turned off -->
+    <string name="turbo_mode_disabled">Turbo Mode disabled</string>
     <string name="onboarding_turbo_mode_title">Turbo Mode helps you browse faster</string>
     <string name="onboarding_turbo_mode_body">Turbo Mode automatically blocks trackers and ads to help get you where you\'re going faster. If there\'s a speed bump – like a site that isn\'t behaving the way you expect – just turn Turbo Mode off.</string>
     <string name="onboarding_turbo_mode_button_on">Keep Turbo Mode Enabled</string>
     <string name="onboarding_turbo_mode_button_off">Turn Turbo Mode Off</string>
```
mcomella: The context of onboarding is different than "turbo_mode_enable", where I think we're using it as opposites – why don't we just add a new string for the a11y use case?

app/src/main/res/values/strings.xml
```diff
@@ -171,21 +171,34 @@ be temporary, and you can try again later.</li>
...
+    <!-- Spoken to screen reader users to explain action of Turbo mode button -->
+    <string name="turbo_mode_enable">Turn Turbo Mode On</string>
 
     <!-- Text displayed in a toast to explain to the user how to unpin a tile from the homescreen.
          SELECT refers to a button on the remote: the button at the center of the 4-way directional navigation. -->
     <string name="homescreen_unpin_tutorial_toast">Press and hold SELECT to unpin from homescreen</string>
     <string name="homescreen_tile_remove">Remove</string>
+    <!-- Spoken to screen reader users to explain action of deactivating Pin button state -->
+    <string name="homescreen_unpin">Unpin</string>
```
mcomella: As I mentioned in person, I think it'd be ideal if we could get the Components to use the default checkable announcements here.

mcomella: nit: homescreen_unpin_a11y, or similar

app/src/main/res/values/strings.xml
```diff
@@ -171,21 +171,34 @@ be temporary, and you can try again later.</li>
     <string name="settings_cookies_dialog_content">Selecting ‘Clear Now’ will clear all cookies and site data stored by Firefox for Fire TV. This may sign you out of websites and remove offline web content.</string>
 
     <string name="turbo_mode">Turbo Mode</string>
+    <!-- Displayed as notification when Turbo mode is turned on -->
+    <string name="turbo_mode_enabled">Turbo Mode enabled</string>
+    <!-- Displayed as notification when Turbo mode is turned off -->
+    <string name="turbo_mode_disabled">Turbo Mode disabled</string>
```
mcomella: nit: -> turbo_mode_disabled_toast or similar, same above

app/src/main/res/values/strings.xml
```diff
@@ -171,21 +171,34 @@ be temporary, and you can try again later.</li>
...
+    <!-- Spoken to screen reader users to explain action of deactivating Pin button state -->
+    <string name="homescreen_unpin">Unpin</string>
 
+    <!-- Spoken to screen reader users to describe Firefox homescreen -->
+    <string name="homescreen_title">Homescreen</string>
     <string name="pin_label">Pin to homescreen</string>
+    <!-- Short version of notification displayed when a site is added to homescreen -->
+    <string name="notification_pinned_general">Pinned to homescreen</string>
+    <!-- Short version of notification displayed when a site is removed from homescreen -->
+    <string name="notification_unpinned_general">Unpinned from homescreen</string>
```
mcomella: Amin proposed "Removed from|Pinned to Firefox homescreen" - we could use the substitution there too.

app/src/main/res/values/strings.xml
liuche: Hmm, I guess we don't use _a11y anywhere else, and it's not clear to me that localizers would understand what that means. But okay, I'll suffix this.

app/src/main/res/values/strings.xml
liuche: Agreed, @mcomella you should file a bug. I think it's better to use system a11y strings as well because they'll be more consistent.

app/src/main/res/values/strings.xml
mcomella: I was thinking this was for us, not for localizers: it helps us know where this string was intended to be used.

Issue #930, opened 2018-06-08
Implement MediaSession API to support voice interactions for video/media playback


mcomella:
## User benefit
On the new Fire TV cube device, you can interact with video/media using voice and [the MediaSession API](https://developer.amazon.com/blogs/appstore/post/623addd5-6184-4e7f-a969-123a653cfb1a/implementing-voice-control-with-the-media-session-api-on-amazon-fire-tv). This would be convenient for users.

## Requirements
- Investigate what we'd have to do to integrate
- Integrate if possible

mcomella:
An additional link: https://developer.amazon.com/docs/fire-tv/mediasession-api-integration.html

mcomella:
Looking briefly into this, it looks like we need to instantiate a `MediaSessionCompat` instance and add some callbacks. However, WebView doesn’t appear to have APIs to query/modify audio/video playback so we might need to do this through JavaScript.

mcomella:
fwiw, there is a [`MediaSession` JS API](https://developer.mozilla.org/en-US/docs/Web/API/Media_Session_API) which makes this a little harder to google. :)

mcomella:
The code we write could be usable by others trying to integrate `MediaSession` with `WebView`: we should consider making it a library.

mcomella:
I'd estimate this is a size M though it could turn into a size L if the JS turns out to be difficult/fragile.

mcomella:
I think this is as straight-forward as creating a `MediaSession` and registering callbacks. However, one unclear thing: when do we need to call `release`? The docs seem to assume that media Activities will playback for their whole lifecycle (thus in `onDestroy`) but videos can be created on each page: should we create a new session for each page? Here's the code I wrote in onCreate to test without a device:
```kotlin
        mediaSession = MediaSessionCompat(this, "lol")
        val pb = PlaybackStateCompat.Builder()
                .setActions(PlaybackStateCompat.ACTION_PLAY or
                        PlaybackStateCompat.ACTION_PAUSE)
                .build()

        mediaSession.setPlaybackState(pb)
        mediaSession.setCallback(object : MediaSessionCompat.Callback() {
            val browserFragment get() = supportFragmentManager.findFragmentByTag(BrowserFragment.FRAGMENT_TAG) as BrowserFragment?
            override fun onPlay() {
                Log.d("lol", "play called")
[...]

Issue #240, opened 2017-12-20
Provide APK download links


mcomella:
Like we do for Focus.

sorryusernameisalreadytaken:
That would be really cool! +vote+

LoneFenris:
When this is done will builds also appear on the Mozilla FTP?

mcomella:
We intend to also upload these builds to archive.mozilla.org – if it doesn't get done in this issue, we'll file another issue to do it.

mcomella:
We added an Amazon-specific permission in #937: `<uses-permission-sdk-23 android:name="com.amazon.permission.media.session.voicecommandcontrol" />`

We should make sure this doesn't create problems on non-Amazon devices.

liuche:
Puffin also has a "send link to tv" using QR codes.

![screenshot_2018-08-10 puffin tv - fast web browser - apps on google play](https://user-images.githubusercontent.com/1082253/43980327-a0952b02-9ca2-11e8-8a7e-b974f7315128.png)

mcomella:
@liuche wrt previous comment, is this the wrong issue?

yoasif:
"Me too" comment here. 

I picked up a Nvidia Shield TV over black friday this year and was surprised to not find Firefox available in the Play Store, having seen the news about Firefox TV a while back. I managed to get it installed from one of the apk mirror sites while writing this comment, and it seems to work without issue.

mcomella:
@yoasif Which APK mirror site did you use? We're curious where our APKs are being hosted. :)

yoasif:
Downloaded from https://www.apkmirror.com/apk/mozilla/firefox-android-tv/firefox-android-tv-3-1-1-release/firefox-android-tv-3-1-1-android-apk-download/

Noiwex:
Don't really trust apkmirror.

mcomella:
Note: we currently post our release APKs to [our GitHub Releases](https://github.com/mozilla-mobile/firefox-tv/releases) about every two weeks and while these APKs can be used to sideload, they are intended for project developers. One side effect of this is that we sign them with different keys each time, which users from being able to install the app without first uninstalling it. We're also not officially supporting builds outside of the Amazon store (though feel free to file issues).

In the near future (end of Q3, early Q4?), we hope to post consistent signed APKs to Mozilla's archive and GitHub releases.

sorryusernameisalreadytaken:
> 
> 
[...]

PR #937, opened 2018-06-21
Issue #930, Closes #936: Add play/pause voice command support.

### LINKED ISSUE #936: [Voice] "Alexa, pause" on youtube.com/tv doesn't work consistently, shows black screen
### Steps to reproduce
On cube:
- Open youtube.com/tv, select video
- Say, "Alexa pause"

### Expected behavior
Alexa pauses

### Actual behavior
Alexa pauses while you speak but starts the video again. The video seems to do a reload thing and sometimes will pause.

If we comment out the `onPause/pauseTimers` WebView methods (and resume counterparts), the video will pause when speaking, start playing again, and pause a few moments later (I assume Alexa sends media button events). I think we can fix the pause afterwards by implementing [this optimization suggested by the MediaSession docs](https://developer.amazon.com/docs/fire-tv/mediasession-api-integration.html#enhanced-experience-for-voice-enabled-playback-controls-using-media-session-optional).

We should investigate what the expected behavior for Alexa is: should we be calling onPause in onPause? What about pauseTimers? Will it all just work if we implement the optimization above?

Note how this functionality will intersect with #930, which implements the MediaSession APIs to support this functionality.

---

Note that fixing this *does not* implement other MediaSession functionality:
- Rewind/fast-forward
- Next/previous
- Restart

mcomella:
For time's sake, we:
- Only do play/pause
- Do not handle audio (b/c mute key events, see below)

After this change, we handle play/pause events ourselves (through
MediaSession) and mute these key events from reaching the web page (who
could have handled these themselves, such as youtube). Benefits:
- Voice and hardware remotes use the same code path for the same actions
- We can play/pause videos on pages that did not previously support it

Downside is that pages may have extra special handling of the media
buttons that we're disabling, but we have yet to see this.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -10,6 +10,12 @@
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" /> <!-- Used by sentry. -->
     <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
     <uses-permission android:name="com.android.launcher.permission.INSTALL_SHORTCUT" />
+
+    <!-- Required for voice control. It's only available on the Cube (so far) which is API 25+
+         so we use sdk-23 to prevent users from manually having to update the app.
+         https://developer.amazon.com/blogs/appstore/post/623addd5-6184-4e7f-a969-123a653cfb1a/implementing-voice-control-with-the-media-session-api-on-amazon-fire-tv -->
+    <uses-permission-sdk-23 android:name="com.amazon.permission.media.session.voicecommandcontrol" />
+
     <uses-feature android:name="android.hardware.touchscreen" android:required="false"/>
     <uses-feature android:name="android.software.leanback" android:required="true" />
 
--- a/app/src/main/java/org/mozilla/focus/browser/BrowserFragment.kt
+++ b/app/src/main/java/org/mozilla/focus/browser/BrowserFragment.kt
@@ -69,6 +69,8 @@ class BrowserFragment : IWebViewLifecycleFragment() {
     override val initialUrl get() = session.url.value
     override val iWebViewCallback get() = SessionCallbackProxy(session, BrowserIWebViewCallback(this))
 
+    private lateinit var videoVoiceCommandMediaSession: VideoVoiceCommandMediaSession
+
     /**
      * The current URL.
      *
@@ -95,6 +97,7 @@ class BrowserFragment : IWebViewLifecycleFragment() {
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
         initSession()
+        initMediaSession()
     }
 
     private fun initSession() {
@@ -115,6 +118,11 @@ class BrowserFragment : IWebViewLifecycleFragment() {
                 }
             }
         })
+    }
+
+    private fun initMediaSession() {
+        videoVoiceCommandMediaSession = VideoVoiceCommandMediaSession(activity!!)
+        lifecycle.addObserver(videoVoiceCommandMediaSession)
     }
 
     private val onNavigationEvent = { event: NavigationEvent, value: String?,
@@ -165,6 +173,8 @@ class BrowserFragment : IWebViewLifecycleFragment() {
     override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View {
         val layout = inflater.inflate(R.layout.fragment_browser, container, false)
 
+        videoVoiceCommandMediaSession.iWebView = layout.webview
+
         cursor = CursorController(this, cursorParent = layout.browserFragmentRoot,
                 view = layout.cursorView)
         lifecycle.addObserver(cursor!!)
@@ -210,6 +220,9 @@ class BrowserFragment : IWebViewLifecycleFragment() {
 
     override fun onDestroyView() {
         super.onDestroyView()
+
+        videoVoiceCommandMediaSession.iWebView = null
+
         lifecycle.removeObserver(cursor!!)
         cursor = null
         overlayVisibleCached = browserOverlay.visibility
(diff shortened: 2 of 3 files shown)


app/src/main/AndroidManifest.xml
```diff
@@ -10,7 +10,12 @@
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" /> <!-- Used by sentry. -->
     <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
     <uses-permission android:name="com.android.launcher.permission.INSTALL_SHORTCUT" />
-    <uses-feature android:name="android.hardware.touchscreen" android:required="false"/>
+
+    <!-- Required for voice control. It's only available on the Cube (so far) which is API 25+
+         so we use sdk-23 to prevent users from manually having to update the app.
+         https://developer.amazon.com/blogs/appstore/post/623addd5-6184-4e7f-a969-123a653cfb1a/implementing-voice-control-with-the-media-session-api-on-amazon-fire-tv -->
+    <uses-permission-sdk-23 android:name="com.amazon.permission.media.session.voicecommandcontrol" />
```
liuche: Interesting - I guess if we start to ship on Android TV with #240 for non-Amazon use, this would be a permission that we'd have to make a flavor for (unless extraneous permissions have no effect).

app/src/main/java/org/mozilla/focus/browser/BrowserFragment.kt
```diff
@@ -117,6 +120,11 @@ class BrowserFragment : IWebViewLifecycleFragment() {
         })
     }
 
+    private fun initMediaSession() {
+        videoVoiceCommandMediaSession = VideoVoiceCommandMediaSession(activity!!)
```
liuche: Are `!!` a case to use NonFatalAssertionException? Not saying you need to, just wondering.

app/src/main/java/org/mozilla/focus/browser/VideoVoiceCommandMediaSession.kt
```diff
@@ -0,0 +1,141 @@
...
+        lifecycleOwner.lifecycle.removeObserver(this)
+
+        mediaSession.release()
+    }
+
+    fun dispatchKeyEvent(event: KeyEvent): Boolean = when (event.keyCode) {
+        // If we do nothing, these keys will be received twice: on key down by the MediaSession API
+        // and on key up by the WebView/web page, causing playback state updates to be undone.
+        // Thus we mute the WebView key up events and handle it all ourselves through MediaSession.
+        KEYCODE_MEDIA_PLAY_PAUSE, KEYCODE_MEDIA_PLAY, KEYCODE_MEDIA_PAUSE -> event.action == KeyEvent.ACTION_UP
```
liuche: Hmmm tricky - so "voice commands" come through as key events? Do using the buttons work too?

app/src/main/java/org/mozilla/focus/browser/VideoVoiceCommandMediaSession.kt
```diff
@@ -0,0 +1,141 @@
...
+                |    if (!videoToOperateOn) {
+                |        videoToOperateOn = videos[0];
+                |    }
+                |
+                |    if (videoToOperateOn.paused) {
+                |        videoToOperateOn.play();
+                |    } else {
+                |       videoToOperateOn.pause();
+                |   }
+                |})();
```
liuche: wow nice javascript snippet - is this the intended way to do this? Or does the Amazon MediaSessions provide other APIs (rather than writing JS)?

mcomella:
> although I wonder if there's a better way than using injected JS...?

I don't think so: there are no WebView APIs for handling media playback and Amazon doesn't add any special functionality.

app/src/main/AndroidManifest.xml
mcomella: I think extraneous permissions have no effect but I don't know: added a note to the issue.

app/src/main/java/org/mozilla/focus/browser/BrowserFragment.kt
mcomella: The activity is guaranteed to be attached here: the lifecycle is `onAttach`, `onCreate`.

app/src/main/java/org/mozilla/focus/browser/VideoVoiceCommandMediaSession.kt
mcomella: No other way to do this, as far as I know.

MediaSession is an Android API, not Amazon, and is intended to be an agnostic way to communicate directly with the back-end player (while MediaController communicates with the UI and MediaController communicates with MediaSession). Since the player is in the WebView, we have to communicate directly with the player and, afaik, that means JS.

app/src/main/java/org/mozilla/focus/browser/VideoVoiceCommandMediaSession.kt
mcomella: I should clarify:
- Voice commands call MediaSession callbacks
- Hardware media buttons are swallowed for key down – they call MediaSession callbacks – but on key up they propagate as standard key events (hence they're handled twice)

I'll clarify in the commend.

app/src/main/AndroidManifest.xml
```diff
@@ -10,7 +10,12 @@
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" /> <!-- Used by sentry. -->
     <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
     <uses-permission android:name="com.android.launcher.permission.INSTALL_SHORTCUT" />
-    <uses-feature android:name="android.hardware.touchscreen" android:required="false"/>
```
mcomella: Didn't mean to remove this: undoing this change.

mcomella:
Taskcluster failure was timeout – it's probably fine because I didn't change anything that'd timeout.

Issue #939, opened 2018-06-22
Move MediaSession instance to MainActivity


mcomella:
There is expected to be a single instance per process

PR #945, opened 2018-06-22
Closes #930: complete initial MediaSession implementation

### LINKED ISSUE #930: Implement MediaSession API to support voice interactions for video/media playback

mcomella:
We still need to complete #936 and #939 before this is shippable, however.

### DIFF
--- a/app/src/main/java/org/mozilla/focus/browser/VideoVoiceCommandMediaSession.kt
+++ b/app/src/main/java/org/mozilla/focus/browser/VideoVoiceCommandMediaSession.kt
@@ -33,16 +33,7 @@ private const val SUPPORTED_ACTIONS = ACTION_PLAY_PAUSE or ACTION_PLAY or ACTION
         ACTION_SKIP_TO_NEXT or ACTION_SKIP_TO_PREVIOUS or
         ACTION_SEEK_TO // "Alexa, rewind/fast-forward <num> <unit-of-time>"
 
-/**
- * A workaround for playback position. The playback position increments in the MediaSession APIs
- * automatically based on the given playback speed. The position is used to provide an absolute
- * position to seek to in `onSeekTo` (triggered by rewind/fast-forward commands). Since these
- * commands are currently relative ("Alexa fast-forward 10 minutes"), we can avoid syncing with the
- * JS playback state, simplifying the code, and set the the position to an absolute value with
- * playback speed 0: when onSeekTo is called, the constant value will be added to the offset we need
- * so we can subtract to get the offset. See onSeekTo for details. This will break if absolute
- * seeking is implemented (#941.
- */
+// See `onSeekTo` for details on HACKED_*.
 private const val HACKED_PLAYBACK_POSITION: Long = Long.MAX_VALUE / 2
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mozilla/focus/browser/VideoVoiceCommandMediaSession.kt
```diff
@@ -16,20 +16,35 @@ import android.support.v4.media.session.PlaybackStateCompat
...
+
+/**
+ * A workaround for playback position. The playback position increments in the MediaSession APIs
+ * automatically based on the given playback speed. The position is used to provide an absolute
+ * position to seek to in `onSeekTo` (triggered by rewind/fast-forward commands). Since these
+ * commands are currently relative ("Alexa fast-forward 10 minutes"), we can avoid syncing with the
+ * JS playback state, simplifying the code, and set the the position to an absolute value with
+ * playback speed 0: when onSeekTo is called, the constant value will be added to the offset we need
+ * so we can subtract to get the offset. See onSeekTo for details. This will break if absolute
+ * seeking is implemented (#941.
```
liuche: Okay, I found this kind of confusing, so I'd rewrite it with things like:

- We want to get the unit of time delta that the user has said (e.g., 1 min in "fast forward 1 min") so we can advance the video in JS by that relative amount.

- Since the MediaSession API tries to be helpful and calculates the playback position for you (not the delta), to get only the delta from MediaSession API, we initialize the Playback.Builder to a constant and then set the playback speed to 0. Since we want to handle rewinds too, we use MAX_VALUE / 2 as our constant. To get the delta, we just subtract the position returned by MediaSession from our known constant.

mcomella: I thought it'd be easier to explain near the code so I moved my comment to `onSeekTo` and took your advice:
```kotlin
            // This method is called for "fast-forward/rewind X <time-unit>" where the args are an
            // absolute position. The system calculates the absolute position from the current
            // playback position and the user-provided offset.
            //
            // The MediaSession API calculates the current playback time with the last "current time"
            // value and the playback speed we provide it (in MediaSession.setPlaybackState):
            //   current_time = time_passed * playback_speed + initial_time
            //
            // However, we force STATE_PLAYING for the duration the app is open so this calculated
            // playback time will be inaccurate to the actual video. We hack around this by setting
            // HACKED_PLAYBACK_SPEED to 0 and never updating the current playback time so the
            // calculated position never changes: we can use this to calculate the offset (see the code).
            //
            // The system will never provide a negative absolute position so if HACKED_PLAYBACK_POSITION
            // is 0, we cannot calculate negative offsets and thus can't rewind). To avoid this,
            // we set it to the middle-most value (MAX_VALUE / 2) to support the widest range of
            // playback.
```

PR #946, opened 2018-06-22
Closes #939: Move MediaSession to MainActivity

### LINKED ISSUE #939: Move MediaSession instance to MainActivity

mcomella:


app/src/main/java/org/mozilla/focus/MainActivity.kt
```diff
@@ -169,13 +182,12 @@ class MainActivity : LocaleAwareAppCompatActivity(), OnUrlEnteredListener {
 
     override fun dispatchKeyEvent(event: KeyEvent): Boolean {
         val fragmentManager = supportFragmentManager
-        val browserFragment = fragmentManager.findFragmentByTag(BrowserFragment.FRAGMENT_TAG) as BrowserFragment?
+        val foundBrowserFragment = fragmentManager.findFragmentByTag(BrowserFragment.FRAGMENT_TAG) as BrowserFragment?
+        val maybeBrowserFragment = if (foundBrowserFragment?.isVisible == true) foundBrowserFragment else null
```
liuche: Okay, the found/maybeBrowserFragemnt might be confusing in the future, so maybe add a comment here about trying to consolidate to a single null-safe variable.

mcomella: Finally found a good use for let! :P Swapped with:
```kotlin
val maybeBrowserFragment = (fragmentManager.findFragmentByTag(BrowserFragment.FRAGMENT_TAG) as BrowserFragment?)?.let {
    if (it.isVisible) it else null
}
```

PR #947, opened 2018-06-22
Closes #586: Disable WebSettings.mediaPlaybackRequiresUserGesture.

### LINKED ISSUE #586: Youtube Videos Don't Always Autoplay
1. Visit Youtube
2. Choose a video

Actual: Video doesn't autoplay, have to press select and then the play button to start it up
Expected: Video begins playing

This also happens if you pin a youtube video and click on it's custom tile.

FYI sometimes it autoplays, but more often than not it doesn't.

mcomella:


app/src/amazonWebview/java/org/mozilla/focus/iwebview/WebViewProvider.kt
```diff
@@ -67,6 +67,8 @@ private fun initWebSettings(context: Context, settings: AmazonWebSettings) = wit
     javaScriptEnabled = true
     domStorageEnabled = true
 
+    mediaPlaybackRequiresUserGesture = false // Allow YouTube to auto-play.
```
liuche: Nit: I'd say "Allows auto-play" - unless you really want to emphasize Youtube with something like "Allows auto-play (which improves YouTube experience)"

PR #948, opened 2018-06-22
Closes #936: Move WebView.onPause to onStop

### LINKED ISSUE #936: [Voice] "Alexa, pause" on youtube.com/tv doesn't work consistently, shows black screen

mcomella:
Unfortunately, there are a lot of unknowns here about side effects.

### DIFF
--- a/app/src/main/java/org/mozilla/focus/iwebview/IWebViewLifecycleFragment.kt
+++ b/app/src/main/java/org/mozilla/focus/iwebview/IWebViewLifecycleFragment.kt
@@ -66,17 +66,23 @@ abstract class IWebViewLifecycleFragment : LocaleAwareFragment() {
 
     override fun onPause() {
         webView!!.saveWebViewState(session)
-        webView!!.onPause()
         webView!!.pauseTimers()
-
         super.onPause()
     }
 
     override fun onResume() {
-        webView!!.onResume()
         webView!!.resumeTimers()
-
         super.onResume()
+    }
+
+    override fun onStop() {
+        super.onStop()
+        webView!!.onStop() // internally calls WebView.onPause: see impl for details.
+    }
+
+    override fun onStart() {
+        super.onStart()
+        webView!!.onStart() // internally calls WebView.onResume: see impl for details.
     }
 
     override fun onDestroy() {
--- a/app/src/amazonWebview/java/org/mozilla/focus/webview/FirefoxAmazonWebView.kt
+++ b/app/src/amazonWebview/java/org/mozilla/focus/webview/FirefoxAmazonWebView.kt
@@ -63,6 +63,28 @@ internal class FirefoxAmazonWebView(
         webViewDelegate.setOnFocusChangeListener(::onFocusChange)
     }
 
+    override fun onStop() {
+        // NB: onStop unexpectedly calls onPause: see below.
+        //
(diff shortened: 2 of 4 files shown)


app/src/main/java/org/mozilla/focus/iwebview/IWebViewLifecycleFragment.kt
```diff
@@ -66,17 +66,17 @@ abstract class IWebViewLifecycleFragment : LocaleAwareFragment() {
 
     override fun onPause() {
         webView!!.saveWebViewState(session)
-        webView!!.onPause()
-        webView!!.pauseTimers()
```
liuche: Can we verify if the video pausing is because of `pauseTimers` vs `onPause`? Looking at the documentation for `pauseTimers` it seems like that should actually be called in `onStop` rather than `onPause`.

If it's in fact `onPause` I'm fine with this change, but I do wonder...

`Pauses all layout, parsing, and JavaScript timers for all WebViews. This is a global requests, not restricted to just this WebView. This can be useful if the application has been paused.`

^ This sounds like it might be the culprit, not `onPause`.

mcomella: > Can we verify if the video pausing is because of pauseTimers vs onPause

I definitely needed to comment out `onPause` to get this behavior to stop happening but I don't remember if `pauseTimers` needed to be moved too or not: maybe we should test this before landing.

> Looking at the documentation for pauseTimers it seems like that should actually be called in onStop rather than onPause

The docs make me think `pauseTimers` could be called from `onPaused`: "This can be useful if the application has been paused." But it doesn't specify: might be fine to call from `onStop` too.

liuche: Specifically commenting out `webView!!.onPause()`, right?

mcomella: Yes. I want to reverify the behaviors, however, before landing.

mcomella: I verified the video pauses because of `WebView.onPause`. In fact, I can still call `pauseTimers` in `onPause` if I move `WebView.onPause` to `onStop`: I pushed an amended commit with this change.

PR #953, opened 2018-06-29
Closes #952: Move string references to a separate file.

### LINKED ISSUE #952: Move string references (or arrays) to a separate resource file
Like https://github.com/mozilla-mobile/focus-android/issues/2826

In order to make it easier for our L10N tools we should only keep strings in strings.xml that we actually want to translate. We should move the other strings (e.g. arrays or string references) into a separate file.

pocmo:


app/src/main/res/values/strings.xml
```diff
@@ -88,11 +93,7 @@
     ]]>
     </string>
 
-    <string name="error_connect_title">@string/error_connectionfailure_title</string>
```
mcomella: Why do these even exist? Why not just reference the source strings directly? Looks like ahunt did this a while ago...

pocmo: Yeah, I don't know. When I first saw this blocking the new L10N tooling I thought about just removing them. But then I decided to just move them (in Focus we also have a string array).

Issue #941, opened 2018-06-22
Sync position/speed MediaSession playback state with the DOM


mcomella:
Depends on #940. Speculating, this playback information could be used by the Alexa API to:
- Allow absolute seeking, ("Alexa, go to one hour").
- Display playback data to the user ("Alexa, how much longer in the video?") so it'd be ideal if we synced it up.

With #940 done, this is likely straight forward but size M to be conservative.

mcomella:
There is no duration.

PR #956, opened 2018-07-03
Closes #940: sync play/pause playback state between JS and Java

### LINKED ISSUE #940: Sync playing/paused MediaSession playback state with the DOM
Our current implementation forces itself to be in the PLAYING state. Since the Alexa API acts as the MediaController and can make playback decisions based on such state, we're not acting as a team player in the Alexa ecosystem: we should fix that.

This is complicated: we'll need to reset playback state when loading new pages, sync up with any video tags on the current page, etc.

fwiw, we don't need to be perfectly in sync: syncing every 10s or so should be sufficinet.

mcomella:
Testing on older devices now. This is the required part; the more optional part is #941.

### DIFF
--- a/app/src/main/java/org/mozilla/focus/browser/VideoVoiceCommandMediaSession.kt
+++ b/app/src/main/java/org/mozilla/focus/browser/VideoVoiceCommandMediaSession.kt
@@ -10,6 +10,7 @@ import android.arch.lifecycle.Lifecycle.Event.ON_START
 import android.arch.lifecycle.Lifecycle.Event.ON_STOP
 import android.arch.lifecycle.LifecycleObserver
 import android.arch.lifecycle.OnLifecycleEvent
+import android.support.annotation.UiThread
 import android.support.v4.media.session.MediaSessionCompat
 import android.support.v4.media.session.PlaybackStateCompat
 import android.support.v4.media.session.PlaybackStateCompat.ACTION_PAUSE
@@ -18,14 +19,19 @@ import android.support.v4.media.session.PlaybackStateCompat.ACTION_PLAY_PAUSE
 import android.support.v4.media.session.PlaybackStateCompat.ACTION_SEEK_TO
 import android.support.v4.media.session.PlaybackStateCompat.ACTION_SKIP_TO_NEXT
 import android.support.v4.media.session.PlaybackStateCompat.ACTION_SKIP_TO_PREVIOUS
+import android.support.v4.media.session.PlaybackStateCompat.STATE_PAUSED
 import android.support.v4.media.session.PlaybackStateCompat.STATE_PLAYING
 import android.view.KeyEvent
 import android.view.KeyEvent.KEYCODE_MEDIA_PAUSE
 import android.view.KeyEvent.KEYCODE_MEDIA_PLAY
 import android.view.KeyEvent.KEYCODE_MEDIA_PLAY_PAUSE
+import android.webkit.JavascriptInterface
+import kotlinx.coroutines.experimental.android.UI
+import kotlinx.coroutines.experimental.launch
 import org.mozilla.focus.iwebview.IWebView
 import java.util.concurrent.TimeUnit
 
+private const val JS_INTERFACE_IDENTIFIER = "_firefoxTV_playbackStateObserverJava"
 private const val MEDIA_SESSION_TAG = "FirefoxTVMedia"
 
 private const val SUPPORTED_ACTIONS = ACTION_PLAY_PAUSE or ACTION_PLAY or ACTION_PAUSE or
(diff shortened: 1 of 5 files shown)


mcomella:
I have confirmed:
- Play/pause hardware button works on Fire Stick
- Microphone does not have effect on Fire Stick

app/src/main/java/org/mozilla/focus/browser/VideoVoiceCommandMediaSession.kt
```diff
@@ -57,25 +63,29 @@ private val KEY_EVENT_ACTIONS_DOWN_UP = listOf(KeyEvent.ACTION_DOWN, KeyEvent.AC
...
  */
-class VideoVoiceCommandMediaSession(
+class VideoVoiceCommandMediaSession @UiThread constructor(
         private val activity: Activity,
         private val getIWebView: () -> IWebView?
 ) : LifecycleObserver {
 
+    @get:UiThread // mediaSession is not thread safe.
     private val mediaSession = MediaSessionCompat(activity, MEDIA_SESSION_TAG)
+    private val cachedPlaybackState = PlaybackStateCompat.Builder()
```
liuche: What is the cachedPlaybackState for?

app/src/main/java/org/mozilla/focus/browser/VideoVoiceCommandMediaSession.kt
```diff
@@ -116,11 +118,25 @@ class VideoVoiceCommandMediaSession @UiThread constructor(
...
+
+        // Videos playing when the app was backgrounded get into an inconsistent state: `paused`
+        // will return false but they won't actually be playing. This makes any user "play" commands
+        // take two clicks: one to "pause" and the second to play for real. We get around this by
+        // pausing all videos before being backgrounded. This issue doesn't affect pages with
+        // autoplay, like YouTube.
+        //
+        // The videos may send playback state update events to , which we're forced to ignore: see
+        // JavascriptVideoPlaybackStateSyncer for the code.
+        webView?.evalJS("document.querySelectorAll('video').forEach(v => v.pause());")
```
liuche: Ugh, the webview doesn't automatically do this, even though the video stops playing? That's annoying.

This is something that I'd flag QA to look at, because it seems weird and easily broken. Please add a QA-needed flag with extra info when this gets landed.

app/src/main/java/org/mozilla/focus/browser/VideoVoiceCommandMediaSession.kt
```diff
@@ -188,3 +212,82 @@ class VideoVoiceCommandMediaSession(
...
+        addPlaybackStateListeners();
+        syncPlaybackState();
+    }
+
+    function addPlaybackStateListeners() {
+        document.querySelectorAll('video').forEach(videoElement => {
+            if (videosWithListeners.has(videoElement)) { return; }
+            videosWithListeners.add(videoElement);
+
+            PLAYBACK_STATE_CHANGE_EVENTS.forEach(event => {
```
liuche: huh, I didn't know JS had => carets

...maybe I'm not the right one to be reviewing this code heh

mcomella: Since Firefox 22: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions

I think new JS features aren't often widespread because of slow browser adoption (though that's less of an issue with Babel now...).

app/src/main/java/org/mozilla/focus/iwebview/IWebView.kt
```diff
@@ -39,6 +39,9 @@ interface IWebView {
     fun canGoForward(): Boolean
     fun canGoBack(): Boolean
 
+    fun addJavascriptInterface(obj: Any?, interfaceName: String?)
```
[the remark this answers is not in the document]
mcomella: This is identical to the WebView documentation.

Verified IRL that this doesn't need a reference.

app/src/main/java/org/mozilla/focus/browser/VideoVoiceCommandMediaSession.kt
mcomella: The WebView stops playback but the video does not consider itself paused (perhaps it's in a state similar to when it's buffering?).

QA flags sounds like a good idea.

mcomella:
> I'm a little confused why this wasn't the case before, because I thought the play/pause worked already.

The current implementation correctly plays/pauses the video but fails to set the PlaybackState in the MediaSession such that it accurately represents the current playback state (i.e. it's always playing, instead of pausing/playing).

app/src/main/java/org/mozilla/focus/browser/VideoVoiceCommandMediaSession.kt
mcomella: I'll add a comment.

mcomella:
Passes locally, previous revision passed.

PR #958, opened 2018-07-04
Closes #941: Sync MediaSession position/playback rate

### LINKED ISSUE #941: Sync position/speed MediaSession playback state with the DOM

mcomella:


app/src/main/java/org/mozilla/focus/browser/VideoVoiceCommandMediaSession.kt
```diff
@@ -320,11 +327,33 @@ var _firefoxTV_isPlaybackStateObserverLoaded;
...
+            playbackRate = null;
+        }
+
+        _firefoxTV_playbackStateObserverJava.syncPlaybackState(
+            isVideoPresent, isPlaying, positionSeconds, playbackRate);
     }
 
-    function isAnyVideoPlaying() {
-        return Array.from(document.querySelectorAll('video')).some(video => !video.paused);
+    function getMaybeTargetVideo() {
```
liuche: Why "MaybeTarget"? How about getActiveVideo? or getMaybeActiveVideo?

app/src/main/java/org/mozilla/focus/browser/VideoVoiceCommandMediaSession.kt
```diff
@@ -336,8 +345,30 @@ var _firefoxTV_isPlaybackStateObserverLoaded;
...
 
-        _firefoxTV_playbackStateObserverJava.syncPlaybackState(
-            isVideoPresent, isPlaying, positionSeconds, playbackRate);
+        schedulePlaybackStateSyncInterval(isPlaying);
+
+        javaInterface.syncPlaybackState(isVideoPresent, isPlaying, positionSeconds, playbackRate);
+    }
+
+    /**
+     * When a video is playing, schedules a function to repeatedly sync the playback state;
```
liuche: Nice, this makes sense.

app/src/main/java/org/mozilla/focus/browser/VideoVoiceCommandMediaSession.kt
```diff
@@ -298,7 +298,8 @@ private fun PlaybackStateCompat.Builder.setStateHacked(state: Int) =
 private val JS_OBSERVE_PLAYBACK_STATE = """
 var _firefoxTV_isPlaybackStateObserverLoaded;
 (function () {
-    const PLAYBACK_STATE_CHANGE_EVENTS = ['play', 'pause'];
+    /* seeking will send "pause, play" and so is covered here. */
+    const PLAYBACK_STATE_CHANGE_EVENTS = ['play', 'pause', 'ratechange'];
```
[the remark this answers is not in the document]
mcomella: Yes, it's a standard HTMLMediaElement event. Unfortunately, the docs suck and don't having a listing of relevant events but I found them on the w3 HTML 5 video test page: https://www.w3.org/2010/05/video/mediaevents.html

I'll add a link to the page in this file somewhere.

app/src/main/java/org/mozilla/focus/browser/VideoVoiceCommandMediaSession.kt
mcomella: I'm picking the video based on a heuristic: I think I used "target" unconsciously as way of indicating this because it's hard to quantify what the heuristic is briefly. It's not an "active" video because if there is no playing video, we get the first one in the page.

Maybe, `getMaybePlayingVideoOrFirstInPage()`? :)

Issue #883, opened 2018-05-03
(necessary?) Upgrade support library & SDK to 27.0.*


mcomella:
It's used by Picasso and we're on 26* so it blocks #867 unless we upgrade it or switch to glide (which doesn't have such dependencies).

mcomella:
See https://groups.google.com/a/mozilla.com/forum/#!topic/firefox-tv/fF1CNIOsyCQ for an explanation.

mcomella:
We might not have to upgrade the SDK all the time with:

> For Android Components we use an image based on sdkmanager and there we can just do the following to accept everything: `yes | sdkmanager --licenses`

mcomella:
Looks like Sebastian fixed this in #961.

PR #961, opened 2018-07-06
Issue #883: Update Android SDK and support libraries to version 27.

pocmo:
The SDK update was painless. But the new support libraries seem to have changed some things. Especially `activity` and `context` being nullable (Well, technically they could have been `null` before too. Kotlin code just didn't know.).

### DIFF
--- a/app/src/main/java/org/mozilla/focus/SettingsFragment.kt
+++ b/app/src/main/java/org/mozilla/focus/SettingsFragment.kt
@@ -25,16 +25,17 @@ class SettingsFragment : Fragment() {
         updateForAccessibility()
     }
 
-    override fun onCreateView(inflater: LayoutInflater?, container: ViewGroup?, savedInstanceState: Bundle?) =
-            inflater!!.inflate(R.layout.fragment_settings, container, false)
+    override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?) =
+            inflater.inflate(R.layout.fragment_settings, container, false)
 
-    override fun onViewCreated(view: View?, savedInstanceState: Bundle?) {
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        val context = view.context
 
         ic_lock.setImageResource(R.drawable.ic_lock)
-        telemetryButton.isChecked = DataUploadPreference.isEnabled(activity)
+        telemetryButton.isChecked = DataUploadPreference.isEnabled(context)
         telemetryView.setOnClickListener { _ ->
-            val newTelemetryState = !DataUploadPreference.isEnabled(activity)
-            DataUploadPreference.setIsEnabled(activity, newTelemetryState)
+            val newTelemetryState = !DataUploadPreference.isEnabled(context)
+            DataUploadPreference.setIsEnabled(context, newTelemetryState)
             telemetryButton.isChecked = newTelemetryState
         }
 
@@ -72,13 +73,13 @@ class SettingsFragment : Fragment() {
 
     override fun onStart() {
         super.onStart()
-        context.getAccessibilityManager().addTouchExplorationStateChangeListener(voiceViewStateChangeListener)
+        context?.getAccessibilityManager()?.addTouchExplorationStateChangeListener(voiceViewStateChangeListener)
         updateForAccessibility()
     }
 
     override fun onStop() {
         super.onStop()
-        context.getAccessibilityManager().removeTouchExplorationStateChangeListener(voiceViewStateChangeListener)
+        context?.getAccessibilityManager()?.removeTouchExplorationStateChangeListener(voiceViewStateChangeListener)
     }
 
     /**
@@ -94,7 +95,8 @@ class SettingsFragment : Fragment() {
         // cleared from this setting and nothing is selected. This is fine: the user can press
         // left-right to focus something else and it's an edge case that I don't think it is worth
         // adding code to fix.
-        val shouldFocus = !context.isVoiceViewEnabled()
+        val context = context ?: return
+        val shouldFocus = context.isVoiceViewEnabled()
         telemetryView.isFocusable = shouldFocus
         if (!shouldFocus) { telemetryView.clearFocus() }
     }
--- a/build.gradle
+++ b/build.gradle
@@ -1,7 +1,7 @@
 // Top-level build file where you can add configuration options common to all sub-projects/modules.
 
 buildscript {
-    ext.support_libraries_version = '26.1.0'
+    ext.support_libraries_version = '27.1.1'
     ext.architecture_components_version = '1.1.1'
     ext.espresso_version = '3.0.1'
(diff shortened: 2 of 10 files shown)


pocmo:
Aye, forgot about the taskcluster Docker image. This one doesn't update automatically.

pocmo:
I created a new Dockerfile based on the android-components one (with sdkmanager). However executing the tests fail (testAmazonWebViewBetaUnitTest) while building the image. This doesn't happen when executing the tests locally.

pocmo:
Okay, I think I have a working image now. I'll update the PR tomorrow and push the image with a version number so that it doesn't break other tasks.

mcomella:
k, will review when the PR is updated.

pocmo:
@mcomella Okay, I think the PR is ready now!

app/src/main/java/org/mozilla/focus/SettingsFragment.kt
```diff
@@ -94,9 +95,11 @@ class SettingsFragment : Fragment() {
         // cleared from this setting and nothing is selected. This is fine: the user can press
         // left-right to focus something else and it's an edge case that I don't think it is worth
         // adding code to fix.
-        val shouldFocus = !context.isVoiceViewEnabled()
-        telemetryView.isFocusable = shouldFocus
-        if (!shouldFocus) { telemetryView.clearFocus() }
+        context?.let {
```
mcomella: I don't love `?.let`: it's very easy to `.let` and break your program.

In this case, I think I'd prefer:
```kotlin
val context = context ?: return
...
```

Or something like that.

build.gradle
```diff
@@ -1,7 +1,7 @@
 // Top-level build file where you can add configuration options common to all sub-projects/modules.
 
 buildscript {
-    ext.support_libraries_version = '26.1.0'
+    ext.support_libraries_version = '27.1.1'
```
mcomella: btw, I discovered you can actually define these values inside the `dependencies { }` block in the other file, e.g.
```groovy
dependencies {
    def support_libraries_version = '27.1.1'
    implementation "...:$support_libraries_version"
```

For a code example, see https://github.com/mcomella/NonContactCallBlocker/blob/dce463a70a09932b5e7ee7a199215a184a80b1fe/app/build.gradle#L36

The downside to this approach is that you can only use the value in the dependencies block, which is fine for most deps but not kotlin_version (which is used by another plugin).

tools/docker/Dockerfile
```diff
@@ -5,27 +5,36 @@
 # Inspired by:
 # https://hub.docker.com/r/runmymind/docker-android-sdk/~/dockerfile/
 
-FROM ubuntu:17.10
+FROM ubuntu:18.04
 
-MAINTAINER Sebastian Kaspari "skaspari@mozilla.com"
+ENV ANDROID_BUILD_TOOLS "27.0.3"
+ENV ANDROID_SDK_VERSION "3859397"
```
mcomella: Where does this number come from? Perhaps add a comment where to get it.

tools/docker/Dockerfile
```diff
@@ -34,50 +43,38 @@ ENV LANG en_US.UTF-8
...
 WORKDIR /opt
 
 # Checkout source code
-RUN git clone https://github.com/mozilla-mobile/firefox-tv.git
+RUN git clone ${PROJECT_REPOSITORY}
 
 # Build project and run gradle tasks once to pull all dependencies
 WORKDIR /opt/firefox-tv/
-RUN ./gradlew clean assemble lint checkstyle ktlint pmd test
+RUN ./gradlew --no-daemon clean assemble lint checkstyle ktlint pmd detektCheck
```
mcomella: We don't want to run the unit tests too?

app/src/main/java/org/mozilla/focus/SettingsFragment.kt
pocmo: I do not really understand how `.let` will break the program. In this case the compiler would again complain that you are dealing with a nullable inside the block?

But I agree that your approach is nicer to read in this case. I'll update!

build.gradle
pocmo: > The downside to this approach is that you can only use the value in the dependencies block, which is fine for most deps but not kotlin_version (which is used by another plugin).

Yeah, and we may want to separate our code better by using multiple gradle modules in the future.

tools/docker/Dockerfile
pocmo: 👍

build.gradle
pocmo: Btw. in the Android components repository (where we have a bunch of gradle modules) I'm planning to move the whole dependency definition into the root build.gradle.

Something like:
```Groovy
ext.dependencies = [
    appcompat: "com.android.support:appcompat-v7:27.1.1"
    // ...
]
```

And then in the modules that use the dependency something like:
```Groovy
implementation rootProject.ext.dependencies['appcompat']
```

That's something I want to look into once we switch to the `androidx` builds of the support libraries where every library is going to have its own version.

tools/docker/Dockerfile
pocmo: I'm running into an issue where gradle gets killed (exit code 137) during test run - but only when building the Docker image. Running the tests in the built image works without any issues though. I'll file a follow-up issue. I think we have fixed a similar issue in Focus some time ago.

pocmo: #980

app/src/main/java/org/mozilla/focus/SettingsFragment.kt
pocmo: But why would I use `let` without using the object inside the block?

mcomella: Two cases I'd been concerned about are the code being refactored or a developer just accidentally starts using the optional type without thinking about it, e.g.:
```kotlin
maybeBrowserFragment.let {
    val webView = it?.webView
    return webView?.canGoBack() ?: false
}
```

I think this isn't that extreme of an example but I have hesitations it can get more confusing.

I think I also dislike let because its intentions are unclear. It stands in for multiple features that already exist:
- let without ? is reassignment, i.e. `val whatever = <let value>`
- let with ? is casting to non-optional types

One advantage it does have is that it groups together related concepts (with a scope; I'm assuming it actually returns a value though), e.g.:
```kotlin
val firstActiveToolbarButton = browserFragment?.let {
    val toolbar = it?.toolbar
    toolbar.children.firstOrNull { it.isActive() }
}
// Unrelated code.
```

But I think putting newlines between related concepts in code or pushing this functionality to helper functions does this just as well.

So overall, it feels like let is lacking intention and is redundant. I'm also concerned with how refactoring `let`s in and out of the code destroys blame (a problem I'm also having with `apply` and `also`) even if I do actually really like their conceptual groupings, assuming others are using them like I do.

I could be wrong about this, let me know what you think! :)

@colintheshots I'm wondering if you have an opinion on `let` too! :)

Issue #934, opened 2018-06-18
Confusing string for clearing cookies/data


liuche:
The string for clearing data may be confusing because it refers to strings that aren't present in the app.

Current string:
Selecting ‘Clear Now’ will clear all cookies and site data stored by Firefox for Fire TV. This may sign you out of websites and remove offline web content.

This string refers to a **Clear Now** that isn't present in the app (and also not one of our strings). The options in-app are "Ok" and "Cancel".

![device-2018-06-18-105251](https://user-images.githubusercontent.com/1082253/41553270-125ab900-72e6-11e8-9170-bfaf51ce7735.png)

PR #972, opened 2018-07-10
Issue #934 - Add L10N note describing button usage.

liuche:
Add some changes requested by Delphine, and also use these new strings.

### DIFF
--- a/app/src/main/res/layout/content_onboarding.xml
+++ b/app/src/main/res/layout/content_onboarding.xml
@@ -82,8 +82,7 @@
                 android:layout_marginStart="10dp"
                 android:layout_marginEnd="10dp"
                 android:background="@drawable/nav_button_selector"
-                android:text="@string/onboarding_turbo_mode_button_on"
-                tools:text="@string/onboarding_turbo_mode_button_on"
+                android:text="@string/button_turbo_mode_keep_enabled"
                 android:textColor="@color/tv_white"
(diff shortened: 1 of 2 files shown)


app/src/main/res/layout/content_onboarding.xml
```diff
@@ -98,8 +98,8 @@
                 android:layout_marginStart="10dp"
                 android:layout_marginEnd="10dp"
                 android:background="@drawable/nav_button_selector"
-                android:text="@string/onboarding_turbo_mode_button_off"
-                tools:text="@string/onboarding_turbo_mode_button_off"
+                android:text="@string/button_turbo_mode_turn_off"
+                tools:text="@string/button_turbo_mode_turn_off"
```
mcomella: nit: I don't think tools is necessary if we already have text.

app/src/main/res/layout/content_onboarding.xml
```diff
@@ -82,8 +82,8 @@
                 android:layout_marginStart="10dp"
                 android:layout_marginEnd="10dp"
                 android:background="@drawable/nav_button_selector"
-                android:text="@string/onboarding_turbo_mode_button_on"
-                tools:text="@string/onboarding_turbo_mode_button_on"
+                android:text="@string/button_turbo_mode_keep_enabled"
+                tools:text="@string/button_turbo_mode_keep_enabled"
```
mcomella: nit: I don't think tools is necessary if we already have text.

PR #974, opened 2018-07-10
Closes #965: Send play when paused and vice versa.

### LINKED ISSUE #965: Pause Command Is Resuming Playback
We tested the 2.2 APK on FireTV Cube.
Steps:

    "Alexa, open Youtube" - This will launch the YouTube.com app.
    Select Firefox. (version 2.2)
    Then, select a video to play and click on it to start playback.
    "Alexa, pause" - video paused
    Again say "Alexa pause"
    Expect Behavior: Nothing happens
    Observed Behavior: Video resumes
    Summary: It doesn't take utterances accurately and toggles between pause and play.
    Sometimes FF was crashing while doing pause and play.

https://github.com/mozilla-mobile/firefox-tv/issues/930#issuecomment-403700455

mcomella:
We used to toggle the state, which could lead to videos playing when you
say, "Alexa pause", and this is not acceptable.

### DIFF
--- a/app/src/main/java/org/mozilla/focus/browser/VideoVoiceCommandMediaSession.kt
+++ b/app/src/main/java/org/mozilla/focus/browser/VideoVoiceCommandMediaSession.kt
@@ -247,10 +247,21 @@ class VideoVoiceCommandMediaSession @UiThread constructor(
                 """.trimMargin())
         }
 
+        /**
+         * Called to play media.
+         *
+         * If a play or pause voice command is issued, the corresponding method
+         * will be called. In JS, if `play` is called on an already playing video, it's a no-op.
+         *
+         * If a play/pause hardware media button (it's one button, not two) is hit, it will send
+         * either [onPlay] or [onPause] as a toggle, based on current MediaSession playback state.
+         * We've already synced playback state from JS so the correct method should be chosen.
+         */
         override fun onPlay() {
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mozilla/focus/browser/VideoVoiceCommandMediaSession.kt
```diff
@@ -233,32 +236,25 @@ class VideoVoiceCommandMediaSession @UiThread constructor(
...
-                |    } else {
-                |       $ID_TARGET_VIDEO.pause();
-                |   }
+                |    $expressionToEval
                 |})();
                 """.trimMargin())
         }
 
+        override fun onPlay() {
+            evalJSWithTargetVideo { videoId -> "$videoId.play();" }
```
liuche: You don't think you need to check for paused() state anymore? What happens when people say the same thing twice "pause....pause" does it still work?

mcomella: No, I don't think we need to check: the API is designed to handle this.

Voice commands will do what you say. Calling a redundant `play/pause` on a DOM video will be a no-op.

If it's play/pause media button, it'll send the appropriate play/pause based on the current Java playback state, which we've already synced from JS.

I'll add a comment to explain this.

Issue #929, opened 2018-06-08
Use browser-session component for browser state


pocmo:
The component has not all features we need yet. However it's close already. I'll experiment a bit to see if there's something missing. This issue is mostly for tracking this work for now.

bbinto:
@pocmo will you be able to write and connect/integrate with the FFTV code? 

If so, could you please let us know what t-shirt size this issue will land at?

pocmo:
> @pocmo will you be able to write and connect/integrate with the FFTV code?

Yes, I'm doing the integration to see what parts are missing in our component. I'll open a PR at the end of it.

> If so, could you please let us know what t-shirt size this issue will land at?

That's definitely a Large.

mcomella:
Note: it might be good for library consumers to integrate the components to get feedback from a different perspective than the library author but it could also be helpful for Sebastian to implement the components himself to understand the requirements first hand.

@pocmo btw, we're not using the "multi-session" parts of our code, but they're still in the codebase, dangling (e.g. iirc, opening external intents will create more sessions but the back stack makes it act like one session), so if it's easier for you to remove those parts of the code, please do!

PR #978, opened 2018-07-11
Convert FocusApplication from Java to Kotlin.

pocmo:
Something I needed in #929. Just thought I could land this independently.


mcomella:
Failure was codecov which shouldn't regress this.

PR #979, opened 2018-07-11
Closes #963: Prevent infinite loop on `onSeekToNext`

### LINKED ISSUE #963: With Cube, 'Alexa, next/previous' causes the screen to go black, load continuously
### Steps to reproduce
- Open Firefox TV, launch youtube, play a video
- Play several videos, then say 'Alexa, next' or 'Alexa, previous' 

### Expected behavior
- Jumps to next or previous video

### Actual behavior
- Shows black screen, and periodically shows the loading indicator, but does not load any video
- This does not happen when remote is used to skip to next / previous video via graphics UI

### Device information
- Fire TV cube
* Firefox version: 2.2 release branch local build

mcomella:


app/src/main/java/org/mozilla/focus/browser/VideoVoiceCommandMediaSession.kt
```diff
@@ -157,6 +161,14 @@ class VideoVoiceCommandMediaSession @UiThread constructor(
         mediaSession.release()
     }
 
+    /**
+     * Potentially handles key events before they are received by the system for standard handling
```
liuche: Good comments here.

mcomella: I've been spending as much time writing the comments as the code, I think. :P

Issue #960, opened 2018-07-06
Use browser-domains component


pocmo:
This will allow us to remove the code and assets that deal with domains (e.g. autocomplete). Depends on Android SDK 27: #883.

### Why

Using browser-domains allows us to use the latest top domains for autocompletion and to have the same experience across our products.

### What

* Remove domain lists from assets
* Remove custom code in UrlAutoCompleteFilter

### Acceptance Criteria

* browser-domains is integrated and app contains no assets/code related to domains/autocompletion.

Issue #959, opened 2018-07-06
Use browser-search component


pocmo:
We can replace the search code with the browser-search component. We did the same some time ago in Focus. Nice side effect is that this will import the latest search plugins.

Depends on SDK 27: #883.

PR #982, opened 2018-07-12
Issue #929: Use browser-session component for browser state.

### LINKED ISSUE #929: Use browser-session component for browser state

pocmo:
Closes #929: This PR removes the custom session code from the app and replaces it with the `browser-session` component.

There are three missing pieces in the component (webview state, session source, blocking state) that I added via property extensions for now. I created upstream issues to get that into the component and once that has happened we can remove the extensions here.

For now this is almost a 1:1 replacement. In further iterations we can start to clean some things up. I want to look into how we can make usage of the component for "single-session" users a bit easier.

I did test this on a Fire TV 4k stick a bit. I seems to work fine. However as this touches a lot of areas I think this should get good test coverage by QA too.


app/src/main/java/org/mozilla/focus/Components.kt
```diff
@@ -0,0 +1,38 @@
...
+ */
+class Components {
+    val sessionManager by lazy { SessionManager(DummyEngine()) }
+}
+
+/**
+ * We are not using an "Engine" implementation yet. Therefore we create this dummy that we pass to
+ * the <code>SessionManager</code> for now.
+ */
+class DummyEngine: Engine {
```
sputnikci: \[ktlint\] WARNING: \[colon\-spacing\] Missing spacing before ":" in column 18

app/src/main/java/org/mozilla/focus/browser/FirefoxProgressBar.kt
```diff
@@ -26,23 +27,34 @@ private const val HIDE_ANIMATION_DURATION_MILLIS = 250L
...
-                if (isLoading) {
-                    showBar()
-                } else {
-                    scheduleHideBar()
-                }
+        session = browserFrag.session.also {
+            it.register(observer = this, owner = browserFrag)
+        }
+
+        browserFrag.session.register(observer =  this, owner = browserFrag)
```
sputnikci: \[ktlint\] WARNING: \[no\-multi\-spaces\] Unnecessary space\(s\) in column 49

pocmo:
Codecov: Unfortunately the coverage went down because there were some tests for the removed code. On the positive side the components have a good/better coverage. :)

Sputnik: That's something we saw in other repos too: If we remove files then sputnik fails because it cannot look at those files (obviously). This seems to be a sputnik bug.

pocmo:
> Sputnik: That's something we saw in other repos too: If we remove files then sputnik fails because it cannot look at those files (obviously). This seems to be a sputnik bug.

Filed an issue for that: https://github.com/TouK/sputnik-ci/issues/23

mcomella:
I'm trying to land some MediaSession stuff which needs to get uplifted to v2.2 so I'm going to review that after the core work is done (which should be today or tomorrow).

app/src/main/java/org/mozilla/focus/session/SessionCallbackProxy.java
```diff
@@ -24,7 +27,7 @@ public SessionCallbackProxy(Session session, IWebView.Callback delegate) {
     @Override
     public void onPageStarted(@NonNull String url) {
         session.setLoading(true);
-        session.setSecure(false);
+        session.setSecurityInfo(new Session.SecurityInfo(false, "", ""));
```
mcomella: No change necessary but I wonder what the performance effects of allocating here (and in onFinished is). Probably nothing because WebView (both Java and JS) probably does much worse than us :).

app/src/main/java/org/mozilla/focus/browser/FirefoxProgressBar.kt
```diff
@@ -26,23 +25,34 @@ private const val HIDE_ANIMATION_DURATION_MILLIS = 250L
...
 class FirefoxProgressBar @JvmOverloads constructor(
-        context: Context, attrs: AttributeSet? = null, defStyle: Int = 0)
-    : LinearLayout(context, attrs, defStyle) {
+    context: Context,
+    attrs: AttributeSet? = null,
+    defStyle: Int = 0
+) : LinearLayout(context, attrs, defStyle), Session.Observer {
 
     private val hideHandler = FirefoxProgressBarHideHandler(this)
+    private var session: Session? = null
```
mcomella: `lateinit var`? A progress bar shouldn't exist without a session so we should assert that state.

mcomella: fwiw, if the observers passed in the value they change (i.e. isLoading, url), we wouldn't need to keep a reference to session and potentially mismanage it.

app/src/main/java/org/mozilla/focus/browser/FirefoxProgressBar.kt
```diff
@@ -26,23 +25,34 @@ private const val HIDE_ANIMATION_DURATION_MILLIS = 250L
...
-        browserFrag.session.loading.observe(browserFrag, object : NonNullObserver<Boolean>() {
-            public override fun onValueChanged(value: Boolean) {
-                val isLoading = value
-                if (isLoading) {
-                    showBar()
-                } else {
-                    scheduleHideBar()
-                }
+        session = browserFrag.session.also {
+            it.register(observer = this, owner = browserFrag)
```
mcomella: Do lifecycleOwners work correctly with fragment's view hierarchies? e.g. A fragment creates a view hierarchy, removes the view hierarchy, and adds another one, what happens? Do we still receive events to the listeners of the invalid view hierarchy? Are those objects prevented from being GC'd?

app/src/main/java/org/mozilla/focus/browser/FirefoxProgressBar.kt
```diff
@@ -26,23 +25,34 @@ private const val HIDE_ANIMATION_DURATION_MILLIS = 250L
...
+        }
+
+        browserFrag.session.register(observer = this, owner = browserFrag)
+    }
+
+    override fun onLoadingStateChanged() {
+        session?.let {
```
mcomella: nit: In this case, to avoid indentation, I prefer:
```kotlin
val session = session ?: return
if (session.isLoading) { ...
```

app/src/main/java/org/mozilla/focus/browser/VideoVoiceCommandMediaSession.kt
```diff
@@ -101,16 +101,16 @@ class VideoVoiceCommandMediaSession @UiThread constructor(
...
+        session.register(sessionIsLoadingObserver, owner = activity)
         this.sessionIsLoadingObserver = sessionIsLoadingObserver
     }
 
     fun onDestroyWebView(webView: IWebView, session: Session) {
         webView.removeJavascriptInterface(JS_INTERFACE_IDENTIFIER)
         this.webView = null
 
-        session.loading.removeObserver(sessionIsLoadingObserver!!)
+        sessionIsLoadingObserver?.let { session.unregister(it) }
```
mcomella: This should be non-null so I think we should assert that.

app/src/gecko/java/org/mozilla/focus/iwebview/WebViewProvider.java
```diff
@@ -254,5 +244,13 @@ public boolean isYoutubeTV() {
             // TODO: Default implementation
             return false;
         }
+
+        @Override
+        public void saveWebViewState(@NonNull Session session) {
```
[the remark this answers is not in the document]
pocmo: Correct! I asked Android Studio to add the methods from the interface (different Session) and they ended up there. I'll move them back.

app/src/main/java/org/mozilla/focus/FocusApplication.kt
```diff
@@ -18,6 +18,8 @@ class FocusApplication : LocaleAwareApplication() {
     lateinit var visibilityLifeCycleCallback: VisibilityLifeCycleCallback
         private set
 
+    val components by lazy { Components() }
```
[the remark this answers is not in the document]
pocmo: > Is it because we expect it to take the ApplicationContext as an argument in the future? That seems fine then.

Yes, that's exactly the reason. It's going to be `Components(applicationContext)` because some components will need the application context (#959, #960).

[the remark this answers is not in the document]
pocmo: It's `lazy` because at the time `FocusApplication` gets created it's not a valid `Context` yet that can be used as a `Context` for various things. So I delay the creation until it's actually needed.

app/src/amazonWebview/java/org/mozilla/focus/webview/FirefoxAmazonWebView.kt
```diff
@@ -135,7 +136,7 @@ internal class FirefoxAmazonWebView(
         super.saveState(stateData)
         client.saveState(this, stateData)
 
-        session.saveWebViewState(stateData)
+        session.webViewState = stateData
```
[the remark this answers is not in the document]
pocmo: Okay, I'll rename it to `savedWebViewState`.

app/src/main/java/org/mozilla/focus/MainActivity.kt
```diff
@@ -43,12 +43,26 @@ interface MediaSessionHolder {
...
     override lateinit var videoVoiceCommandMediaSession: VideoVoiceCommandMediaSession
 
+    private val sessionObserver = object : SessionManager.Observer {
+        override fun onSessionSelected(session: Session) {
+            ScreenController.showBrowserScreenForCurrentSession(supportFragmentManager, session)
+        }
+
+        override fun onSessionRemoved(session: Session) {
+            // There's no active session. Start a new session with "homepage".
+            ScreenController.showBrowserScreenForUrl(this@MainActivity, supportFragmentManager, APP_URL_HOME, Source.NONE)
```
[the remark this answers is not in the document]
pocmo: Sounds good. I'll update this.

app/src/main/java/org/mozilla/focus/MainActivity.kt
```diff
@@ -43,12 +43,26 @@ interface MediaSessionHolder {
...
     // We crash if we init MediaSession at init time, hence lateinit.
     override lateinit var videoVoiceCommandMediaSession: VideoVoiceCommandMediaSession
 
+    private val sessionObserver = object : SessionManager.Observer {
+        override fun onSessionSelected(session: Session) {
+            ScreenController.showBrowserScreenForCurrentSession(supportFragmentManager, session)
+        }
+
+        override fun onSessionRemoved(session: Session) {
+            // There's no active session. Start a new session with "homepage".
```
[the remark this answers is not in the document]
pocmo: Okay, added the non-empty check.

app/src/main/java/org/mozilla/focus/browser/BrowserFragment.kt
```diff
@@ -103,21 +99,21 @@ class BrowserFragment : IWebViewLifecycleFragment() {
     private fun initSession() {
         val sessionUUID = arguments?.getString(ARGUMENT_SESSION_UUID)
                 ?: throw IllegalAccessError("No session exists")
-        session = if (sessionManager.hasSessionWithUUID(sessionUUID))
-            sessionManager.getSessionByUUID(sessionUUID)
-        else
-            NullSession()
+        session = components.sessionManager.findSessionById(sessionUUID) ?: Session("about:blank")
```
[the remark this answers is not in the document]
pocmo: Yes, this should be `NullSession`. This is from an earlier state where I didn't add `NullSession` back. I'll update.

app/src/main/java/org/mozilla/focus/browser/BrowserFragment.kt
```diff
@@ -103,21 +99,21 @@ class BrowserFragment : IWebViewLifecycleFragment() {
...
-                if (browserOverlay.isVisible) {
-                    browserOverlay.updateOverlayForCurrentState()
-                }
-            }
-        })
+
+        session.register(observer = this, owner = this)
+    }
+
+    override fun onUrlChanged() {
```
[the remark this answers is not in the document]
pocmo: Yep, that's what I thought too when refactoring this. And this already landed in the components repository and will be in the next release (https://github.com/mozilla-mobile/android-components/issues/406, https://github.com/mozilla-mobile/android-components/commit/46d47d14ad2b357b8a91e267948af98fd32c8a12)

app/src/main/java/org/mozilla/focus/browser/BrowserFragment.kt
```diff
@@ -242,7 +238,7 @@ class BrowserFragment : IWebViewLifecycleFragment() {
                 TelemetryWrapper.browserBackControllerEvent()
             }
             else -> {
-                SessionManager.getInstance().removeCurrentSession()
+                components.sessionManager.removeAll()
```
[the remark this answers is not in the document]
pocmo: Okay, changed it back to just remove the current session. Let's look at the multiple session bug separately.

app/src/main/java/org/mozilla/focus/ext/Fragment.kt
```diff
@@ -0,0 +1,16 @@
...
+import android.support.v4.app.Fragment
+import org.mozilla.focus.Components
+
+// Extension functions for the Fragment class
+
+/**
+ * Get the components of this application.
+ */
+val Fragment.components: Components
+    get() = context!!.components
```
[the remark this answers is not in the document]
pocmo: Okay, I made `Fragment.components` nullable and moved `!!` to the callers (There are just three).

app/src/main/java/org/mozilla/focus/session/NullSession.kt
```diff
@@ -0,0 +1,19 @@
...
+
+import mozilla.components.browser.session.Session
+
+/**
+ * Helper class for migrating to browser-session component. Eventually this class should get
+ * refactored away. It only exists to make code depending on a "NullSession" easier to migrate.
+ * At the latest when introducing the browser-engine component this should not be needed anymore.
+ */
+class NullSession private constructor() {
+    companion object {
```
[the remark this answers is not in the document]
pocmo: Currently `Session` is final.

app/src/test/java/org/mozilla/focus/session/SessionCallbackProxyTest.kt
```diff
@@ -0,0 +1,69 @@
...
+        val session = Session("about:blank")
+        val proxy = SessionCallbackProxy(session, mock(IWebView.Callback::class.java))
+
+        proxy.onURLChanged(TEST_URL)
+
+        assertEquals(TEST_URL, session.url)
+    }
+
+    companion object {
+        private val TEST_URL = "https://github.com/mozilla-mobile/focus-android/"
```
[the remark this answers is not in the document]
pocmo: OK.

app/src/main/java/org/mozilla/focus/ext/Session.kt
```diff
@@ -0,0 +1,62 @@
...
+
+/**
+ * The source of a session: How was this session created? E.g. did we receive an Intent or did the
+ * user start this session?
+ *
+ * Component upstream issue:
+ * https://github.com/mozilla-mobile/android-components/issues/407
+ */
+var Session.source: Source
+    get() = extensions.get(this)?.source ?: Source.NONE
```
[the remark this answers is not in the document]
pocmo: Aye, that's a good idea!

app/src/main/java/org/mozilla/focus/session/SessionCallbackProxy.java
pocmo: Yeah, probably negligible. onPageStarted() / onPageFinished() doesn't happen that often. We _could_ create constants as we are ignoring the `host` and `issuer` parts here.

app/src/main/java/org/mozilla/focus/browser/FirefoxProgressBar.kt
pocmo: 👍

pocmo: Yep. Comes with the next release of the components. :)

app/src/main/java/org/mozilla/focus/browser/FirefoxProgressBar.kt
pocmo: This works the same as the observe() on the LiveData in the code before: The observer stays registered until the lifecycle of the owner changes to DESTROYED. So it's at least not a regression.
Attaching observers to the "lifecycle" of a View is something that is so common that we added something for that too. We can attach it to FirefoxProgressBar instead of the Fragment. I'll update the PR.

app/src/main/java/org/mozilla/focus/browser/FirefoxProgressBar.kt
pocmo: With the `lateinit` change session is not nullable anymore. I'l update that.

app/src/main/java/org/mozilla/focus/browser/VideoVoiceCommandMediaSession.kt
pocmo: 👍

pocmo:
@mcomella Updated the PR and added a separate commit to address your review comments.

To address your review summary:

>  Components as a member of ApplicationContext

As mentioned in the comments: Components will need application context going forward. Especially for #959 and #960. We use a similar concept in our sample browser:
https://github.com/mozilla-mobile/android-components/blob/master/samples/browser/src/main/java/org/mozilla/samples/browser/Components.kt#L28

> Can we have multiple sessions? We should avoid changing logic around that interaction if it's possible

Yes, I changed everything back to handle multiple sessions. Let's address that separately. We are also looking into making single session use cases easier in the components.

> The session observer callbacks not passing in a reference to the changed values

As mentioned above: I had the same feeling when using the component here and a patch for that already landed in the components repository.
* https://github.com/mozilla-mobile/android-components/issues/406
* https://github.com/mozilla-mobile/android-components/commit/46d47d14ad2b357b8a91e267948af98fd32c8a12

> How lifecycle owners work for Fragments with view hierarchies

This was a 1:1 conversion from the previous code where the LiveData observer was bound to the fragment too. I updated this code to bind the observer to the view instead of the Fragment.

app/src/main/java/org/mozilla/focus/FocusApplication.kt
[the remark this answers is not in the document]
mcomella: I think I'd prefer `lateinit` with initialization in `onCreate`: that way it's explicit after which point it's intended to be used.

But if you prefer lazy, please add a comment explaining why it's necessary (not a valid Context yet).

pocmo:
@mcomella Updated the PR (added the comment to FocusApplication)!

pocmo:
... and rebased the PR. Let me know if there are more things. Otherwise I'll squash the commits.

mcomella:
Sorry, I forgot to get to this today: I'll get to it tomorrow.

app/src/main/java/org/mozilla/focus/MainActivity.kt
```diff
@@ -53,13 +53,13 @@ class MainActivity : LocaleAwareAppCompatActivity(), OnUrlEnteredListener, Media
         }
 
         override fun onSessionRemoved(session: Session) {
-            // There's no active session. Start a new session with "homepage".
-            ScreenController.showBrowserScreenForUrl(this@MainActivity, supportFragmentManager, APP_URL_HOME, Source.NONE)
+            if (components.sessionManager.sessions.isEmpty()) {
```
mcomella: The current logic is:
```kotlin
if (sessionsIsEmpty) 
    showBrowserScreenForHome()
else
    showBrowserScreenForCurrentUrl()
```

The logic in this method isn't the same. Is the logic the same because `onSessionSelected` will be called if `onSessionRemoved` is called? If so, this is good; if not, please fix it!

app/src/main/java/org/mozilla/focus/MainActivity.kt
```diff
@@ -204,6 +204,11 @@ class MainActivity : LocaleAwareAppCompatActivity(), OnUrlEnteredListener, Media
                 super.dispatchKeyEvent(event)
     }
 
+    private fun onNoActiveSession() {
```
mcomella: nit: you can define this in the session observer (for clearer encapsulation of who wants to call this)

app/src/main/java/org/mozilla/focus/ext/Fragment.kt
```diff
@@ -12,5 +12,5 @@ import org.mozilla.focus.Components
 /**
  * Get the components of this application.
  */
-val Fragment.components: Components
-    get() = context!!.components
+val Fragment.components: Components?
```
mcomella: nit: I still don't love that it's unclear why `Fragment.components` can be null - that it's tied to the Context. I think I might actually prefer it to be `val Context.components: Components` (i.e. non-optional) to make it clear that it's the context we need to look out for.

app/src/test/java/org/mozilla/focus/session/SessionCallbackProxyTest.kt
```diff
@@ -12,6 +12,8 @@ import org.mockito.Mockito.mock
 import org.mozilla.focus.iwebview.IWebView
 import org.robolectric.RobolectricTestRunner
 
+private val TEST_URL = "https://github.com/mozilla-mobile/focus-android/"
```
mcomella: nit: `private const val` :P I'm not sure what the value of `const` is though, tbh... compiler note?

app/src/main/java/org/mozilla/focus/MainActivity.kt
pocmo: >  Is the logic the same because onSessionSelected will be called if onSessionRemoved is called? 

Yes! If the removed session was the selected session then `SessionManager` will determine the next selected session (like in Fennec) and will notify observers via `onSessionSelected()`.

app/src/main/java/org/mozilla/focus/MainActivity.kt
pocmo: I'll update.

app/src/test/java/org/mozilla/focus/session/SessionCallbackProxyTest.kt
pocmo: Was curious and decompiled to Java:

WIthout `const`:

```Java
public final class SessionCallbackProxyTestKt {
   private static final String TEST_URL = "https://github.com/mozilla-mobile/focus-android/";

   // $FF: synthetic method
   @NotNull
   public static final String access$getTEST_URL$p() {
      return TEST_URL;
   }
}
```

With `const`:

```Java
public final class SessionCallbackProxyTestKt {
   private static final String TEST_URL = "https://github.com/mozilla-mobile/focus-android/";
}
```

app/src/main/java/org/mozilla/focus/ext/Fragment.kt
pocmo: Okay, I removed this extension completely and access `components` via a `Context` in the fragments now.

pocmo:
@mcomella Adressed your last comments. Updated and squashed the commit. Rebased. And just tested again a build of that a bit on a Fire TV 4k stick. I think that's ready to merge now.

pocmo:
✅ Taskcluster and travis seem to agree. Landing!

PR #986, opened 2018-07-13
Closes #983, Closes #962: Pause video immediately after play if paused issued whil…

### LINKED ISSUE #962: On 2.2, it briefly resumes half a second after 'Alexa pause' command. (Fire TV cube)
### Steps to reproduce
- Open Firefox TV app
- Go to youtube, play a video
- Give voice command 'Alexa, pause'

### Expected behavior
- Pauses immediately, and does not resume

### Actual behavior
- Pauses immediately, then it resumes for a split second, then it pauses again and stays paused
- This does not happen when button UI is used

### Device information

* Fire TV device: 
Fire TV cube
* Firefox version: 
Locally built 2.2 release branch

### LINKED ISSUE #983: "Pause" with hardware remote on 4K does not work correctly
### Steps to reproduce
- Install MediaSession build on 4K
- Click microphone button on hardware remote
- Say "pause"

### Expected behavior
Video pauses

### Actual behavior
Video plays

The Fire TV is sending the MediaSession pause event while the video is paused for the voice command so our pause command is undone when the video resumes:
```
07-12 16:29:07.519 8362-8436/org.mozilla.tv.firefox.debug I/chromium: [INFO:HTMLMediaElement.cpp(2479)] Pause video
07-12 16:29:09.704 8362-8362/org.mozilla.tv.firefox.debug D/lol: pause called
07-12 16:29:09.710 8362-8436/org.mozilla.tv.firefox.debug I/chromium: [INFO:HTMLMediaElement.cpp(2479)] Pause video
07-12 16:29:09.752 8362-8362/org.mozilla.tv.firefox.debug D/GeckoLocales: No selected locale. No correction needed.
07-12 16:29:09.788 8362-8362/org.mozilla.tv.firefox.debug D/AudioManager: AudioManager dispatching onAudioFocusChange(1) for com.amazon.media.AmazonAudioManager@642c7f8org.chromium.content.browser.AudioFocusDelegate@52ce4d1
07-12 16:29:09.806 8362-8436/org.mozilla.tv.firefox.debug I/chromium: [INFO:HTMLMediaElement.cpp(2436)] Play video
```

This is not the behavior we see on the Cube, except when the manifest is changed to implement #962.

mcomella:
…e paused.

See comments for details.

This solution isn't perfect: as expressed in the comments, there is an
opportunity for us to pause a video that was intentionally played.

One possible alternative would be to pause all videos when voice
commands are issued and resume them when the voice command completes
but, if my theory is correct that audio focus changes are what cause
the videos to pause, this is called before any callbacks we have access
to and it's not possible to implement.

### DIFF
--- a/app/src/main/java/org/mozilla/focus/browser/VideoVoiceCommandMediaSession.kt
+++ b/app/src/main/java/org/mozilla/focus/browser/VideoVoiceCommandMediaSession.kt
@@ -94,8 +94,8 @@ class VideoVoiceCommandMediaSession @UiThread constructor(
     private var webView: IWebView? = null
     private var sessionIsLoadingObserver: SessionIsLoadingObserver? = null
 
-    private var isResumed = false
-    private var isStarted = false
+    private var isLifecycleResumed = false
+    private var isLifecycleStarted = false
 
     init {
         mediaSession.setCallback(MediaSessionCallbacks())
@@ -123,17 +123,17 @@ class VideoVoiceCommandMediaSession @UiThread constructor(
 
     @OnLifecycleEvent(ON_RESUME)
     fun onResume() {
-        isResumed = true
+        isLifecycleResumed = true
     }
 
     @OnLifecycleEvent(ON_PAUSE)
     fun onPause() {
-        isResumed = false
+        isLifecycleResumed = false
     }
 
     @OnLifecycleEvent(ON_START)
     fun onStart() {
-        isStarted = true
+        isLifecycleStarted = true
 
         // We want to make our MediaSession active: state buffering is more accurate than state
         // playing. For an explanation of MediaSession (in)active states, see class javadoc.
@@ -149,7 +149,7 @@ class VideoVoiceCommandMediaSession @UiThread constructor(
 
     @OnLifecycleEvent(ON_STOP)
     fun onStop() {
-        isStarted = false
+        isLifecycleStarted = false
 
         // Videos playing when the app was backgrounded get into an inconsistent state: `paused`
         // will return false but they won't actually be playing. This makes any user "play" commands
@@ -215,7 +215,7 @@ class VideoVoiceCommandMediaSession @UiThread constructor(
             // method. In theory, since the JS is async, this could undo the playback state we set
             // in onStop so we ignore these updates. In practice, this method doesn't appear to be
             // called from that JS but we leave this in for safety.
-            if (!isStarted) { return }
+            if (!isLifecycleStarted) { return }
 
             val playbackStateInt: Int
             val positionMillis: Long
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/mozilla/focus/browser/VideoVoiceCommandMediaSession.kt
```diff
@@ -300,7 +312,41 @@ class VideoVoiceCommandMediaSession @UiThread constructor(
 
         // See onPlay for details.
         override fun onPause() {
-            evalJSWithTargetVideo { videoId -> "$videoId.pause();" }
+            fun getJS(videoId: String) = if (isResumed) {
+                "$videoId.pause();"
+            } else {
+                // If we receive a MediaSession callback will the app is paused, it's coming from a
```
liuche: will => while

app/src/main/java/org/mozilla/focus/browser/VideoVoiceCommandMediaSession.kt
```diff
@@ -118,6 +121,16 @@ class VideoVoiceCommandMediaSession @UiThread constructor(
         this.sessionIsLoadingObserver = null
     }
 
+    @OnLifecycleEvent(ON_RESUME)
```
liuche: Maybe include a comment that onResume and onPause are related to interruptions of the voice commands, because I think it can be confusing since we also have onStart onStop, for actual play/pause events.

app/src/main/java/org/mozilla/focus/browser/VideoVoiceCommandMediaSession.kt
```diff
@@ -92,6 +94,7 @@ class VideoVoiceCommandMediaSession @UiThread constructor(
     private var webView: IWebView? = null
     private var sessionIsLoadingObserver: SessionIsLoadingObserver? = null
 
+    private var isResumed = false
```
liuche: Hm, I wonder if `isInterrupted` is clearer?

app/src/main/java/org/mozilla/focus/browser/VideoVoiceCommandMediaSession.kt
mcomella: This feels like overkill to me. The way I see it, if someone wants to know why these methods exist, they'd see what flag changes in it (`isResumed`) and see what uses that flag for why the flag needs to exist (and thus the methods).

Another concerns with adding a comment are:
- The boolean flag stands on its own and represents a state (`isResumed`) such that others can use it: the comment makes less sense if there are multiple consumers
- If a consumer is added and this consumer is removed, the comment may not be removed and then the comment is out-of-date. I'd rather comment directly on where the flag is used (as I have done).

app/src/main/java/org/mozilla/focus/browser/VideoVoiceCommandMediaSession.kt
mcomella: I think this should stay `isResumed`: see my above comment. I could clarify this in the place where it's used (onPause) by giving it another name, e.g.
```kotlin
fun onPause() {
    val isInterruptedByVoiceCommand = !isResumed
    fun getJS(videoId: String) = if (isInterruptedByVoiceCommand) {
    }
}
```

I'll do that.

PR #993, opened 2018-07-17
Closes #992: Don't require SDK23 for voice control.

### LINKED ISSUE #992: Support Fire Stick by lowering permissions
We have permissions set to 23+ but Fire Stick is lower (22?) so we should remove that restriction.

mcomella:


app/src/main/AndroidManifest.xml
```diff
@@ -11,10 +11,9 @@
     <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
     <uses-permission android:name="com.android.launcher.permission.INSTALL_SHORTCUT" />
 
-    <!-- Required for voice control. It's only available on the Cube (so far) which is API 25+
-         so we use sdk-23 to prevent users from manually having to update the app.
-         https://developer.amazon.com/blogs/appstore/post/623addd5-6184-4e7f-a969-123a653cfb1a/implementing-voice-control-with-the-media-session-api-on-amazon-fire-tv -->
-    <uses-permission-sdk-23 android:name="com.amazon.permission.media.session.voicecommandcontrol" />
+    <!-- Required for media playback voice control.
+           https://developer.amazon.com/blogs/appstore/post/623addd5-6184-4e7f-a969-123a653cfb1a/implementing-voice-control-with-the-media-session-api-on-amazon-fire-tv -->
+    <uses-permission android:name="com.amazon.permission.media.session.voicecommandcontrol" />
```
liuche: Okay, so is it is available on FireStick, API 22?

mcomella: Yes, but it's buggy: we're landing so we can test and I'm going to file a follow-up to decide whether we want it or not.

mcomella: Filed https://github.com/mozilla-mobile/firefox-tv/issues/997

mcomella:
> (I don't remember what the conclusion on permissions bumps was, did we figure it out or is that something we should also file to make sure people can actually get the new update?)

I don't think there's much we can do here; users get the notifications they normally get when they need to update an app and that's that. I don't like that we're potentially further fragmenting our user base but it's good that users will get an additional update notification (I assume) for the new version. I'll leave this decision up to #997.

PR #1014, opened 2018-08-14
Closes #957: Update okhttp to 3.11.0

### LINKED ISSUE #957: java.lang.NullPointerException: Attempt to invoke virtual method 'void okhttp3.Cache.evictAll()' on a null object reference
The type signature is incorrect, it seems: `Cache` can return null:
```kotlin
try { client.cache().evictAll() } catch (_: IOException) { /* We don't care. */ }
```

```
    java.lang.NullPointerException: Attempt to invoke virtual method 'void okhttp3.Cache.evictAll()' on a null object reference
        at org.mozilla.focus.utils.OkHttpWrapper.onLowMemory(OkHttpWrapper.kt:16)
        at org.mozilla.focus.FocusApplication.onLowMemory(FocusApplication.java:60)
        at android.app.ActivityThread.handleLowMemory(ActivityThread.java:4343)
        at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1446)
```

This happened on a Fire Stick while fast-forwarding and rewinding Youtube shortly after rebooting the device.

psymoon:
Add non-null assertion operator

### DIFF
--- a/app/src/main/java/org/mozilla/focus/utils/OkHttpWrapper.kt
+++ b/app/src/main/java/org/mozilla/focus/utils/OkHttpWrapper.kt
@@ -13,6 +13,12 @@ object OkHttpWrapper {
 
     @JvmStatic
     fun onLowMemory() {
-        try { client.cache().evictAll() } catch (_: IOException) { /* We don't care. */ }
+        try { client.cache()!!.evictAll() }
+        catch (ex:Exception) { /* We don't care. */
+            when (ex) {
+                is IOException -> {}
+                is NullPointerException -> {} // Issue 957
+            }
+        }
     }
 }
--- a/app/src/main/java/org/mozilla/focus/utils/OkHttpWrapper.kt
+++ b/app/src/main/java/org/mozilla/focus/utils/OkHttpWrapper.kt
@@ -13,12 +13,8 @@ object OkHttpWrapper {
 
     @JvmStatic
     fun onLowMemory() {
-        try { client.cache()!!.evictAll() }
-        catch (ex:Exception) { /* We don't care. */
-            when (ex) {
-                is IOException -> {}
-                is NullPointerException -> {} // Issue 957
-            }
-        }
+        try {
+            client?.cache()?.evictAll()
+        } catch (_: IOException) { /* We don't care. */ }
     }
 }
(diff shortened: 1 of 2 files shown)


mcomella:
As mentioned on Slack, will review when this is fixed or this is split apart. The commit format isn't quite right though: before sure to read through https://github.com/mozilla-mobile/shared-docs/blob/master/android/CONTRIBUTING_code.md !

Also, we rebase and merge, not squash commits, so we generally don't want commits that undo themselves - can you remove the two ktlint commits from this PR? Thanks!

app/src/main/java/org/mozilla/focus/utils/OkHttpWrapper.kt
```diff
@@ -13,6 +13,12 @@ object OkHttpWrapper {
 
     @JvmStatic
     fun onLowMemory() {
-        try { client.cache().evictAll() } catch (_: IOException) { /* We don't care. */ }
+        try { client.cache()!!.evictAll() }
```
mcomella: We can write more correct, concise code with null chaining. Check out the docs to figure it out :):  https://kotlinlang.org/docs/reference/null-safety.html Let me know if you have questions.

In general, it's bad to execute unsafe calls like `!!`. You really only want to use `!!` when you're sure the value will never be null. And even then, you generally want to avoid it because if the code changes, your assumptions about the value never being null can be invalidated and the program can crash.

mcomella:
btw, the issue number should be in the commit summary, not the extended body, e.g.: `Closes #957: Update okhttp to 3.11.0 and add non-null assertion operator `

app/src/main/java/org/mozilla/focus/utils/OkHttpWrapper.kt
psymoon: Please review https://github.com/psymoon/firefox-tv/commit/905d4996d6fa17d547ad3ff09d092789b2520475

mcomella:
@psymoon FYI: to ask me to review again after rejection, it's best to click on the "reviewers" column and select me again: that'll put the review in my dashboard.

mcomella:
I took care of it this time (so I don't forget about this) but for future notice!

Issue #837, opened 2018-04-20
Use "focus ring" for focused state in pocket feed view (others?)


mcomella:
### Why/User Benefit/User Problem
- Users expect a level of polish and consistent design experience: this is the design Amin wants to move forward with

### What / Requirements
In https://github.com/mozilla-mobile/firefox-tv/issues/759#issuecomment-382419513, Amin suggested we change the focused state to use a "focus ring":
<img width="1512" alt="screen shot 2018-04-18 at 6 04 03 pm" src="https://user-images.githubusercontent.com/1721875/38943973-f54908ae-4332-11e8-9e05-6adf0be062d0.png">

Amin mentioned:

> If it requires more time than estimated we can add the focus ring in a later release.

This is more work than the original mock so I finished the original implementation and filed this follow-up. Also, Amin mentioned:

> Also this addition can be part of a bigger polish where we bring the same focus state across the whole app.

So perhaps we'll want to create a UI widget that we can re-use throughout the app so many views can have a "focus ring".

### Acceptance Criteria (how do I know when I’m done?)
- Implement it

aminalhazwani:
@mcomella thanks for filing this!

> What were you thinking for "a bigger polish where we bring the same focus state across the whole app"?

> Do you expect the home screen view focused to change too?

Yes, I believe that there's a good opportunity for bringing the same focus state across the whole app - and across different firefox products. I chatted with @shorlander and we shared some ideas on how that might look. For now, I only played with the focus ring but maybe we would also want to add a subtle zoom to the focused elements, in the meantime here some explorations:

Focused card

<img width="323" alt="screen shot 2018-04-20 at 2 27 08 pm" src="https://user-images.githubusercontent.com/1721875/39050781-f107a8d4-44a6-11e8-97a6-5b03560b2c7e.png">

Focused address bar

<img width="563" alt="screen shot 2018-04-20 at 2 22 32 pm" src="https://user-images.githubusercontent.com/1721875/39050635-6a7b6468-44a6-11e8-828e-f088ec01b881.png">

Focused mega tile

<img width="551" alt="screen shot 2018-04-20 at 2 23 01 pm" src="https://user-images.githubusercontent.com/1721875/39050636-6a94b1b6-44a6-11e8-8633-13ff53b6bff7.png">

Focused tile

[...]

PR #1024, opened 2018-08-20
Closes #818: Change colours on homescreen

### LINKED ISSUE #818: Revisit colors on overlay/homescreen
*edit*: Make sure the colors in the overlay match the colors in Amin's https://github.com/mozilla-mobile/firefox-tv/issues/818#issuecomment-413206116

---

Now that the homescreen is combined with the overlay which has a darker background, we should revisit/audit these colors.

- Navigation urlbar background color is currently #cc0c0c15 which blends in completely with the overlay
- Tile "blue highlight" is slightly different from the nav button highlight
- "Press" color for tiles
- default "non-selected" tint for tiles (it's slightly less bright than the selected white)
- Any other colors (greyed-out navigation button state looks different in the mocks from what he have)

For reference, here is the [overlay layout file](https://github.com/mozilla-mobile/firefox-tv/blob/master/app/src/main/res/layout/browser_overlay.xml), and the colors can be found in [colors](https://github.com/mozilla-mobile/firefox-tv/blob/master/app/src/main/res/values/colors.xml) and [tv_colors](https://github.com/mozilla-mobile/firefox-tv/blob/master/app/src/main/res/values/tv_colors.xml)

dnarcese:


dnarcese:
@shorlander Would you mind reviewing the UX, since Amin is on PTO?

app/src/main/res/drawable/ic_search.xml
```diff
@@ -8,6 +8,6 @@
         android:viewportWidth="24.0"
         android:viewportHeight="24.0">
     <path
-        android:fillColor="#ffffff"
+        android:fillColor="#99F9F9FA"
```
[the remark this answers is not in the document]
liuche: I think the search icon is used in other places, so we should definitely only change the color for this place where the spec asks for it.

app/src/main/res/drawable/pocket_video_mega_tile_background.xml
```diff
@@ -17,7 +17,7 @@
 
     <item android:state_focused="true">
         <shape>
-            <solid android:color="@color/pocket_coral"/>
+            <solid android:color="@color/tv_grey80"/>
```
liuche: Same as previous comments about wrapping colors.

app/src/main/res/values/colors.xml
```diff
@@ -27,13 +27,12 @@
...
-    <color name="button_default">@color/translucent_tv_ink</color>
-    <color name="button_disabled">@color/tv_ink</color>
-    <color name="button_focused">@color/tv_blue50</color>
+    <color name="button_highlight">#00ffffff</color>
+    <color name="button_default">@android:color/transparent</color>
+    <color name="button_disabled">@android:color/transparent</color>
     <color name="button_pressed">@color/tv_blue60</color>
-    <color name="button_checked">@color/translucent_tv_ink2</color>
-    <color name="button_border">@color/translucent_tv_gray40</color>
+    <color name="button_checked">@color/tv_grey90</color>
```
liuche: Same, just use this color directly.

app/src/main/res/values/colors.xml
```diff
@@ -27,13 +27,12 @@
 
     <color name="pocket_onboarding_background">#e60c0c0d</color>
 
-    <color name="button_highlight">#64ffffff</color>
-    <color name="button_default">@color/translucent_tv_ink</color>
-    <color name="button_disabled">@color/tv_ink</color>
-    <color name="button_focused">@color/tv_blue50</color>
+    <color name="button_highlight">#00ffffff</color>
```
liuche: #00xxxxxx just means that the color is transparent, because it's at 00 transparency. If a hex color has 8 characters, the first two indicate the transparency. If there are only 6, then the transparency is assumed to be FF or 100%.

app/src/main/res/color/overlay_button_border.xml
```diff
@@ -5,6 +5,7 @@
 
 <selector xmlns:android="http://schemas.android.com/apk/res/android">
     <item android:state_pressed="true" android:color="@color/button_pressed" />
-    <item android:state_focused="true" android:color="@color/button_focused" />
+    <item android:state_focused="true" android:color="@color/tv_blue50" />
```
liuche: Same, just use these colors directly from the photon colors without wrapping/aliasing them.

app/src/main/res/color/overlay_button_selector.xml
```diff
@@ -6,7 +6,6 @@
 <selector xmlns:android="http://schemas.android.com/apk/res/android">
     <item android:state_enabled="false" android:color="@color/button_disabled"/>
     <item android:state_pressed="true" android:color="@color/button_pressed" />
-    <item android:state_focused="true" android:color="@color/button_focused" />
```
liuche: Okay, so you're deleting this line because in this case, focusing a button doesn't change it's background color ✅

app/src/main/res/drawable/turbo_off.xml
```diff
@@ -4,6 +4,6 @@
...
+    <path android:fillColor="#FFFFFFFF" android:pathData="M22,3a1,1 0,0 0,-1 -1C16.3,2 10.873,5.66 7.089,10H4a1,1 0,0 0,-0.895 0.553l-1,2a1,1 0,0 0,0.579 1.395l1.67,0.557a1,1 0,0 0,0.111 0.212,21.8 21.8,0 0,0 4.818,4.818 0.791,0.791 0,0 0,0.082 0.042l0.7,1.789a1,1 0,0 0,1.378 0.529l2,-1A1,1 0,0 0,14 20V16.911C18.34,13.127 22,7.705 22,3ZM12.326,15.729A24.182,24.182 0,0 1,9.909 17.5,19.674 19.674,0 0,1 6.5,14.09a24.65,24.65 0,0 1,1.761 -2.4C11.623,7.659 16.23,4.6 19.921,4.079 19.4,7.765 16.349,12.365 12.326,15.729Z"/>
+    <path android:fillColor="#FFFFFFFF" android:pathData="M17.33,7.578a4.906,4.906 0,0 0,-0.915 -0.913,0.416 0.416,0 0,0 -0.378,-0.061A10.162,10.162 0,0 0,13.829 7.92a0.417,0.417 0,0 0,-0.041 0.625l1.667,1.667a0.42,0.42 0,0 0,0.295 0.121h0.029a0.414,0.414 0,0 0,0.3 -0.164,11.312 11.312,0 0,0 1.309,-2.189A0.422,0.422 0,0 0,17.33 7.578Z"/>
```
liuche: Solid white is okay, because it's easy to tint over white.

app/src/main/res/values/colors.xml
```diff
@@ -27,13 +27,12 @@
...
+    <color name="button_default">@android:color/transparent</color>
```
liuche: I'm a little on the fence about whether we should wrap/alias these if they're just `transparent`.

On the one hand, if we decide to change the button default color from transparent, we can just change it here and it does the right thing. But on the other hand, aliasing a color to transparent doesn't really make a ton of sense because we don't "need" the color to be defined in `colors.xml` we already know what it is.

I took a quick look at Focus, and I don't see any colors that wrap `transparent`, so I'm going to say that we should follow the pattern and not include transparent aliases in `colors.xml` but just use them directly.

app/src/main/res/color/overlay_button_border.xml
```diff
@@ -5,6 +5,7 @@
...
+    <item android:state_checked="true" android:color="@color/button_checked"/>
```
liuche: Do you need to include `checked` here? I think it should just fall through to the default color (below).

liuche (changes requested):
Good patch, I think it's almost there!

I tried to be thorough in my comments, explaining both my thinking process and providing background context, so please don't be alarmed that there are so many long comments!

Also, so it's easier for @shorlander to review, can you include a screenshot that match the state of the mock in the bug? i.e. refresh focused, back enabled, forward disabled, pinned+, turbo- (turbo is the only difference from the mock, but I'd like to see what it looks like to get all the states).

app/src/main/res/values/tv_colors.xml
```diff
@@ -11,7 +11,9 @@
     <color name="tv_blue50" tools:keep="@color/tv_blue50">@color/photonBlue50</color>
     <color name="tv_blue60" tools:keep="@color/tv_blue60">@color/photonBlue60</color>
     <color name="tv_ink" tools:keep="@color/tv_ink">#0c0c15</color>
-    <color name="translucent_tv_ink" tools:keep="@color/translucent_tv_ink">#cc0c0c15</color>
+    <color name="translucent_tv_ink" tools:keep="@color/translucent_tv_ink">#2a2a2e</color>
```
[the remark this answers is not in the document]
dnarcese: So, since we are now using the photon colour palette, can I just delete the `tv_colors.xml` file and use all those colours directly? And then `colors.xml` will just be for any colours not already in the photon palette.

app/src/main/res/color/overlay_button_border.xml
dnarcese: Yes- the default border is transparent, but the checked border needs to be black.

dnarcese:
Here is a screenshot with every state:
image

liuche (approved):
lgtm! Just one question about the ic_search tinting, but if that's addressed this is good to land.

liuche:
@dnarcese one last thing - I don't see the second blue border on the "highlighted" button, but you can file something for that, bc it looks like we might not already have that feature and this is mainly a color change (unless you want to add that second ring).

The colors for icons also visually look a little different from the mock, but they match in hex so that is fine - we can go over it again with @aminalhazwani when he gets back, but the code matches the spec.

dnarcese:
@liuche I think the double border would be addressed in #837

liuche:
@dnarcese one last thing - I also noticed that the pin button and rocket button are "facing" different ways - can you mirror the icon so they match the mock?

Then I think we can ping @shorlander again to take a look and get this landed.

README.md
```diff
@@ -39,7 +39,7 @@ we're available Monday-Friday, GMT and PST working hours.
   ./gradlew clean app:assembleAmazonWebviewDebug
```
psymoon: Can you change this to ./gradlew clean app:debug?

liuche:
Rebased and re-pushed as part of #1074, so closing this PR because it's the same so as not to be confusing.

PR #1026, opened 2018-08-23
Closes #995: Add DpadKeyMovementMethod

### LINKED ISSUE #995: URL moves when moving back up to URL bar from Pocket tile
### Steps to reproduce

1. Visit a page with a long URL
2. Press Menu
3. URL bar is focused
4. Press Down / Go to pocket tile
5. Press Up / Go back to URL bar

### Expected behavior

URL stays in the same position showing the "root" domain

### Actual behavior

URL show usually unhelpful last characters

### Device information

* Fire TV device: Cube
* Firefox version: Master branch, latest commit 6c9e41a

psymoon:
Extends from ArrowKeyMovementMethod, remove end cursor selection call from directional onTakeFocus()

### DIFF
--- a/app/src/main/java/org/mozilla/focus/widget/DpadKeyMovementMethod.kt
+++ b/app/src/main/java/org/mozilla/focus/widget/DpadKeyMovementMethod.kt
@@ -1,5 +1,4 @@
-/* -*- Mode: Java; c-basic-offset: 4; tab-width: 20; indent-tabs-mode: nil; -*-
- * This Source Code Form is subject to the terms of the Mozilla Public
+/* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
--- a/app/src/main/java/org/mozilla/focus/widget/DpadKeyMovementMethod.kt
+++ b/app/src/main/java/org/mozilla/focus/widget/DpadKeyMovementMethod.kt
@@ -5,20 +5,13 @@
 
 package org.mozilla.focus.widget
 
-import android.text.Selection
 import android.text.Spannable
 import android.text.method.ArrowKeyMovementMethod
-import android.view.View
 import android.widget.TextView
 
+/**
+ * An ArrowKeyMovementMethod that eliminates end cursor selection onFocus
+ */
 class DpadKeyMovementMethod : ArrowKeyMovementMethod() {
-
-    override fun onTakeFocus(view: TextView?, text: Spannable?, dir: Int) {
-        if (dir and (View.FOCUS_FORWARD or View.FOCUS_DOWN) != 0) {
-            if (view?.layout == null) {
-                // This shouldn't be null, but do something sensible if it is.
-                Selection.setSelection(text, text!!.length)
-            }
-        }
-    }
+    override fun onTakeFocus(view: TextView, text: Spannable, dir: Int) {}
 }
--- a/app/src/main/java/org/mozilla/focus/widget/DpadKeyMovementMethod.kt
+++ b/app/src/main/java/org/mozilla/focus/widget/DpadKeyMovementMethod.kt
@@ -1,0 +1,24 @@
+/* -*- Mode: Java; c-basic-offset: 4; tab-width: 20; indent-tabs-mode: nil; -*-
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.focus.widget
+
+import android.text.Selection
+import android.text.Spannable
+import android.text.method.ArrowKeyMovementMethod
+import android.view.View
+import android.widget.TextView
+
+class DpadKeyMovementMethod : ArrowKeyMovementMethod() {
+
+    override fun onTakeFocus(view: TextView?, text: Spannable?, dir: Int) {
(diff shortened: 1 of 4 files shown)


psymoon:
@liuche I wonder if I should refactor out setMovementMethod() from InlineAutocompleteEditText scope. If we decide to use android components, we should probably stay away from custom changes to the framework as much as possible

app/src/main/java/org/mozilla/focus/widget/DpadKeyMovementMethod.kt
```diff
@@ -0,0 +1,24 @@
...
+
+package org.mozilla.focus.widget
+
+import android.text.Selection
+import android.text.Spannable
+import android.text.method.ArrowKeyMovementMethod
+import android.view.View
+import android.widget.TextView
+
+class DpadKeyMovementMethod : ArrowKeyMovementMethod() {
```
liuche: Let's get a javadoc here, explaining why we're overriding this MM and what it does.

app/src/main/java/org/mozilla/focus/widget/DpadKeyMovementMethod.kt
```diff
@@ -0,0 +1,24 @@
...
+
+    override fun onTakeFocus(view: TextView?, text: Spannable?, dir: Int) {
```
liuche: Is it possible to just leave off the option in Spannable? That way we don't have to force-deref text later on.

app/src/main/java/org/mozilla/focus/widget/DpadKeyMovementMethod.kt
```diff
@@ -0,0 +1,24 @@
...
+        if (dir and (View.FOCUS_FORWARD or View.FOCUS_DOWN) != 0) {
+            if (view?.layout == null) {
+                // This shouldn't be null, but do something sensible if it is.
```
liuche: I know this is from the existing code, but what's the reasoning for keeping this part only?

app/src/main/java/org/mozilla/focus/widget/DpadKeyMovementMethod.kt
```diff
@@ -0,0 +1,24 @@
...
+                Selection.setSelection(text, text!!.length)
```
liuche: I was looking at the code for ArrowKey, and it looks like setting the selection to the end is what causes the text to scroll to the end.

http://androidxref.com/8.1.0_r33/xref/frameworks/base/core/java/android/text/method/ArrowKeyMovementMethod.java#277

So maybe here (and in the other original selection code), we can not set the selection to the end, which is causing the problem ^

Or just remove this whole `onTakeFocus` completely bc we don't want to change anything when the textview is focused.

liuche (dismissed):
Want a bit more clarification, but this would certainly work!

liuche:
> @liuche I wonder if I should refactor out setMovementMethod() from InlineAutocompleteEditText scope. If we decide to use android components, we should probably stay away from custom changes to the framework as much as possible

True - we'll want to switch to using the a-c InlineAutocomplete, hmmm. What if on focus, you call setSelection to 0, so it doesn't need to scroll to the end on "up" arrow?

psymoon:
> True - we'll want to switch to using the a-c InlineAutocomplete, hmmm. What if on focus, you call setSelection to 0, so it doesn't need to scroll to the end on "up" arrow?

@liuche How about setting DpadKeyMovementMethod in setupUrlInput()?

app/src/main/java/org/mozilla/focus/widget/DpadKeyMovementMethod.kt
```diff
@@ -0,0 +1,17 @@
...
+package org.mozilla.focus.widget
+
+import android.text.Spannable
+import android.text.method.ArrowKeyMovementMethod
+import android.widget.TextView
+
+/**
+ * An ArrowKeyMovementMethod that eliminates end cursor selection onFocus
+ */
+class DpadKeyMovementMethod : ArrowKeyMovementMethod() {
```
Baron-Severin: Nit: Suggest changing the class name to more closely describe what it's doing.  Maybe something like `IgnoreFocusMovementMethod`?  I feel that would make the code easier to read from `BrowserNavigationOverlay`.

app/src/main/java/org/mozilla/focus/widget/DpadKeyMovementMethod.kt
```diff
@@ -0,0 +1,17 @@
+/* -*- Mode: Java; c-basic-offset: 4; tab-width: 20; indent-tabs-mode: nil; -*-
```
Baron-Severin: Nit: We can delete this top line, it's an outdated convention that we don't need to track anymore.

severinrudie (approved):
Minor changes requested.

app/src/main/java/org/mozilla/focus/widget/DpadKeyMovementMethod.kt
psymoon: Sounds reasonable. Just for food for thought though - cross referencing ArrowKeyMovementMethod suggests that it is designed to accommodate onTouchEvent more so than onKeyEvent. The reason I named it DpadKeyMovementMethod is so that we have more controls over manipulating movement methods specific to remote control needs moving forward. What do you think is more ideal naming convention for classes (vs. methods) - immediate needs or scalability?

Baron-Severin: Personally, I don't see it as a scalability issue.  If we update the class to handle other aspects of movement we can change the name to match.  But right now the only use of this is to fix this bug.  IMO, if we can make it clear what's happening at the top level, so when reading this I don't have to drill down to find out what's going on, that's a win for readability.

PR #1027, opened 2018-08-23
Closes #906: Updated Dpad navigation

### LINKED ISSUE #906: Combined overlay: specify dpad navigation
Right now dpad navigation does whatever Android does by default: it's not great. Here are some problems:
1) Pressing down from Pocket tile goes to third home tile
2) Pressing up from url bar sometimes goes to settings
3) Selecting a button that is then disabled (e.g. press back and the next page is last in the back stack) will focus the url bar while another action (e.g. forward if you pressed back) maybe be more appropriate

### What / Requirements
- A logical criteria for how the dpad should navigation through these views.

@aminalhazwani Can you go through the current implementation, see what feels wrong, and provide some criteria here? Thanks.

psymoon:


psymoon:
@aminalhazwani please try this build and let me know if everything makes sense!

psymoon:
@shorlander @antlam 

Can you please take a look at the build (while Amin's away) and give feedback? Thanks

antlam:
@psymoon unfortunately I don't have access to the device in the YVR office... :(

@shorlander might have though?

mcomella:
@antlam Does Megan or Roland have a device?

psymoon:
Update from @shorlander via slack:

> I was on a page, I brought up the menu, went back (which was the top of the stack), the focus went from the back button to the forward button, I went forward, and the focus disappeared and won't come back.
> I have to exit the menu and open it again to get it back

I am currently investigating

app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt
```diff
@@ -266,18 +266,30 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
...
         val isFocusLost = focusedView != null && findFocus() == null
         if (isFocusLost) {
             navUrlInput.requestFocus()
+            viewToFocus.requestFocus()
         }
 
         // Update nextFocusUpId of URL bar to leftmost active item in topNavContainer
         topNavContainer?.getFocusables(FOCUS_LEFT)?.get(0)?.let {
             navUrlInput.nextFocusUpId = it.id
+            maybeUpdateOverlayURLForCurrentState()
```
liuche: I think this is misplaced in the wrong bracket - in this case, we're calling maybeUpdate only when we're successfully setting a focus, which shouldn't be a dependency.

app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt
```diff
@@ -266,18 +266,30 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
...
+        }
+    }
+
+    private fun getLeftmostActiveItemInTopNav(): View {
+        topNavContainer?.getFocusables(View.FOCUS_LEFT)?.get(0)?.let {
+            return it
         }
 
         maybeUpdateOverlayURLForCurrentState()
```
liuche: Do we still need this maybeUpdate?

liuche (changes requested):
Some questions about usage of maybeUpdate.

app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt
```diff
@@ -266,18 +266,30 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
...
         }
 
+        var viewToFocus = getLeftmostActiveItemInTopNav()
+        navUrlInput.nextFocusUpId = viewToFocus.id
+
         // We may have lost focus when disabling the focused view above.
         val isFocusLost = focusedView != null && findFocus() == null
         if (isFocusLost) {
             navUrlInput.requestFocus()
+            viewToFocus.requestFocus()
```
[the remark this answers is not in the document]
psymoon: Ah, meant to delete `navUrlInput.requestFocus()`

app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt
```diff
@@ -272,6 +272,11 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
             navUrlInput.requestFocus()
         }
 
+        // Update nextFocusUpId of URL bar to leftmost active item in topNavContainer
+        topNavContainer?.getFocusables(FOCUS_LEFT)?.get(0)?.let {
```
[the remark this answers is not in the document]
psymoon: It looks that way

app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt
psymoon: No, we do not

app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt
psymoon: The whole block is unnecessary actually, it's already handled above. Github rebase issue perhaps? Removed!

psymoon:
Update: back/forward navigation flow seems to have concurrency issue... investigating

liuche:
So it's weird that codecov is failing now with the code removal, but maybe there were unit/UI tests that covered that code - I'd probably see if something in the tests changed/is failing. Maybe @npark-mozilla might have some ideas, but they're at a work week at the moment.

mcomella:
> So it's weird that codecov is failing now with the code removal

Codecov currently fails if the project's entire coverage is reduced by 0.2% or greater in a single PR: https://github.com/mozilla-mobile/firefox-tv/blob/1dd746fb2069f14d0a957d544de2d739330f518c/codecov.yml#L9-L12

I'm not sure this is very helpful considering we don't have agreed upon standards for what the coverage should be and that some PRs can add code that can't be tested or remove code that is tested (i.e. this case).

npark-mozilla:
Just curious firefox-tv isn't using kotlin yet is it? If so I wonder the kotlin bug is affecting it here as well.  Also, checking codecov.yml, I used to think 0.2 means 20% (and I was wrong)? It looks like @pocmo wanted strict coverage here, but I am okay with setting this to 10%.

pocmo:
> Just curious firefox-tv isn't using kotlin yet is it? If so I wonder the kotlin bug is affecting it here as well. Also, checking codecov.yml, I used to think 0.2 means 20% (and I was wrong)? It looks like @pocmo wanted strict coverage here, but I am okay with setting this to 10%.

* If the setup here doesn't cover Kotlin code then those numbers are meaningless.

* Finding a good percentage for codecov is hard. And it depends on the size of the project. In a small code base one uncovered line will have a higher impact on the code coverage than the same line in a gigantic repository.

* In *android-components* we have a very minimal codecov config and do not change the threshold:
https://github.com/mozilla-mobile/android-components/blob/master/codecov.yml

* In *android-components* we use codecov more as a helpful metric than a required check that needs to pass: For project coverage we try to not decrease it. If it decreases we can look at the report and see if we missed things. For the patch coverage (this seems to be disabled here?) codecov generates some dynamic threshold. Again if this fails we check the report to see if we missed testing big parts of our code. To summarize: If the coverage goes up -> Everything's great. If it goes down: Can we improve this PR to keep the coverage up? If there's no way to test or this is code that we cannot test (that can happen with Kotlin): We will have to live with a lower code coverage.

> but I am okay with setting this to 10%.

I don't have a good number. But I'd consider a (close to) 10% coverage drop to be something severe:
* Either we removed a lot of code that had tests and the remaining code unfortunately doesn't -> Okay, that's fine.
* Or we added a lot of code that doesn't have tests -> Red flag.

Overall it seems to be impossible to come up with a threshold that works in all cases: You almost never hit 100% patch coverage and with that the overall coverage can go up/down and you need to decide whether that's okay or not. And even worse: Even with 100% patch coverage you may have missed important test cases. :)

psymoon:
@aminalhazwani Can you try this build out and give UX feedback? Thanks!

aminalhazwani:
hey @psymoon I'm testing your build, it's looking great, I only have a couple of question/considerations: if they are not related to this bug feel free to ignore them 🙂 

**On launch**
- the first element focused is the reload button while it should be the search/url bar
- reload and pin button shouldn't be active on launch since there is no webpage open
- when I click on the turbo mode button the focus jumps back to the reload button while it should remain on the turbo mode button

**While browsing**
- clicking on the menu button opens the overlay homescreen, the focus is on the back button while it should be on the search/url bar. here we are making the assumption that if users want to go back they will simply hit the back button on their remote.
- this is more of an open question: clicking on the back/forward button in the overlay homescreen does not close the overlay, I wonder if we should close it and show the prev/next page in the history stack, gonna dig this more in #1071

psymoon:
@aminalhazwani Thanks for the feedback! I've made the changes as per your feedback. Please try the new build out and let me know if anything else needs to be updated

aminalhazwani:
@psymoon almost there! I noticed only two minor things:

**On launch**

1. Launch Firefox
2. Search/URL bar is focused
3. Go up with the remote
4. Focus is on the Settings button instead of the Turbo Mode button

**Go to/Back from Settings**

1. Go to the settings button
2. Open Settings
3. Press Back on the remote
4. Search bar is focused instead of the settings button

psymoon:
@aminalhazwani Updated. Please try this build and let me know if anything else!

aminalhazwani:
@psymoon I found two minor issues, here the steps to reproduce the first one:

1. Launch Firefox
2. Select a tile, eg Rotten Tomatoes
3. Open Menu
4. Go to Settings button
5. Open Settings
6. Press Back with the remote
7. Search/URL bar is focused instead of Settings button

And this is the second one:

1. Launch Firefox
2. Go to Settings button
3. Open Settings
4. Press Back with the remote
5. Select a tile, eg Rotten Tomatoes
6. Open Menu
7. Settings button is focused instead of the Search/URL bar

Sorry for the back and forth Simon 😓

psymoon:
@aminalhazwani No problem! Fix is out now, please try again :)

aminalhazwani:
I think we are there @psymoon 🙌

psymoon:
@liuche UX is happy, please review!

app/src/main/java/org/mozilla/focus/MainActivity.kt
```diff
@@ -149,13 +151,36 @@ class MainActivity : LocaleAwareAppCompatActivity(), OnUrlEnteredListener, Media
...
+
+            if (browserFragment.arguments == null) {
+                browserFragment.arguments = Bundle()
+            }
+
+            // Set ParentFragment flag to the BrowserFragment based on currFragment and let
+            // fragment lifecycle handle the rest
+            when (currFragment) {
+                is SettingsFragment -> {
+                    browserFragment.arguments!!.putSerializable("PARENT_FRAGMENT", BrowserNavigationOverlay.ParentFragment.SETTINGS)
```
liuche: Nit: can you reuse this string as a constant? Just make it a companion object so there's less chance of typos causing other bugs.

liuche (approved):
Talked about this irl - passing Bundles around isn't ideal, but using `setTargetFragment` and timing with `setVisibility` and the different flows from which you can access the overlay (Fragment back stack, menu button) are not ideal.

However, this is just one more reason for rearchitecting our Fragment/screens story, because we've clearly outgrown our simple BF/BFO situation (*shakes fist at past-liuche*)

app/src/main/java/org/mozilla/focus/MainActivity.kt
psymoon: Updated @liuche

psymoon:
The implementation will further need to be addressed in #1085

PR #1032, opened 2018-08-27
Use version 0.20 of Mozilla Android components.

pocmo:


app/src/main/java/org/mozilla/focus/Components.kt
```diff
@@ -24,7 +25,9 @@ class Components {
  * the <code>SessionManager</code> for now.
```
Baron-Severin: Nit: Not related to your changes, but could we update '<code>' to '[' so that this links properly?

pocmo: Will update!

PR #1043, opened 2018-08-29
Closes #1037: fix unable to click pocket videos

### LINKED ISSUE #1037: Clicking on Pocket videos does nothing
## Steps to reproduce
- On home screen
- Select Pocket
- Select video

### Expected behavior
- Navigate to video

### Actual behavior
- Nothing happens

### Device information
- Master 2a20bb98f1a6e1f5eb0693e7828223a79162b8d0
- Device (Samsung 4k in SF office)
- Fire OS 6.2.5.3

severinrudie:
Root cause:
- Browser screen fragment transaction is started from a SessionManager.Observer#onSessionSelected callback, defined in MainActivity
- ScreenController#showBrowserScreenForUrl added a session, but 'selected' defaulted to false. This caused the above callback to not be triggered

### DIFF
--- a/app/src/main/java/org/mozilla/focus/ScreenController.kt
+++ b/app/src/main/java/org/mozilla/focus/ScreenController.kt
@@ -92,7 +92,7 @@ object ScreenController {
             browserFragment.loadUrl(url)
         } else {
             val session = Session(url, source = source)
-            context.components.sessionManager.add(session)
+            context.components.sessionManager.add(session, selected = true)
         }
     }
 


app/src/main/java/org/mozilla/focus/ScreenController.kt
```diff
@@ -92,7 +92,7 @@ object ScreenController {
             browserFragment.loadUrl(url)
         } else {
             val session = Session(url, source = source)
-            context.components.sessionManager.add(session)
+            context.components.sessionManager.add(session, true)
```
pocmo: nit: Do you think it would be more readable to use a named parameter like `selected = true` here?

pocmo (approved):
Good catch! 🚢

app/src/main/java/org/mozilla/focus/ScreenController.kt
Baron-Severin: Good call, I'll make that change.

PR #1047, opened 2018-08-29
Clean up licenses ordering, add missing licenses

liuche:
We had a bunch of extraneous non-top-level dependency licenses so I removed them (and reordered them to match `<dependencies>`).

Were missing Sentry, okhttp, and picasso, so I added them.


severinrudie (approved):
LGTM

PR #1051, opened 2018-08-30
Closes #895: Add placeholder tiles when Pocket API key not provided.

### LINKED ISSUE #895: Add placeholder tiles when Pocket API key is not provided
## User benefit
None, but it might save us confusion and time. Also, devs hit the pocket servers unnecessarily, which costs them server time and thus money.

## Details
Without a key provided, the code defaults to the error state (which right now is nothing). This is fine but misleading if you don't know that we're not handling the pocket feed when you don't have an API key. We also hit the Pocket servers each time the app is started when the key is not provided (which isn't great).

We should put in a placeholder if the key is null (e.g. use a ColorDrawable or a drawable already in the app with some generic names and URLs).

dnarcese:


app/src/main/java/org/mozilla/focus/home/pocket/Pocket.kt
```diff
@@ -79,7 +81,8 @@ object Pocket {
      */
     @AnyThread // videosCache is synchronized.
     fun getRecommendedVideos() = videosCache
-    private fun getRecommendedVideosNoCacheAsync() = async { PocketEndpoint.getRecommendedVideos() }
+    private fun getRecommendedVideosNoCacheAsync(placeholderFlag: Boolean) =
```
Baron-Severin: Personal opinion, but I usually think that there are more readable alternatives to passing flags to change method behavior.  If we broke this into two methods I think we would get a few benefits.

- At the top level (inside the init block) it would be clear what was going on.  `getRecommendedVideosNoCacheAsync(true)` can be ambiguous.  `getRecommendedVideosNoCacheAsync` vs `getPlaceholderVideos` is immediately clear
- We wouldn't have to pass this flag through multiple levels of code, which makes maintenance simpler.
- We could pull the conditional statement out of PocketEndpoint.  We would then be less likely to confuse developer ergonomics code with prod code when debugging.

severinrudie (changes requested):
Approach looks good, just have a readability request.

Baron-Severin:
@dnarcese Sorry, I started reviewing before @liuche was tagged

app/src/main/java/org/mozilla/focus/home/pocket/PocketVideoMegaTile.kt
```diff
@@ -23,11 +25,21 @@ class PocketVideoMegaTile(
...
+                        .into(thumbnailView)
+            }
+            Toast.makeText(context, "Pocket API key was not found.", Toast.LENGTH_LONG).show()
+        } else {
+            if (newVideos == null) return@observable
+            thumbnailViews.forEachIndexed { i, thumbnailView ->
+                PicassoWrapper.client.load(newVideos[i].thumbnailURL)
+                        .transform(roundCornerTransformation)
+                        .into(thumbnailView)
+            }
```
Baron-Severin: Optional nit: I think the readability is more or less a wash between these two, but we could reduce some of the repetition by doing something like the following.  Up to you which you prefer.

```kotlin
    var pocketVideos by Delegates.observable<List<PocketVideo>?>(null) { _, _, newVideos ->
        val thumbnails = if (BuildConfig.POCKET_KEY == null) {
            Toast.makeText(context, "Pocket API key was not found.", Toast.LENGTH_LONG).show()
            (0 until thumbnailViews.size)
                    .map { "https://blog.mozilla.org/firefox/files/2017/12/Screen-Shot-2017-12-18-at-2.39.25-PM.png" }
        } else {
            newVideos?.map { it.thumbnailURL } ?: return@observable
        }
        thumbnailViews.forEachIndexed { i, thumbnailView ->
            PicassoWrapper.client.load(thumbnails[i])
                    .transform(roundCornerTransformation)
                    .into(thumbnailView)
        }
    }
```

severinrudie (approved):
LGTM, with an optional nit.

PR #1061, opened 2018-08-31
Closes #1060: Add clear cookies copy.

### LINKED ISSUE #1060: Add strings for clear cookies copy

mcomella:


app/src/main/res/values/strings.xml
```diff
@@ -50,6 +50,9 @@
     <!-- In-app link (in settings) to Firefox for Fire TV privacy notice. -->
     <string name="preference_privacy_notice">Privacy Notice</string>
 
+    <!-- This title is shown on a dialog that prompts the user if they want to clear their cookies and site data.  -->
+    <string name="preference_clear_cookies_title">Clear Cookies?</string>
```
liuche: I wouldn't call it a preference - that implies it's in the list of preferences and that's not where this string is. Maybe just "clear_cookies_title"?

Issue #1063, opened 2018-09-03
Follow-up work from browser-engine-system migration


pocmo:
Follow-up from #1039. This is a collection of things that are missing in the first iteration of the migration to `browser-engine-system`. For now I'll just collect everything here and will file separate issues as needed.

### Components 0.21

* [x] ~~Setup tracking protection and make sure toggle behaves correctly~~
* [x] ~~Add interception implementation for firefox:home and other internal pages~~

### Components 0.22

* [x] (https://github.com/mozilla-mobile/android-components/issues/696) "X-Requested-With" header is cleared

### Components 0.23

* [x] (https://github.com/mozilla-mobile/android-components/issues/643) Add fullscreen API behavior 
* [x] (https://github.com/mozilla-mobile/android-components/issues/718) Error pages
* [x] (https://github.com/mozilla-mobile/android-components/issues/728) Call onNavigationStateChange() earlier (Improvement)

### Misc

* [x] Re-add `onFocusChange` behavior: https://github.com/mozilla-mobile/firefox-tv/blob/ec097ab058862129f0b5baa0294983ee32a67ab5/app/src/amazonWebview/java/org/mozilla/focus/webview/FirefoxAmazonWebView.kt#L90-L103
To test this is fixed:
> To test this is fixed:
>  Open YT.com/tv
>  Open menu
>  Close menu
>  move selection
>    expected: selection moves
>    actual: user cannot move selection
* [x] ~~InfoFragment: Do we need to hide the WebView if we are loading an internal page?~~
* [x] ~~InfoFragment: Do we need to re-add to onPageStart():  `progressView.announceForAccessibility(getString(R.string.accessibility_announcement_loading));`~~
* [x] ~~The back/forward state in the toolbar updates very slowly. It looks like you need to move the cursor around to make it refresh at times (?!).~~  (Improved this and filed an a-c issue to improve this).
* [x] Internal pages are now black text on a white background?

### QA

* [x] Make sure EngineView.scrollByClamped performs fast. Consider adding API to EngineView.
* [x] ~~Make sure tracking protection is disabled in InfoFragment (Is that strictly required?)~~
* [X] ~~Make sure back/forward navigation is working properly.~~ Fixed it to behave consistently - but actual behavior is to be defined in # #1071.
* [ ] Make sure handling external URLs (with fallback) are working properly (Do we even support that?)

### Crashes

[...]

PR #1066, opened 2018-09-05
browser-engine integration follow-up

pocmo:
This fixes a bunch of follow-up items from the list in #1063.

The Pocket patch is an odd one. I didn't change that code but AS started to complain about the condition being always `true`. This depends on the local configuration and the compiler obviously doesn't know about that... so I just suppressed the warning.

### DIFF
--- a/app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt
+++ b/app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt
@@ -25,6 +25,7 @@ import kotlinx.coroutines.experimental.android.UI
 import kotlinx.coroutines.experimental.launch
 import org.mozilla.focus.R
 import org.mozilla.focus.autocomplete.UrlAutoCompleteFilter
+import org.mozilla.focus.ext.components
 import org.mozilla.focus.ext.forEachChild
 import org.mozilla.focus.ext.updateLayoutParams
 import org.mozilla.focus.home.HomeTilesManager
@@ -109,7 +110,15 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
     private var isTurboEnabled: Boolean
         get() = Settings.getInstance(context).isBlockingEnabled
         set(value) {
-            Settings.getInstance(context).isBlockingEnabled = value
+            val settings = Settings.getInstance(context)
+            settings.isBlockingEnabled = value
+
+            val engineSession = context.components.sessionManager.getOrCreateEngineSession()
+            if (value) {
+                engineSession.enableTrackingProtection(settings.trackingProtectionPolicy)
+            } else {
+                engineSession.disableTrackingProtection()
+            }
         }
 
     private val pocketVideos = Pocket.getRecommendedVideos()
--- a/app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt
+++ b/app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt
@@ -25,13 +25,12 @@ import kotlinx.coroutines.experimental.android.UI
 import kotlinx.coroutines.experimental.launch
 import org.mozilla.focus.R
 import org.mozilla.focus.autocomplete.UrlAutoCompleteFilter
-import org.mozilla.focus.ext.components
 import org.mozilla.focus.ext.forEachChild
 import org.mozilla.focus.ext.updateLayoutParams
 import org.mozilla.focus.home.HomeTilesManager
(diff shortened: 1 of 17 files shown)


app/src/main/java/org/mozilla/focus/engine/CustomContentRequestInterceptor.kt
```diff
@@ -0,0 +1,29 @@
...
+import org.mozilla.focus.browser.LocalizedContent
+
+/**
+ * [RequestInterceptor] implementation to inject custom content for firefox:* pages.
+ */
+class CustomContentRequestInterceptor(
+    private val context: Context
+) : RequestInterceptor {
+    override fun onLoadRequest(session: EngineSession, uri: String): RequestInterceptor.InterceptionResponse? {
+        return when (uri) {
```
Baron-Severin: I like the approach of updating LocalizedContent#generateAboutPage to remove side effects, and instead applying them here.
👍

app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt
```diff
@@ -108,7 +109,15 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
     private var isTurboEnabled: Boolean
         get() = Settings.getInstance(context).isBlockingEnabled
         set(value) {
-            Settings.getInstance(context).isBlockingEnabled = value
+            val settings = Settings.getInstance(context)
+            settings.isBlockingEnabled = value
+
+            val engineSession = context.components.sessionManager.getOrCreateEngineSession()
+            if (value) {
+                engineSession.enableTrackingProtection(settings.trackingProtectionPolicy)
```
Baron-Severin: What would happen if `settings.isBlockingEnabled` and `engineSession.trackingProtectionEnabled` went out of sync?  This looks like a chunk of code where bugs could pretty easily worm their way in.  Testing setters is usually overkill, but I think it might be worthwhile here just in case someone who is unfamiliar with this setup accidentally tweaks it.

More broadly, it seems like we're doing an awesome job of breaking out components, but if these two both need this information I wonder if they might benefit from some adjustment.  Are there any concerns that could be moved from one to another?  If not, would it be possible to place some of this behind a facade?

app/src/test/java/org/mozilla/focus/engine/CustomContentRequestInterceptorTest.kt
```diff
@@ -0,0 +1,51 @@
...
+    @Test
+    fun `Interceptor should return content for firefox-about`() {
+        val result = testInterceptor("firefox:about")
+
+        assertNotNull(result)
+        assertTrue(result!!.data.isNotEmpty())
+        assertEquals("text/html", result.mimeType)
+        assertEquals("UTF-8", result.encoding)
+    }
+
```
Baron-Severin: Optional nit: Might be worth adding a small test that asserts that "firefox:home" and "firefox:about" return values are not equal, to avoid false positives.

app/src/main/java/org/mozilla/focus/browser/InfoFragment.kt
```diff
@@ -32,14 +32,20 @@ class InfoFragment : EngineViewLifecycleFragment(), Session.Observer {
         val session = Session(arguments!!.getString(ARGUMENT_URL))
         session.register(this, owner = this)
 
-        webView!!.render(requireComponents.sessionManager.getOrCreateEngineSession(session))
+        val engineSession = requireComponents.sessionManager.getOrCreateEngineSession(session)
+        engineSession.disableTrackingProtection()
```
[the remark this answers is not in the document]
pocmo: That's a good question.

I re-added this because the initial integration PR removed this.

If I remember correctly this part is inherited from Focus, which has changed since then. I think we didn't want use tracking protection because it may break some of the linked pages - but I do not remember.

app/src/test/java/org/mozilla/focus/engine/CustomContentRequestInterceptorTest.kt
pocmo: That's a great idea. I'll add that.

app/src/main/java/org/mozilla/focus/browser/InfoFragment.kt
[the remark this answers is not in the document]
pocmo: Ah, found the issue where I added this in Focus:
https://github.com/mozilla-mobile/focus-android/issues/717

Will add a comment.

pocmo:
@Baron-Severin Okay, updated the PR! :)

PR #1067, opened 2018-09-05
Closes #1010: Modify SessionLoadedIdlingResource  to use session manager from browser-session component..

### LINKED ISSUE #1010: Browser-session integration broke SessionLoadedIdlingResource
As mentioned in https://github.com/mozilla-mobile/firefox-tv/issues/998, this is also blocking the UI test development as the tests will not compile.  When this bug is fixed, ideally #998 should be resolved as well.

pocmo:
With this patch IWebViewExecuteJavascriptTest passes.

I noticed that other tests do not compile anymore and this seems to be unrelated to browser-session and may be caused by the overlay refactoring:
* `BrowserOverlayTest` (Referencing navButtonHome)
* `PageLoadTest` (Referencing navButtonHome)
* `PinnedTileTests` (Referencing navButtonHome)
* `TVScreenshots` (Referencing multiple now unknown ids; most tests that compile fail with the new UI though)


app/src/androidTest/java/org/mozilla/focus/integration/IWebViewExecuteJavascriptTest.kt
```diff
@@ -27,6 +27,7 @@ import org.junit.runner.RunWith
 import org.mozilla.focus.R
 import org.mozilla.focus.MainActivity
 import org.mozilla.focus.browser.BrowserFragment
+import org.mozilla.focus.ext.evalJS
```
npark-mozilla: is this import being used? the body isn't changed...

pocmo: Heh. Yes. The diff is confusing!

It is used here:
https://github.com/mozilla-mobile/firefox-tv/blob/f84f8f6ff98dd7b9a2b8b5eb97b07bede9ce8224/app/src/androidTest/java/org/mozilla/focus/integration/IWebViewExecuteJavascriptTest.kt#L79-L80

Previously this was a method on IWebView. Now it's an extension function on EngineView. Extension functions need to be imported if they are in a different package.

https://kotlinlang.org/docs/reference/extensions.html

PR #1068, opened 2018-09-06
browser-engine integration follow-up II

pocmo:
Part II. :)

Those patches fix some of the back/forward behavior and a crash if we start loading a URL after removing the session.

The whole cycle of how we create new sessions, destroy them and re-create fragments is very complex and I spend quite some time debugging it. Tomorrow I'm meeting with @aminalhazwani to understand how this is supposed to work and will try to simplify (and add tests) that behavior. For now that's the best I can come up with.

### DIFF
--- a/app/src/main/java/org/mozilla/focus/browser/BrowserFragment.kt
+++ b/app/src/main/java/org/mozilla/focus/browser/BrowserFragment.kt
@@ -66,16 +66,7 @@ class BrowserFragment : EngineViewLifecycleFragment(), Session.Observer {
 
     private val mediaSessionHolder get() = activity as MediaSessionHolder? // null when not attached.
 
-    /**
-     * The current URL.
-     *
-     * Use this instead of the WebView's URL which can return null, return a null URL, or return
-     * data: URLs (for error pages).
-     */
-    var url: String? = null
-        private set
-
-    val isUrlEqualToHomepage: Boolean get() = url == APP_URL_HOME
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/mozilla/focus/browser/BrowserFragment.kt
```diff
@@ -268,16 +257,36 @@ class BrowserFragment : EngineViewLifecycleFragment(), Session.Observer {
...
     }
 
     fun loadUrl(url: String) {
         if (url.isNotEmpty()) {
-            requireComponents.sessionUseCases.loadUrl.invoke(url)
+            val session = requireComponents.sessionManager.selectedSession
+
+            if (session != null) {
+                // We already have an active session, let's just load the URL.
+                requireComponents.sessionUseCases.loadUrl.invoke(url)
```
Baron-Severin: Nit: Is this logic particular to our use case, or will this be replicated by other projects?  If the latter, could we open an issue to have SessionManager handle it, then add a todo here to replace this when we bump our components version?

pocmo: This is more or less a workaround to get this working with the current logic here.

Ideally I would like this whole back and forth between BrowserFragment, MainActivity and SessionManager.Observer callbacks to disappear: In this app there should always be only one session or none (initial state). Hopefully after the meeting tomorrow I'll know if that's the way to go.

PR #1069, opened 2018-09-06
Update pre push hook to reflect change in build variant name

dnarcese:


dnarcese:
...made a mistake with the branch

PR #1082, opened 2018-09-10
Closes #864: adds loading state to Pocket megatile

### LINKED ISSUE #864: Pocket megatile loading screen
User benefit: without a loading state, it looks weird and it's unclear to the user that we're loading. What we have in the mocks is a transition between gradient images in the thumbnails: seems straightforward.

fwiw, we cache our data and never replace the cache with bad data so users will only see this each time the app is started after OOM.

Baron-Severin:
This seems to be just as effective as defining a crossfade using two ImageViews, but is less work and easier to read.

tv-loading


Baron-Severin:
@psymoon Good point, pulling it out.

Baron-Severin:
@psymoon The Picasso placeholder won't show until the Pocket response comes in, so I added the src back in as a default to guard against the (hopefully) rare case that Pocket is the bottleneck.

@aminalhazwani Good catch!  Better?
tv-rounded-pocket

aminalhazwani:
> @aminalhazwani Good catch! Better?

Yes 😬, thanks!

PR #1090, opened 2018-09-11
Closes #1008: correct Kotlin codecov

### LINKED ISSUE #1008: codecov is not aware of Kotlin
See https://github.com/mozilla-mobile/focus-android/issues/3029:

e.g. Kotlin files don't appear in the codecov results directories: https://codecov.io/gh/mozilla-mobile/firefox-tv/tree/master/app/src/main/java/org/mozilla/focus/browser

Our Jacoco android plugin doesn't support Kotlin apparently – https://github.com/arturdm/jacoco-android-gradle-plugin/issues/37 – but codecov does ([supported languages](https://docs.codecov.io/docs), [kotlin example](https://github.com/codecov/example-kotlin), [android example](https://github.com/codecov/example-android), [gradle example](https://github.com/codecov/example-gradle)) We can:
- Look in this issue for solutions
- Look for people who've forked and improved it the jacoco android plugin
- Write our own build code to run codecov (or make a PR upstream to the plugin)

mcomella:


app/build.gradle
```diff
@@ -2,7 +2,6 @@ apply plugin: 'com.android.application'
 apply plugin: 'kotlin-android'
 apply plugin: 'findbugs'
 apply plugin: 'jacoco'
-apply plugin: 'jacoco-android'
```
Baron-Severin: Nit: Regarding this commit message: 
```
This plugin doesn't support Kotlin so we're going to configure it by
hand.
```

It might be good to drop a comment into the section that does this configuration explaining this.  Otherwise this will look like a good place to clean things up by replacing the custom code with using the plugin.

mcomella: Good idea, adding:
```kotlin
// Set up jacoco for code coverage. We originally used the jacoco-android plugin
// but it doesn't support kotlin so we're configuring it manually.

Issue #1109, opened 2018-09-13
Menu press on Pocket video feed should open overlay


Baron-Severin:
## Steps to reproduce
From clean start
- Open pinned site
- Click menu
- Open pocket
- Click menu

### Expected behavior
@aminalhazwani Should this go back to the main menu, or to the previously opened site?

### Actual behavior
Nothing happens

### Device information
Device: Fire Cube, smaller TV in office
Fire OS: 6.2.2.3
Commit: e0ae14e04f35f94a600385c8e34d5d62886d52c1

aminalhazwani:
Menu button should always open the menu unless otherwise indicated. So in this case it can open the overlay menu with the Pocket tile focused.

PR #1115, opened 2018-09-13
Fixes i1108: fixes blank WebView after Pocket carousel

Baron-Severin:
Root cause:
- BrowserFragment was replaced with PocketVideoFragment when Pocket screen was shown
- This destroyed the BrowserFragment instance
- Our session was up to date, but when we returned to the BrowserFragment it was with a fresh WebView

@liuche I have a deep fear of messing with Fragment transactions.  If you have a better solution to this, please let me know.

### DIFF
--- a/app/src/main/java/org/mozilla/focus/ScreenController.kt
+++ b/app/src/main/java/org/mozilla/focus/ScreenController.kt
@@ -96,9 +96,10 @@ object ScreenController {
         }
     }
 
-    fun showPocketScreen(fragmentManager: FragmentManager) {
+    fun showPocketScreen(fragmentManager: FragmentManager, browserFragment: BrowserFragment) {
         fragmentManager.beginTransaction()
-                .replace(R.id.container, PocketVideoFragment.create(Pocket.getRecommendedVideos()))
+                .hide(browserFragment)
+                .add(R.id.container, PocketVideoFragment.create(Pocket.getRecommendedVideos()))
                 .addToBackStack(null)
                 .commit()
     }
--- a/app/src/main/java/org/mozilla/focus/browser/BrowserFragment.kt
+++ b/app/src/main/java/org/mozilla/focus/browser/BrowserFragment.kt
@@ -14,6 +14,7 @@ import android.view.View
 import android.view.ViewGroup
 import kotlinx.android.synthetic.main.browser_overlay.*
 import kotlinx.android.synthetic.main.browser_overlay.view.*
+import kotlinx.android.synthetic.main.firefox_progress_bar.*
 import kotlinx.android.synthetic.main.fragment_browser.*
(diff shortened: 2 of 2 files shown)


app/src/main/java/org/mozilla/focus/ScreenController.kt
```diff
@@ -96,9 +96,10 @@ object ScreenController {
         }
     }
 
-    fun showPocketScreen(fragmentManager: FragmentManager) {
+    fun showPocketScreen(fragmentManager: FragmentManager, browserFragment: BrowserFragment) {
         fragmentManager.beginTransaction()
-                .replace(R.id.container, PocketVideoFragment.create(Pocket.getRecommendedVideos()))
+                .hide(browserFragment)
```
liuche: Interesting, looking into `hide` it's definitely better than `replace` here because we always want to return to the BrowserFragment, so there's no need to tear down BF. I don't *think* there will be bad side effects, since this transaction is in the backstack and can reversed. We also shouldn't have to worry about the the problem where `commit` causes crashes because the Activity is torn down because we basically only have one activity so it'll be alive while we're doing all of this.

liuche: This could make #1109 more complicated, because if, maybe, we want to show the BrowserFragment on top of the Pocket feed we'd have to make sure to also call `show` on BF because it'll be hidden from this call.

One thing I'm concerned about is (and I don't know what happens when we find a fragment by id): when you click on a video from the Pocket feed, does it show the BF correctly, or is it still hidden? I don't see us ever killing the PocketFragment (remove, replace) so I think....unless we back out of it by tapping "back" it just hangs around forever?

Can you test to make sure it works? but yeah we should fix this blasted architecture now that we have all these screens.

liuche: @Baron-Severin two things I can think of, if returning to BF doesn't work.

a) updating showBrowserForUrl in `ScreenController` to show the fragment when it's loaded
a) setting the session info in the BF when it's created (but keeping `replace` instead of `hide`/`add`

Baron-Severin: Setting session info kinda works, but we still lose a lot of state in the WebView.  We have to go through loading again, and if the user was midway through filling out a form, or had expanded an image, they'll lose all of that.  

I'm experimenting with changes I can make to `showBrowserForUrl`, but making the entire jigsaw puzzle fit is proving challenging.

liuche:
@Baron-Severin one thing that I find weird is that we used to be able to background Firefox, and returning to the foreground didn't cause the white screen.

liuche:
One problem is that if you have a video playing, we do lose all playing/progress state, which is not ideal.

PR #1116, opened 2018-09-14
Severin/i865 pocket video placeholders

Baron-Severin:


app/src/main/java/org/mozilla/focus/home/pocket/PocketVideoFragment.kt
```diff
@@ -86,6 +88,11 @@ class PocketVideoFragment : Fragment() {
         }
     }
 
+    override fun onStop() {
+        super.onStop()
+        job?.cancel()
```
mcomella: I don't think this will work: I think this job either:
- Needs to be restarted in onStart if it does not complete
- Needs to be destroyed in onDestroyView instead

The former is probably more correct but more complex.

app/src/main/java/org/mozilla/focus/home/pocket/PocketVideoFragment.kt
```diff
@@ -103,10 +108,19 @@ class PocketVideoFragment : Fragment() {
 
 private class PocketVideoAdapter(
     context: Context,
-    private val pocketVideos: List<PocketVideo>,
     private val fragmentManager: FragmentManager
 ) : RecyclerView.Adapter<PocketVideoViewHolder>() {
 
+    private val pocketVideos = mutableListOf<PocketVideo>()
+
+    inner class VideoObserver {
```
mcomella: nit: `*Observer` isn't the right name here: that conflicts with the Observer pattern. Is there any reason `setVideos` can't be a method directly on the adapter? `adapter.setVideos` makes sense to me.

app/src/main/java/org/mozilla/focus/home/pocket/PocketVideoFragment.kt
```diff
@@ -57,25 +56,31 @@ class PocketVideoFragment : Fragment() {
...
+            return adapter
         }
 
-        val layout = inflater.inflate(R.layout.fragment_pocket_video, container, false)
+        val videoObserver = initAdapter().VideoObserver()
 
         job = launch(UI) {
             if (deferredVideos.isNotCompleted()) {
-                //TODO set up loading UI
+                val placeholders = (1..20).map { i ->
```
mcomella: nit: 20 because we get 20 items from the server? Add a comment (or constant that is commented)

mcomella: nit: slightly more readable:
```kotlin
List(20) { i ->
    PocketVideo(i, ...)
}
```

app/src/main/java/org/mozilla/focus/home/pocket/PocketVideoFragment.kt
```diff
@@ -57,25 +56,31 @@ class PocketVideoFragment : Fragment() {
     }
 
     override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View {
-        fun displayFeed(layout: View, videos: List<PocketVideo>) {
-            layout.videoFeed.gridView.adapter = PocketVideoAdapter(context!!, videos, fragmentManager!!)
+        val layout = inflater.inflate(R.layout.fragment_pocket_video, container, false)
+
+        fun initAdapter(): PocketVideoAdapter {
```
mcomella: nit: I don't feel this improves readability much over doing it in-line like:
```kotlin
val adapter = PocketVideoAdapter(context!!, fragmentManager!!)
layout.videoFeed.gridView.adapter = adapter
```

Or maybe (is this how `also` is supposed to be used?)
```kotlin
val adapter = PocketVideoAdapter(context!!, fragmentManager!!).also {
    layout.videoFeed.gridView.adapter = it
}

File: app/src/main/java/org/mozilla/focus/home/pocket/PocketVideoFragment.kt (hunk 3 above)
Comment (mcomella) [inline app/src/main/java/org/mozilla/focus/home/pocket/PocketVideoFragment.kt]: nit: maybe slightly more high-level/readable if `getLoadingPlaceholders()` was it's own function so it's just
```kotlin
videoObserver.setVideos(getLoadingPlaceholders())

File: app/src/main/java/org/mozilla/focus/home/pocket/Pocket.kt [hunk 5]
```diff
@@ -110,7 +110,7 @@ object Pocket {
 data class PocketVideo(
     val id: Int,
     val title: String,
-    val url: String,
+    val url: String?,
```
mcomella: I don't like representing `PocketVideoPlaceholder` as a `PocketVideo` with a null URL, especially if it's undocumented: I'd prefer to use a sealed class, a "PocketVideo" kind enum, or similar.

app/src/main/java/org/mozilla/focus/home/pocket/PocketVideoFragment.kt
```diff
@@ -36,6 +39,7 @@ import java.net.URISyntaxException
 class PocketVideoFragment : Fragment() {
 
     private lateinit var deferredVideos: PocketVideosDeferred
+    private var job: Job? = null
```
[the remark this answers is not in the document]
Baron-Severin: This is something that occurred to me, but as we currently only have one long running task in this file, I didn't think it was worth the extra complexity.  That said, I don't feel strongly about it.  Would you like me to add it in?

app/src/main/java/org/mozilla/focus/home/pocket/PocketVideoFragment.kt
Baron-Severin: Hm.  We could move the existing setup logic from `onCreateView` into `onStart`.  That does feel more correct, wouldn't be a big task, and shouldn't impact functionality (we're creating a new instance every time anyway).  Are you ok with that solution?

app/src/main/java/org/mozilla/focus/home/pocket/PocketVideoFragment.kt
Baron-Severin: Yeah, that's much better.  I need to get into the habit of using that.

Baron-Severin: Good call, but after replacing it with its own class this is no longer necessary.

Baron-Severin:
@mcomella I think I forgot to ping you for rereview.

mcomella:
I feel the parent job approach is less complex because it separates
concerns (i.e. doesn't mix job cancellation and job creation) and doesn't
force anyone to refactor existing jobs to create new jobs.

app/src/main/java/org/mozilla/focus/home/pocket/PocketVideoFragment.kt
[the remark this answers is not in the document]
mcomella: I feel the parent job approach is less complex because it separates concerns (i.e. doesn't mix job cancellation and job creation) and doesn't force anyone to refactor existing jobs to create new jobs.

Baron-Severin:
@mcomella Updated Job lifecycle handling as requested.

app/src/main/java/org/mozilla/focus/home/pocket/PocketVideoFragment.kt
```diff
@@ -53,31 +56,38 @@ class PocketVideoFragment : Fragment() {
...
-            launch(UI) {
-                val videos = deferredVideos.await()
-                if (videos != null) {
-                    displayFeed(layout, deferredVideos)
-                } else {
-                    // TODO: #769: display error screen.
-                }
+    override fun onStart() {
+        super.onStart()
+        val adapter = PocketVideoAdapter(context!!, fragmentManager!!).also {
```
mcomella: Is there any reason this should be in `onStart` instead of `onCreate`? I feel like one-time set-up code usually goes in onCreate.

app/src/main/java/org/mozilla/focus/home/pocket/PocketVideoFragment.kt
```diff
@@ -53,31 +56,38 @@ class PocketVideoFragment : Fragment() {
...
+            videoFeed.gridView.adapter = it
+        }
+        fun setupLoadingIfApplicable() {
+            if (deferredVideos.isNotCompleted()) {
+                // We show 20 placeholders because our feed will be updated to show 20 items
+                val placeholders = List(20) { PocketPlaceholder }
```
mcomella: nit: constant

app/src/main/java/org/mozilla/focus/home/pocket/PocketVideoFragment.kt
```diff
@@ -53,31 +56,38 @@ class PocketVideoFragment : Fragment() {
...
         }
+        fun displayFeedOrError(videos: List<PocketVideo>?) {
+            if (videos != null) {
+                adapter.setVideos(videos)
+            } else {
+                // TODO: #769: display error screen.
+            }
+        }
+
+        launch(UI + uiLifecycleCancelJob) {
```
mcomella: nit: `launch(UI + uiLifecycleCancelJob, UNDISPATCHED) {` It's slightly more efficient because we don't have to dispatch to the UI thread, though it will mean the flow is more likely to be different between loading and finished (i.e. **Currently:** SUSPEND loading SUSPEND feed / SUSPEND feed **vs.** loading SUSPEND feed / feed).

app/src/main/java/org/mozilla/focus/home/pocket/PocketVideoFragment.kt
```diff
@@ -53,31 +56,38 @@ class PocketVideoFragment : Fragment() {
...
+            setupLoadingIfApplicable()
+            val videos = deferredVideos.await()
+            displayFeedOrError(videos)
+        }
 
         // SVGs can have artifacts if we set them in XML so we set it in code.
-        layout.pocketWordmarkView.setImageDrawableAsPocketWordmark()
-        layout.pocketHelpButton.setImageDrawable(context!!.getDrawable(R.drawable.pocket_onboarding_help_button))
-        return layout
+        pocketWordmarkView.setImageDrawableAsPocketWordmark()
```
mcomella: Put this above the coroutine: depending on how the code is written/refactored, the coroutine could suspend so there will be a break in our execution on the UI thread and the user may see these items pop in.

app/src/main/java/org/mozilla/focus/home/pocket/PocketVideoFragment.kt
```diff
@@ -86,6 +96,11 @@ class PocketVideoFragment : Fragment() {
         }
     }
 
+    override fun onStop() {
+        super.onStop()
+        uiLifecycleCancelJob.cancel()
```
mcomella: After this point, this job is perpetually cancelled and we never attach a new job to it. I think you should be creating/cancelling new parent jobs in onStart/onStop or `onCreateView/onDestroyView`.

app/src/main/java/org/mozilla/focus/home/pocket/PocketVideoFragment.kt
```diff
@@ -108,48 +124,68 @@ private class PocketVideoAdapter(
...
-        } catch (e: Exception) { // Apparently Kotlin doesn't have multi-catch.
-            when (e) {
-                is URISyntaxException, is NullPointerException -> null
-                else -> throw e
+    }
+
+    private fun bindPocketVideo(holder: PocketVideoViewHolder, item: PocketVideo) {
+        with(holder) {
+            holder.itemView.setOnClickListener {
+                // If the [PocketVideo] has no url field, assume it is a placeholder
```
mcomella: nit: outdated

app/src/main/java/org/mozilla/focus/home/pocket/PocketVideoFragment.kt
```diff
@@ -108,48 +124,68 @@ private class PocketVideoAdapter(
...
-            URI(item.url)
-        } catch (e: Exception) { // Apparently Kotlin doesn't have multi-catch.
-            when (e) {
-                is URISyntaxException, is NullPointerException -> null
-                else -> throw e
+    }
+
+    private fun bindPocketVideo(holder: PocketVideoViewHolder, item: PocketVideo) {
+        with(holder) {
+            holder.itemView.setOnClickListener {
```
mcomella: nit: unnecessary `holder`

app/src/main/java/org/mozilla/focus/home/pocket/PocketVideoFragment.kt
```diff
@@ -108,48 +124,68 @@ private class PocketVideoAdapter(
...
-        val itemURI = try {
-            URI(item.url)
-        } catch (e: Exception) { // Apparently Kotlin doesn't have multi-catch.
-            when (e) {
-                is URISyntaxException, is NullPointerException -> null
-                else -> throw e
+    }
+
+    private fun bindPocketVideo(holder: PocketVideoViewHolder, item: PocketVideo) {
+        with(holder) {
```
mcomella: nit: Since `bind*` is only likely to use `holder` as a receiver, make this function equal an expression to reduce indentation? It'd better preserve blame too. Here and above.

app/src/main/java/org/mozilla/focus/home/pocket/PocketVideoFragment.kt
Baron-Severin: I'm not sure I agree.  Since the value is only used here, it seems like an additional level of indirection that we don't need.

app/src/main/java/org/mozilla/focus/home/pocket/PocketVideoFragment.kt
Baron-Severin: We need access to a view here, but it could be moved to `onCreateView`.  Would you prefer that I move it there?  I don't really have any preference one way or the other.

Issue #1125, opened 2018-09-17
[a11y] No accessibility string for "help/more information" button on Pocket feed


liuche:
We definitely won't have time to localize this for 3.0, but we should get a string for this "help/info" button. @BrianNJones what do you think?

Right now, a11y just announces "Button".
Here's a screenshot of the button highlighted - VoiceView announces "Button".
![help](https://user-images.githubusercontent.com/1082253/45626453-e7644d00-ba8f-11e8-8706-f54de431d27d.png)

Clicking this button will re-show the onboarding screen that introduces the Pocket video feed feature.
![pocket-intro](https://user-images.githubusercontent.com/1082253/45626964-17602000-ba91-11e8-86d8-d76c1fcd4026.png)

See this comment for more context:
https://github.com/mozilla-mobile/firefox-tv/issues/828#issuecomment-382710975

liuche:
In the meantime, @BrianNJones could we use the string "Recommended videos" or "Video recommendations by Pocket"? Neither really explains this "more info" button very well.

liuche:
@eejay can you also let us know how important this is? The "?" button re-shows the onboarding UI that was already shown so the button isn't "critical" UI, but the lack of labeling is bad. Also, any thoughts on non-localized (for 3.0) or localized?

bbinto:
I'd suggest to go with a) for first launch and follow-up with a bug fix soon after (3.1)

BrianNJones:
@liuche @bbinto , is there a reason we can't use: 
The most interesting videos on the Web. Recommended by Pocket, now part of Mozilla.

liuche:
@BrianNJones sorry for this being so out of the loop: the string for this is for the "help" button (which is the question mark), so the use case is:

(in accessibility mode)
1. Open Pocket Fragment
2. Move selection to "?" button

I would think having the "?" announce "The most interesting videos on the Web. Recommended by Pocket, now part of Mozilla." seems redundant. 
Maybe something like "More information about Pocket" or "Pocket help"?

liuche:
@BrianNJones any thoughts on this?

BrianNJones:
I'd vote for redundant here. It may be redundant, but it's contextual and probably helpful.

BrianNJones:
> 
> 
> Okay, sounds good.
> 
> So just to confirm, @BrianNJones, the flow would be like this:
> 
>     1. move to "?" button on homescreen
[...]

Issue #827, opened 2018-04-18
[a11y] on Pocket video feed


mcomella:
I forget how something like the Pocket header, which says "Pocket" but is generally not focusable, should be handled.

+any content descriptions for UI (that will also need to be localized)

liuche:
When opening the Pocket video feed, it goes immediately to the feed and lists off what is in the feed. 

Some problems:
- [ ] The "help" button doesn't have any accessibility text just "button"
- [ ] In Review mode (where you can select non-selectable items, read by char/word/item), the "Recommended videos" header is read, but the Pocket logo isn't selectable (because it's not text)

liuche:
Filed #1125 for the "help" button.

PR #1126, opened 2018-09-17
Issue #827 - Add accessibility to Pocket wordmark on feed

liuche:
Removing the "ignore content description" because we should add that back in in #1125 .


liuche:
ahh, I guess we need that `tools:ignore` until we get the new string so adding it back in.

PR #1135, opened 2018-09-18
Closes #1138: Fix travis CI

### LINKED ISSUE #1138: Travis CI failure
Currently Travis CI is failing, causing PR merge blocks

```Exception in thread "main" java.lang.RuntimeException: Gradle distribution '/home/travis/.gradle/wrapper/dists/gradle-4.4-all/9br9xq1tocpiv8o6njlyu5op1' does not contain any directories. Expected to find exactly 1 directory.```

### PR DESCRIPTION
**Pull request by psymoon**

### CODE REVIEW

File: .travis.yml [hunk 1]
```diff
@@ -20,9 +20,9 @@ before_install:
   - touch local.properties
 
 script:
-  - ./gradlew clean assemble
+  - ./gradlew -Pcoverage --no-daemon clean assembledebug
```
mcomella: nit: While gradle accepts case insensitivity, I'd prefer to stick to their naming for scripts: -> `assembleDebug`

.travis.yml
```diff
@@ -34,6 +34,5 @@ before_cache:
 cache:
   directories:
     - $HOME/.gradle/caches/
-    - $HOME/.gradle/wrapper/
```
mcomella: Do you have more context on why this works? Because:
- the focus team hasn't run into issues
- The official docs recommend putting this in here: https://docs.travis-ci.com/user/languages/android/#caching

I wonder if our cache some how got messed up and we actually just need to purge it (here: https://travis-ci.org/mozilla-mobile/firefox-tv/caches ?) and this change actually is otherwise not necessary.

mcomella: Here's a bit on clearing caches: https://docs.travis-ci.com/user/caching/#clearing-caches

"Sometimes you spoil your cache by storing bad data in one of the cached directories, or your cache can become invalid when language runtimes change."

psymoon:
Purging cache worked. No need for any changes

Issue #1099, opened 2018-09-12
Navigation history not preserved


AaronMT:
Build: 6dc3e493daba124e58d8d4af397c3ba5cb9aa75c

Run in Android AVD (1080P) API 27

1. Visit http://example.com
2. Tap the menu button
3. Go to Settings, and return (hit back)
4. Visit http://mozilla.org
5. Tap the menu button

Expected to see the back button accessible. It is greyed out.

I should be able to return to example.com as my first visited URL

AaronMT:
Reproducible on a Fire TV Cube device too

liuche:
Looks like this is blocking #1083 which is in the 3.0 sprint.

npark-mozilla:
Another STR:
go to google.com
go to imdb.com
go to youtube and open a video, play it
Open tiles page, press back button to traverse the history

Expected:
All sites should be shown
Actual:
google.com and imdb.com is not shown while pressing back button.  But when it reaches firefox:home, press forward, google.com and imdb.com will be shown.  This is on Basic TV Stick.

AaronMT:
This should block 3.0

PR #1163, opened 2018-09-20
Fixes #1083 - UITest fixes and test cleanup

### LINKED ISSUE #1083: Failing UI tests
UI Tests are currently broken.

Four compilation failures
- [ ] PinnedTileTests#testCustomPinnedTile
- [ ] TvScreenshots#defaultHomeScreen
- [ ] TvScreenshots#browserNavOverlay
- [ ] BrowserOverlayTest#testBrowserOverlay
- [ ] PageLoadTest#PageLoadTest

When these compilation failures are commented out, 11 of our 12 tests fail.

We need to go through all tests, determine whether or not each is still a valid test case, then update them.  For this story to be considered complete, all tests should compile and pass.

AaronMT:
- Renames BrowserOverlayTest to BrowserBFT
- UI test fixes
- Fixes Screengrabfile default
- Removes `pinTileFromOverlay` from screenshots until a less flaky solution is found, it was causing all sorts of session issues

Addresses issue #1083

### DIFF
--- a/app/src/androidTest/java/org/mozilla/focus/screenshots/TVScreenshots.java
+++ b/app/src/androidTest/java/org/mozilla/focus/screenshots/TVScreenshots.java
@@ -11,10 +11,10 @@ import android.content.SharedPreferences;
 import android.preference.PreferenceManager;
 import android.support.test.InstrumentationRegistry;
 import android.support.test.espresso.ViewInteraction;
+import android.support.test.espresso.matcher.ViewMatchers;
 import android.support.test.rule.ActivityTestRule;
 import android.support.test.runner.AndroidJUnit4;
 import android.support.test.uiautomator.UiDevice;
-import android.support.test.uiautomator.UiObjectNotFoundException;
 
 import org.junit.After;
 import org.junit.Before;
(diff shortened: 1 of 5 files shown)


AaronMT:
The BFT's will pass with with https://github.com/mozilla-mobile/firefox-tv/issues/1099 fixed

app/src/androidTest/java/org/mozilla/focus/screenshots/TVScreenshots.java
```diff
@@ -11,40 +11,40 @@
 import android.preference.PreferenceManager;
```
npark-mozilla: Can we also add test to check for "page not found" error here?  i.e. going to invalid url like "firefox:aaa" would open the dialog.

AaronMT: There's many errors, I don't know how important these are they seem low priority https://github.com/mozilla-mobile/firefox-tv/blob/master/app/src/main/res/values/strings.xml

npark-mozilla: Right, this reminds me, supposedly it is using the focus string for error message, and if so we won't need to check here again.

PR #1164, opened 2018-09-20
Closes #1148: Match pocket button styles

### LINKED ISSUE #1148: Use "Try Again" button styling for Pocket videos onboarding screen
"Okay, got it" button here:
screenshot_1537371119

should match "Try again" button here:

homescreen pocket tile error

dnarcese:


app/src/main/res/layout/pocket_onboarding.xml
```diff
@@ -49,7 +49,7 @@
         android:layout_width="88dp"
```
liuche: Can you also match the other layout params of the Button? Mainly, here the widths and height are hard-coded, but they shouldn't be.

PR #1169, opened 2018-09-20
No issue: disable codecov patch check.

mcomella:


mcomella:
After the Android meeting, we should probably set this to some minimal threshold so it creates discussions.

PR #1176, opened 2018-09-21
Closes #1146: Add focus refresh check

### LINKED ISSUE #1146: Homescreen loses focus and can't navigate when all tiles deleted
## Steps to reproduce
Delete all tiles

### Expected behavior
Closest above item is focused

### Actual behavior
Nothing is focused, focus doesn't come back with arrow buttons.

Toggling menu can bring the focus back though.

### Device information

* Fire TV device: Cube
* Firefox version: master c6d19ac8

psymoon:
For pinned tiles removal via onContextItemSelected()

### DIFF
--- a/app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt
+++ b/app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt
@@ -367,6 +367,20 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
         adapter.removeTile(tileId)
     }
 
+    /**
+     * Focus may be lost if all pinned items are removed via onContextItemSelected()
+     */
+    fun checkIfTilesFocusNeedRefresh() {
+        val adapter = tileContainer.adapter as HomeTileAdapter
+        if (adapter.itemCount == 0) {
+            if (pocketVideosContainer.isVisible) {
+                pocketVideoMegaTileView.requestFocus()
+            } else {
+                megaTileTryAgainButton.requestFocus()
+            }
+        }
+    }
+
     inner class HomeTileManager(
         context: Context,
         colCount: Int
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt
```diff
@@ -367,6 +367,20 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
         adapter.removeTile(tileId)
     }
 
+    /**
+     * Focus may be lost if all pinned items are removed via onContextItemSelected()
+     */
+    fun checkIfTilesFocusNeedRefresh() {
+        val adapter = tileContainer.adapter as HomeTileAdapter
```
Baron-Severin: Nit: We could be a bit more generic and cast to `RecyclerView.Adapter` instead, since we're only using `itemCount`.

app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt
```diff
@@ -367,6 +367,20 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
...
+        if (adapter.itemCount == 0) {
+            if (pocketVideosContainer.isVisible) {
+                pocketVideoMegaTileView.requestFocus()
+            } else {
+                megaTileTryAgainButton.requestFocus()
+            }
+        }
+    }
```
Baron-Severin: Optional nit: we could remove a bit of nesting here with something like the following:

```kotlin
val currentTiles = (tileContainer.adapter as RecyclerView.Adapter).itemCount
if (currentTiles == 0) return
val targetView = if (pocketVideosContainer.isVisible) {
    pocketVideoMegaTileView
  } else {
  megaTileTryAgainButton
}
targetView.requestFocus()
```

Up to you though, it's not a huge difference.

app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt
psymoon: Makes sense

PR #1181, opened 2018-09-22
Severin/1086 dont show firefox:home

Baron-Severin:


app/src/test/java/org/mozilla/focus/utils/UrlUtilsTest.java
```diff
@@ -112,6 +116,18 @@ public void testIsSearchQuery() {
         assertFalse(UrlUtils.isSearchQuery("mozilla"));
     }
 
+    @Test public void testCreateSearchUrl() {
+        assertCreatedUrlContainsBase("dogs are cool", "https://www.google.com/search?q=dogs%20are%20cool");
+        assertCreatedUrlContainsBase("how can mirrors be real if our eyes arent real?",
+                "https://www.google.com/search?q=how%20can%20mirrors%20be%20real%20if%20our%20eyes%20arent%20real");
+    }
+
+    private void assertCreatedUrlContainsBase(String searchTerm, String base) {
```
mcomella: nit: base -> `baseUrl` or similar.

app/src/test/java/org/mozilla/focus/utils/UrlUtilsTest.java
```diff
@@ -126,6 +142,11 @@ public void testStripUserInfo() {
         assertEquals("ftp://mozilla.org", UrlUtils.stripUserInfo("ftp://user:password@mozilla.org"));
 
         assertEquals("öäü102ß", UrlUtils.stripUserInfo("öäü102ß"));
+
+        String badUri = "https://user:password@www.i/have/percentage/signs/%.org%";
```
mcomella: This fails because it's not encoded correctly right (e.g. % should be associated with a numeric encoding). Can you change the uri to make that clearer? "I have percent signs" wasn't clear.

app/src/main/java/org/mozilla/focus/search/SearchEngine.java
```diff
@@ -61,7 +61,7 @@ public String buildSearchUrl(final String searchTerm) {
 
         final String template = Uri.decode(searchUri.toString());
         final String urlWithSubstitutions = paramSubstitution(template, Uri.encode(searchTerm));
-        return UrlUtils.normalize(urlWithSubstitutions); // User-entered search engines may need normalization.
+        return UrlUtils.INSTANCE.normalize(urlWithSubstitutions); // User-entered search engines may need normalization.
```
mcomella: fwiw, it'd be helpful to mention in the comment summary that this was an auto conversion with no modifications.

app/src/test/java/org/mozilla/focus/utils/UrlUtilsTest.java
```diff
@@ -16,104 +16,104 @@
 public class UrlUtilsTest {
     @Test
     public void isValidSearchQueryUrl() {
-        assertTrue(UrlUtils.isValidSearchQueryUrl("https://example.com/search/?q=%s"));
-        assertTrue(UrlUtils.isValidSearchQueryUrl("http://example.com/search/?q=%s"));
-        assertTrue(UrlUtils.isValidSearchQueryUrl("http-test-site.com/search/?q=%s"));
-        assertFalse(UrlUtils.isValidSearchQueryUrl("httpss://example.com/search/?q=%s"));
+        assertTrue(UrlUtils.INSTANCE.isValidSearchQueryUrl("https://example.com/search/?q=%s"));
```
mcomella: Not worth changing now given all that's piled on top of this but I don't love this churn. `INSTANCE` shouldn't be necessary with `@JvmStatic`, right?

app/src/main/java/org/mozilla/focus/utils/UrlUtils.kt
```diff
@@ -151,4 +151,9 @@ object UrlUtils {
 
         return host.substring(start)
     }
+
+    @JvmStatic
+    fun toDisplayUrl(url: String): String {
+        return ""
```
mcomella: nit: when adding code that is going to change, I usually add a TODO to notify my reviewer that I'm going to update it.

app/src/test/java/org/mozilla/focus/utils/UrlUtilsTest.kt
```diff
@@ -186,4 +186,27 @@ class UrlUtilsTest {
         assertEquals("random.mozilla.org", UrlUtils.stripCommonSubdomains("random.mozilla.org"))
         assertEquals(null, UrlUtils.stripCommonSubdomains(null))
     }
+
+    @Test
+    fun testToDisplayUrl() {
+        infix fun String.shouldBecome(output: String) {
```
mcomella: This is modifying our assertion style and it's potentially easy to overlook: I would go back to calling `assert*` directly or encapsulating that style in a helper function, e.g. `assertToDisplayUrlIsUnchanged(input)`.

Two other options to keep syntax clean are:
- Put the assertions in a loop. This complicates test code and obfuscates line numbers though
- Create a parameterized test. I haven't tried this before and JUnit's syntax seems verbose, but it solves the complexity of loops. This also encourages the style of unit testing where you have one unit test for each "case".

Parameterized tests would look something like:
```kotlin
@Test
@Parameters("error:-8", "hello world")
fun `test displayUrl is unchanged for X style parameter`(input: String) {
    assertEquals(input, UrlUtils.toDisplayUrl(input)
}
```

If you're interested in changing our assertion style, I'd propose it to the team. Personally, I've been missing jasmine's (it's behavior driven, I think?) hierarchical test setup and naming structure: it works well with GIVEN, WHEN, THEN.

app/src/test/java/org/mozilla/focus/utils/UrlUtilsTest.kt
```diff
@@ -186,4 +186,27 @@ class UrlUtilsTest {
         assertEquals("random.mozilla.org", UrlUtils.stripCommonSubdomains("random.mozilla.org"))
         assertEquals(null, UrlUtils.stripCommonSubdomains(null))
     }
+
+    @Test
+    fun testToDisplayUrl() {
```
mcomella: I wonder if we should be more expressive with our test names (and testing style), even if it's inconsistent with the rest of the file. I think the win for readability when tests fail is worth it.

app/src/main/java/org/mozilla/focus/utils/UrlUtils.kt
```diff
@@ -151,4 +151,9 @@ object UrlUtils {
 
         return host.substring(start)
     }
+
+    @JvmStatic
+    fun toDisplayUrl(url: String): String {
```
mcomella: Writing the test first: cool idea! :)

mcomella: nit: -> `toToolbarDisplayUrl`.

I wonder if this should be packaged with other Toolbar related methods because it's unlikely we'll be using this more generally.

app/src/main/java/org/mozilla/focus/utils/UrlUtils.kt
```diff
@@ -154,6 +155,9 @@ object UrlUtils {
 
     @JvmStatic
     fun toDisplayUrl(url: String): String {
-        return ""
+        return when (url) {
+            BrowserFragment.APP_URL_HOME -> ""
```
mcomella: I think we could do better than using the empty string to represent the intent of "use the hint text". e.g. `Either<String, HintText>`?

Given that we're pressed for time, we can leave this as is for now and at least document:
- Document in the function javadoc what the return values are
- Document inline why the empty string (because it forces TextView to show the intent)

app/src/test/java/org/mozilla/focus/utils/UrlUtilsTest.java
Baron-Severin: Whoops.

app/src/main/java/org/mozilla/focus/utils/UrlUtils.kt
Baron-Severin: Good idea.  I'll do that in the future.

app/src/main/java/org/mozilla/focus/utils/UrlUtils.kt
Baron-Severin: I don't think we actually have a toolbar in Fire TV.  At least there don't seem to be any Toolbar related classes.

Issue #1005, opened 2018-07-30
Verify history is being cleared correctly


mcomella:
I think we need to run clear on the WebView used by the app, not just the one in settings.

psymoon:
So it looks like webView instance in BrowserFragment is null after the static cleanup methods get called from SettingsFragment. This suggests that we are clearing the history properly from the original implementation. Adding addition checks to make sure Browser webView instance calls cleanup() IFF not null. See https://github.com/mozilla-mobile/firefox-tv/pull/1016

psymoon:
Looks to be working as intended with the current implementation

Baron-Severin:
We've changed how we're handling WebView creation, so we need to reopen this issue.

PR #1182, opened 2018-09-23
Fixes #1099: retains WebView between fragments

### LINKED ISSUE #1099: Navigation history not preserved

Baron-Severin:
- Creates WebViewCache to store SystemEngineView instances
- Updates MainActivity#onCreateView to retrieve from WebViewCache instead of creating an instance

### DIFF
--- a/app/src/main/java/org/mozilla/focus/MainActivity.kt
+++ b/app/src/main/java/org/mozilla/focus/MainActivity.kt
@@ -21,6 +21,7 @@ import org.mozilla.focus.browser.BrowserNavigationOverlay
 import org.mozilla.focus.browser.WebViewCache
 import org.mozilla.focus.browser.VideoVoiceCommandMediaSession
 import org.mozilla.focus.ext.components
+import org.mozilla.focus.ext.setupForApp
 import org.mozilla.focus.ext.toSafeIntent
 import org.mozilla.focus.home.pocket.Pocket
 import org.mozilla.focus.home.pocket.PocketOnboardingActivity
@@ -45,6 +46,7 @@ class MainActivity : LocaleAwareAppCompatActivity(), OnUrlEnteredListener, Media
     // There should be at most one MediaSession per process, hence it's in MainActivity.
     // We crash if we init MediaSession at init time, hence lateinit.
     override lateinit var videoVoiceCommandMediaSession: VideoVoiceCommandMediaSession
+    private lateinit var webViewCache: WebViewCache
 
     private val sessionObserver = object : SessionManager.Observer {
         override fun onSessionSelected(session: Session) {
@@ -75,6 +77,7 @@ class MainActivity : LocaleAwareAppCompatActivity(), OnUrlEnteredListener, Media
         Pocket.init()
         PublicSuffix.init(this) // Used by Pocket Video feed & custom home tiles.
         initMediaSession()
+        webViewCache = WebViewCache()
 
         val intent = SafeIntent(intent)
 
@@ -139,14 +142,11 @@ class MainActivity : LocaleAwareAppCompatActivity(), OnUrlEnteredListener, Media
         TelemetryWrapper.stopMainActivity()
     }
 
-    override fun onDestroy() {
-        super.onDestroy()
-        WebViewCache.clear()
-    }
-
     override fun onCreateView(name: String, context: Context, attrs: AttributeSet): View? {
         return if (name == EngineView::class.java.name) {
-            return WebViewCache.getWebView(context, attrs)
+            webViewCache.getWebView(context, attrs) {
+                setupForApp(context)
+            }
         } else super.onCreateView(name, context, attrs)
     }
 
--- a/app/src/main/java/org/mozilla/focus/MainActivity.kt
+++ b/app/src/main/java/org/mozilla/focus/MainActivity.kt
@@ -12,16 +12,15 @@ import android.util.AttributeSet
 import android.view.KeyEvent
 import android.view.View
 import kotlinx.android.synthetic.main.activity_main.*
-import mozilla.components.browser.engine.system.SystemEngineView
 import mozilla.components.browser.session.Session
 import mozilla.components.browser.session.SessionManager
 import mozilla.components.concept.engine.EngineView
 import org.mozilla.focus.browser.BrowserFragment
 import org.mozilla.focus.browser.BrowserFragment.Companion.APP_URL_HOME
 import org.mozilla.focus.browser.BrowserNavigationOverlay
+import org.mozilla.focus.browser.WebViewCache
 import org.mozilla.focus.browser.VideoVoiceCommandMediaSession
 import org.mozilla.focus.ext.components
-import org.mozilla.focus.ext.setupForApp
 import org.mozilla.focus.ext.toSafeIntent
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/mozilla/focus/browser/WebViewCache.kt
```diff
@@ -0,0 +1,49 @@
...
+        fun View?.removeFromParentIfAble() {
+            (this?.parent as? ViewGroup)?.removeView(cachedView)
+        }
+        fun createAndCacheEngineView(): SystemEngineView {
+            return SystemEngineView(context, attrs).apply {
+                setupForApp(context)
+            }.also { cachedView = it }
+        }
+
+        cachedView?.removeFromParentIfAble()
```
mcomella: Why do we need to remove this from the parent? Is this to avoid having the WebView added to the view hierarchy twice? Can that happen in practice or is this just a safety measure?

Add a comment.

app/src/test/java/org/mozilla/focus/browser/WebViewCacheTest.kt
```diff
@@ -0,0 +1,46 @@
...
+import org.mockito.Mockito.mock
+import org.robolectric.RobolectricTestRunner
+import org.robolectric.RuntimeEnvironment
+
+@RunWith(RobolectricTestRunner::class)
+class WebViewCacheTest {
+
+    @Test
+    fun `WHEN getWebView is called multiple times THEN the same WebView is returned`() {
+        val cachRefs = List(5) { getWebView() }
```
mcomella: nit: -> cacheRefs

Or cashRefs, for the lols. ;)

app/src/test/java/org/mozilla/focus/browser/WebViewCacheTest.kt
```diff
@@ -0,0 +1,46 @@
...
+        cachRefs.forEach { assertEquals(cachRefs.first(), it) }
```
mcomella: Should we use reference equality here (`===`) instead of structural equality?

app/src/test/java/org/mozilla/focus/browser/WebViewCacheTest.kt
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    @Test
+    fun `GIVEN cached WebView has a parent WHEN WebView is returned THEN WebView should be removed from parent`() {
+        for (i in 0..5) {
```
mcomella: nit: I think this test would make more sense if it was written without a loop: in order to read the core existence of the test (removed from parent), you need to read the bottom assertion and go back to the top assertion.

We don't need to test this more than once or twice so I don't think the loop is necessary.

app/src/test/java/org/mozilla/focus/browser/WebViewCacheTest.kt
```diff
@@ -0,0 +1,46 @@
...
+            val webView = getWebView()
+            assertEquals(null, webView.parent)
+            val parent = FrameLayout(RuntimeEnvironment.application)
```
mcomella: nit: If we do set-up inside the method, I'd use newlines to separate setup (GIVEN) from WHEN and THEN, unless the function is trivial. In this case, I think GIVEN should be separated but WHEN and THEN can probably be put together.

app/src/test/java/org/mozilla/focus/browser/WebViewCacheTest.kt
```diff
@@ -0,0 +1,46 @@
...
+            parent.addView(webView)
+            assertEquals(parent, webView.parent)
+        }
+    }
+
+    @Test
+    fun `GIVEN cache has been cleared WHEN WebView is retrieved from cache THEN it should be a different instance`() {
+        val webView = getWebView()
+        assertEquals(webView, getWebView())
```
mcomella: This assertion seems to be testing "WHEN getWebView is called multiple times THEN the same WebView is returned". I would remove it: it creates a dependency on that test. If `getWebView` is returning a new WebView each time, this should pass because it meets the conditions for passing and the other test should fail to catch it.

app/src/test/java/org/mozilla/focus/browser/WebViewCacheTest.kt
```diff
@@ -0,0 +1,46 @@
+package org.mozilla.focus.browser
```
mcomella: nit: license

app/src/main/java/org/mozilla/focus/browser/BrowserFragment.kt
```diff
@@ -225,11 +225,6 @@ class BrowserFragment : EngineViewLifecycleFragment(), Session.Observer {
 
         if (session.url == APP_URL_HOME) {
             browserOverlay?.visibility = View.VISIBLE
-        } else {
```
mcomella: We also need to do #1005: since the WebView survives, it needs to have `clearData` called on it after we clear data in the settings (we need a test for that...).

This could be a follow-up but needs to happen for v3.0.

app/src/main/java/org/mozilla/focus/MainActivity.kt
```diff
@@ -140,11 +139,14 @@ class MainActivity : LocaleAwareAppCompatActivity(), OnUrlEnteredListener, Media
...
+        super.onDestroy()
+        WebViewCache.clear()
+    }
+
     override fun onCreateView(name: String, context: Context, attrs: AttributeSet): View? {
         return if (name == EngineView::class.java.name) {
-            SystemEngineView(context, attrs).apply {
-                setupForApp(context)
-            }
+            return WebViewCache.getWebView(context, attrs)
```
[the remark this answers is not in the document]
Baron-Severin: I'm honestly not sure if it should be a permanent solution.  I think changing the fragment transactions would be better, but I'm not sure the cost-benefit of changing that really makes sense.

If we were to move to lazy WV initialization, I don't see one method being particularly easier than the other to set up.

app/src/main/java/org/mozilla/focus/browser/WebViewCache.kt
Baron-Severin: If we try to add a view to the view hierarchy that already has a parent it'll throw an exception.  Until I added this method, it crashed 100% of the time.  I'll add a comment.

app/src/test/java/org/mozilla/focus/browser/WebViewCacheTest.kt
Baron-Severin: Removed this test completely after changing `WebViewCache` to no longer be a singleton.

app/src/main/java/org/mozilla/focus/browser/BrowserFragment.kt
Baron-Severin: Reopened #1005.

Issue #878, opened 2018-05-01
[a11y] Make sure it's not broken on overlay


mcomella:
We did a lot of refactoring so let's make sure it all still works, especially the tile we added.

*edit: behavior differences I see:*
- [x] 1) On the initial overlay, "firefox:home" is announced instead of "Search or enter address"
- [x] 2) There is a bug when opening the overlay over web content where the web content will continue to be announced after the url bar is read
- [x] 3) Pocket videos in not announced
- [x] 4) Recycler view now has "row x, column y" announcements, unlike before
- [x] 5) Pocket error state, blocked on #889 
- [x] 6) Press and hold to SELECT toast is sometimes announced when menu is opened *I'm having trouble reproducing this*

mcomella:
@aminalhazwani For 1, "firefox:home" isn't recognizable as a URL even for visual users so it could be confusing what users can do here. Should we replace that text on the initial screen with "Search or enter address" and display a url when on a real web page like before?

---

I fixed 3 already and 4 actually sounds like an improvement.

mcomella:
I'm having difficulty with 2: I can't seem to get the announcements to stop. I tried `interrupt`, changing focusability of the webView when the overlay is shown, and overriding accessibility delegates. One set of code I had seemed to work but I lost that revision (d'oh!). I'm concerned that the events are getting added to a queue which we don't have access to which makes it hard to stop the events, or that the page is getting updated asynchronously in the background, which draws focus and announcements.

One fragile option we can try is changing visibility state of the WebView when accessibility is enabled/disabled. It's fragile because if we have a bug in our code, the browser could stop working altogether and it's pretty easy to break with code refactors (because we don't test these code paths often).

I'll revisit this after I play around with the other issues.

aminalhazwani:
> Should we replace that text on the initial screen with "Search or enter address" and display a url when on a real web page like before?

@mcomella yes, let's use the same placeholder "Search or enter address". 

I noticed that now on Desktop we use "Search with {name of search engine} or enter address". Is this something we also want to follow? Maybe in the future where you could choose among multiple search engines.

liuche:
1) firefox:home is announced - we don't have a "Search or enter address" so this is expected
2) Overlay is working - Verified that the following are announced: nav buttons, url, pocket recommendations tile, individual tiles (title, row, column)
3) Pocket videos megatile is announced
4) Recycler view now has "row x, column y" announcements
5) Pocket error state can be covered in #827
[...]

PR #1183, opened 2018-09-24
Issue #878: Fix accessibility on overlay

dnarcese:
- Fix 'Press and hold SELECT...' toast not being announced 
    - Added a short delay for displaying the toast when VoiceView is enabled. This ensures that it doesn't disappear before VoiceView gets around to announcing it.
- Fix Pocket megatile error message not being announced
    - Now, the error message will be announced first followed by the 'Try again' button.

### DIFF
--- a/app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt
+++ b/app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt
@@ -40,7 +40,6 @@ import org.mozilla.focus.widget.InlineAutocompleteEditText
 import kotlin.properties.Delegates
 import org.mozilla.focus.ext.isVoiceViewEnabled
 
-
 private const val NAVIGATION_BUTTON_ENABLED_ALPHA = 1.0f
 private const val NAVIGATION_BUTTON_DISABLED_ALPHA = 0.3f
 
@@ -69,7 +68,7 @@ enum class NavigationEvent {
         const val VAL_UNCHECKED = "unchecked"
     }
 }
-
+@Suppress("LargeClass")
 class BrowserNavigationOverlay @JvmOverloads constructor(
     context: Context,
     attrs: AttributeSet? = null,
@@ -169,8 +168,7 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
                 // is finished. This adds a short delay to the display of the toast so it will always be announced
                 if (context.isVoiceViewEnabled())
                     uiHandler.postDelayed(
-                        { Toast.makeText(context, R.string.homescreen_unpin_tutorial_toast, Toast.LENGTH_LONG).show() }
-                        , 1500)
+                        { Toast.makeText(context, R.string.homescreen_unpin_tutorial_toast, Toast.LENGTH_LONG).show() }, 1500)
 
                 canShowUpinToast = false
             }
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt
```diff
@@ -159,7 +163,13 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
                         .edit()
                         .putInt(SHOW_UNPIN_TOAST_COUNTER_PREF, prefInt + 1)
                         .apply()
-                Toast.makeText(context, R.string.homescreen_unpin_tutorial_toast, Toast.LENGTH_LONG).show()
+
+                // With VoiceView enabled, the toast is not announced if it disappears before the home tile description
+                // is finished. This adds a short delay to the display of the toast so it will always be announced
+                if (context.isVoiceViewEnabled())
+                    uiHandler.postDelayed(
+                        { Toast.makeText(context, R.string.homescreen_unpin_tutorial_toast, Toast.LENGTH_LONG).show() }, 1500)
```
Baron-Severin: We want to avoid capturing context in this lambda, since it could potentially outlive the fragment.  That would briefly leak the fragment, and I'm not sure what happens when we call `makeText` with a destroyed context but I'm sure it's not good.

Suggest wrapping the context in a WeakReference, then unwrapping it in the lambda.  Make sure to set up the WeakReference before the call to `postDelayed`, or the lambda will implicitly capture the context anyway.

app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt
```diff
@@ -159,7 +164,15 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
                         .edit()
                         .putInt(SHOW_UNPIN_TOAST_COUNTER_PREF, prefInt + 1)
                         .apply()
-                Toast.makeText(context, R.string.homescreen_unpin_tutorial_toast, Toast.LENGTH_LONG).show()
+
+                val contextReference = WeakReference(context)
+                val showToast = postDelayed@{
```
Baron-Severin: Nit: We don't always delay this one, only when VV is enabled.  Might be better to name it `showToast`, same as the val?  Or something similar.

app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt
```diff
@@ -36,14 +38,17 @@ import org.mozilla.focus.utils.TurboMode
...
 import kotlin.properties.Delegates
+import org.mozilla.focus.ext.isVoiceViewEnabled
+import java.lang.ref.WeakReference
 
 private const val NAVIGATION_BUTTON_ENABLED_ALPHA = 1.0f
 private const val NAVIGATION_BUTTON_DISABLED_ALPHA = 0.3f
 
 private const val SHOW_UNPIN_TOAST_COUNTER_PREF = "show_upin_toast_counter"
-private const val MAX_UNPIN_TOAST_COUNT = 3
+private const val MAX_UNPIN_TOAST_COUNT = 100
```
Baron-Severin: Nit: remember to change this one back

PR #1185, opened 2018-09-24
Fixes #1155: fixes white screen on nav to firefox:home

### LINKED ISSUE #1155: Going to firefox:home shows white screen, makes app unusuable
## Steps to reproduce
Open Firefox app
go to URL bar, and either type firefox:home or press Enter if the bar is prepopulated with 'firefox:home' text

### Expected behavior
Shows main tiles page

### Actual behavior
White screen, app becomes unusuable. need to force stop

### Device information

* Fire TV device: Basic TV stick
* Firefox version: master branch

Baron-Severin:
I think that making this check in the BrowserFragment is wrong, but setting the visibility here is probably also wrong.  Without refactoring how we handle NavigationEvents, I don't see a better way to do this.

### DIFF
--- a/app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt
+++ b/app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt
@@ -26,6 +26,7 @@ import kotlinx.android.synthetic.main.pocket_video_mega_tile.view.*
 import kotlinx.coroutines.experimental.Job
 import kotlinx.coroutines.experimental.android.UI
 import kotlinx.coroutines.experimental.launch
+import mozilla.components.support.ktx.android.view.hideKeyboard
 import org.mozilla.focus.R
 import org.mozilla.focus.autocomplete.UrlAutoCompleteFilter
 import org.mozilla.focus.ext.forEachChild
@@ -70,7 +71,8 @@ enum class NavigationEvent {
         const val VAL_UNCHECKED = "unchecked"
     }
 }
-@Suppress("LargeClass")
+
+@Suppress("LargeClass") // TODO remove this. See https://github.com/mozilla-mobile/firefox-tv/issues/1187
 class BrowserNavigationOverlay @JvmOverloads constructor(
     context: Context,
     attrs: AttributeSet? = null,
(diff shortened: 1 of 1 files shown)


Baron-Severin:
FWIW, I spent around an hour trying to add tests, but was unable to get BrowerFragment into a testing harness.

app/src/main/java/org/mozilla/focus/browser/BrowserFragment.kt
```diff
@@ -128,7 +128,7 @@ class BrowserFragment : EngineViewLifecycleFragment(), Session.Observer {
             NavigationEvent.SETTINGS -> ScreenController.showSettingsScreen(fragmentManager!!)
             NavigationEvent.LOAD_URL -> {
                 (activity as MainActivity).onTextInputUrlEntered(value!!, autocompleteResult!!, UrlTextInputLocation.MENU)
-                setOverlayVisible(false)
+                if (value != Urls.APP_HOME) setOverlayVisible(false)
```
mcomella: Rather than stop this at the LOAD_URL layer, can we stop it at the input validation layer? i.e. in the commit listener, don't call load url if the URL is invalid? This seems less reactive to me and doesn't require us to try to undo state (e.g. we're recording a telemetry event for this page load).

Baron-Severin: Good point.

app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt
```diff
@@ -70,6 +71,9 @@ enum class NavigationEvent {
         const val VAL_UNCHECKED = "unchecked"
     }
 }
+
+// TODO remove this suppression.
```
mcomella: nit: inline, e.g.:
```kotlin
@Suppress("LargeClass") // #1187: we'll address this later

File: app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt [hunk 3]
```diff
@@ -238,6 +242,11 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
     private fun setupUrlInput() = with(navUrlInput) {
         setOnCommitListener {
             val userInput = text.toString()
+            if (userInput == BrowserFragment.APP_URL_HOME) {
```
mcomella: Much better! Add a comment to explain why this is necessary.

PR #1186, opened 2018-09-25
Fixes #1005: clear history on clear data

### LINKED ISSUE #1005: Verify history is being cleared correctly

Baron-Severin:


app/src/main/java/org/mozilla/focus/SettingsFragment.kt
```diff
@@ -52,6 +52,9 @@ class SettingsFragment : Fragment() {
                     requireComponents.engine.deleteData(this)
                     requireComponents.sessionManager.removeAll()
                     dialog.cancel()
+                    // The call to recreate destroys state being maintained in the
+                    // WebView (including navigation history) and Activity
+                    activity?.recreate()
```
mcomella: This will stop working if we add session restoration logic. From the docs:

> This results in essentially the same flow as when the Activity is created due to a configuration change

If we weren't pressed for time, I would say let's fix this the right way but since we are, let's leave it as is. Just:
- Add a comment mentioning ^
- File a follow-up to write a regression test so we catch this when it changes (perhaps a UI smoke test since BF is difficult to harness). Mention this is high priority: 3.1?

I don't think there's a point in filing a bug to implement the right solution if we have a regression test.

Baron-Severin: Filed https://github.com/mozilla-mobile/firefox-tv/issues/1192

PR #1196, opened 2018-09-25
Severin/i1193 dont display data url

Baron-Severin:


Baron-Severin:
@mcomella Had a mistake in the commit history.  Fixed, and WIP tag removed.

app/src/main/java/org/mozilla/focus/browser/Urls.kt
```diff
@@ -0,0 +1,14 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.focus.browser
+
+/**
+ * A collection of known URLs and parts of URLs.
+ */
+object Urls {
+    const val APP_HOME = "firefox:home"
+    const val APP_PREFIX = "firefox:"
```
mcomella: nit: `private const val SCHEME_APP = "firefox"`

- private: unused elsewhere
- It's more flexible to insert the colon via string interpolation.
- Scheme is more specific

app/src/main/java/org/mozilla/focus/browser/Urls.kt
```diff
@@ -0,0 +1,14 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.focus.browser
+
+/**
+ * A collection of known URLs and parts of URLs.
+ */
+object Urls {
+    const val APP_HOME = "firefox:home"
```
mcomella: nit: Should interpolate `APP_PREFIX`

app/src/main/java/org/mozilla/focus/browser/Urls.kt
```diff
@@ -0,0 +1,14 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.focus.browser
+
+/**
+ * A collection of known URLs and parts of URLs.
+ */
+object Urls {
```
mcomella: nit: -> `URLs`.

I'm not opinionated about whether it should be URL or Url but we should be consistent with our other products.

app/src/main/java/org/mozilla/focus/browser/Urls.kt
```diff
@@ -10,5 +10,6 @@ package org.mozilla.focus.browser
 object Urls {
     const val APP_HOME = "firefox:home"
     const val APP_PREFIX = "firefox:"
+    const val DATA_PREFIX = "data:"
```
mcomella: nit: -> `SCHEME_DATA = "data"`

Might want to group separately from the other things too:
```kotlin
SCHEME_...
SCHEME_...

APP_HOME
...
```

app/src/main/java/org/mozilla/focus/utils/UrlUtils.kt
```diff
@@ -161,6 +161,7 @@ object UrlUtils {
     fun toUrlBarDisplay(url: String): String {
         return when {
             url == Urls.APP_HOME -> "" // Empty string forces the EditText to show hint text
+            url.startsWith(Urls.DATA_PREFIX) -> ""
```
mcomella: Maybe now is not the time but if `url` was a `Uri`, you could do:
```kotlin
url.scheme == URLs.SCHEME_DATA
```

mcomella: Actually, we can't do this for security reason. Imagine the following attack vector:
- User goes to malicious site
- Malicious site sends user to `data:...` site that is google login page
- User sees "Search or enter address", much less suspicious than a `data:` URI

Options:
- Components can use the `loadDataWithBaseUrl` method instead of `loadData`: this would let us set our own URL (in [this code](https://github.com/mozilla-mobile/android-components/blob/14fdb2d78a009bc0dc782cb28898a6f5d3128c23/components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineSession.kt#L71))
- We could whitelist our exact error page though I'd prefer to fix it the right way ^

mcomella: I filed https://github.com/mozilla-mobile/android-components/issues/865 against components.

Baron-Severin:
Closed due to security issue with implementation. https://github.com/mozilla-mobile/android-components/issues/865 opened.

PR #1201, opened 2018-09-26
Fixes #1195, Fixes #1193: updates A-C to 25

### LINKED ISSUE #1193: On error page, URL in menu, progress bar is replaced with data:...
## Steps to reproduce
- Open browser
- Type "firefox:gnome" or other URL which will never resolve
- Click menu

### Expected behavior
URL is "firefox:gnome"

### Actual behavior
URL is `data:...`

screenshot_1537908683

This also affects the progress bar when changing sites:
screenshot_1537908720

Can we write a regression test for this? Or file a follow-up to do so.

### LINKED ISSUE #1195: java.net.URISyntaxException: Expected scheme-specific part at index
## Steps to reproduce
- User types in invalid URL like "fgh:"

### Expected behavior
Error page loads.

### Actual behavior
Crash! Under some conditions: looks like we may need to a "security change".

This is a components issue: see https://github.com/mozilla-mobile/android-components/issues/863

### Device information

* Fire TV device: ?
* Firefox version: ?

Baron-Severin:
- 1195: Fixes crash when user inputs bad URI
- 1193: Fixes URL field being filled with data on error
- A-C #825: Fixes URL field/ProgressBar URL not being updated until a page has almost finished loading

### DIFF
--- a/app/src/main/java/org/mozilla/focus/browser/BrowserFragment.kt
+++ b/app/src/main/java/org/mozilla/focus/browser/BrowserFragment.kt
@@ -340,14 +340,7 @@ class BrowserFragment : EngineViewLifecycleFragment(), Session.Observer {
 
         if (event.keyCode == KeyEvent.KEYCODE_DPAD_CENTER && actionIsDown) MenuInteractionMonitor.selectPressed()
 
-        /**
-         * [Session.url] is currently set late in the loading process, so while loading the first
-         * site of a session [isUrlEqualToHomepage] will return true for some time (often several
-         * seconds).  We check [Session.loading] to prevent the menu button from being unresponsive
-         * during that time
-         */
-        val overlayCanBeInvisible = !isUrlEqualToHomepage || session.loading
-        if (event.keyCode == KeyEvent.KEYCODE_MENU && overlayCanBeInvisible) {
+        if (event.keyCode == KeyEvent.KEYCODE_MENU && !isUrlEqualToHomepage) {
             if (actionIsDown) {
                 val toShow = !browserOverlay.isVisible
                 setOverlayVisible(toShow)
--- a/app/src/main/java/org/mozilla/focus/engine/CustomContentRequestInterceptor.kt
+++ b/app/src/main/java/org/mozilla/focus/engine/CustomContentRequestInterceptor.kt
@@ -37,7 +37,13 @@ class CustomContentRequestInterceptor(
     }
 
     @Suppress("NestedBlockDepth")
-    override fun onErrorRequest(session: EngineSession, errorCode: Int, uri: String?) {
+    override fun onErrorRequest(session: EngineSession, errorCode: Int, uri: String?): RequestInterceptor.ErrorResponse? {
+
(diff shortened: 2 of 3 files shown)


app/src/main/java/org/mozilla/focus/browser/BrowserFragment.kt
```diff
@@ -340,14 +340,7 @@ class BrowserFragment : EngineViewLifecycleFragment(), Session.Observer {
...
 
-        /**
-         * [Session.url] is currently set late in the loading process, so while loading the first
-         * site of a session [isUrlEqualToHomepage] will return true for some time (often several
-         * seconds).  We check [Session.loading] to prevent the menu button from being unresponsive
-         * during that time
-         */
-        val overlayCanBeInvisible = !isUrlEqualToHomepage || session.loading
-        if (event.keyCode == KeyEvent.KEYCODE_MENU && overlayCanBeInvisible) {
+        if (event.keyCode == KeyEvent.KEYCODE_MENU && !isUrlEqualToHomepage) {
```
Baron-Severin: Session.url no longer loads slowly, so this hack can be cleaned out.

mcomella: For future reference, this type of comment is best put into a small, standalone commit with the comment in the commit summary so that this information is captured in version control.

mcomella: This also would have been a great commit to backout. :)

app/src/main/java/org/mozilla/focus/engine/CustomContentRequestInterceptor.kt
```diff
@@ -37,7 +37,13 @@ class CustomContentRequestInterceptor(
     }
 
     @Suppress("NestedBlockDepth")
-    override fun onErrorRequest(session: EngineSession, errorCode: Int, uri: String?) {
+    override fun onErrorRequest(session: EngineSession, errorCode: Int, uri: String?): RequestInterceptor.ErrorResponse? {
+
+        fun createErrorResponse(withUri: String, withErrorCode: Int): RequestInterceptor.ErrorResponse {
```
mcomella: nit: remove `with` -> that's not a java/kotlin convention. fwiw, it makes sense in swift where labels are required in some circumstances, becoming part of the method name, and you can supply alternative names for use within the function, e.g. a call in Swift might be `createUri(forString: string)` instead of Java: `createUriForString(string)`

Baron-Severin: I wanted names that didn't shadow the params from the surrounding method, and really struggled to find anything better.  Do you have any suggestions?

mcomella: Huh, I didn't notice that. Maybe `errorResponseUri` and `errorResponseCode`? They provide more accurate specificity than `uri` and `errorCode`. But it's possible someone will refactor those to be "uri" and "errorCode".

PR #1221, opened 2018-09-27
CLoses #1204: correct web settings

### LINKED ISSUE #1204: Video content from Pocket feed does not autoplay; user agent & other settings are changed
## Steps to reproduce
- Open pocket
- Click a video (I chose the second one)

### Expected behavior
Video plays

### Actual behavior
Black screen.

Autoplay was originally implemented in #586. It appears to work when visiting YouTube directly but doesn't work for Pocket videos though I wouldn't expect different behavior.

Autoplay requires `mediaPlaybackRequiresUserGesture` to be false - maybe it's not getting set correctly?

### Device information

* Fire TV device: 4k
* Firefox version: v3.0 on e869530c2be44344627723c83defdf3af395a2be + #1201

mcomella:


app/src/main/java/org/mozilla/focus/FocusApplication.kt
```diff
@@ -6,18 +6,24 @@ package org.mozilla.focus
...
+import android.webkit.WebSettings
 import org.mozilla.focus.locale.LocaleAwareApplication
 import org.mozilla.focus.search.SearchEngineManager
 import org.mozilla.focus.session.VisibilityLifeCycleCallback
 import org.mozilla.focus.telemetry.TelemetryWrapper
 import org.mozilla.focus.utils.AppConstants
 import org.mozilla.focus.utils.OkHttpWrapper
 
-class FocusApplication : LocaleAwareApplication() {
+open class FocusApplication : LocaleAwareApplication() {
```
Baron-Severin: Optional nit: do you think it might be a good idea to add "// Open for testing"?

mcomella: I figured when someone removed it and the tests broke, they'd know. @Baron-Severin What do you think?

PR #1223, opened 2018-09-28
Fixes #1216: fixes bug that caused WebView to freeze

### LINKED ISSUE #1216: WebView fails to load any content after hitting back on initial website (white screen)
Running e869530c2be44344627723c83defdf3af395a2be on a FireTV Cube in the office.

- Clicked on IMDB
- _IMDB loads_
- Hit back
- Clicked on Google Video Search
- _white screen of death_

Reproduced also via:

- Visit Mozilla.org
- Hit back on the remote
- Clicked on IMDB
- _white screen of death_

Baron-Severin:


app/src/main/java/org/mozilla/focus/browser/WebViewCache.kt
```diff
@@ -42,4 +42,9 @@ class WebViewCache {
         cachedView?.removeFromParentIfAble()
         return cachedView ?: createAndCacheEngineView()
     }
+
+    fun clear() {
```
mcomella: Two ideas:
- I wonder if this should be called `onDestroy` to make it clear that it should be called in `onDestroy`
- If we're doing ^, then maybe we should use a lifecycle observer to make it automatic

*Then again*, a `WebViewCache` can be used outside the context of an Activity, in theory (though, it is dealing with views whose lifecycles are the same as contexts...) so maybe it doesn't make sense to name it as if the lifecycle is the same.

app/src/main/java/org/mozilla/focus/browser/WebViewCache.kt
```diff
@@ -42,4 +45,10 @@ class WebViewCache {
         cachedView?.removeFromParentIfAble()
         return cachedView ?: createAndCacheEngineView()
     }
+
+    @OnLifecycleEvent(Lifecycle.Event.ON_DESTROY)
+    private fun clear() {
```
mcomella: nit: call `onDestroy`; calling it clear will discourage other people from using `onDestroy` as `onDestroy`, thinking it might be just for `clear`

Issue #1147, opened 2018-09-19
Enhance Turbo Mode onboarding screen


aminalhazwani:
### Why/User Benefit/User Problem

We introduce Turbo Mode on first app launch but we don't show how Turbo Mode looks in the app - aka rocket icon.

### What / Requirements

Add a rocket icon to the Turbo Mode onboarding. Potentially we could also add a section of the screen so that users see where they are going to find the Turbo Mode button.

aminalhazwani:
@liuche I uploaded the two proposal with the web inspector, see "Turbo Mode Onboarding" here https://github.com/mozilla-mobile/firefox-tv/wiki/Design#mocks-with-a-web-interface

And in case we opt for option 2 I added the asset here https://github.com/mozilla-mobile/firefox-tv/wiki/Design#assets

Option B can be animated (we could create a video that we bundle with the app) or can also be a static image.

Both option A and B use the same buttons and typography as the Pocket onboarding, just posting this for reference:

<img width="792" alt="screen shot 2018-09-20 at 3 57 44 pm" src="https://user-images.githubusercontent.com/1721875/45823420-f811fe80-bced-11e8-9aeb-2c80afe69581.png">

I lean towards option B (with or without video) as it displays the product, the brand, the turbo mode button/icon and sets the right expectations. Option A is a bit too anonymous, we could add the Firefox brand there as well but it might still be detached to the product.

Baron-Severin:
@BrianNJones We're updating this layout to match the following link.  It includes a large image over the top half of the screen, what should its content description (i.e., accessibility string) be?

https://firefoxux.github.io/people/aalhazwani/firefox-for-fire-tv/turbo-mode-onboarding/index.html#artboard1

Baron-Severin:
> @Baron-Severin @aminalhazwani
> Turbo Mode helps you browse faster
> Turbo Mode automatically blocks ads and trackers. If a site doesn't act the way you expect, try turning it off.

@BrianNJones I don't understand, is this entire string the image's content description?  Do we need the line break?  I'm not sure how line breaks are handled by Talk Back.

aminalhazwani:
@Baron-Severin this is how we could set the line break:

<img width="703" alt="image" src="https://user-images.githubusercontent.com/1721875/46461492-13393f80-c7bf-11e8-8ce1-4e751ce3dd2c.png">

The distance between the text and the buttons remains unvaried.

Baron-Severin:
[...]

PR #1228, opened 2018-09-28
Issue #1147: add strings for Turbo Mode onboarding

Baron-Severin:


app/src/main/res/values/strings.xml
```diff
@@ -181,10 +181,13 @@ be temporary, and you can try again later.</li>
     <string name="turbo_mode_disabled_toast">Turbo Mode disabled</string>
     <string name="onboarding_turbo_mode_title">Turbo Mode helps you browse faster</string>
     <string name="onboarding_turbo_mode_body">Turbo Mode automatically blocks trackers and ads to help get you where you\'re going faster. If there\'s a speed bump – like a site that isn\'t behaving the way you expect – just turn Turbo Mode off.</string>
+    <string name="onboarding_turbo_mode_body2">Turbo Mode automatically blocks trackers and ads to help get you where you’re going faster. If there’s a speed bump – like a site that isn’t behaving the way you expect – just turn Turbo Mode off.</string>
```
mcomella: Did you try building this? I think you need to escape the apostrophes.

mcomella: How is this different from the string that's above it? The strings should be named to indicate *where* in the UI they're used.

Are the old strings no longer used? In that case you should remove them.

mcomella: Then again, if these strings are identical and used in the same context (just with a redesigned layout), maybe we should just use the first string.

Baron-Severin: Yeah, I was surprised as well, but it looks like the curved apostrophes don't need to be escaped.

Baron-Severin: The only difference is using curved apostrophes instead of normal ones.

Baron-Severin: To be clear, I did build to double check.

Baron-Severin: I can remove the old strings when I merge the layout changes, but until then we need them or compilation will break.

mcomella: @Baron-Severin I feel like delphine has previously said we shouldn't have curly apostrophes in strings, but I'm checking right now.

app/src/main/res/values/strings.xml
```diff
@@ -181,10 +181,13 @@ be temporary, and you can try again later.</li>
...
     <!-- Button text for keeping Turbo Mode enabled in onboarding -->
     <string name="button_turbo_mode_keep_enabled">Keep enabled</string>
+    <string name="button_turbo_mode_keep_enabled2">Keep Turbo Mode enabled</string>
     <!-- Button text for turning off Turbo Mode in onboarding -->
     <string name="button_turbo_mode_turn_off">Turn off</string>
+    <string name="button_turbo_mode_turn_off2">Turbo Turbo Mode off</string>
```
dnarcese: Should this say '_Turn_ Turbo Mode off'?

app/src/main/res/values/strings.xml
mcomella: Why are these strings merging independently of the code? In general, it's better to land strings with the code so localizers can see where their strings will be used.

mcomella: Sounds like it was the opposite: everything should be curly. lgtm.

app/src/main/res/values/strings.xml
Baron-Severin: Good catch

Baron-Severin:
Closed, adding the strings in along with the PR in which they're used.

PR #1232, opened 2018-09-29
Severin/i1192 session regression test

Baron-Severin:


app/src/androidTest/java/org/mozilla/focus/session/ClearSessionTest.kt
```diff
@@ -0,0 +1,68 @@
...
+        }
+
+        settings {
+            clearAllDataAndReturnHome()
+        }
+
+        home {
+            assertCannotGoBack()
+
+            assertCannotGoForward()
```
Baron-Severin: @mcomella suggested checking some other types of data to make sure that they are cleared as well, so I'll be adding some additional checks after the weekend.

AaronMT:
lgtm, thanks for this

app/src/androidTest/java/org/mozilla/focus/session/ClearSessionTest.kt
Baron-Severin: @AaronMT I wanted to add a test where I navigate to a page and verify that the cookies have been cleared, but I'm having trouble working out how to do that without having any direct references to the WebView.  Do you have any suggestions on how to accomplish this?

Baron-Severin: It looks like they're digging through the file system to make sure no trace of anything was saved, I don't think we need to be that extreme here.  I did find what looks to be a good solution though.  Do you think you could double check https://github.com/mozilla-mobile/firefox-tv/pull/1232/commits/2772dcf0a4be9627f4fca1e55fa11bc496d6edf3 for me?  Then I think we're good to merge.

npark-mozilla: FYI, that folder check test in Focus fails currently, it seems that there are web caches not being cleared. It's pretty extreme, but was useful.  But I agree that checking for cookies might be sufficient for Fire TV for now.

Baron-Severin: That's an interesting idea @npark-mozilla.  It's probably not a bad idea for clear data to work that way for all of our projects, I wonder if we should consider breaking part of that test out into a component.  The checks they were doing were nontrivial, it would be good to share the code.

npark-mozilla: https://github.com/npark-mozilla/focus-android/blob/master/app/src/androidTest/java/org/mozilla/focus/privacy/WebViewDataTest.java is one test, what it does is to load a page, delete it, and checks to see whether there are folders/files containing the texts from the web are not there.  But the list of folders may be quite specific, different in case of Fire TV.

Baron-Severin: That test has comments that mention it doesn't work with GV, which we're currently working on integrating.  I think we should move forward with what we have for now, but consider expanding this test in the future.  Does that sound reasonable to you, @npark-mozilla?

In the meantime, I'm pinging the GV team about the statuses of this and other related testing problems.

npark-mozilla: yes sounds good

Issue #1249, opened 2018-10-03
Unpinning via menu long press does not update pin icon


Baron-Severin:
## Steps to reproduce
- From home screen
- Navigate to pinned tile
- Open menu
- Long press the same tile as above
- Select remove

### Expected behavior
- Pin button icon should update immediately

### Actual behavior
- Pin button icon does not update until either of the following:
  - Menu is closed and opened
  - URL bar gains then loses focus

### Device information
* Fire TV device: Emulator
* Firefox version: 043631c5b61e08b668d1cbe386331fbeef7aab7a

![update-unpin](https://user-images.githubusercontent.com/13170306/46445719-27346100-c72d-11e8-8c52-f299e134091a.gif)

dnarcese:
Dupe of #1210 ?

Baron-Severin:
Closed by https://github.com/mozilla-mobile/firefox-tv/pull/1250

Issue #1085, opened 2018-09-11
BrowserNavigationOverlay to maintain UI state


psymoon:
### Why/User Benefit/User Problem
- Currently UI updates and manipulation are being handled case by case
- UI updates are also being handled outside its apparent scope (i.e BrowserFragment)
- This results issues when a specific case is not being handled/missed
- Integration takes longer because you have to search through the codebase to make sure there is a relevant case that is being handled

### What / Requirements
- Separating concerns from UI updating to maintaining
- Possible solution include use of livedata, RxJava and/or kotlin delegation

### Acceptance Criteria (how do I know when I’m done?)
- UI state test cases

bbinto:
- Meeting with @aminalhazwani and eng, QA team to define UX states
- Meeting with engineers to define new architecture

athomasmoz:
@psymoon to set up meeting to discuss next steps for this.

Baron-Severin:
Whatever we wind up moving towards, this is a really helpful pattern for incrementally updating architecture that we should consider using.
https://www.martinfowler.com/bliki/StranglerApplication.html

PR #1250, opened 2018-10-03
Closes #1178, #1249: make empty tile container unfocusable

### LINKED ISSUE #1178: Empty tileContainer is focusable
## Steps to reproduce
- Open firefox
- Remove all pins from tileContainer

### Expected behavior
- tileContainer is not focusable

### Actual behavior
- tileContainer is focusable

Baron-Severin:
- Also fixes 1249, which at times led to the unpin icon from being out of sync with the state

### DIFF
--- a/app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt
+++ b/app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt
@@ -335,6 +335,11 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
             else -> R.id.turboButton
         }
 
+        pocketVideoMegaTileView.nextFocusDownId = when {
+            tileContainer.adapter.itemCount == 0 -> R.id.pocketVideoMegaTileView
+            else -> R.id.tileContainer
+        }
+
         // We may have lost focus when disabling the focused view above.
         val isFocusLost = focusedView != null && findFocus() == null
         if (isFocusLost) {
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt
```diff
@@ -406,6 +411,7 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
                 megaTileTryAgainButton.requestFocus()
             }
         }
+        updateOverlayForCurrentState()
```
psymoon: Think we need to call this for `refreshTilesForInsertion()` as well, for when inserting a pin into empty container

Baron-Severin: Is there any way to pin a tile aside from clicking the icon itself?  The icon is already updated when we do that.

psymoon: Interesting, so we call `updateOverlyCurrentState()` in `navUrlInput.onFocusChangeListener {}`. So you're right, technically we are updating `pocketVideoMegaTileView.nextFocusDownId` when we navigate down from pin button focus. Seems very fragile though 🤔

Baron-Severin: Oh, I thought you were talking about #1249 , not #1178.  Right, good point.  I'll add it in.

Baron-Severin: This is made a little complicated by how many ways there are to add and remove tiles.

psymoon: Hopefully gets cleaned up with https://github.com/mozilla-mobile/firefox-tv/issues/1085

PR #1251, opened 2018-10-04
Closes i1224: Fixes bug when backing from fullscreen

Baron-Severin:
This is blocked until the next A-C release.  `SystemEngineSession#exitFullScreenMode` is currently a no-op.  Otherwise, the code is complete.

### DIFF
--- a/app/src/main/java/org/mozilla/focus/browser/BrowserFragment.kt
+++ b/app/src/main/java/org/mozilla/focus/browser/BrowserFragment.kt
@@ -53,6 +53,7 @@ private const val TOAST_Y_OFFSET = 200
 /**
  * Fragment for displaying the browser UI.
  */
+@Suppress("LargeClass") // TODO this will be removed as part of the upcoming architecture refactor
 class BrowserFragment : EngineViewLifecycleFragment(), Session.Observer {
     companion object {
         const val FRAGMENT_TAG = "browser"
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/mozilla/focus/browser/BrowserFragment.kt
```diff
@@ -278,14 +278,21 @@ class BrowserFragment : EngineViewLifecycleFragment(), Session.Observer {
         activity?.menuInflater?.inflate(R.menu.menu_context_hometile, menu)
     }
 
+    private fun exitFullScreenAndBack() {
+        // Backing while full-screened can lead to unstable behavior (see #1224),
+        // so we always attempt to exit full-screen before backing
+        requireComponents.sessionManager.getEngineSession()?.exitFullScreenMode()
+        if (session.canGoBack) requireComponents.sessionUseCases.goBack.invoke()
```
psymoon: Interesting. So if the user perform `NavigationEvent.BACK` but nothing in the back stack, it will always exitFullScreenMode(). But not for `onBackPressed()`?

Baron-Severin: When we hit the `browserOverlay.isVisible && !isUrlEqualToHomepage` branch of `onBackPressed`, we close the overlay and go back to the content without changing the URL.  I think in this case we should probably keep the video full-screened.  I'm not super happy about how many similar code paths we have here though.

PR #1252, opened 2018-10-04
Fixes #1159: catches telemetry lifecycle crash

### LINKED ISSUE #1159: Fatal exception on out of order Telemetry events
See https://sentry.prod.mozaws.net/operations/firefox-tv/issues/4270217/

We are seeing ~250 crashes per day on v2.1 due to `TelemetryWrapper#stopSession` being called before `TelemetryWrapper#startSession`.  This implies that, on some small number of device/OS combinations, `MainActivity#onPause` is being called before `MainActivity.onResume`.

Baron-Severin:
Blocked until we integrate the next A-C release.  Otherwise finished.

### DIFF
--- a/app/src/main/java/org/mozilla/focus/telemetry/SentryIntegration.kt
+++ b/app/src/main/java/org/mozilla/focus/telemetry/SentryIntegration.kt
@@ -51,36 +51,7 @@ object SentryIntegration {
         Sentry.getContext().user = SentryUnknownUser()
     }
 
-    /**
-     * One way to test calls to static methods is to pass instances of their objects into the
-     * class being tested. This allows tests to pass spies or mocks instead.
-     *
-     * Example:
-     *
-     *  class SimpleClass(private val singleton: Singleton = Singleton) {
-     *      fun testThis() {
-     *          singleton.callMethod()
-     *      }
-     *  }
-     *
-     *  class SimpleClassTest {
-     *
-     *      @Test
-     *      fun testSimpleClass() {
-     *          val spy = spy(Singleton)
-     *          val simpleClass = SimpleClass(spy)
-     *          simpleClass.testThis()
-     *          verify(spy).callMethod()
-     *      }
-     *  }
-     *
-     * [Sentry] has a private constructor, disallowing this pattern, so instead we delegate
-     * [SentryIntegration.capture] to [Sentry.capture]. By injecting [SentryIntegration] into
-     * classes that we test, we can now verify that this method is being called in our tests.
-     *
-     * See [TelemetryIntegrationTest.setup] for an example.
-     */
-    internal fun capture(exception: Exception) {
+    fun capture(exception: Exception) {
         Sentry.capture(exception)
     }
 }
--- a/app/src/main/java/org/mozilla/focus/telemetry/SentryIntegration.kt
+++ b/app/src/main/java/org/mozilla/focus/telemetry/SentryIntegration.kt
@@ -27,7 +27,7 @@ import org.mozilla.focus.BuildConfig
  * As opposed to:
  * "Sentry DSN (amazonWebviewRelease): X_X"
  */
-object SentryWrapper {
+object SentryIntegration {
 
     fun init(context: Context) {
         onIsEnabledChanged(context, DataUploadPreference.isEnabled(context))
--- a/app/src/main/java/org/mozilla/focus/telemetry/SentryIntegration.kt
+++ b/app/src/main/java/org/mozilla/focus/telemetry/SentryIntegration.kt
@@ -50,6 +50,39 @@ object SentryIntegration {
         // they're personally identifiable information.
         Sentry.getContext().user = SentryUnknownUser()
     }
+
+    /**
+     * One way to test calls to static methods is to pass instances of their objects into the
+     * class being tested. This allows tests to pass spies or mocks instead.
+     *
+     * Example:
+     *
+     *  class SimpleClass(private val singleton: Singleton = Singleton) {
+     *      fun testThis() {
+     *          singleton.callMethod()
+     *      }
+     *  }
+     *
+     *  class SimpleClassTest {
+     *
+     *      @Test
+     *      fun testSimpleClass() {
+     *          val spy = spy(Singleton)
+     *          val simpleClass = SimpleClass(spy)
+     *          simpleClass.testThis()
+     *          verify(spy).callMethod()
+     *      }
+     *  }
+     *
+     * [Sentry] has a private constructor, disallowing this pattern, so instead we delegate
+     * [SentryIntegration.capture] to [Sentry.capture]. By injecting [SentryIntegration] into
+     * classes that we test, we can now verify that this method is being called in our tests.
+     *
+     * See [TelemetryIntegrationTest.setup] for an example.
+     */
+    internal fun capture(exception: Exception) {
+        Sentry.capture(exception)
+    }
 }
 
 /**
--- a/app/src/main/java/org/mozilla/focus/telemetry/TelemetryIntegration.kt
+++ b/app/src/main/java/org/mozilla/focus/telemetry/TelemetryIntegration.kt
@@ -32,7 +32,7 @@ private const val KEY_CLICKED_HOME_TILE_IDS_PER_SESSION = "clickedHomeTileIDsPer
         "TooManyFunctions",
         "LargeClass"
 )
-object TelemetryWrapper {
+object TelemetryIntegration {
 
     private object Category {
         val ACTION = "action"
@@ -161,9 +161,9 @@ object TelemetryWrapper {
     @JvmStatic
     fun urlBarEvent(isUrl: Boolean, autocompleteResult: AutocompleteResult, inputLocation: UrlTextInputLocation) {
         if (isUrl) {
-            TelemetryWrapper.browseEvent(autocompleteResult, inputLocation)
+            TelemetryIntegration.browseEvent(autocompleteResult, inputLocation)
         } else {
-            TelemetryWrapper.searchEnterEvent(inputLocation)
+            TelemetryIntegration.searchEnterEvent(inputLocation)
         }
     }
 
@@ -353,7 +353,7 @@ private object TelemetryHomeTileUniqueClickPerSessionCounter {
 
         val uniqueClickCount = getSharedPrefs(context)
                 .getStringSet(KEY_CLICKED_HOME_TILE_IDS_PER_SESSION, null)?.size ?: 0
-        TelemetryWrapper.homeTileUniqueClickCountPerSessionEvent(uniqueClickCount)
+        TelemetryIntegration.homeTileUniqueClickCountPerSessionEvent(uniqueClickCount)
     }
 
     fun resetSessionData(context: Context) {
--- a/app/src/main/java/org/mozilla/focus/telemetry/TelemetryIntegration.kt
+++ b/app/src/main/java/org/mozilla/focus/telemetry/TelemetryIntegration.kt
@@ -34,7 +34,7 @@ private const val KEY_CLICKED_HOME_TILE_IDS_PER_SESSION = "clickedHomeTileIDsPer
         "LargeClass"
 )
 open class TelemetryIntegration protected constructor(
-        private val sentryIntegration: SentryIntegration = SentryIntegration
+    private val sentryIntegration: SentryIntegration = SentryIntegration
 ) {
 
     companion object {
--- a/app/src/main/java/org/mozilla/focus/telemetry/TelemetryIntegration.kt
+++ b/app/src/main/java/org/mozilla/focus/telemetry/TelemetryIntegration.kt
@@ -32,7 +32,11 @@ private const val KEY_CLICKED_HOME_TILE_IDS_PER_SESSION = "clickedHomeTileIDsPer
         "TooManyFunctions",
         "LargeClass"
 )
-object TelemetryIntegration {
+open class TelemetryIntegration protected constructor() {
+
+    companion object {
+        val INSTANCE: TelemetryIntegration by lazy { TelemetryIntegration() }
+    }
 
     private object Category {
         val ACTION = "action"
@@ -98,7 +102,6 @@ object TelemetryIntegration {
     // Available on any thread: we synchronize.
     private val pocketUniqueClickedVideoIDs = Collections.synchronizedSet(mutableSetOf<Int>())
 
-    @JvmStatic
     fun init(context: Context) {
         // When initializing the telemetry library it will make sure that all directories exist and
         // are readable/writable.
@@ -150,7 +153,6 @@ object TelemetryIntegration {
         pocketUniqueClickedVideoIDs.clear()
     }
 
-    @JvmStatic
     fun stopMainActivity() {
         TelemetryHolder.get()
                 .queuePing(TelemetryCorePingBuilder.TYPE)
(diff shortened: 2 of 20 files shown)


app/src/main/java/org/mozilla/focus/telemetry/TelemetryWrapper.kt
```diff
@@ -162,7 +122,15 @@ object TelemetryWrapper {
 
     @UiThread // via TelemetryHomeTileUniqueClickPerSessionCounter
     fun stopSession(context: Context) {
-        TelemetryHolder.get().recordSessionEnd()
+        fun sendEndSessionFailedEvent() {
+            // In rare instances, we have seen stopSession called before startSession.
+            // It is unknown at this time how this happens, but it is very infrequent
+            // so we are logging the event and otherwise dropping the error
+            // See https://bugzilla.mozilla.org/show_bug.cgi?id=1405192
+            TelemetryEvent.create(Category.ERROR, Method.ONPAUSE_CALLED_BEFORE_ONRESUME, Object.APP).queue()
```
mcomella: Alternatively, I think we can record events in Sentry that include a stack trace.

app/src/test/java/org/mozilla/focus/telemetry/TelemetryWrapperTest.kt
```diff
@@ -0,0 +1,43 @@
...
+
+    @Before
+    fun setup() {
+        val telemetry = TelemetryFactory.createTelemetry(RuntimeEnvironment.application)
+        telemetrySpy = spy(telemetry)
+        TelemetryHolder.set(telemetrySpy)
+    }
+
+    @Test
+    fun `WHEN session events are called on TelemetryWrapper THEN associated Telemetry methods should be called`() {
```
mcomella: nit: which session events? Which Telemetry methods are expected to be called?

app/src/test/java/org/mozilla/focus/telemetry/TelemetryWrapperTest.kt
```diff
@@ -0,0 +1,43 @@
...
+        TelemetryWrapper.stopSession(RuntimeEnvironment.application)
+        verify(telemetrySpy, times(1)).recordSessionStart()
+        verify(telemetrySpy, times(1)).recordSessionEnd(any())
+    }
+
+    /**
+     * See comment in [TelemetryWrapper.stopSession]
+     */
+    @Test
+    fun `WHEN TelemetryWrapper is called out of order THEN we should not crash`() {
```
mcomella: Should we additionally verify:
- when stopSession is called multiple times in a row (e.g. in theory, could happen in real world).
- That we create a telemetry event when this happens?

mcomella:
@Baron-Severin Please fill out [a data request review form](https://github.com/mozilla/data-review/blob/master/request.md) and update the telemetry documentation!

Baron-Severin:
@liuche 

* How frequently does our session start / session end code fail to run properly?

* We need to swallow these exceptions to prevent a crash, but should be able to keep an eye on how often this is failing.

* There is no other way to see how frequently this code fails.

* No existing telemetry exists to serve this need.

| Event                                    | category | method     | object                   | value |
|------------------------------------------|----------|------------|-----------|---------------|
| onPause called before onResume               | error   | pause_before_resume       | app         |  |


* Permanently (Barbara Bermes)

* All users

* Users may opt out of data collection by turning off data collection from the settings menu.

* If we see this number increase significantly, we will have to rework our session code to allow for this framework lifecycle bug.

* On re-dash and with mobile teams.

app/src/main/java/org/mozilla/focus/telemetry/TelemetryWrapper.kt
Baron-Severin: @liuche do you know how we can do that?

app/src/test/java/org/mozilla/focus/telemetry/TelemetryWrapperTest.kt
Baron-Severin: Hm.  I can see how startSession could theoretically be called twice in a row (and I'll add that case), but I'm not sure how stopSession could be.

I'll check that telemetry events are being sent in this case too.

app/src/main/java/org/mozilla/focus/telemetry/TelemetryWrapper.kt
mcomella: You'll probably have to look in the Sentry docs and potentially try out some methods to see which just send String text and which might send stacktraces.

app/src/main/java/org/mozilla/focus/telemetry/TelemetryWrapper.kt
```diff
@@ -113,43 +104,12 @@ object TelemetryWrapper {
...
-            val serializer = JSONPingSerializer()
-            val storage = FileTelemetryStorage(configuration, serializer)
-            val client = HttpURLConnectionTelemetryClient()
-            val scheduler = JobSchedulerTelemetryScheduler()
-
-            TelemetryHolder.set(Telemetry(configuration, storage, client, scheduler)
-                    .addPingBuilder(TelemetryCorePingBuilder(configuration))
-                    .addPingBuilder(TelemetryMobileEventPingBuilder(configuration))
-                    .setDefaultSearchProvider(createDefaultSearchProvider(context)))
+            TelemetryHolder.set(TelemetryFactory.createTelemetry(context))
```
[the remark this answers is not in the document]
Baron-Severin: > typealias TelemetryEntryPoint = Telemetry and make the factory based on that. EntryPoint isn't a great name so maybe something better.

I think having global typealiases seems like it could get confusing.  I would rather try to open a PR against A-C that changes the class name itself.
> TelemetryHolder -> TelemetryProvider (more conventional)

Unfortunately this is part of A-C.
> TelemetryWrapper -> TelemetryIntegration (to better distinguish from Holder/Provider)

Agreed, and done.

> could also try putting TelemetryWrapper behind a module so the only thing exposed to the app is TelemetryIntegration, simplifying this problem.

Although I haven't needed to yet, keeping TelemetryFactory available and using it in our Application to assemble TelemetryIntegration allows us to substitute in different values for testing.  I think that's worth preserving.

liuche:
unassigning myself from review now that this bug doesn't require data review and is using Sentry to send a stacktrace instead.

app/src/test/java/org/mozilla/focus/utils/AnyNonNull.kt
```diff
@@ -0,0 +1,22 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.focus.utils
+
+import org.mockito.Mockito
+
+/**
+ * A normal [Mockito.any] is a nullable type, so this allows us to test nonnull
```
mcomella: You should mention that the reason this is required is because it crashes otherwise, i.e. "use this when `any()` crashes".

app/src/test/java/org/mozilla/focus/telemetry/TelemetryIntegrationTest.kt
```diff
@@ -0,0 +1,72 @@
...
+import org.robolectric.RobolectricTestRunner
+import org.robolectric.RuntimeEnvironment
+import org.robolectric.annotation.Config
+
+class NoTelemetryApplication : FocusApplication() {
+    override fun initTelemetry() {}
+}
+
+@RunWith(RobolectricTestRunner::class)
+@Config(application = NoTelemetryApplication::class)
```
mcomella: I realized that robolectric could, but highly improbably, be uploading telemetry data: I filed https://github.com/mozilla-mobile/firefox-tv/issues/1300

Baron-Severin:
@mcomella I changed my approach considerably, the changes should be a lot less invasive now.  Ready for re-review.

app/src/main/java/org/mozilla/focus/telemetry/TelemetryIntegration.kt
```diff
@@ -97,16 +98,15 @@ object TelemetryWrapper {
 
     // Available on any thread: we synchronize.
     private val pocketUniqueClickedVideoIDs = Collections.synchronizedSet(mutableSetOf<Int>())
+    private lateinit var sentryWrapper: SentryWrapper
 
     @JvmStatic
-    fun init(context: Context) {
+    fun init(sentryWrapper: SentryWrapper, telemetry: Telemetry) {
+        this.sentryWrapper = sentryWrapper
         // When initializing the telemetry library it will make sure that all directories exist and
```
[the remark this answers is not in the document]
mcomella: I meant that `//` comments typically have a new line above them:
```kotlin
fun init(sentryWrapper: SentryWrapper, telemetry: Telemetry) {
        this.sentryWrapper = sentryWrapper

        // When initializing the telemetry library
```

GitHub comments give context above the line the author comments on so I was commenting specifically on the `//` comment line.

app/src/main/java/org/mozilla/focus/telemetry/TelemetryIntegration.kt
```diff
@@ -41,8 +33,13 @@ private const val KEY_CLICKED_HOME_TILE_IDS_PER_SESSION = "clickedHomeTileIDsPer
...
         "LargeClass"
 )
-object TelemetryWrapper {
-    private const val TELEMETRY_APP_NAME_FOCUS_TV = "FirefoxForFireTV"
+open class TelemetryIntegration protected constructor(
+    private val sentryIntegration: SentryIntegration = SentryIntegration
+) {
+
+    companion object {
+        val INSTANCE: TelemetryIntegration by lazy { TelemetryIntegration() }
```
mcomella: I don't love that we have to do this but it looks like the best of all of our options:
- (old solution) Top-level object with static `init()`: can't replace the instance under testing; strange mechanics when testing the object
- Give TelemetryIntegration instance to Application to own: need context to add telemetry

app/src/main/java/org/mozilla/focus/telemetry/SentryIntegration.kt
```diff
@@ -50,6 +50,39 @@ object SentryWrapper {
...
+     *      }
+     *  }
+     *
+     * [Sentry] has a private constructor, disallowing this pattern, so instead we delegate
+     * [SentryIntegration.capture] to [Sentry.capture]. By injecting [SentryIntegration] into
+     * classes that we test, we can now verify that this method is being called in our tests.
+     *
+     * See [TelemetryIntegrationTest.setup] for an example.
+     */
+    internal fun capture(exception: Exception) {
```
mcomella: nit: internal isn't doing anything here, right?

app/src/main/java/org/mozilla/focus/telemetry/SentryIntegration.kt
```diff
@@ -50,6 +50,39 @@ object SentryWrapper {
         // they're personally identifiable information.
         Sentry.getContext().user = SentryUnknownUser()
     }
+
+    /**
+     * One way to test calls to static methods is to pass instances of their objects into the
```
mcomella: See my other comment about the first line of kdoc comments.

In general, this comment seems unrelated to what the method does so is inappropriate for kdoc. In situations like these, prefer comments that are not exposed in API documentation, like `//`. (we generally don't use `/*` comments).

app/src/main/java/org/mozilla/focus/telemetry/SentryIntegration.kt
```diff
@@ -50,6 +50,39 @@ object SentryWrapper {
...
+     *      @Test
+     *      fun testSimpleClass() {
+     *          val spy = spy(Singleton)
+     *          val simpleClass = SimpleClass(spy)
+     *          simpleClass.testThis()
+     *          verify(spy).callMethod()
+     *      }
+     *  }
+     *
+     * [Sentry] has a private constructor, disallowing this pattern, so instead we delegate
```
mcomella: Another reason we force everyone to go through `SentryIntegration` (same for Telemetry) is modularity: if we need to swap out the implementation of the Integration class, consumers can make the same calls. No one in the app code should know about `Sentry`, just `SentryIntegration`.

Since this is the primary reason for having SentryIntegration, I don't think this comment is necessary.

I don't know where to comment about why we do these `Integration` classes like that – maybe a README in an `integrations` package?

mcomella:
that is, lgtm w/ nits. Thanks for improving our testing story too!

app/src/test/java/org/mozilla/focus/utils/AnyNonNull.kt
```diff
@@ -0,0 +1,21 @@
...
+/**
+ * Use this when [Mockito.any] crashes.
+ *
+ * A normal [Mockito.any] is a nullable type, so this allows us to test nonnull
+ * code.
+ *
+ * Taken from https://medium.com/@elye.project/befriending-kotlin-and-mockito-1c2e7b0ef791
+ */
+@Suppress("UNCHECKED_CAST")
+fun <T> anyNonNull(): T {
```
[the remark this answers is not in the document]
Baron-Severin: Unfortunately no.  This is weird, but take a look at `org.mockito.ArgumentMatchers#notNull`.  It produces a side effect that somehow results in the test passing, but actually returns `null`.  This causes Kotlin's null checks to throw an exception.

app/src/test/java/org/mozilla/focus/utils/AnyNonNull.kt
```diff
@@ -0,0 +1,21 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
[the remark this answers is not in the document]
Baron-Severin: Makes sense to me.

app/src/test/java/org/mozilla/focus/telemetry/TelemetryIntegrationTest.kt
```diff
@@ -0,0 +1,64 @@
...
+    private lateinit var telemetrySpy: Telemetry
+    private lateinit var sentrySpy: SentryIntegration
+
+    @Before
+    fun setup() {
+        val telemetry = TelemetryFactory.createTelemetry(RuntimeEnvironment.application)
+        telemetrySpy = spy(telemetry)
+        TelemetryHolder.set(telemetrySpy)
+        sentrySpy = spy(SentryIntegration)
+        telemetryIntegration = TestTelemetryIntegration(sentrySpy)
```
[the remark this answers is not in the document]
Baron-Severin: `TelemetryIntegration` has a protected constructor, so the rest of the app isn't able to pass `Sentry` values to it.  If we relaxed that visibility, then we wouldn't need the subclass.  I don't strongly prefer either method, what would your preference be?

app/src/main/java/org/mozilla/focus/telemetry/SentryIntegration.kt
Baron-Severin: Hm.  I'm not sure if we need to include this in the project at all, now that you mention it.  What do you think about adding a page in our shared documentation repo?  I could take the time to write one.

mcomella: Good idea. Maybe we should link to it from each of the `Integration` pages, just so it's not overlooked? Like:
```
/* License
 */

// See <url> for dependency integration best practices.

class TelemetryIntegration {

app/src/test/java/org/mozilla/focus/telemetry/TelemetryIntegrationTest.kt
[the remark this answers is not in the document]
mcomella: Ah, I didn't notice that. `protected` constructor sounds better to me. No need to add comments on the `Test*` then: I think it's a known pattern.

app/src/main/java/org/mozilla/focus/telemetry/SentryIntegration.kt
Baron-Severin: I've made a note to myself to write it, and I'll open a new PR adding the link once it's done.

PR #1254, opened 2018-10-05
Closes #1253: updates A-C to v0.26

### LINKED ISSUE #1253: Update to A-C v0.26
- Necessary to fix #1159 

### What / Requirements
- Upgrade Android Components to v0.26
- No functional changes

Baron-Severin:
- Updates to new A-C repository
- Changes to new A-C module names
- Removes custom error code logic in favor of A-C equivelent

### DIFF
--- a/app/src/main/java/org/mozilla/focus/browser/ErrorPage.java
+++ b/app/src/main/java/org/mozilla/focus/browser/ErrorPage.java
@@ -6,91 +6,21 @@ package org.mozilla.focus.browser;
 
 import android.content.Context;
 import android.content.res.Resources;
+import android.support.annotation.NonNull;
 import android.support.v4.util.ArrayMap;
-import android.support.v4.util.Pair;
 
-import android.util.SparseArray;
-import android.webkit.WebViewClient;
 import org.mozilla.focus.R;
 import org.mozilla.focus.utils.HtmlLoader;
 
 import java.util.Map;
 
+import mozilla.components.browser.errorpages.ErrorType;
+
 public class ErrorPage {
 
-    private static final SparseArray<Pair<Integer, Integer>> errorDescriptionMap;
-
-    static {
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/mozilla/focus/browser/ErrorPage.java
```diff
@@ -6,91 +6,21 @@
...
-    private static final SparseArray<Pair<Integer, Integer>> errorDescriptionMap;
-
-    static {
-        errorDescriptionMap = new SparseArray<>();
-
-        // Chromium's mapping (internal error code, to Android WebView error code) is described at:
-        // https://chromium.googlesource.com/chromium/src.git/+/master/android_webview/java/src/org/chromium/android_webview/ErrorCodeConversionHelper.java
-
-        errorDescriptionMap.put(WebViewClient.ERROR_UNKNOWN,
-                new Pair<>(R.string.error_connectionfailure_title, R.string.error_connectionfailure_message));
```
mcomella: Should we delete all these strings too? I assume they're unused.

pocmo: Note that components do not contain any translations yet. We are waiting for the new L10N tooling to. be ready. You may want to keep the strings to keep the translations for now - I think?

mcomella: So does that mean we can't land this because the current strings are being pulled from components, not strings.xml?

pocmo: You can land this. We kept the string names the same so that the strings from Focus / FFFFTV will override them.

pocmo: But I assume that if you remove the Strings here then the next string export/import will remove all the translations!

PR #1293, opened 2018-10-12
Fix blank screen on re-open

Baron-Severin:


app/src/main/java/org/mozilla/focus/ext/EngineView.kt
```diff
@@ -109,6 +110,16 @@ fun EngineView.scrollByClamped(vx: Int, vy: Int) {
 val EngineView.focusedDOMElement: FocusedDOMElementCache
     get() = getOrPutExtension(this).domElementCache
 
+fun EngineView.saveState(): Bundle {
+    val bundle = Bundle()
+    getOrPutExtension(this).webView.saveState(bundle)
```
liuche: Looks like from the docs that this saves the back/forward state, but not the display state.

app/src/main/java/org/mozilla/focus/browser/WebViewCache.kt
```diff
@@ -39,6 +46,7 @@ class WebViewCache : LifecycleObserver {
         fun createAndCacheEngineView(): SystemEngineView {
             return SystemEngineView(context, attrs).apply {
                 initialize()
+                state?.also { this.restoreState(it) }
```
liuche: Why `also` if you're not using the return value of state? If you just want to make sure state is not null, I think `let` will do the trick.

liuche: Looking at the docs, it also says that `restoreState` should be called before any back/forward, page loading happens - I assume we don't do any of that in initialize, so what do you think about calling `restoreState` before `initialize` in case that changes?

https://developer.android.com/reference/android/webkit/WebView.html#restoreState(android.os.Bundle)

app/src/main/java/org/mozilla/focus/ext/EngineView.kt
Baron-Severin: Yeah, I was hesitant about that too.  We had a big talk about it on slack (https://mozilla.slack.com/archives/C8P1Z6JSZ/p1539383487000100).  I'll add a comment about it here.

app/src/main/java/org/mozilla/focus/browser/WebViewCache.kt
Baron-Severin: Semantically speaking, I see `also` as being used for side effects and `let` for transformations.  Definitely either work here, and I'm willing to change it if you feel strongly about it.  Just let me know.

mcomella: fwiw, I see `let` as a short-hand for: `if (item != null) { /* do something with item */ }`

I don't have a strong understanding of when `also` should be used but I think the name, "also" is indicative that it should follow some other action. `state?.also { ... }` doesn't do some other action first so it seems weird to me.

---

My very weak understanding of `also` was two cases. It's like `apply` in that it handles side effects but does so when the receiver is not the object you're calling the function on. This is especially helpful for immutable values, e.g.:
```kotlin
val immutableApple = Apple().also {
    webViewTarget = it
    println("apple is confirmed")
}
```

Another use case is chaining operations (especially during debugging):
```kotlin
val builderResult = SomeBuilder().apply {
    food = "bars"
    foo = bar
}.also { println(it.hash) }
.build()
.also { println(it.hash) }
```

mcomella: My experience from clojure teaches me that "let" is "let this value be known as this identifier": `(let [four (+ 1 3)] ...)`. I assume functional programming languages use the functions similarly but that's my only real functional exposure to let so far.

As such, in kotlin, I consider, `let` to be similar: `obj.let { newIdentifierForObj -> ... }`. And similar, "if obj is non-null, let it be known as...": `obj?.let { nonNullNewIdentifier -> ... }`.

`let` in kotlin also provides an accompanying grouping effect: in a new scope, you can execute multiple statements in a row and return a single result. This lets you encapsulate related functionality and variables without letting them leak. It's like inlining a function by hand, I guess?

This doesn't limit it semantically to transformations or non-side effects (though, from a functional heritage I guess it implies no side effects).

I dunno. :P I don't like that there isn't a prescriptive way to use these so we constantly have these discussions, unlike map, filter, fold. :\ As such, I tend to favor not using them if I'm in doubt on the semantics (e.g. I rarely use `also`).

Baron-Severin: I know I see `let` used that way a lot, but I've started to wonder if it's entirely appropriate.  It does seem to me that Kotlin best practices are still a little bit in flux, although I'm far from certain that I'm correct in this case.

I know in Haskell and Elm, `let` is used to define local variables to make expressions easier to read (maybe @csadilek can correct any gross inaccuracies here?).  To me that would imply that let should be used to define a value, and that its return value should always be captured.
```haskell
-- The most advanced Haskell code @Baron-Severin knows how to write
value =
    let
        x = 5
        y = 6
    in
        x + y
```

PR #1295, opened 2018-10-13
Closes #1280: Add basic validations for release tasks

### LINKED ISSUE #1280: Create release build validation task (to prevent releasing the same version twice)
Sometimes for pre-release versions, we mishandle the version numbers and release two builds with the same ID or forget a git tag. This is also a threat for release builds.

Not having a clear idea of which release is which can convolute our data (did this report come from 1.1 or the second 1.1?), making it harder to solve reproduce issues and solve problems for our users when that is dependent on version.

At compile-time, we should ensure:
- Bonus-ish: The user has no local changes
- The user's checked out git commit is tagged with the same version as the build name

And give users an escape hatch to build local builds without validation. Since git tags are unique, we can never release two of the same version number.

Part of this will be solved by https://github.com/mozilla-mobile/firefox-tv/issues/1173 which will create the builds on CI from a git tag, but without this fix, we may still forget to bump the version (unless we can get CI to do that for us and that's hard).

### What / Requirements
- Create gradle task to validate release builds
  - Does checkout out git commit's tag == buildName?
  - Bonus: Confirm no changes in working directory
- assembleRelease.dependsOn(validationTask)
- A way for users to disable this validation, e.g. `./gradlew assembleRelease -Pno-validation`

mcomella:


mcomella:
Travis failure:
```
Execution failed for task ':app:validateReleaseConfiguration'.
> Expected top-most commit to have a git tag
```

Taskcluster:
```
Execution failed for task ':app:validateReleaseConfiguration'.
> Unable to build release builds with uncommitted changes
```

Travis is building off of master so this makes sense. Some options:
- Don't validate the build on CI. This seems *okay* but we miss out on:
  - If the final release build gets built by CI, it won't catch buildVersion != git tag version
  - If the CI is making changes to the source before we build (accidentally or not), we won't catch that
- Run a different set of validations on CI, or for particular CI modes (e.g. for PRs and master branch, don't check git tags but do for release builds if generated on CI)
- Don't do these checks with `assembleRelease`: create a new task to do them; it's likely someone will build a release build without this task then (unless we disable the standard tasks...)

I think I like option 2 (`./gradlew assembleRelease -Pcontinuous-integration` ?). It's weird that TaskCluster has uncommitted changes: we should investigate that. Also, there is probably no need to build release builds on travis either.

mcomella:
I see `echo \"--\" > .adjust_token ` in the taskcluster config: since we don't gitignore this anymore, it could be causing problems.

mcomella:
With this next push, I still expect TaskCluster to fail: I haven't implemented one of the proposed solutions at the top yet.

buildSrc/src/main/java/org/mozilla/gradle/Git.kt
```diff
@@ -0,0 +1,37 @@
...
+
+/** Raw Git commands. */
+object Git {
+    private val runtime: Runtime
+        get() = Runtime.getRuntime()
+
+    fun status(): String = runtime.execWaitForStdOut("git status --porcelain")
+
+    fun nameRev(vararg args: String): String =
+            Runtime.getRuntime().execWaitForStdOut("git name-rev ${args.joinToString(" ")}")
```
Baron-Severin: Nit: This can use the above val.  Suggest replacing with:
```
runtime.execWaitForStdOut("git name-rev ${args.joinToString(" ")}")
```

buildSrc/src/main/java/org/mozilla/gradle/tasks/ValidateAndroidAppReleaseConfiguration.kt
```diff
@@ -0,0 +1,83 @@
...
+ * - The buildVersion in gradle matches the checked out git tag
+ *
+ * This task can be skipped by supplying the `-Pno-validate` argument.
+ */
+open class ValidateAndroidAppReleaseConfiguration : DefaultTask() {
+    init {
+        group = "Verification"
+        description = "Validates an Android app is correctly configured for release"
+
+        this.onlyIf { !project.hasProperty("no-validate") }
```
Baron-Severin: Optional nit: I think the following would be more clear.  Might be worth a simple extension function.
```
project.doesntHaveProperty("no-validate")
```

buildSrc/src/main/java/org/mozilla/gradle/tasks/ValidateAndroidAppReleaseConfiguration.kt
```diff
@@ -0,0 +1,83 @@
...
+
+        project.gradle.projectsEvaluated {
+            getAssembleReleaseTasks().forEach {
+                it.dependsOn(this@ValidateAndroidAppReleaseConfiguration)
+            }
+
+            // The assemble release tasks run late during execution so we make
+            getCompileReleaseTasks().forEach {
```
Baron-Severin: I don't understand this comment, what is happening here?

buildSrc/src/main/java/org/mozilla/gradle/tasks/ValidateAndroidAppReleaseConfiguration.kt
```diff
@@ -0,0 +1,83 @@
...
+open class ValidateAndroidAppReleaseConfiguration : DefaultTask() {
+    init {
+        group = "Verification"
+        description = "Validates an Android app is correctly configured for release"
+
+        this.onlyIf { !project.hasProperty("no-validate") }
+
+        project.gradle.projectsEvaluated {
+            getAssembleReleaseTasks().forEach {
+                it.dependsOn(this@ValidateAndroidAppReleaseConfiguration)
```
Baron-Severin: Hooking into build order seems unrelated to this class' main responsibility (validating git and version line up).  Suggest breaking these out of this class and putting them into extension functions.  They seem fairly likely to be reused in other contexts.
```
project.assembleReleaseDependsOn(this)
project.compileReleaseMustRunAfter(this)
```

buildSrc/src/main/java/org/mozilla/gradle/tasks/ValidateAndroidAppReleaseConfiguration.kt
mcomella: It got cut off:
```
// The assemble release tasks run late during execution but the compile tasks run
// early and assemble depends on them so we make our new task run before these compile tasks.
```

buildSrc/src/main/java/org/mozilla/gradle/tasks/ValidateAndroidAppReleaseConfiguration.kt
mcomella: I'm not a fan of having a methods that wrap another method with a boolean state inversion for slightly more english readability because it bloats the number of method calls you can make on an object (reducing effectiveness of autocomplete) and requires additional maintenance overhead by violating the DRY principle (e.g. if a method is changed to support additional arguments).

If anything, I think this is on the language to make boolean inversion more readable (e.g. python's `if not condition:`). I guess you could make a function in kotlin such that, `if (not(condition())) {`...

As such, I'm going to keep it as is.

buildSrc/src/main/java/org/mozilla/gradle/tasks/ValidateAndroidAppReleaseConfiguration.kt
mcomella: I don't think hooking into build order is out of scope – a task is its execution and non-parameterized configuration –  but I think it's out of scope to iterate through the tasks to determine what an "assembleRelease" task is. It also pains me how overly specific a method like, `project.assembleReleaseDependsOn(this)` is. As such, I took the middle ground and changed the code to this:
```kotlin
        project.gradle.projectsEvaluated { project.tasks.let { tasks ->
            // The assemble release tasks run late during execution but the compile tasks run
            // early and assemble depends on them so we make our new task run before these compile tasks.
            tasks.getAssembleReleaseTasks().forEach { it.dependsOn(this@ValidateAndroidAppReleaseConfiguration) }
            tasks.getCompileReleaseTasks().forEach { it.mustRunAfter(this@ValidateAndroidAppReleaseConfiguration) }
        } }

// TaskContainer.kt:
fun TaskContainer.getAssembleReleaseTasks(): List<Task> = filter { task -> task.name.let {
    it.startsWith("assemble") && it.endsWith("Release")
} }

fun TaskContainer.getCompileReleaseTasks(): List<Task> = filter { task -> task.name.let {
    it.startsWith("compile") && it.endsWith("ReleaseSources")
} }
```

mcomella: Thanks for always thinking about "whose responsibility it is to do this"!

PR #1306, opened 2018-10-16
Severin/1283 en only pocket

Baron-Severin:
### Pull Request checklist

- isEffectivelyVisible is tested
- I have not yet found a good way to test focus behavior

### DIFF
--- a/app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt
+++ b/app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt
@@ -41,6 +41,7 @@ import org.mozilla.focus.widget.IgnoreFocusMovementMethod
 import org.mozilla.focus.widget.InlineAutocompleteEditText
 import kotlin.properties.Delegates
 import org.mozilla.focus.ext.isVoiceViewEnabled
+import org.mozilla.focus.locale.LocaleManager
 import java.lang.ref.WeakReference
 
 private const val NAVIGATION_BUTTON_ENABLED_ALPHA = 1.0f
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/mozilla/focus/locale/LocaleManager.java
```diff
@@ -283,6 +283,16 @@ public Locale getCurrentLocale(@NonNull Context context) {
         return currentLocale = Locales.parseLocaleCode(current);
     }
 
+    public boolean currentLanguageIsEnglish(@NonNull Context context) {
+        Locale current = getCurrentLocale(context);
+        // If locale hasn't been updated (i.e., 'current' is null), use system default
+        if (current == null) {
+            current = context.getResources().getConfiguration().locale;
```
liuche: I'm not sure why this file has so many ways to get locales, but this seems reasonable to get the system locale (from looking at other methods here).

app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt
```diff
@@ -337,8 +338,10 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
         }
 
         navUrlInput.nextFocusDownId = when {
-            pocketVideosContainer.isVisible -> R.id.pocketVideoMegaTileView
-            else -> R.id.megaTileTryAgainButton
+            pocketVideosContainer.isEffectivelyVisible -> R.id.pocketVideoMegaTileView
+            megaTileTryAgainButton.isEffectivelyVisible -> R.id.megaTileTryAgainButton
+            tileContainer.adapter.itemCount == 0 -> R.id.navUrlInput
+            else -> R.id.tileContainer
```
liuche: What does focusing the tileContainer do? Does it pass correctly to its descendants?

app/src/main/java/org/mozilla/focus/locale/LocaleManager.java
Baron-Severin: There's a pretty crazy comment on line 89 above a line using this, apparently other code in this class mutates some other ways of getting the locale.

app/src/main/java/org/mozilla/focus/browser/BrowserNavigationOverlay.kt
Baron-Severin: I was as surprised as you are, but it does.  This is borrowed from `pocketVideoMegaTileView.nextFocusDownId` a few lines down.

PR #1329, opened 2018-10-22
Add service locator

Baron-Severin:
### Pull Request checklist

Non-functional change. This supports future testing, though.

### DIFF
--- a/app/src/main/java/org/mozilla/focus/browser/BrowserFragment.kt
+++ b/app/src/main/java/org/mozilla/focus/browser/BrowserFragment.kt
@@ -38,6 +38,7 @@ import org.mozilla.focus.ext.requireComponents
 import org.mozilla.focus.ext.toUri
 import org.mozilla.focus.ext.isYoutubeTV
 import org.mozilla.focus.ext.focusedDOMElement
+import org.mozilla.focus.ext.serviceLocator
 import org.mozilla.focus.session.NullSession
 import org.mozilla.focus.telemetry.MenuInteractionMonitor
 import org.mozilla.focus.telemetry.TelemetryIntegration
@@ -144,7 +145,12 @@ class BrowserFragment : EngineViewLifecycleFragment(), Session.Observer {
                 (activity as MainActivity).onNonTextInputUrlEntered(value!!)
                 setOverlayVisible(false)
             }
-            NavigationEvent.POCKET -> ScreenController.showPocketScreen(fragmentManager!!)
+            NavigationEvent.POCKET -> {
+                val (fragmentManager, activity) = fragmentManager to activity
+                if (fragmentManager != null && activity != null) {
+                    ScreenController.showPocketScreen(fragmentManager, activity.serviceLocator)
+                }
+            }
             NavigationEvent.PIN_ACTION -> {
                 this@BrowserFragment.session.url.let { url ->
                     when (value) {
--- a/app/src/main/java/org/mozilla/focus/browser/BrowserFragment.kt
+++ b/app/src/main/java/org/mozilla/focus/browser/BrowserFragment.kt
@@ -146,7 +146,7 @@ class BrowserFragment : EngineViewLifecycleFragment(), Session.Observer {
                 setOverlayVisible(false)
             }
             NavigationEvent.POCKET -> {
-                val (fragmentManager, activity) = fragmentManager to activity
+                val (fragmentManager, activity) = Pair(fragmentManager, activity)
                 if (fragmentManager != null && activity != null) {
                     ScreenController.showPocketScreen(fragmentManager, activity.serviceLocator)
                 }
--- a/app/src/main/java/org/mozilla/focus/ext/ServiceLocator.kt
+++ b/app/src/main/java/org/mozilla/focus/ext/ServiceLocator.kt
@@ -20,5 +20,5 @@ val Activity.serviceLocator: ServiceLocator
 val Context.serviceLocator: ServiceLocator
     get() = this.application.serviceLocator
 
-val Fragment.serviceLocator: ServiceLocator
-    get() = this.context!!.serviceLocator
+val Fragment.serviceLocator: ServiceLocator?
+    get() = this.context?.serviceLocator
--- a/app/src/main/java/org/mozilla/focus/ext/ServiceLocator.kt
+++ b/app/src/main/java/org/mozilla/focus/ext/ServiceLocator.kt
@@ -1,0 +1,24 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.focus.ext
+
+import android.app.Activity
+import android.app.Application
+import android.content.Context
+import android.support.v4.app.Fragment
+import org.mozilla.focus.FocusApplication
+import org.mozilla.focus.utils.ServiceLocator
+
+val Application.serviceLocator: ServiceLocator
+    get() = (this as FocusApplication).serviceLocator
+
+val Activity.serviceLocator: ServiceLocator
+    get() = this.application.serviceLocator
+
+val Context.serviceLocator: ServiceLocator
+    get() = this.application.serviceLocator
+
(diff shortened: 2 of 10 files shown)


app/src/main/java/org/mozilla/focus/ext/ServiceLocator.kt
```diff
@@ -0,0 +1,24 @@
...
+    get() = (this as FocusApplication).serviceLocator
+
+val Activity.serviceLocator: ServiceLocator
+    get() = this.application.serviceLocator
+
+val Context.serviceLocator: ServiceLocator
+    get() = this.application.serviceLocator
+
+val Fragment.serviceLocator: ServiceLocator
+    get() = this.context!!.serviceLocator
```
mcomella: I don't like that this ext fn obscures a `!!`

app/src/main/java/org/mozilla/focus/home/pocket/PocketVideoFragment.kt
```diff
@@ -53,7 +54,7 @@ class PocketVideoFragment : Fragment() {
         // being reconstructed, the user probably cares more about an up-to-date feed rather than
         // consistency with the home screen they saw a while ago.
         if (!this::deferredVideos.isInitialized) {
-            deferredVideos = Pocket.getRecommendedVideos()
+            deferredVideos = context!!.application.serviceLocator.pocket.getRecommendedVideos()
```
mcomella: nit: `context!!.serviceLocator`?

app/src/main/java/org/mozilla/focus/home/pocket/Pocket.kt
```diff
@@ -39,7 +39,7 @@ typealias PocketVideosDeferred = Deferred<List<PocketVideo>?>
  * We cache in memory only. We chose not to cache on disk because it adds complexity and the cache
  * is likely to be severely outdated between runs of the application anyway.
  */
-object Pocket {
+open class Pocket {
```
mcomella: This change seems unnecessary atm.

documentation/architecture/ADR1:dependency-injection.md
```diff
@@ -0,0 +1,76 @@
...
+
+We use a service locator pattern. This is similar to, but distinct from dependency injection.
+
+In dependency injection, client code declares that they use an object (either in their constructor, or a field), and it is provided from outside. The client code has no idea how this occurs. In the service locator pattern, client code has a hard dependency upon a service locator, which is used to fulfill dependencies. The above example can be rewritten as a service locator as follows:
+```kotlin
+class Repository(serviceLocator: ServiceLocator) {
+    val endpoint = serviceLocator.endpoint()
+}
+```
+
```
liuche: Could the below be put in a `Tradeoffs` section?
I'd mention lifecycle management here, as well as any other problems, and how we're choosing to accept those problems/mitigate them.

documentation/architecture/ADR1:dependency-injection.md
```diff
@@ -0,0 +1,76 @@
+## Context
```
mcomella: This should be in `docs/`

mcomella: nit: You should probably make the directory sort by default. Maybe: `adr-0001-dependency-injection.md`

I don't like the colon: it's an atypical character in file names and it may need special handling in some applications.

app/src/main/java/org/mozilla/focus/home/pocket/PocketVideoFragment.kt
liuche: Can this just be `context!!.serviceLocator`, because you've added an ext to `Context` already?

documentation/architecture/ADR1:dependency-injection.md
mcomella: nit: This needs a title; include the ADR number.

documentation/architecture/ADR1:dependency-injection.md
```diff
@@ -0,0 +1,76 @@
...
+##### What is DI?
```
mcomella: I think the formatting of this document is a little confusing if you're coming in with no context:
- it doesn't separate the "Context" from the "Decision" (it's missing the "decision" section entirely)
- It doesn't clearly identify the problems *our team* is having
- It doesn't clearly identify why the chosen solution better solves our problems than another solution
- I think it leaves out some potential future problems with ServiceLocator (e.g. when SL grows large, it's hard to maintain, no compile-time checks for dependency graph, etc.) and why we don't care about these problems

In general, I think the formatting doesn't make it clear what options there are and the trade-offs between them (e.g. DI vs. ServiceLocator talks trade-offs but then is following by "Why not DI framework?" and "Why not Singletons?" where DI leaks into two sections, the two "Why not" sections aren't comparing similar things, and Singletons wasn't mentioned as an option sooner).

I think you should start with the specific problems that *our team* is having which is why we implemented this solution and be clear to enumerate what our options are (e.g. in lists), e.g.:

> ## Context
> Our code is currently using singletons for blah blah. We're having problems testing our code because they make hard-coded to make network calls and are tightly coupled to their dependencies.
>
> The industry-standard way of solving this problem is the Dependency Injection (wikipedia-link) pattern. The DI pattern solves this issue by passing the dependencies into a class, wherein they can be mocked. There are no other standard solutions to this problem. (sample of with and without DI).
> 
> Currently, there are X standard ways to do the dependency injection pattern:
> - ServiceLocator
> - Dependency Injection framework
> - Application holds singletons
>
> (Briefly describe differences, trade-offs)
>
> ## Decision
> Given the problems we were having, we decided on X because these trade-offs are worth it and we don't care about trade-offs Y, and Z.

mcomella: nit: You use DI before you say what the acronym stands for

mcomella: This is missing the "Decision" section.

app/src/main/java/org/mozilla/focus/browser/BrowserFragment.kt
```diff
@@ -145,7 +146,12 @@ class BrowserFragment : EngineViewLifecycleFragment(), Session.Observer {
                 (activity as MainActivity).onNonTextInputUrlEntered(value!!)
                 setOverlayVisible(false)
             }
-            NavigationEvent.POCKET -> ScreenController.showPocketScreen(fragmentManager!!)
+            NavigationEvent.POCKET -> {
+                val (fragmentManager, activity) = fragmentManager to activity
```
liuche: Is it possible to just do this as a pair? I think the `to` is confusing.

documentation/architecture/ADR1:dependency-injection.md
```diff
@@ -0,0 +1,76 @@
...
+
+##### Why inject?
+
+Injection makes testing simpler. When testing, we don't want our repos to make normal network calls (these make tests flaky, as calls can fail or return unexpected data). So we provide fake endpoints that return the data we need. This is much simpler with DI.
+
+It also makes it easier to follow the adage "program to an interface, not an implementation." Client code (in this case, the Repository) can define the interface upon which they depend, and then never even import the actual implementation. This helps us to separate concerns.
+
+##### Dependency injection vs service locators
+
+We use a service locator pattern. This is similar to, but distinct from dependency injection.
```
mcomella: This should be part of the `decision` section but sections after this are more appropriate for the context.

It also sounds like what we currently do, as opposed to what we will do.

documentation/architecture/ADR1:dependency-injection.md
```diff
@@ -0,0 +1,76 @@
+## Context
+##### What is DI?
+
+In DI, client code does not create it's own dependencies.
```
mcomella: nit: "its"

mcomella:
To follow-up, after reading the ADR, from my perspective, I know the theory of SL vs. DI but **I don't understand them in practice:** what problems people run into after using them for multiple years on large projects? We should try to figure that out (someone must have written about them!) and include it in our decision before we run into these issues ourselves.

liuche:
Merging this because the TC failure is in fetching resources, and it might be bc of github outages, but local lint passes and this is not important enough to block our hackday.

PR #1335, opened 2018-10-23
Package rename: focus -> tv.firefox

liuche:
### Pull Request checklist

Package rename, tests run


app/src/main/java/org/mozilla/tv/firefox/generated/LocaleList.java
```diff
@@ -0,0 +1,9 @@
+package org.mozilla.tv.firefox.generated;
```
mcomella: I think we need to remove org.mozilla.focus.generated (I wonder why this gets generated into the src dir at all though...).

app/src/main/AndroidManifest.xml
```diff
@@ -39,7 +39,7 @@
                 android:name="android.support.FILE_PROVIDER_PATHS"
                 android:resource="@xml/provider_paths"/>
         </provider>
-        <activity android:name=".MainActivity"
+        <activity android:name="org.mozilla.tv.firefox.MainActivity"
```
mcomella: I think this can be `.MainActivity`, right?

app/src/main/AndroidManifest.xml
```diff
@@ -26,7 +26,7 @@
         android:label="@string/app_name"
         android:supportsRtl="true"
         android:theme="@style/AppTheme"
-        android:name=".FocusApplication"
```
mcomella: I think `.FirefoxApplication`, right?

app/src/main/AndroidManifest.xml
```diff
@@ -76,13 +76,13 @@
             </intent-filter>
         </activity>
 
-        <activity android:name=".browser.InfoActivity"
+        <activity android:name="org.mozilla.tv.firefox.browser.InfoActivity"
```
mcomella: I don't think these need to change either.

liuche:
Merging this because the TC failure is in fetching resources, and it might be bc of github outages, but local lint passes and this is not important enough to block our hackday.

PR #1346, opened 2018-10-23
Issue #1344: break VM out of PocketVideoFragment

Baron-Severin:
### Pull Request checklist


Baron-Severin:
I tried to reconstruct as reasonable a Git history as possible.  The original version was a mess of back and forth changes.

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketRepo.kt
```diff
@@ -23,4 +35,72 @@ open class PocketRepo(private val pocketEndpoint: PocketEndpoint) {
...
+        // mutableState.value should always be initialized at the top of init, because we treat
+        // it as non-null
+        @Suppress("SENSELESS_COMPARISON")
+        // Values of BuildConfig can change but the compiler doesn't know that
+        mutableState.value = when {
+            BuildConfig.POCKET_KEY == null -> PocketRepoState.NoKey
+            else -> PocketRepoState.Loading
+        }
+
+        update()
```
mcomella: I don't like that this has side effects in the constructor: this generally mocks mocking difficult. I'd rather have a state factory method has the initialization method that constructs and calls `update()`.

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketDrawable.kt
```diff
@@ -0,0 +1,20 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.tv.firefox.pocket
+
+import android.support.v4.content.ContextCompat
+import android.support.v4.graphics.drawable.DrawableCompat.setTint
+import android.widget.ImageView
+import org.mozilla.tv.firefox.R
+
+object PocketDrawable {
```
mcomella: nit: javadoc

app/src/test/java/org/mozilla/tv/firefox/utils/PreventLiveDataMainLooperCrashRule.kt
```diff
@@ -0,0 +1,9 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.tv.firefox.utils
+
+import android.arch.core.executor.testing.InstantTaskExecutorRule
+
+typealias PreventLiveDataMainLooperCrashRule = InstantTaskExecutorRule
```
mcomella: nit: kdoc. Why is this necessary?

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketViewModel.kt
```diff
@@ -50,6 +59,6 @@ class PocketViewModel(pocketRepo: PocketRepo) : ViewModel() {
         url = "https://www.mozilla.org/en-US/",
         thumbnailURL = "https://blog.mozilla.org/firefox/files/2017/12/Screen-Shot-2017-12-18-at-2.39.25-PM.png,",
         popularitySortId = it
-        )
+    )
```
mcomella: nit: maybe `) }`? We open the braces on the same lines so it's intuitive to close on the same line, right?

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketVideoMegaTile.kt
```diff
@@ -10,42 +10,17 @@ import android.view.Gravity
 import android.view.LayoutInflater
```
mcomella: nit: issue number in commit?

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/BrowserNavigationOverlay.kt
```diff
@@ -238,24 +240,26 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
...
-            // TODO: #864 show loading screen
-            launch(UI) {
-                pocketVideoMegaTileView.pocketVideos = pocketVideos.await()
+    fun observeForMegaTile(fragment: Fragment) {
+        // TODO remove this
+        // This function is necessary because BrowserNavigationOverlay is not a LifecycleOwner,
+        // and so cannot subscribe to PocketRepo's LiveData. When the overlay is turned into a
+        // fragment, this can be moved into the overlay fragment
+        val serviceLocator = fragment.context!!.serviceLocator
+        val factory = ViewModelFactory(serviceLocator)
```
mcomella: `ViewModelFactory` is stateless: should we just store one instance in the serviceLocator instead to make this (neglibly) more efficient?

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/BrowserNavigationOverlay.kt
```diff
@@ -238,24 +240,26 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
...
+        val pocketViewModel = ViewModelProviders.of(fragment, factory).get(PocketViewModel::class.java)
+
+        pocketViewModel.state.observe(fragment.viewLifecycleOwner, Observer { state ->
+            state ?: return@Observer
```
mcomella: nit: maybe put this in else case instead? I'd rather `when` be exhaustive because it better displays intent.

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketRepo.kt
```diff
@@ -0,0 +1,106 @@
...
+    object Loading : PocketRepoState()
+    object NoKey : PocketRepoState()
+    object FetchFailed : PocketRepoState()
+}
+
+/**
+ * Manages backing state for Pocket data, as well as any logic related to
+ * retrieving or storing that data.
+ */
+open class PocketRepo(private val pocketEndpoint: PocketEndpoint) {
```
mcomella: nit: PocketVideoRepo?

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketRepo.kt
```diff
@@ -0,0 +1,106 @@
...
+import kotlinx.coroutines.experimental.Job
+import kotlinx.coroutines.experimental.delay
+import kotlinx.coroutines.experimental.launch
+import kotlinx.coroutines.experimental.withContext
+import org.mozilla.tv.firefox.BuildConfig
+import java.util.concurrent.TimeUnit
+
+private val CACHE_UPDATE_FREQUENCY_MILLIS = TimeUnit.MINUTES.toMillis(45)
+
+sealed class PocketRepoState {
```
mcomella: nit: `PocketVideoRepo.FeedState`? I think it's unclear what "state" will represent.

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketRepo.kt
```diff
@@ -0,0 +1,106 @@
...
+
+    private fun startBackgroundUpdatesInner() = launch {
+        while (true) {
+            val delayForMillis = nextUpdateMillis - SystemClock.elapsedRealtime()
+            if (delayForMillis > 0) {
+                delay(delayForMillis, TimeUnit.MILLISECONDS)
+            }
+
+            val newState = requestVideos()
+            // Delay next request if the previous was successful, otherwise try again immediately
```
mcomella: So we retry infinitely? That doesn't seem right. Maybe try to follow the previous logic, which I don't think was this?

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketRepo.kt
```diff
@@ -0,0 +1,106 @@
...
+        return videos?.let { PocketRepoState.LoadComplete(it) } ?: PocketRepoState.FetchFailed
+    }
+
+    private fun updateState(newState: PocketRepoState) {
+        val oldState = mutableState.value!! // State is initialized at the top of init
+
+        val computedState = PocketRepoStateMachine(newState, oldState).computedState()
+
+        if (oldState != computedState) {
+            mutableState.postValue(computedState)
```
mcomella: If a user is looking at the Pocket feed and we do a background update, it's going to update the feed right underneath them, right? Do we handle this case? If we don't, it's going to be a bad user experience. Similar for the mega tile: if a user sees a video they want and it changes, they might be annoyed.

iirc, the previous implementation had a cached version of the feed so it'd never update when the user is looking at it (but, for better or worse, could update if they backed out).

This comment should probably be in the view layer, but curiously, it's only necessary because of an implementation detail (background updates) of the repo. 🤷‍♀️ Then again, this is an important concern for most UIs that used cached data.

app/src/test/java/org/mozilla/tv/firefox/pocket/PocketViewModelTest.kt
```diff
@@ -0,0 +1,102 @@
+package org.mozilla.tv.firefox.pocket
```
mcomella: nit: license

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketVideoFragment.kt
```diff
@@ -19,74 +21,52 @@ import android.widget.ImageView
...
-        // here. Typically, the data would be provided with `Fragment.getArguments`, which would be
-        // provided to reconstructed fragment but we intentionally don't do that: if the feed is
-        // being reconstructed, the user probably cares more about an up-to-date feed rather than
-        // consistency with the home screen they saw a while ago.
-        if (!this::deferredVideos.isInitialized) {
-            deferredVideos = context!!.serviceLocator.pocket.getRecommendedVideos()
-        }
+        serviceLocator = context!!.serviceLocator
+        val factory = ViewModelFactory(serviceLocator)
+        viewModel = ViewModelProviders.of(this, factory).get(PocketViewModel::class.java)
```
mcomella: Why not do this in `onCreateView` and avoid separating this logic?

In fact, this might break because the viewModel's lifecycle owner is viewLifecycleOwner (I'm not sure if a ViewModel can be re-used at that point).

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/BrowserNavigationOverlay.kt
```diff
@@ -238,24 +240,26 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
...
-                pocketVideosContainer.visibility = View.VISIBLE
-                pocketErrorContainer.visibility = View.GONE
-                pocketVideosContainer.contentDescription = resources.getString(R.string.pocket_home_a11y_tile_focused,
-                        resources.getString(R.string.pocket_brand_name))
-            }
-        } else {
-            // TODO: #864 show loading screen
-            launch(UI) {
-                pocketVideoMegaTileView.pocketVideos = pocketVideos.await()
+    fun observeForMegaTile(fragment: Fragment) {
```
mcomella: Why not take a context and a lifecycle owner instead of a fragment?

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/BrowserNavigationOverlay.kt
```diff
@@ -216,7 +218,7 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
                 resources.getString(R.string.pocket_brand_name)) + " " + resources.getString(R.string.pocket_video_feed_reload_button)
 
         megaTileTryAgainButton.setOnClickListener { _ ->
-            pocketVideos = context!!.serviceLocator.pocket.getRecommendedVideos()
+            context!!.serviceLocator.pocketRepo.update()
```
mcomella: Did we decide as a team that we shouldn't go through the ViewModel to change these?

We should touch base with the team on our desired approach and be consistent. Maybe testability (e.g. swapping in a custom ViewModel during a UI test, we can have control over any side effects) is a concern.

mcomella: CC @psymoon ^ Since your implementation has these decisions too.

Baron-Severin: No need for a meeting, this is just a mistake on my part.  I'll update to go through the VM.

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketRepo.kt
Baron-Severin: I agree that we should change this, but this is how it currently works in master.  I can make that change here or we can add a follow-up issue, either way is fine.

```kotlin
    private fun startBackgroundUpdatesInner() = launch {
        while (true) {
            val delayForMillis = nextUpdateMillis - SystemClock.elapsedRealtime()
            if (delayForMillis > 0) {
                delay(delayForMillis, TimeUnit.MILLISECONDS)
            }

            val deferredVideoUpdate = getRecommendedVideosNoCacheAsync()

            // We only want to update the cache 1) after the request completes so the user never has
            // to see a loading screen except for the initial load and 2) if the request has been
            // successful (i.e. non-null) so the user never has to see an error screen (unless we
            // never got data initially).
            val videoUpdate = deferredVideoUpdate.await()
            if (videoUpdate != null) {
                videosCache = deferredVideoUpdate
                lastUpdateMillis = SystemClock.elapsedRealtime()
            }
        }
    }
```

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketViewModel.kt
```diff
@@ -0,0 +1,55 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.tv.firefox.pocket
+
+import android.arch.lifecycle.MutableLiveData
+import android.arch.lifecycle.ViewModel
+
+sealed class PocketViewModelState {
```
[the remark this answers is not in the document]
Baron-Severin: Observing LiveData is pretty noisy, and we're generally going to be observing all of it in the same place anyway, so I'm thinking VMs should wrap all of their state into a single object and just emit that.  In this case the state is either a list of tiles or error, and I'm not sure what else I would call that.  I'm open to suggestions, but I think this is appropriate.

I do agree on moving inside the VM for namespacing, and I'll shorten the name accordingly.

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketViewModel.kt
```diff
@@ -35,4 +39,17 @@ class PocketViewModel(pocketRepo: PocketRepo) : ViewModel() {
 
     val state = MutableLiveData<PocketViewModelState>()
 
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    val loadingPlaceholders: List<PocketFeedItem> =
+        List(POCKET_VIDEO_COUNT) { PocketFeedItem.Loading(R.color.photonGrey50) }
+
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    val noKeyPlaceholders: List<PocketFeedItem> = (1..POCKET_VIDEO_COUNT).map { PocketFeedItem.Video(
```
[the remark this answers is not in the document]
Baron-Severin: Correct me if I'm wrong, but I don't believe I can iterate the `ID` and `popularitySortId` with a `List` constructor.

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketRepoStateMachine.kt
```diff
@@ -0,0 +1,50 @@
...
+import org.mozilla.tv.firefox.pocket.PocketRepoState.Loading
+import org.mozilla.tv.firefox.pocket.PocketRepoState.NoKey
+
+/**
+ * Manages all logic for updating cached [PocketRepoState] in a testable way.
+ *
+ * When instantiated with cached state and a new value, [computedState] will return
+ * whatever the new state should be.
+ */
+class PocketRepoStateMachine(private val inputState: PocketRepoState, private val cachedState: PocketRepoState) {
```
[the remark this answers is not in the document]
Baron-Severin: Cached state is whatever the LiveData's previous value is, input state is a fetch, an error, or a loading.  Output is whatever should be pushed to the LiveData.  I'll try to improve the naming here, and update the kdoc to be more explicit.

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/BrowserNavigationOverlay.kt
Baron-Severin: `ViewModelProviders#of` requires either a `Fragment` or `Activity`.

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketEndpoint.kt
```diff
@@ -37,26 +37,26 @@ object PocketEndpoint {
...
-    suspend fun getRecommendedVideos(): List<PocketVideo>? {
+    suspend fun getRecommendedVideos(): List<PocketFeedItem.Video>? {
         val videosJSON = PocketEndpointRaw.getGlobalVideoRecommendations() ?: return null
         return convertVideosJSON(videosJSON)
     }
     /** Developer Ergonomics
      * @return A placeholder list of videos when Pocket API key is not provided. */
-    fun getPlaceholderVideos(): List<PocketVideo>? {
-        val placeholderList = mutableListOf<PocketVideo>()
+    fun getPlaceholderVideos(): List<PocketFeedItem.Video>? {
```
dnarcese: I don't think this function is needed anymore since it is in the VM and not being used anywhere.

app/src/test/java/org/mozilla/tv/firefox/utils/PreventLiveDataMainLooperCrashRule.kt
Baron-Severin: I think it's unclear what problem `@Rule val rule = InstantTaskExecutorRule()` is solving.  If someone is having trouble testing LiveData and looks back to one of these files, it will be immediately obvious how to fix it.  I don't know of any other situation in which we would use this rule, but if we run across one we can rename the alias.

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketRepo.kt
mcomella: So it is: I apologize for my assumption. 😞

I think one thing that makes this code unclear is that `nextUpdateMillis` is calculated from `lastUpdateMillis` so this infinite retry happens at most at intervals of 45 minutes, which seems fine to me (I had assumed it there was no delay because `nextUpdateMillis` wasn't updated). Since it's only used in one place, maybe we should calculate `nextUpdateMillis` at the top of the loop instead?

Baron-Severin: All good, it was a good call out.  Agreed, I'll make that change.

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketViewModel.kt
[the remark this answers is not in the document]
mcomella: > Observing LiveData is pretty noisy,

What do you mean?

> and we're generally going to be observing all of it in the same place anyway, so I'm thinking VMs should wrap all of their state into a single object and just emit that.

This is purely hypothetical at the moment because Pocket only has one state (Loading), but I think have one global state loses the advantages (and disadvantages) of encapsulation: isolating unrelated state and their transitions so that the smaller amount of code is easier to understand and making it more difficult to allow state to unintentionally couple.

In the short term, I'd prefer to be specific and name it after the one thing it is - a loading state - so it's clearer to someone what exactly code manipulated the `PocketViewModelState` entails. If someone expands that object to encapsulate more state because they feel it's a better design, then it can take a more general name.

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketViewModel.kt
```diff
@@ -20,24 +20,7 @@ sealed class PocketFeedItem {
         val url: String,
         val thumbnailURL: String,
         val popularitySortId: Int
-    ) : PocketFeedItem() {
-
-        companion object {
-            fun fromJSONObject(jsonObj: JSONObject): PocketVideo? = try {
```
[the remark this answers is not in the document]
mcomella: We could keep the method on the VM but have it call out to a separate file:
```kotlin
data class PocketViewModel {
    companion object {
        fun fromJSONObject(json) = PocketViewModelFactory.fromJSONObject(json)

File: app/src/test/java/org/mozilla/tv/firefox/utils/PreventLiveDataMainLooperCrashRule.kt (hunk 3 above)
Comment (mcomella) [reply app/src/test/java/org/mozilla/tv/firefox/utils/PreventLiveDataMainLooperCrashRule.kt]: Maybe a separate issue but I wonder if we could subclass RobolectricTestRunner (or similar) and apply this rule to all tests.

File: app/src/main/java/org/mozilla/tv/firefox/pocket/PocketRepoStateMachine.kt (hunk 18 above)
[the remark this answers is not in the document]
Comment (mcomella) [reply app/src/main/java/org/mozilla/tv/firefox/pocket/PocketRepoStateMachine.kt]: Those sound like different things – should they be different types?

File: app/src/main/java/org/mozilla/tv/firefox/pocket/PocketViewModel.kt (hunk 17 above)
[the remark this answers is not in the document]
Comment (mcomella) [reply app/src/main/java/org/mozilla/tv/firefox/pocket/PocketViewModel.kt]: afaik, the `it` value in the `List` constructor is the index so it should work the same way.

File: app/src/main/java/org/mozilla/tv/firefox/pocket/PocketViewModel.kt (hunk 16 above)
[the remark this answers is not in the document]
Comment (Baron-Severin) [reply app/src/main/java/org/mozilla/tv/firefox/pocket/PocketViewModel.kt]: >> Observing LiveData is pretty noisy,

> What do you mean?

There's enough boilerplate around it that if, for example, the VM carried three different values, I wouldn't want each to have its own LiveData.
```kotlin
personVm.name.observe(fragment.viewLifecycleOwner, Observer { name ->
  nameTv.text = name
}
personVm.age.observe(fragment.viewLifecycleOwner, Observer { age ->
  ageTv.text = age
}
personVm.height.observe(fragment.viewLifecycleOwner, Observer { height ->
  heightTv.text = height
}
```
as opposed to
```kotlin
personVm.state.observe(fragment.viewLifecycleOwner, Observer { state ->
  setName(state.name)
  setAge(state.age)
  setHeight(state.height)
}
```

> Pocket only has one state (Loading)

Tbh, I'm not sure I would agree with this assessment.  As is, Pocket VM state covers:
- Loading vs complete
- Error state
- Feed contents (images, urls, etc)
- Feed is clickable vs not

I don't feel that `LoadingState` properly conveys that.

> I think have one global state loses the advantages (and disadvantages) of encapsulation

True, but imo we're encapsulating state by breaking it up into relatively small view models.  I'm not sure there's much of a difference between organizing the interiors of the VMs to contain several independent LiveDatas and having one LiveData that delivers all of the state.  If we see any VMs that are managing too much state, I think the prefered solution would be to break them into smaller, composable VMs wherever possible.

app/src/test/java/org/mozilla/tv/firefox/utils/PreventLiveDataMainLooperCrashRule.kt
Baron-Severin: Probably.  I guess it couldn't really do any harm, although my knee-jerk reaction is that I like explicitly specifying custom test behavior where possible.

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketEndpoint.kt
Baron-Severin: Good catch!  I'll pull that out.

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketRepo.kt
Baron-Severin: Added an exponential backoff in a recent commit.

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketViewModel.kt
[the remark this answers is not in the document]
Baron-Severin: That's something I should know.  Fixed, thanks.

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketViewModel.kt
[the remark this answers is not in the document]
Baron-Severin: Good call.  Done.

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketRepo.kt
Baron-Severin: A few possible solutions.

VM solution
- Update VM to stop listening to updates after a successful one
- Modify mega tile and Pocket screen to use the same VM instance, and make sure it falls out of scope when neither is active.  This would mean retrieving the VM with a `LifecycleOwner` that they share, but that the browser doesn't.  I'm not sure what that would be.

Repo solution
- Pause background updates whenever the menu or the Pocket screen are displayed

There has to be a better solution, both of these are terrible.  I'm thinking about it.

app/src/main/java/org/mozilla/tv/firefox/MainActivity.kt
```diff
@@ -135,12 +134,12 @@ class MainActivity : LocaleAwareAppCompatActivity(), OnUrlEnteredListener, Media
 
     override fun onStart() {
         super.onStart()
-        serviceLocator.pocket.startBackgroundUpdates()
+        serviceLocator.pocketRepo.startBackgroundUpdates()
```
psymoon: I wonder if there is a need for another ViewModel here; MainActivityViewModel? May work with `CursorDispatcher` or any other lifecycle subscriptions that's more on the application level

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/BrowserNavigationOverlay.kt
```diff
@@ -238,24 +240,26 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
...
+        // TODO initialize pocketViewModel in onCreate when overlay is refactored into a fragment
+        pocketViewModel = ViewModelProviders.of(fragment, factory).get(PocketViewModel::class.java)
+
+        pocketViewModel.state.observe(fragment.viewLifecycleOwner, Observer { state ->
+            when (state) {
+                is PocketViewModelState.Error -> showMegaTileError()
+                is PocketViewModelState.Feed -> {
+                    pocketVideoMegaTileView.setContent(state.feed)
+                }
+                null -> return@Observer
```
psymoon: Why would the state be null?

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketRepo.kt
```diff
@@ -0,0 +1,125 @@
...
+ * Manages backing state for Pocket data, as well as any logic related to
+ * retrieving or storing that data.
+ */
+open class PocketRepo(
+    private val pocketEndpoint: PocketEndpoint,
+    private val pocketRepoStateMachine: PocketRepoStateMachine,
+    buildConfigDerivables: BuildConfigDerivables
+) {
+
+    private val mutableState = MutableLiveData<PocketRepoState>().apply {
```
psymoon: Curious about the purpose of having both `mutableState` and `state`. Why did you decide to do this?

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketRepo.kt
```diff
@@ -0,0 +1,125 @@
...
+    @Volatile private var lastUpdateAttemptMillis = -1L
+    @Volatile private var retryDelayMillis = 1_000L
+    @Volatile private var lastAttemptWasSuccessful = false
+
+    fun update() {
+        launch { updateInner() }
+    }
+
+    @UiThread // update backgroundUpdates.
+    fun startBackgroundUpdates() {
```
psymoon: nit: could use a comment about the lifecycle of background update process

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketRepoStateMachine.kt
```diff
@@ -0,0 +1,31 @@
...
+        return when {
+            videos?.isNotEmpty() == true -> LoadComplete(videos)
+            currentState === Loading -> FetchFailed
+            else -> currentState
+        }
+    }
+
+    fun setLoading(currentState: PocketRepoState): PocketRepoState {
+        return when {
+            currentState === FetchFailed -> Loading
```
psymoon: So `Loading` state can only be achieved from `FetchFailed`? kdoc of pseudo diagram representing each possible state transitions would be great

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketVideoMegaTile.kt
```diff
@@ -54,7 +29,31 @@ class PocketVideoMegaTile(
...
+
+    private fun showLocalDrawable(view: ImageView, drawable: Int) {
+        view.setImageResource(drawable)
+    }
+
+    private fun showFetchedImage(imageView: ImageView, url: String) {
+        val roundCornerTransformation =
+            RoundCornerTransformation(resources.getDimension(R.dimen.pocket_video_mega_tile_thumbnail_corner_radius))
+
+        PicassoWrapper.client.load(url)
```
psymoon: nit: maybe we can provide better UX by using `onImageLoadFailed`

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketVideoParser.kt
```diff
@@ -0,0 +1,28 @@
...
+            id = jsonObj.getInt("id"),
+            title = jsonObj.getString("title"),
+            // Note that the 'url' property of our domain object can be retrieved from
+            // either of two JSON fields, and we make no distinction internally
+            url = jsonObj.optString("tv_url", null) ?: jsonObj.getString("url"),
+            thumbnailURL = jsonObj.getString("image_src"),
+            popularitySortId = jsonObj.getInt("popularity_sort_id")
+        )
+    } catch (e: JSONException) {
+        null
```
psymoon: This is fragile because if pocket API decides to change JSON key then the users will end up with unhandled state until next app store submission. Generally risky not handling error when parsing raw JSON. Good news is that we are using MVVM. One possible solution is to introduce a `PARSING_ERROR` state

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketViewModel.kt
```diff
@@ -0,0 +1,72 @@
...
+    data class Loading(val thumbnailResource: Int) : PocketFeedItem()
+    data class Video(
+        val id: Int,
+        val title: String,
+        val url: String,
+        val thumbnailURL: String,
+        val popularitySortId: Int
+    ) : PocketFeedItem() {
+        companion object {
+            fun fromJSONObject(jsonObject: JSONObject) = PocketVideoParser.parse(jsonObject)
```
psymoon: See above for `PocketVideoParser`

app/src/main/java/org/mozilla/tv/firefox/utils/ServiceLocator.kt
```diff
@@ -38,6 +41,12 @@ import org.mozilla.tv.firefox.pocket.Pocket
  *   open val telemetry: TelemetryInterface by lazy { SentryWrapper() }
  *   ```
  */
-open class ServiceLocator {
-    open val pocket by lazy { Pocket() }
+class ServiceLocator {
+    private val pocketEndpoint by lazy { PocketEndpoint }
+    private val buildConfigDerivables by lazy { BuildConfigDerivables() }
+    private val pocketRepoStateMachine by lazy { PocketRepoStateMachine() }
+    val pocketRepo = PocketRepo(pocketEndpoint, pocketRepoStateMachine, buildConfigDerivables).apply {
```
psymoon: Do you think all of pocket related instances need to be exposed here? I understand for testability it makes sense to pass them into PocketRepo's constructors since we can mock the dependencies easier. But I even wonder if some of these instance memory need to be allocated for the entire lifecycle of the application

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/BrowserNavigationOverlay.kt
Baron-Severin: Not sure, but `Observer#onChanged` passes a nullable argument.  Possibly if an object observes before one has been set, I'm not sure how LiveData handles that case.

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketRepo.kt
Baron-Severin: It's a backing field, equivalent to `_pinnedTiles` in your PR.  I'll update the name to make that more clear.

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketVideoParser.kt
Baron-Severin: We are handling failed parsing, but not in this layer.  In the repo this is treated the same way as any IO exception, which displays either the most recently cached data or a 'try again' button to the user.

app/src/main/java/org/mozilla/tv/firefox/utils/ServiceLocator.kt
Baron-Severin: > Do you think all of pocket related instances need to be exposed here?

I think it makes sense to do object setup in the `ServiceLocator`.  They aren't part of its public API, so I wouldn't consider them exposed.  If setup gets big we can break these into individual factories (and if it gets huge we can introduce a framework), but that's a pretty easy refactor so I would prefer to keep it simple until we need to.

> But I even wonder if some of these instance memory need to be allocated for the entire lifecycle of the application

Good point.  A lot of those 'by lazy's are useless, I'll pull them out.

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketRepo.kt
Baron-Severin: Intermediate cache solution
- Write a thin cache that lives between the repo and the VM.  Once a LoadComplete has passed through, stop updating the cache.  Have the ScreenController clear the cache whenever the active screen is not overlay or Pocket screen.

app/src/main/java/org/mozilla/tv/firefox/MainActivity.kt
Baron-Severin: Yeah, that makes sense to me.  Out of scope for this refactor, but I'll add a note on this line to move it to an eventual VM.

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketViewModel.kt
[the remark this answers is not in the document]
mcomella: > If we see any VMs that are managing too much state, I think the prefered solution would be to break them into smaller, composable VMs wherever possible.

My concern with this is that we lose specificity on the naming of the state objects, which makes it hard to understand what the object does without looking at its definition, hurting abstraction ability.

I think the naming/encapsulation trade-offs seems to be on a case-by-case basis so I don't feel too strongly here.

app/src/test/java/org/mozilla/tv/firefox/utils/PreventLiveDataMainLooperCrashRule.kt
mcomella: I think the benefits are preventing violating DRY and making it clearer what is special about *this test* but we lose transparency.

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketRepo.kt
Baron-Severin: Implemented the intermediate cache.  Please take a look and let me know what you think.

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketVideoMegaTile.kt
Baron-Severin: Hm, I don't see the option.  Does `placeholder` take care of that?

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketVideoRepo.kt
```diff
@@ -0,0 +1,129 @@
...
+        fun List<PocketViewModel.FeedItem>?.wasSuccessful() = this?.isNotEmpty() == true
+
+        fun updateRequestTimers(requestSuccessful: Boolean) {
+            lastAttemptWasSuccessful = requestSuccessful
+            val now = SystemClock.elapsedRealtime()
+            lastUpdateAttemptMillis = now
+            when (requestSuccessful) {
+                true -> {
+                    lastSuccessfulUpdateMillis = now
+                    retryDelayMillis = 1_000L
```
liuche: Nit: pull out this default retry time into a constant at the top of the file.

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketVideoRepo.kt
```diff
@@ -0,0 +1,129 @@
...
+    @get:UiThread
+    @set:UiThread
+    private var backgroundUpdates: Job? = null
+    @Volatile private var lastSuccessfulUpdateMillis = -1L
+    @Volatile private var lastUpdateAttemptMillis = -1L
+    @Volatile private var retryDelayMillis = 1_000L
+    @Volatile private var lastAttemptWasSuccessful = false
+
+    fun update() {
+        launch { updateInner() }
```
liuche: Should a manual `update` reset the exponential backoff? Say a user clicks the "try again" button.

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketVideoRepo.kt
```diff
@@ -0,0 +1,129 @@
...
+    @UiThread // update backgroundUpdates.
+    fun startBackgroundUpdates() {
+        backgroundUpdates?.cancel(CancellationException("Cancelling unexpectedly active job to ensure only one is running"))
+        backgroundUpdates = startBackgroundUpdatesInner()
+    }
+
+    // When we the app is not in use, we don't want to hit the network for no reason, so we cancel updates
+    @UiThread // stop updating backgroundUpdates.
+    fun stopBackgroundUpdates() {
+        backgroundUpdates?.cancel(CancellationException("Stopping background updates"))
```
liuche: Does the CancellationException include the stack? Might want to make the message a tiny bit more explicit and mention Pocket feed, if only for easier log grepping.

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketVideoRepo.kt
liuche: Does this launch need an explicit scope with Kotlin 1.3?

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketRepoCache.kt
```diff
@@ -0,0 +1,34 @@
...
+    private val _feedState = MutableLiveData<PocketVideoRepo.FeedState>() // Mutable backer for feedState
+    open val feedState: LiveData<PocketVideoRepo.FeedState> = _feedState
+
+    private val observer = Observer<PocketVideoRepo.FeedState> {
+        _feedState.postValue(it)
+    }
+
+    // This should only be called when Pocket videos are not visible to the user.
+    // See class kdoc
+    fun unfreeze() = repo.feedState.observeForever(observer)
```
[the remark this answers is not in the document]
Baron-Severin: Added this to the kdoc

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketVideoRepo.kt
Baron-Severin: I think when we migrate it will make sense to make the repo itself the enclosing scope here, since we don't have any real case where we cancel this.  I'm not certain though, I need to read more about the changes.

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketVideoRepo.kt
Baron-Severin: Not sure, this is taken from the previous `Pocket` implementation.  The kdoc on that exception includes ...
```
 * It indicates _normal_ cancellation of a coroutine.
 * **It is not printed to console/log by default uncaught exception handler**.
```
... so I'm not sure it'll show up in logs at all.  I'll update the message to mention Pocket regardless.

PR #1351, opened 2018-10-24
Issue #1345: Introduce MVVM to PinnedTile

psymoon:
### Pull Request checklist


app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/BrowserNavigationOverlay.kt
```diff
@@ -134,6 +136,9 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
 
     private lateinit var tileAdapter: PinnedTileAdapter
 
+    lateinit var pinnedTileViewModel: PinnedTileViewModel
+    lateinit var lifeCycleOwner: LifecycleOwner
```
Baron-Severin: I just noticed that we don't use this anywhere in this class except to start observing it.  Suggest making `PinnedTileViewModel` a parameter of `#initTiles` and not maintaining a reference outside of that function.  

FWIW I'm doing this too and will be making the same change on the Pocket refactor.

app/src/main/java/org/mozilla/tv/firefox/pinnedtile/PinnedTileRepo.kt
```diff
@@ -0,0 +1,131 @@
...
+/**
+ * TODO
+ * Some methods require applicationContext in order to access /assets/
+ */
+class PinnedTileRepo(private val applicationContext: Application) {
+    private val _pinnedTiles = MutableLiveData<LinkedHashMap<String, PinnedTile>>() // FIXME: URI? String?
+    val tileCount get() = _pinnedTiles.value?.size
+
+    // TODO:  applicationContext.getSharedPreferences(PREF_HOME_TILES, Context.MODE_PRIVATE)
+    private var _sharedPreferences: SharedPreferences = applicationContext.getSharedPreferences(PREF_HOME_TILES, Context.MODE_PRIVATE)
```
Baron-Severin: Unless I'm mistaken, it looks like this can be a `val`.

app/src/main/java/org/mozilla/tv/firefox/pinnedtile/PinnedTileViewModel.kt
```diff
@@ -0,0 +1,39 @@
...
+import android.arch.lifecycle.ViewModel
+
+/**
+ * TODO
+ *
+ * AndroidViewModel() is a context aware ViewModel => Used for PinnedTileRepo
+ */
+class PinnedTileViewModel(private val pinnedTileRepo: PinnedTileRepo) : ViewModel() {
+
+    private val _tilesList = MediatorLiveData<MutableList<PinnedTile>>()
```
Baron-Severin: Unless I'm mistaken, it looks like the mutable lists in this class can be all normal lists.

app/src/main/java/org/mozilla/tv/firefox/webrender/WebRenderFragment.kt
```diff
@@ -265,8 +268,9 @@ class WebRenderFragment : EngineViewLifecycleFragment(), Session.Observer {
 
                 // This assumes that since we're deleting from a Home Tile object that we created
                 // that the Uri is valid, so we do not do error handling here.
-                HomeTilesManager.removeHomeTile(tileToRemove, context!!)
-                homeTileAdapter.removeTile(tileToRemove.idToString())
+                // FIXME: NavigationOverlayFragment->ViewModel->Repo
+                _pinnedTileViewModel.unpin(tileToRemove.url)
+//                homeTileAdapter.removeTile(tileToRemove.idToString())
```
Baron-Severin: Nit: commented code

app/src/main/java/org/mozilla/tv/firefox/webrender/WebRenderFragment.kt
```diff
@@ -86,9 +86,14 @@ class WebRenderFragment : EngineViewLifecycleFragment(), Session.Observer {
...
     var sessionFeature: SessionFeature? = null
 
+    private lateinit var _pinnedTileViewModel: PinnedTileViewModel
+
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
         initSession()
+
+        val factory = ViewModelFactory(context!!.serviceLocator)
+        _pinnedTileViewModel = ViewModelProviders.of(this, factory).get(PinnedTileViewModel::class.java)
```
Baron-Severin: Optional nit: Since this isn't backing anything in this class, I'm not sure the underscore prefix applies here.  Kind of a grey area though.  Personally, I would take it off.

app/src/main/java/org/mozilla/tv/firefox/pinnedtile/PinnedTileAdapter.kt
```diff
@@ -95,49 +98,39 @@ class PinnedTileAdapter(
...
-            // the first tile that will not match the
-            // previous list of tiles
-            if (tile != pinnedTiles[index]) {
-                tiles = pinnedTiles
-                notifyItemInserted(index)
-                return
+
+        // DiffUtil allows diff calculation between two lists and output a list of update
+        // operations that converts the first list into the second one
+        val diff = DiffUtil.calculateDiff(object : DiffUtil.Callback() {
```
mcomella: I think we can simplify this by using `RecyclerView.widget.ListAdapter`, which uses `DiffUtil` nder the hood.

app/src/main/java/org/mozilla/tv/firefox/pinnedtile/PinnedTileAdapter.kt
```diff
@@ -95,49 +98,39 @@ class PinnedTileAdapter(
...
-    /**
-     * takes in the home tiles cache and updates the adapter's data source
-     * and UI accordingly, assuming only one new tile is added
-     */
-    fun updateAdapterSingleInsertion(pinnedTiles: MutableList<PinnedTile>) {
-        if (pinnedTiles.size == tiles.size) {
-            // The lists must not be the same size in order
-            // for an insertion to be valid
+    fun setTiles(newTiles: MutableList<PinnedTile>) {
+        if (itemCount == 0) {
```
mcomella: Why don't we want DiffUtil to handle this state too?

app/src/main/java/org/mozilla/tv/firefox/pinnedtile/PinnedTileViewModel.kt
```diff
@@ -0,0 +1,39 @@
...
+
+import android.arch.lifecycle.LiveData
+import android.arch.lifecycle.MediatorLiveData
+import android.arch.lifecycle.Transformations
+import android.arch.lifecycle.ViewModel
+
+/**
+ * TODO
+ *
+ * AndroidViewModel() is a context aware ViewModel => Used for PinnedTileRepo
```
mcomella: nit: unnecessary comment

app/src/main/java/org/mozilla/tv/firefox/webrender/WebRenderFragment.kt
```diff
@@ -156,22 +161,15 @@ class WebRenderFragment : EngineViewLifecycleFragment(), Session.Observer {
                 this@WebRenderFragment.session.url.let { url ->
                     when (value) {
                         NavigationEvent.VAL_CHECKED -> {
-                            CustomTilesManager.getInstance(context!!).pinSite(context!!, url,
+                            // FIXME: OverlayVM -> ToolbarVM -> PinnedTileRepo
+                            context!!.serviceLocator.pinnedTileRepo.addPinnedTile(url,
```
mcomella: In this case, we talk to the repo directly but below we talk through the view model: we should pick one and be consistent.

app/src/main/java/org/mozilla/tv/firefox/pinnedtile/PinnedTileRepo.kt
```diff
@@ -0,0 +1,131 @@
...
+
+    fun loadTilesCache(): MutableLiveData<LinkedHashMap<String, PinnedTile>> {
```
mcomella: private?

app/src/main/java/org/mozilla/tv/firefox/pinnedtile/PinnedTileRepo.kt
```diff
@@ -0,0 +1,131 @@
...
+        val pinnedTiles = linkedMapOf<String, PinnedTile>().apply {
```
mcomella: You could also do:
```kotlin
return linkedHashMapOf... {
    // ...
}.also {
    pinnedTiles = it
}

File: app/src/main/java/org/mozilla/tv/firefox/pinnedtile/PinnedTileRepo.kt [hunk 12]
```diff
@@ -0,0 +1,131 @@
...
+private const val BUNDLED_HOME_TILES_DIR = "bundled"
+private const val HOME_TILES_JSON_PATH = "$BUNDLED_HOME_TILES_DIR/bundled_tiles.json"
+
+/**
+ * TODO
+ * Some methods require applicationContext in order to access /assets/
+ */
+class PinnedTileRepo(private val applicationContext: Application) {
+    private val _pinnedTiles = MutableLiveData<LinkedHashMap<String, PinnedTile>>() // FIXME: URI? String?
+    val tileCount get() = _pinnedTiles.value?.size
```
mcomella: nit: should we return 0 if null?

app/src/main/java/org/mozilla/tv/firefox/pinnedtile/PinnedTileRepo.kt
```diff
@@ -0,0 +1,131 @@
...
+
+        _pinnedTiles.value = pinnedTiles
+
+        return _pinnedTiles
+    }
+
+    fun addPinnedTile(url: String, screenshot: Bitmap?) {
+        val newPinnedTile = CustomPinnedTile(url, "custom", UUID.randomUUID()) // TODO: titles
+        _pinnedTiles.value?.put(url, newPinnedTile)
+        _pinnedTiles.value = _pinnedTiles.value
```
mcomella: What does that do?

app/src/main/java/org/mozilla/tv/firefox/pinnedtile/PinnedTileRepo.kt
```diff
@@ -0,0 +1,131 @@
...
+        saveCustomTiles()
```
mcomella: nit: `persistCustomTiles` might be more clear

app/src/main/java/org/mozilla/tv/firefox/pinnedtile/PinnedTileRepo.kt
```diff
@@ -0,0 +1,131 @@
...
+            }
+        }
+
+        return tileToRemove.idToString()
+    }
+
+    @UiThread
+    fun isUrlPinned(url: String): Boolean? = _pinnedTiles.value?.containsKey(url)
+
+    private fun loadBlacklist(): MutableSet<String> {
```
mcomella: I don't think there's any reason to make these mutable: who knows what mutating a set from shared prefs would do!

app/src/main/java/org/mozilla/tv/firefox/pinnedtile/PinnedTileRepo.kt
```diff
@@ -0,0 +1,131 @@
...
+        return _sharedPreferences.getStringSet(BUNDLED_SITES_ID_BLACKLIST, mutableSetOf())!!
+    }
+
+    private fun saveBlackList(blackList: MutableSet<String>) {
+        _sharedPreferences.edit().putStringSet(BUNDLED_SITES_ID_BLACKLIST, blackList).apply()
+    }
+
+    private fun saveCustomTiles() {
+        val tilesJSONArray = JSONArray()
+        for (tile in _pinnedTiles.value!!.values) {
```
mcomella: nit: slightly less performant but perhaps clearer/safer:
```kotlin
pinnedTiles.value.filterIsInstance(CustomPinnedTile::class).forEach { tile ->
    tilesJSONArray.put(tile.toJSONObject())
}

File: app/src/main/java/org/mozilla/tv/firefox/pinnedtile/PinnedTileRepo.kt [hunk 17]
```diff
@@ -0,0 +1,131 @@
...
+        }
+
+        _pinnedTiles.value = pinnedTiles
+
+        return _pinnedTiles
+    }
+
+    fun addPinnedTile(url: String, screenshot: Bitmap?) {
+        val newPinnedTile = CustomPinnedTile(url, "custom", UUID.randomUUID()) // TODO: titles
+        _pinnedTiles.value?.put(url, newPinnedTile)
```
mcomella: Maybe we should:
```kotlin
fun addPinnedTile(...) {
    persistCustomTiles()
    pinnedTiles.value = loadCachedTiles()
}
```

It's less performant but it hides bugs in the duplicated logic of mutating the pinnedTiles in place. Since we don't call this often and our lists are small, it seems acceptable.

Then again, there's already a hiccup when pinning sites due to the screenshot so maybe we don't want to make that worse.

app/src/main/java/org/mozilla/tv/firefox/pinnedtile/PinnedTileViewModel.kt
```diff
@@ -0,0 +1,39 @@
...
+        _tilesList.addSource(pinnedTileRepo.loadTilesCache()) { pinnedTileMap ->
+            val feedTileList: MutableList<PinnedTile>? = pinnedTileMap?.values?.toMutableList()
+            if ((feedTileList) != null) {
+                _tilesList.value = feedTileList
+            }
+        }
+    }
+
+    // For UI injection
+    fun getTileList(): LiveData<MutableList<PinnedTile>> {
```
mcomella: We shouldn't expose this list as mutable. Note that List is a super-class of MutableList so we can use List everywhere MutableList can be used.

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/BrowserNavigationOverlay.kt
```diff
@@ -185,6 +189,14 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
                 canShowUpinToast = false
             }
         })
+
+        pinnedTileViewModel.getTileList().observe(lifeCycleOwner, Observer {
+            if (it != null) {
```
mcomella: Why would the tile list value ever be null? If it's because it's loading, can we add a "Loading" state instead (e.g. with sealed classes) so it's clear what this means?

app/src/main/java/org/mozilla/tv/firefox/pinnedtile/PinnedTileViewModel.kt
```diff
@@ -0,0 +1,39 @@
...
+ */
+class PinnedTileViewModel(private val pinnedTileRepo: PinnedTileRepo) : ViewModel() {
+
+    private val _tilesList = MediatorLiveData<MutableList<PinnedTile>>()
+    val isEmpty: LiveData<Boolean> = Transformations.map(_tilesList) { input -> input.isEmpty() }
+
+    init {
+        _tilesList.addSource(pinnedTileRepo.loadTilesCache()) { pinnedTileMap ->
+            val feedTileList: MutableList<PinnedTile>? = pinnedTileMap?.values?.toMutableList()
+            if ((feedTileList) != null) {
```
mcomella: We check for null in the VM layer: why is this necessary? Can we just expose the tilesList value directly?

app/src/main/java/org/mozilla/tv/firefox/pinnedtile/PinnedTileViewModel.kt
mcomella: nit: Why not `val tileList: LiveData<...>` with a property getter?

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/BrowserNavigationOverlay.kt
psymoon: Like it

app/src/main/java/org/mozilla/tv/firefox/pinnedtile/PinnedTileAdapter.kt
psymoon: For some odd android reason, `notifyDataSetChanged()` is specifically required to refresh the entire list when adding a new pinned item to an empty adapter. I'm not too sure why DiffUtil doesn't handle this

app/src/main/java/org/mozilla/tv/firefox/webrender/WebRenderFragment.kt
psymoon: As the comment suggests, this particular use case shouldn't be handled by `PinnedTileVM`. It should be handled by `ToolbarVM`. For now, we are just calling it directly to Repo (not a permanent solution). We can discuss about this in the upcoming meeting

app/src/main/java/org/mozilla/tv/firefox/pinnedtile/PinnedTileRepo.kt
psymoon: `LiveData` isn't updated unless a new value is set. Mutating the value within the `LiveData` value itself doesn't emit anything (unless I'm missing something) 🤔

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/BrowserNavigationOverlay.kt
psymoon: Technically never. Old Java habit 😞 . Will update

app/src/main/java/org/mozilla/tv/firefox/pinnedtile/PinnedTileViewModel.kt
psymoon: Technically never. Old Java habit 😞 . Will update

app/src/main/java/org/mozilla/tv/firefox/pinnedtile/PinnedTileRepo.kt
```diff
@@ -42,6 +46,11 @@ class PinnedTileRepo(private val applicationContext: Application) {
         return _pinnedTiles
     }
 
+    @VisibleForTesting(otherwise = VisibleForTesting.PROTECTED)
+    fun getPinnedTiles(): MutableLiveData<LinkedHashMap<String, PinnedTile>> {
+        return _pinnedTiles
```
Baron-Severin: Nit: I'd prefer adding @VisibleForTesting to `val _pinnedTiles` and removing this function entirely

app/src/test/java/org/mozilla/tv/firefox/pinnedtile/PinnedTileRepoTest.kt
```diff
@@ -0,0 +1,112 @@
...
+
+    @Before
+    fun setUp() {
+        pinnedTileRepo = PinnedTileRepo(RuntimeEnvironment.application)
+    }
+
+    @Test
+    fun `WHEN repo fetches initial tileList THEN repo should emit a feed of only bundled tile list`() {
+        val observerSpy = spy(Observer<LinkedHashMap<String, PinnedTile>> {
+            assertTrue(it is LinkedHashMap<String, PinnedTile>)
```
Baron-Severin: Nit: replace line with `assertNotNull(it)`

app/src/test/java/org/mozilla/tv/firefox/pinnedtile/PinnedTileRepoTest.kt
```diff
@@ -0,0 +1,112 @@
...
+        pinnedTileRepo.loadTilesCache()
+
+        verify(observerSpy, times(1)).onChanged(any())
+    }
+
+    @Test
+    fun `WHEN repo adds a new pin THEN repo should emit a combined list of bundled and custom`() {
+        val observerSpy = spy(Observer<LinkedHashMap<String, PinnedTile>> {
+            assertTrue(it is LinkedHashMap<String, PinnedTile>)
+            assert((it as LinkedHashMap<String, PinnedTile>).size == 10 || it.size == 11)
```
Baron-Severin: Could we introduce a `BUNDLED_TILE_COUNT` constant and use in place of these ints?  Otherwise any change to bundled tiles is going to involve a lot of updates to these tests.

app/src/test/java/org/mozilla/tv/firefox/pinnedtile/PinnedTileRepoTest.kt
```diff
@@ -0,0 +1,112 @@
...
+        pinnedTileRepo.loadTilesCache()
+        assertEquals(10, pinnedTileRepo.bundledTilesSize)
+        pinnedTileRepo.removePinnedTile("https://example.com/")
+        assertEquals(10, pinnedTileRepo.bundledTilesSize)
+
+        verify(observerSpy, times(1)).onChanged(any())
+    }
+
+    @Test
+    fun `WHEN repo removes a custom pin THEN repo should emit a smaller list and custom size`() {
```
Baron-Severin: I don't think this should block the merge, but in the future we could decouple these tests from the particulars of our bundled tiles by breaking out a separate object for retrieving them from the file system and injecting it through the repos constructor.  That way the test could control how many exist, and we could also give them useful mock data.

app/src/test/java/org/mozilla/tv/firefox/pinnedtile/PinnedTileViewModelTest.kt
```diff
@@ -0,0 +1,83 @@
...
+import org.mockito.ArgumentMatchers
+import org.mockito.Mockito
+import org.robolectric.RobolectricTestRunner
+import org.robolectric.RuntimeEnvironment
+
+/**
+ * Unit tests for the [PinnedTileViewModel].
+ */
+@RunWith(RobolectricTestRunner::class)
+class PinnedTileViewModelTest {
```
Baron-Severin: This is more of an integration test between the repo and VM, which is fine, but if we mocked out the repo we could run this file without Robolectric (which would mean these tests would run _**much**_ more quickly).

app/src/test/java/org/mozilla/tv/firefox/pinnedtile/PinnedTileViewModelTest.kt
```diff
@@ -0,0 +1,83 @@
...
+            Assert.assertEquals(10, it!!.size)
+        })
+
+        pinnedTileViewModel.getTileList().observeForever(observerSpy)
+
+        Mockito.verify(observerSpy, Mockito.times(1)).onChanged(ArgumentMatchers.any())
+    }
+
+    @Test
+    fun `WHEN repo emits an updated list after add THEN view model should emit an updated list`() {
```
Baron-Severin: (if the repo is mocked out / Robolectric removed, ignore this comment)

The name of this test doesn't fully reflect what it's testing.  Suggest updating to `WHEN a new pinned tile is added to repo THEN ...`, to reflect that this is testing how both classes integrate.

Baron-Severin: This applies to the other tests as well.

app/src/test/java/org/mozilla/tv/firefox/pinnedtile/PinnedTileViewModelTest.kt
Baron-Severin: That said, there are advantages to integration tests as well.  This isn't something that necessarily needs to be changed, but we should be careful about using too much Robolectric, since it has a pretty dramatic effect on how long our tests take to run.

app/src/main/java/org/mozilla/tv/firefox/pinnedtile/PinnedTileRepo.kt
```diff
@@ -51,6 +60,7 @@ class PinnedTileRepo(private val applicationContext: Application) {
         if (screenshot != null) {
             PinnedTileScreenshotStore.saveAsync(applicationContext, newPinnedTile.id, screenshot)
         }
+        ++customTilesSize
```
Baron-Severin: Unless I'm mistaken, there is a potential edge case here.

```
-> customTilesSize == 10, _pinnedTiles.value.size == 10
add 'www.example.com'
-> customTilesSize == 11, _pinnedTiles.value.size == 11
add 'www.example.com'
-> customTilesSize == 12, _pinnedTiles.value.size == 11
```

psymoon: Good catch

app/src/main/java/org/mozilla/tv/firefox/pinnedtile/PinnedTileRepo.kt
liuche: Agreed - though it looks like you're using pinnedTiles in the model. In that case, can you make this return type `LiveData` rather than `MutableLiveData` to prevent any future callers from modifying the list from this reference?

app/src/main/java/org/mozilla/tv/firefox/pinnedtile/PinnedTileRepo.kt
psymoon: Actually, this should be mutable because `blackList.add(tileToRemove.id)` when removing a bundled pinned tile

app/src/main/java/org/mozilla/tv/firefox/pinnedtile/PinnedTileRepo.kt
psymoon: I will create an issue about this so we can address it better (optimization)

PR #1354, opened 2018-10-24
Closes #1327: Add BasicNavigationTest & flesh out Robots 🤖

### LINKED ISSUE #1327: [1364] #1: Implement BasicNavigationTest
See #1287, the issue to create a plan, for motivations.

### What / Requirements
- Identify how the tests should be structured (Robot pattern?)
- ^ If so, implement robots
- Create basic navigation test

- Tests pass (or maybe not :d !) pre-refactor

@athomasmoz FYI that I'm working on this during the architecture work day so I'm adding it to the current sprint. We can discuss next steps of the test plan during triage or sprint planning.

mcomella:
### Pull Request checklist

  - This **is a test**!


app/src/androidTest/java/org/mozilla/tv/firefox/helpers/Matchers.kt
```diff
@@ -0,0 +1,20 @@
...
+import android.support.test.espresso.matcher.ViewMatchers.isEnabled as espressoIsEnabled
+import android.view.View
+import org.hamcrest.CoreMatchers.not
+import org.hamcrest.Matcher
+
+/**
+ * The [android.support.test.espresso.matcher.ViewMatchers.isEnabled] function that can also handle
+ * disabled state through the boolean argument.
+ */
+fun isEnabled(isEnabled: Boolean): Matcher<View> = if (isEnabled) {
```
Baron-Severin: Optional nit: I think this would be slightly cleaner using a when expression.  Currently the `if (isEnabled)` is indented really far away from the rest of the function.

```kotlin
fun isEnabled(isEnabled: Boolean): Matcher<View> = when {
    isEnabled -> espressoIsEnabled()
    else -> not(espressoIsEnabled())
}
```

app/src/androidTest/java/org/mozilla/tv/firefox/helpers/ext/ViewInteraction.kt
```diff
@@ -10,7 +10,7 @@ import android.support.test.espresso.assertion.ViewAssertions.matches
 import android.support.test.espresso.matcher.ViewMatchers.isEnabled
 import org.hamcrest.CoreMatchers.not
 
-fun ViewInteraction.click() = this.perform(ViewActions.click())!!
+fun ViewInteraction.click(): ViewInteraction = this.perform(ViewActions.click())
```
Baron-Severin: Personal preference, but I prefer double bangs for converting from platform types.  That conversion is dangerous, and `!!` communicates that to the reader.

app/src/androidTest/java/org/mozilla/tv/firefox/ui/ClearSessionTest.kt
```diff
@@ -42,46 +44,33 @@ class ClearSessionTest {
...
-
-            openMenu()
-
-            assertCanGoBack(true)
-
-            navigateToPage(endpoints[1])
-
-            openMenu()
+        }.enterUrlAndEnterToBrowser(endpoints[0]) {
+        }.openOverlay {
```
Baron-Severin: I find this test a little difficult to read as it currently stands, but after playing around with it I haven't found any formatting that I like better.  I think we should get a third opinion on this file before we merge this in.

app/src/androidTest/java/org/mozilla/tv/firefox/helpers/MockServerHelper.kt
```diff
@@ -24,3 +30,32 @@ object MockServerHelper {
...
     }
 }
+
+/**
+ * A [MockWebServer] [Dispatcher] that will return Android assets in the body of requests.
+ *
+ * If the dispatcher is unable to read a requested asset, it will fail the test by throwing an
+ * Exception on the main thread.
+ *
+ * @sample [org.mozilla.tv.firefox.ui.BasicNavigationTest.whatever]
```
Baron-Severin: Was `.whatever` a placeholder that should be updated?

app/src/androidTest/assets/pages/basic_nav3.html
```diff
@@ -0,0 +1,10 @@
+<html>
+<head>
+    <script>
+        document.title = Math.random(); // any unique identifier per page load.
```
Baron-Severin: This is effectively going to work as expected, but still bothers me a bit.  Could we change this to `new Date().getTime()`?

app/src/androidTest/java/org/mozilla/tv/firefox/ui/BasicNavigationTest.kt
```diff
@@ -141,4 +142,5 @@ class BasicNavigationTest {
             assertNotEquals(preReloadTitle, getPageTitle())
         }
     }
+    /* ktlint-enable no-blank-line-before-rbrace */
```
Baron-Severin: Does this work if we put the suppression at the top of the file, but nothing at the bottom?  It would be really easy to miss this closing comment.

mcomella:
@npark-mozilla @AaronMT As I've alluded to, we want to switch to using the Robot pattern for our tests. I wrote [an Architecture Decision Record](https://github.com/mozilla-mobile/firefox-tv/blob/master/docs/architecture/adr-0002-robot-pattern.md) that will hopefully explain our motivations and provide an introduction to the pattern.

Additionally, in this PR I completed the first UI test using the new Robot pattern: BasicNavigationTest. Please look it over and let me know how it fits in with your expectations and if you have any concerns! If something is unsatisfactory, let's see if we can do something about it.

Baron-Severin:
@mcomella I see that Taskcluster is failing here, we may need to pull in master to get the recent Gradle fixes.

app/src/androidTest/java/org/mozilla/tv/firefox/ui/BasicNavigationTest.kt
```diff
@@ -0,0 +1,146 @@
...
+        }.closeToBrowser {
+            assertTestContent(BODY_TEXT[2]) // different on each reload. Use date? Verify date different?
+
+            // We want to ensure the page is actually reloaded after reload is pressed. This
+            // particular page's title changes each time the page is reloaded so we verify the
+            // current title is different.
+            assertNotEquals(preReloadTitle, getPageTitle())
+        }
+    }
+    /* ktlint-enable no-blank-line-before-rbrace */
```
AaronMT: This test can also replace the former PageLoadTest.java (feel free to delete it).

app/src/androidTest/assets/pages/basic_nav3.html
npark-mozilla: +1 for getTime().  Forgot much about random() use, but I guess with Java we don't have to set the random seed anymore?

app/src/androidTest/java/org/mozilla/tv/firefox/helpers/Matchers.kt
mcomella: Interesting distinction: I agree.

app/src/androidTest/java/org/mozilla/tv/firefox/helpers/ext/ViewInteraction.kt
mcomella: I hadn't considered that perspective before. However, it does goes against a deliberate Kotlin language decision to not treat all platform types as nullable, given what an inconvenient effect this has on developers. Then again, I also disagree with that decision because developers always need to think about it if a valuable is nullable or not and now the system just isn't helping us.

The reason for this change was that I originally found `!!` to be unclear: "why is `!!` necessary? What is this method returning? Oh, it's nullable? Okay, I guess that makes sense. If we wanted a non-nullable `ViewInteraction` to be returned, wouldn't it be clearer just to say that?" I thought the explicit type would help clarify that.

Maybe it's better to be doubly explicit? With a return type and !!? Then again, the final `!!` pattern could be learned. The explicit returns types are always redundant (and never go out of date) but they prevent you from making later mistakes when your expression return type is not what you expect.

I'll do both in this PR but let me know if you disagree.

app/src/androidTest/java/org/mozilla/tv/firefox/helpers/MockServerHelper.kt
mcomella: This is fixed in a later commit.

app/src/androidTest/java/org/mozilla/tv/firefox/ui/BasicNavigationTest.kt
mcomella: Yep. I thought putting it directly above the test though was helpful for people copying this test format to know how to suppress and that it's more clear *why* this was necessary. I'll remove the bottom comments but leave the top one above the tests.

app/src/androidTest/java/org/mozilla/tv/firefox/ui/ClearSessionTest.kt
mcomella: @liuche Any thoughts on making this more readable? If so, please file a new bug to follow-up.

app/src/androidTest/java/org/mozilla/tv/firefox/ui/BasicNavigationTest.kt
mcomella: I wanted to leave in a single test in that handles network page load so I left PageLoadTest around. However, it would be a good idea to make it clear that's why it still exists (@Baron-Severin can you make this part of the non-critical UI test follow-up? Or maybe we should just fix that now...).

app/src/androidTest/java/org/mozilla/tv/firefox/helpers/ext/ViewInteraction.kt
Baron-Severin: I think we should give some more thought to this and see if we come to a better conclusion, but both seems a good immediate solution.  I like having explicit return types on most functions anyway.

app/src/androidTest/java/org/mozilla/tv/firefox/ui/ClearSessionTest.kt
liuche: Looking at the test in-tree, I think it looks fine - each block is a separate step, and the indentation makes it easy to understand the higher-level flow, so I'm fine with it.

PR #1383, opened 2018-10-24
1382 wrap livedata map

Baron-Severin:
### Pull Request checklist

### DIFF
--- a/app/src/androidTest/java/org/mozilla/tv/firefox/helpers/ext/LiveDataMapTest.kt
+++ b/app/src/androidTest/java/org/mozilla/tv/firefox/helpers/ext/LiveDataMapTest.kt
@@ -5,6 +5,7 @@
 package org.mozilla.tv.firefox.helpers.ext
 
 import android.arch.lifecycle.MutableLiveData
+import android.arch.lifecycle.Observer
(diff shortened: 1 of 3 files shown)


app/src/androidTest/java/org/mozilla/tv/firefox/helpers/ext/LiveDataMapTest.kt
```diff
@@ -0,0 +1,63 @@
...
+    @Test
+    fun WHEN_maps_are_chained_THEN_emit_expected_values() {
+        activityTestRule.runOnUiThread {
+            val mapped = source
+                .map { it * 5 }
+                .map { it.toString() }
+            mapped.observeForever { assertEquals("5", it) }
+            source.value = 1
+        }
+    }
```
liuche: One nit: is it worth testing if the source gets updated, that the map persists and emits what is expected? e.g. source.value = 1, then adding another observer that expects when the source.value = 2.

Baron-Severin: Good call, I'll add that in

Issue #1370, opened 2018-10-24
[1396] #5: Implement test for data upload enable toggle


mcomella:
Users expect the data toggle to work so we respect user's choices. We want to ensure:
- The data upload toggle will switch on and off and the state will persist when you leave and return to settings (and may even the app); we may need to enable data upload in debug builds to fix this
- The data upload toggle's state matches the internal representation

mcomella:
May need to fix #384 first.

athomasmoz:
Not included: testing the functionality of the settings that are being toggled.

mcomella:
For the internal representation, I was thinking we'd check the global flag "is data upload enabled" flag rather than validating we're not uploading: unit tests (presumably) will cover the latter.

liuche:
Linking this to #1396 so we don't lose it.

PR #1389, opened 2018-10-25
Fix data toggle

liuche:
The problem here is that there is a hack for accessibility (which can go away once we fix Settings and make it less ugly), that made the code really confusing, and changed whether the click listener was working.

### Pull Request checklist

Since we're changing settings later, and @mcomella says there are lots of test changes landing on master, I'll just fix this in #1370

### DIFF
--- a/app/src/main/java/org/mozilla/tv/firefox/settings/SettingsFragment.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/settings/SettingsFragment.kt
@@ -34,11 +34,15 @@ class SettingsFragment : Fragment() {
 
         ic_lock.setImageResource(R.drawable.mozac_ic_lock)
         telemetryButton.isChecked = DataUploadPreference.isEnabled(context)
-        telemetryView.setOnClickListener { _ ->
+        val dataPreferenceClickListener = { _: View ->
             val newTelemetryState = !DataUploadPreference.isEnabled(context)
             DataUploadPreference.setIsEnabled(context, newTelemetryState)
             telemetryButton.isChecked = newTelemetryState
         }
+        // Due to accessibility hack for #293, either of these views could be unfocusable, so we
+        // need to set the click listener on both.
+        telemetryView.setOnClickListener(dataPreferenceClickListener)
+        telemetryButton.setOnClickListener(dataPreferenceClickListener)
 
         deleteButton.setOnClickListener { _ ->
             val builder1 = AlertDialog.Builder(activity)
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/mozilla/tv/firefox/settings/SettingsFragment.kt
```diff
@@ -34,11 +34,15 @@ class SettingsFragment : Fragment() {
...
-        telemetryView.setOnClickListener { _ ->
+        val dataPreferenceClickListener = { _: View ->
             val newTelemetryState = !DataUploadPreference.isEnabled(context)
             DataUploadPreference.setIsEnabled(context, newTelemetryState)
             telemetryButton.isChecked = newTelemetryState
         }
+        // Due to accessibility hack for #293, either of these views could be unfocusable, so we
+        // need to set the click listener on both.
+        telemetryView.setOnClickListener(dataPreferenceClickListener)
+        telemetryButton.setOnClickListener(dataPreferenceClickListener)
```
Baron-Severin: Nit: so, one of these views is displayed when Talk Back is enabled and the other when it's not?  If that understanding is correct, could we rename them to better describe that?

app/src/main/java/org/mozilla/tv/firefox/settings/SettingsFragment.kt
```diff
@@ -94,18 +98,20 @@ class SettingsFragment : Fragment() {
...
         // adding code to fix.
         val context = context ?: return
-        val shouldFocus = context.isVoiceViewEnabled()
-        telemetryView.isFocusable = shouldFocus
-        if (!shouldFocus) { telemetryView.clearFocus() }
+        val shouldFocusButton = context.isVoiceViewEnabled()
+        telemetryView.isFocusable = !shouldFocusButton
+        // Clear focus so that focus passes to child telemetryButton view.
+        if (shouldFocusButton) {
+            telemetryView.clearFocus()
```
Baron-Severin: Optional nit: this might be more clear if we broke it into a few inner functions, even if it introduces a little bit of code repetition.  Maybe something like this?

```kotlin
fun focusView() = telemetryView.isFocusable = true
fun focusButton() {
  // Clear focus so that focus passes to child telemetryButton view.
  telemetryView.isFocusable = false
  telemetryView.clearFocus
}
when {
  shouldFocusButton -> focusButton()
  else -> focusView()
}
```

app/src/main/java/org/mozilla/tv/firefox/settings/SettingsFragment.kt
liuche: Good point that this is confusing - both are always visible, but `telemetryView` is actually the container. It's not that they're not "displayed", it's that one is focusable and the other is not, so that VoiceView can select the correct one to read aloud. They'll both be visible, all the time.

I'll update the naming to make this more clear, and the comment too.

Issue #1286, opened 2018-10-12
Hints for Youtube Fling


liuche:
### Why/User Benefit/User Problem
After we add #1284 users will be able to cast youtube videos from the Youtube app (but not generic links), but if they didn't already know about this feature, how will they learn about it?

### What / Requirements
Provide a hint so users know about this capability.

@cartertrout do you know what kind of branding requirements we'd have, if this feature is only to highlight "Youtube casting"? Could you suggest on getting Youtube branding, casting icon, etc?

### Acceptance Criteria (how do I know when I’m done?)
* Hint added so users know about casting to Firefox with Youtube.
* Probe added to keep track of how many times the hint is displayed

aminalhazwani:
⚠️ **IN PROGRESS** ⚠️ 

I was able to cast from the YouTube mobile and tablet app to the Fire TV via Silk. I wasn't able to cast from Laptop/Chrome to the Fire TV. For now we could only focus on casting from mobile/tablet YouTube app to Firefox.

If you tap the cast icon in the homescreen of the YouTube app YouTube tells you that there are no videos in the queue:

<img width="2022" alt="image" src="https://user-images.githubusercontent.com/1721875/47025481-de849b00-d163-11e8-9400-2ba18b9b7ed6.png">

then Silk opens and it opens youtube.com and no video is played.

If you do the same process with a Chromecast/Android TV this message is displayed on the TV:

![img_0649](https://user-images.githubusercontent.com/1721875/47025850-8a2deb00-d164-11e8-9598-02945f811181.jpg)

First question **1) can we display a custom message if users simply tap the cast icon in the YouTube app homescreen?**

In order to notify users that this feature is available we can display a small banner/message. The banner will tell users that they can cast videos from their mobile or tablet YouTube app. They would simply have to tap the cast icon on their device and select their Fire TV from the list.

### When can we show this banner

- on launch after update/install, similarly as we do for the Turbo Mode onboarding
  - this is unfortunately not very contextual, user action-related, we might want to avoid this for such specific feature
- people launching youtube.com/tv from the homescreen
  - **question 2) does it go against our policy?**
- people typing youtube.com from the homescreen
  - **question 2) does it go against our policy?**
- people typing on youtube.com/tv (youtube uses a custom keyboard)
  - **question 2) does it go against our policy?**
- people coming from YouTube.com app
  - **question 3) do we know when external apps open Firefox?**

### How can this banner look

[...]

PR #1434, opened 2018-10-31
Issue #1286 - Strings for casting hint.

liuche:
Adding strings to be exported.
### Pull Request checklist

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -252,4 +252,11 @@ be temporary, and you can try again later.</li>
     <string name="notification_request_desktop_site">Requested desktop version</string>
     <!-- Notification displayed when non-desktop version of the site is requested-->
     <string name="notification_request_non_desktop_site">Returned to non-desktop version</string>
+
+    <!-- Notification displayed to introduce a "YouTube casting" feature (similar to "broadcasting")
+         Users who have the YouTube app on their phone or tablet can use it to play YouTube videos directly on their tv.
+
+         %s will be replaced with the icon that matches the button in the YouTube app for this feature.
+         Clicking on that button in the YouTube app will show YouTube's instructions on how to cast. -->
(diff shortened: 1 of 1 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -252,4 +252,11 @@ be temporary, and you can try again later.</li>
     <string name="notification_request_desktop_site">Requested desktop version</string>
     <!-- Notification displayed when non-desktop version of the site is requested-->
     <string name="notification_request_non_desktop_site">Returned to non-desktop version</string>
+
+    <!-- Notification displayed to introduce a "YouTube casting" feature (similar to "broadcasting")
+    Users who have the YouTube app on their phone or tablet can use it to play YouTube videos directly on their tv.
```
mcomella: nit: we usually align the comment blocks so this line should move in

app/src/main/res/values/strings.xml
```diff
@@ -252,4 +252,11 @@ be temporary, and you can try again later.</li>
...
+
+    %s will be replaced with the icon that matches the button in the YouTube app for this feature.
```
mcomella: Can we inline an icon? Does it need to be a unicode character to do that?

app/src/main/res/values/strings.xml
```diff
@@ -252,4 +252,11 @@ be temporary, and you can try again later.</li>
...
+    Clicking on that button in the YouTube app will show YouTube's instructions on how to cast. -->
+    <string name="youtube_casting_onboarding">Cast YouTube videos from your phone or tablet. Open YouTube on your device and tap %s to get started.</string>
```
mcomella: I wonder if this name should be more specific? There can be multiple onboarding/hint experiences. Maybe `youtube_casting_onboarding_toast`

app/src/main/res/values/strings.xml
liuche: We can use ImageSpan for that.

PR #1451, opened 2018-11-02
Closes #1391: Upgrade Robolectric to 4.0.1

### LINKED ISSUE #1391: Upgrade to robolectric 4.0
It came out yesterday: http://robolectric.org/blog/2018/10/25/robolectric-4-0/

And we don't really want to be an 4.0 alpha.

psymoon:
See: http://robolectric.org/migrating/

- `RuntimeEnvironment.application` deprecated => Use [ApplicationProvider](https://developer.android.com/reference/androidx/test/core/app/ApplicationProvider) instead
- Shadow getters are not allowed anymore; use explicit `shadowof(something).set(var)`

### Pull Request checklist


app/src/test/java/org/mozilla/tv/firefox/ext/ContextTest.kt
```diff
@@ -20,35 +20,35 @@ class ContextTest {
...
     fun startUp() {
-        context = RuntimeEnvironment.application
+        context = ApplicationProvider.getApplicationContext()
         accessibilityManager = context!!.getAccessibilityManager()
     }
 
     @Test
     fun `isVoiceViewEnabled() returns true when TouchExploration enabled`() {
-        shadowOf(accessibilityManager).isTouchExplorationEnabled = true
+        shadowOf(accessibilityManager).setTouchExplorationEnabled(true)
```
Baron-Severin: Nit: Was this an automated change?  We should prefer property access syntax here.

app/src/test/java/org/mozilla/tv/firefox/pinnedtile/PinnedTileScreenshotStoreUnitTest.kt
```diff
@@ -25,19 +26,20 @@ import java.util.UUID
...
+        appContext = ApplicationProvider.getApplicationContext()
+        appContext.filesDir.listFiles().forEach { it.deleteRecursively() }
         uuid = UUID.randomUUID()
     }
 
     /** Assumes [PinnedTileScreenshotStore.getFileForUUID] works correctly. */
     @Test
     fun testSaveAsyncDoesNotOverwrite() = runBlocking {
-        val context = RuntimeEnvironment.application
+        val context = appContext
```
Baron-Severin: Nit: I think we should prefer to use `appContext` in these tests to redeclaring the variable in each.

app/src/test/java/org/mozilla/tv/firefox/ext/ContextTest.kt
liuche: I think this is an API requirement, from when I tried this earlier, because I don't think `AccessibilityManager` actually has a property for this for some reason, maybe the getter/setter isn't a simple value assign.

PR #1457, opened 2018-11-02
Fixes #1453: fixes KNPE in voice media session

### LINKED ISSUE #1453: VideoVoiceCommandMediaSession Kotlin NPE
See https://sentry.prod.mozaws.net/operations/firefox-tv/issues/4898695/?query=is:unresolved

This was first seen on 10/31/18, and we're seeing around 3k events per day (good to fix, but not critical).  

Kotlin NPE is thrown when a null value is passed as an argument to a non-nullable param, and in this case it seems to be caused by a double-bang in `WebRenderFragment#onDestroyView` that calls into `VideoVoiceCommandMediaSession#onDestroyWebView`.  Passing a nullable `EngineView` and checking before doing cleanup should fix the problem (no need to clean up a nulled out view).

Baron-Severin:
For RCA, see issue

It seems strange to me that the WebView has occasionally been nulled out before `WebRenderFragment#onDestroyView` is called, but if the view has already been destroyed I think we can safey skip cleaning it up.

### Pull Request checklist

Trivial change
Not user-facing

### DIFF
--- a/app/src/main/java/org/mozilla/tv/firefox/webrender/WebRenderFragment.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/webrender/WebRenderFragment.kt
@@ -276,7 +276,7 @@ class WebRenderFragment : EngineViewLifecycleFragment(), Session.Observer {
     }
 
     override fun onDestroyView() {
-        mediaSessionHolder?.videoVoiceCommandMediaSession?.onDestroyWebView(webView!!, session)
+        mediaSessionHolder?.videoVoiceCommandMediaSession?.onDestroyWebView(webView, session)
 
         super.onDestroyView()
 
--- a/app/src/main/java/org/mozilla/tv/firefox/webrender/VideoVoiceCommandMediaSession.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/webrender/VideoVoiceCommandMediaSession.kt
@@ -120,8 +120,8 @@ class VideoVoiceCommandMediaSession @UiThread constructor(
         this.sessionIsLoadingObserver = sessionIsLoadingObserver
     }
 
-    fun onDestroyWebView(webView: EngineView, session: Session) {
-        webView.removeJavascriptInterface(JS_INTERFACE_IDENTIFIER)
+    fun onDestroyWebView(webView: EngineView?, session: Session) {
+        webView?.removeJavascriptInterface(JS_INTERFACE_IDENTIFIER)
         this.webView = null
 
         session.unregister(sessionIsLoadingObserver!!)


app/src/main/java/org/mozilla/tv/firefox/webrender/WebRenderFragment.kt
```diff
@@ -276,7 +276,7 @@ class WebRenderFragment : EngineViewLifecycleFragment(), Session.Observer {
     }
 
     override fun onDestroyView() {
-        mediaSessionHolder?.videoVoiceCommandMediaSession?.onDestroyWebView(webView!!, session)
```
mcomella: Why would `webView` ever be null here? I'm concerned that, because we're using the webviewcache, the webView will persist outside of BrowserFragment with the javascript interface still attached.

mcomella: > I'm, not sure, but it seems that it very rarely can be.

Well, how often does this crash? :P

When possible, we need to make sure we know why a crash is happening because the assumptions that underlie a crash could be causing other problems too or the fixes can create other side effects in the code. I think switching out !! for ?. seems like an easy fix to avoid a crash but it ignores this understanding. It also may put users into an undefined state (and when these undefined states compound, it becomes harder to fix problems later).

> What if we did the JS removal in the cache instead of in WebRenderFragment?

That seems like a better fix: the interface is supposed to exist for the life of the WebView anyway.

Baron-Severin: > Well, how often does this crash? :P

~3k / day.  Enough that I want to stop the crash, but few enough that I don't want to go too deep into the weeds.

> That seems like a better fix: the interface is supposed to exist for the life of the WebView anyway

Sounds good to me, I'll make the change.

mcomella:
> ~3k / day. Enough that I want to stop the crash, but few enough that I
don't want to go too deep into the weeds.

But is stopping the crash better for users than the alternative undefined
state? What state do users get into if they don't crash? It's often hard to
say. And if they get into a worse state, how easy is it for them to
restart? Fire TV makes it really hard to restart apps so maybe we want to
be more careful than mobile or desktop about getting into undefined states.

I think the solution we ended up with, "Move this over here because it's
more correct", is a good one for this problem but I don't think
side-stepping the NPE with `?.` would have been a good one (which is why
I'm making all these explanations, sorry if it's inappropriate!).

app/src/main/java/org/mozilla/tv/firefox/webrender/WebRenderFragment.kt
Baron-Severin: If we move removal into the `WebViewCache`, we should move the addition there too.  That's going to drag in a lot of surrounding context.  Getting a reference to `VideoVoiceCommandMediaSession`in `WebViewCache` looks like it should be doable, but I don't want to get into what the implications of that could be right now.  Unless you're opposed, I'm inclined to close this PR and push it back for now.

Baron-Severin:
Not at all, that all seems perfectly reasonable.

mcomella:
> Unless you're opposed, I'm inclined to close this PR and push it back for
now.

3k crashes a day doesn't seem very large to me: I'm fine with this.

PR #1458, opened 2018-11-03
Fixes #1408: megatile focus on fresh install

### LINKED ISSUE #1408: On a fresh install, Pocket megatile is skipped when scrolling down
## Steps to reproduce
Uninstall Firefox, this only happens on a fresh install
Install Firefox, and click through onboarding screens
Scroll down from nav bar

### Expected behavior
Pocket megatile gets highlighted

### Actual behavior
Nav highlight skips down to home tiles. Scrolling back up works again.

### Device information

* Fire TV device: Cube
* Firefox version: master 1213142

Baron-Severin:
Root cause:
Previously, whether or not to set focus on the megatile was determined by checking view visibility. When onboarding is up, views that we would normally expect to be visible are not. The solution to this is to check based on VM state, rather than view state

### Pull Request checklist

### DIFF
--- a/app/src/main/java/org/mozilla/tv/firefox/FirefoxApplication.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/FirefoxApplication.kt
@@ -9,7 +9,6 @@ import android.preference.PreferenceManager
 import android.support.annotation.VisibleForTesting
 import android.webkit.WebSettings
 import org.mozilla.tv.firefox.components.locale.LocaleAwareApplication
-import org.mozilla.tv.firefox.components.locale.LocaleManager
 import org.mozilla.tv.firefox.components.search.SearchEngineManager
 import org.mozilla.tv.firefox.webrender.VisibilityLifeCycleCallback
 import org.mozilla.tv.firefox.telemetry.TelemetryIntegration
@@ -43,7 +42,7 @@ open class FirefoxApplication : LocaleAwareApplication() {
 
         PreferenceManager.setDefaultValues(this, R.xml.settings, false)
 
-        initServiceLocator()
+        serviceLocator = ServiceLocator(this)
 
         enableStrictMode()
 
@@ -54,10 +53,6 @@ open class FirefoxApplication : LocaleAwareApplication() {
         visibilityLifeCycleCallback = VisibilityLifeCycleCallback(this).also {
             registerActivityLifecycleCallbacks(it)
         }
-    }
-
-    private fun initServiceLocator() {
-        serviceLocator = ServiceLocator(this)
     }
 
     private fun enableStrictMode() {
--- a/app/src/main/java/org/mozilla/tv/firefox/FirefoxApplication.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/FirefoxApplication.kt
@@ -9,6 +9,7 @@ import android.preference.PreferenceManager
 import android.support.annotation.VisibleForTesting
 import android.webkit.WebSettings
 import org.mozilla.tv.firefox.components.locale.LocaleAwareApplication
+import org.mozilla.tv.firefox.components.locale.LocaleManager
 import org.mozilla.tv.firefox.components.search.SearchEngineManager
 import org.mozilla.tv.firefox.webrender.VisibilityLifeCycleCallback
 import org.mozilla.tv.firefox.telemetry.TelemetryIntegration
@@ -42,7 +43,7 @@ open class FirefoxApplication : LocaleAwareApplication() {
 
         PreferenceManager.setDefaultValues(this, R.xml.settings, false)
 
-        serviceLocator = ServiceLocator(this)
+        initServiceLocator()
 
         enableStrictMode()
 
@@ -53,6 +54,11 @@ open class FirefoxApplication : LocaleAwareApplication() {
         visibilityLifeCycleCallback = VisibilityLifeCycleCallback(this).also {
             registerActivityLifecycleCallbacks(it)
         }
+    }
+
+    private fun initServiceLocator() {
+        val getLanguage = { LocaleManager.getInstance().getCurrentLanguage(this) }
+        serviceLocator = ServiceLocator(this, getLanguage)
     }
 
     private fun enableStrictMode() {
--- a/app/src/main/java/org/mozilla/tv/firefox/FirefoxApplication.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/FirefoxApplication.kt
@@ -57,8 +57,7 @@ open class FirefoxApplication : LocaleAwareApplication() {
     }
 
     private fun initServiceLocator() {
-        val getLanguage = { LocaleManager.getInstance().getCurrentLanguage(this) }
-        serviceLocator = ServiceLocator(this, getLanguage)
+        serviceLocator = ServiceLocator(this)
     }
 
     private fun enableStrictMode() {
--- a/app/src/main/java/org/mozilla/tv/firefox/ViewModelFactory.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/ViewModelFactory.kt
@@ -25,7 +25,7 @@ import org.mozilla.tv.firefox.utils.ServiceLocator
 class ViewModelFactory(private val serviceLocator: ServiceLocator, private val app: Application) :
     ViewModelProvider.Factory {
 
-    private val getLanguage = { LocaleManager.getInstance().getCurrentLanguage(app) }
+    private val getIsEnglishLocale = { LocaleManager.getInstance().currentLanguageIsEnglish(app) }
 
     @Suppress("UNCHECKED_CAST")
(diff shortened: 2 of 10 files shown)


app/src/main/java/org/mozilla/tv/firefox/FirefoxApplication.kt
```diff
@@ -55,6 +56,11 @@ open class FirefoxApplication : LocaleAwareApplication() {
         }
     }
 
+    private fun initServiceLocator() {
+        val getLanguage = { LocaleManager.getInstance().getCurrentLanguage(this) }
```
psymoon: Can this be declared inside `ServiceLocator` (which already has access to `applicationContext`) or even `ViewModelFactory`? It's an unnecessary dependency for the factory class when it's only used for `PocketViewModel`

app/src/main/java/org/mozilla/tv/firefox/ViewModelFactory.kt
```diff
@@ -20,17 +20,24 @@ import org.mozilla.tv.firefox.utils.ServiceLocator
  * val myViewModel = ViewModelProviders.of(this, factory).get(ExampleViewModel::class.java)
  * ```
  */
-class ViewModelFactory(private val serviceLocator: ServiceLocator) : ViewModelProvider.Factory {
+class ViewModelFactory(private val serviceLocator: ServiceLocator, private val getCurrentLanguage: () -> String) :
```
psymoon: See above

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/BrowserNavigationOverlay.kt
```diff
@@ -266,12 +253,17 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
...
+                    showMegaTileError()
+                }
                 is PocketViewModel.State.Feed -> {
+                    pocketVideoMegaTileView.visibility = View.VISIBLE
                     pocketVideoMegaTileView.setContent(state.feed)
                 }
+                is PocketViewModel.State.NotDisplayed -> pocketVideoMegaTileView.visibility = View.GONE
                 null -> return@Observer
-            }
+            }.forceExhaustive
```
psymoon: Is `forceExhaustive` necessary if there is an else statement?

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/BrowserNavigationOverlay.kt
```diff
@@ -372,8 +364,8 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
         }
 
         navUrlInput.nextFocusDownId = when {
-            pocketVideosContainer.isEffectivelyVisible -> R.id.pocketVideoMegaTileView
-            megaTileTryAgainButton.isEffectivelyVisible -> R.id.megaTileTryAgainButton
+            pocketViewModel.state.value is PocketViewModel.State.Feed -> R.id.pocketVideoMegaTileView
```
psymoon: nit: why is one reference equality expression use `is` while the other use `===`?

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketVideoFragment.kt
```diff
@@ -50,10 +50,11 @@ class PocketVideoFragment : Fragment() {
         val viewModel = ViewModelProviders.of(this, factory).get(PocketViewModel::class.java)
 
         viewModel.state.observe(viewLifecycleOwner, Observer<PocketViewModel.State> { state ->
-            state ?: return@Observer
             when (state) {
                 is PocketViewModel.State.Error -> { /* TODO: #769: display error screen */ }
                 is PocketViewModel.State.Feed -> adapter.setVideos(state.feed)
+                is PocketViewModel.State.NotDisplayed -> { }
```
psymoon: What does this do?

app/src/main/java/org/mozilla/tv/firefox/utils/ServiceLocator.kt
```diff
@@ -44,13 +44,13 @@ import org.mozilla.tv.firefox.pocket.PocketVideoRepo
  *   open val telemetry: TelemetryInterface by lazy { SentryWrapper() }
  *   ```
  */
-open class ServiceLocator(val app: Application) {
+open class ServiceLocator(val app: Application, getCurrentLanguage: () -> String) {
```
psymoon: See above about having redundant dependencies. There could be a better resolution while satisfying mockability in testing

app/src/main/java/org/mozilla/tv/firefox/components/locale/LocaleManager.java
```diff
@@ -283,14 +283,14 @@ public Locale getCurrentLocale(@NonNull Context context) {
         return currentLocale = Locales.parseLocaleCode(current);
     }
 
-    public boolean currentLanguageIsEnglish(@NonNull Context context) {
+    @NonNull
+    public String getCurrentLanguage(@NonNull Context context) {
```
psymoon: Why was this changed? It looks like it's only used in `PocketViewModel` to check for English anyway

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketViewModel.kt
```diff
@@ -42,14 +47,21 @@ class PocketViewModel(private val pocketVideoRepo: PocketVideoRepo, pocketRepoCa
...
+    val state = pocketRepoCache.feedState
+        .map { repoState ->
+            when (repoState) {
+                is PocketVideoRepo.FeedState.Loading -> State.Feed(loadingPlaceholders)
+                is PocketVideoRepo.FeedState.LoadComplete -> State.Feed(repoState.videos)
+                is PocketVideoRepo.FeedState.NoAPIKey -> State.Feed(noKeyPlaceholders)
+                is PocketVideoRepo.FeedState.FetchFailed -> State.Error
+            }
+        }.map {
+            if (getCurrentLanguage().toLowerCase().startsWith("en")) it
```
psymoon: See above for `getCurrentLanguage()`

app/src/main/java/org/mozilla/tv/firefox/FirefoxApplication.kt
Baron-Severin: Good point.  I'll pass application into the factory and grab this in there.

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/BrowserNavigationOverlay.kt
Baron-Severin: IMO, `forceExhaustive` should be on every when expression that doesn't return a value (it was a mistake on my part not to have it here in the first place).  Having an else statement will cause an otherwise non-exhaustive when expression to compile, even with `forceExhaustive` on, but for that reason I prefer to avoid else branches when possible.  The advantage here is that, if we ever add a new type to PocketViewModel.State this will force us to update the code everywhere it's used.

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/BrowserNavigationOverlay.kt
Baron-Severin: Yeah, that's annoying and I'm not a big fan of it.  `Feed` is a data class (it carries video information) whereas `Error` is an object, so testing its type and testing referential equality are the same.  Testing `Error` with an `is` results in a linting error.  I'm not sure what the best way to approach this is, tbh.

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketVideoFragment.kt
Baron-Severin: If our state is `notDisplayed` then we shouldn't even be able to reach this fragment.  I can add a comment in explaining that.  Do you think we should throw an exception there as well?

app/src/main/java/org/mozilla/tv/firefox/components/locale/LocaleManager.java
Baron-Severin: Tbh, I don't remember.  Maybe I thought it would be useful in other contexts as well?  I can change it back.

app/src/main/java/org/mozilla/tv/firefox/FirefoxApplication.kt
```diff
@@ -55,6 +55,10 @@ open class FirefoxApplication : LocaleAwareApplication() {
         }
     }
 
+    private fun initServiceLocator() {
```
psymoon: nit: Do you think this method is necessary?

Baron-Severin: Not anymore.  I'll pull it out.

PR #1478, opened 2018-11-07
Closes #1239: Add Pocket author

### LINKED ISSUE #1239: Display Author on Pocket Videos

Right now, when you click into the recommendations, we display the _platform_ attribution (namely, YouTube) directly under the video tile. So even though we have a lineup of videos from, say, Bloomberg, Vox, National Geographic, PBS, HowStuffWorks, and TedEd, they all display as being from YouTube -- since that's simply where the vast majority of them are sourced. The videos I curate from BBC are the only ones that attribute to anything other than YouTube.

Is there any way to make the case that we should show attribution there to the creator of the video, rather than the platform? To credit Bloomberg or National Geographic for instance in the display?

### What / Requirements

* Show author when available

* When available, the youtube author is displayed
* When not available, just youtube is displayed

### Notes from Pocket

The API change is live now: for YouTube videos the authors field is now populated with the YouTube channel. For example:
   ```"authors":{
      "96611324":{
         "author_id":"96611324",
         "name":"Collider Videos",
         "url":"http:\/\/www.youtube.com\/channel\/UCWvMmm_sSdgALpo1Ci4WvtQ"
      }
   }```

dnarcese:
### Pull Request checklist

### DIFF
--- a/app/src/main/java/org/mozilla/tv/firefox/pocket/PocketVideoFragment.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/pocket/PocketVideoFragment.kt
@@ -21,7 +21,6 @@ import android.widget.TextView
 import kotlinx.android.synthetic.main.fragment_pocket_video.*
 import kotlinx.android.synthetic.main.fragment_pocket_video.view.*
 import mozilla.components.browser.session.Session
-import org.json.JSONObject
 import org.mozilla.tv.firefox.R
 import org.mozilla.tv.firefox.ext.forceExhaustive
 import org.mozilla.tv.firefox.ext.serviceLocator
@@ -117,11 +116,8 @@ private class PocketVideoAdapter(
 
         titleView.text = item.title
         PicassoWrapper.client.load(item.thumbnailURL).into(videoThumbnailView)
-        val authorsJSON = JSONObject(item.authors)
-        for (x in authorsJSON.keys()) {
-            val authorObject = JSONObject(authorsJSON[x].toString())
-            domainView.text = authorObject.get("name").toString()
-        }
+
+        domainView.text = item.authors
     }
 
     private fun updateForFocusState(holder: PocketVideoViewHolder, isFocused: Boolean) {
--- a/app/src/main/java/org/mozilla/tv/firefox/pocket/PocketVideoFragment.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/pocket/PocketVideoFragment.kt
@@ -117,9 +117,9 @@ private class PocketVideoAdapter(
 
         titleView.text = item.title
         PicassoWrapper.client.load(item.thumbnailURL).into(videoThumbnailView)
-
-        for (x in item.authors.keys()) {
-            val authorObject = JSONObject(item.authors[x].toString())
+        val authorsJSON = JSONObject(item.authors)
+        for (x in authorsJSON.keys()) {
+            val authorObject = JSONObject(authorsJSON[x].toString())
             domainView.text = authorObject.get("name").toString()
         }
     }
--- a/app/src/main/java/org/mozilla/tv/firefox/pocket/PocketVideoFragment.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/pocket/PocketVideoFragment.kt
@@ -9,7 +9,6 @@ import android.arch.lifecycle.ViewModelProviders
 import android.content.Context
 import android.content.Intent
 import android.os.Bundle
-import android.os.StrictMode
 import android.support.v4.app.Fragment
 import android.support.v4.app.FragmentManager
 import android.support.v4.content.ContextCompat
@@ -22,16 +21,13 @@ import android.widget.TextView
 import kotlinx.android.synthetic.main.fragment_pocket_video.*
 import kotlinx.android.synthetic.main.fragment_pocket_video.view.*
 import mozilla.components.browser.session.Session
+import org.json.JSONObject
 import org.mozilla.tv.firefox.R
 import org.mozilla.tv.firefox.ext.forceExhaustive
-import org.mozilla.tv.firefox.ext.resetAfter
 import org.mozilla.tv.firefox.ext.serviceLocator
 import org.mozilla.tv.firefox.ext.updateLayoutParams
 import org.mozilla.tv.firefox.telemetry.TelemetryIntegration
-import org.mozilla.tv.firefox.utils.FormattedDomain
 import org.mozilla.tv.firefox.utils.PicassoWrapper
-import java.net.URI
-import java.net.URISyntaxException
 
 /** A feed of Pocket videos. */
 class PocketVideoFragment : Fragment() {
@@ -122,23 +118,9 @@ private class PocketVideoAdapter(
         titleView.text = item.title
         PicassoWrapper.client.load(item.thumbnailURL).into(videoThumbnailView)
 
-        @Suppress("TooGenericExceptionCaught") // See below.
-        val itemURI = try {
-            URI(item.url)
-        } catch (e: Exception) { // Apparently Kotlin doesn't have multi-catch.
-            when (e) {
-                is URISyntaxException, is NullPointerException -> null
-                else -> throw e
-            }
-        }
-        domainView.text = if (itemURI == null) {
-            item.url
-        } else {
-            // The first time this method is called ever, it may block until the file is cached on disk.
-            // We pre-cache on startup so I'm hoping this isn't an issue.
-            StrictMode.allowThreadDiskReads().resetAfter {
-                FormattedDomain.format(itemView.context, itemURI, false, 0)
-            }
+        for (x in item.authors.keys()) {
+            val authorObject = JSONObject(item.authors[x].toString())
+            domainView.text = authorObject.get("name").toString()
         }
     }
 
--- a/app/src/main/java/org/mozilla/tv/firefox/pocket/PocketVideoFragment.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/pocket/PocketVideoFragment.kt
@@ -9,6 +9,7 @@ import android.arch.lifecycle.ViewModelProviders
 import android.content.Context
 import android.content.Intent
 import android.os.Bundle
+import android.os.StrictMode
 import android.support.v4.app.Fragment
 import android.support.v4.app.FragmentManager
 import android.support.v4.content.ContextCompat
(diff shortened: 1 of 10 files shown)


app/src/main/java/org/mozilla/tv/firefox/pocket/PocketVideoFragment.kt
```diff
@@ -122,23 +118,9 @@ private class PocketVideoAdapter(
...
-            item.url
-        } else {
-            // The first time this method is called ever, it may block until the file is cached on disk.
-            // We pre-cache on startup so I'm hoping this isn't an issue.
-            StrictMode.allowThreadDiskReads().resetAfter {
-                FormattedDomain.format(itemView.context, itemURI, false, 0)
-            }
+        for (x in item.authors.keys()) {
+            val authorObject = JSONObject(item.authors[x].toString())
+            domainView.text = authorObject.get("name").toString()
```
Baron-Severin: (I know we weren't doing some of this before, but since we're changing this code this is a good time to improve it)

I have two problems with this approach.
1) we aren't handling failures (what if the `JSONObject` doesn't have a "name" field?)
2) we are doing JSON parsing in the middle of our app, rather than at the edge

1: Anything that goes across the network should be considered unsafe.  We can't be sure that these objects will come in at all, or that any particular fields won't be null or malformed.  We want to handle these cases wherever we parse this JSON (in this case, handling probably means providing an empty string on failure).

2: As much as possible, we want our domain objects to have exactly what we need, no more and no less.  This isn't always going to match the shape of the JSON (we might drop unnecessary fields, merge two together, add timestamps, etc).  The goal is to be able to just use these without ceremony in the middle of our app(e.g., `domainView.text = video.author`).  It's the parser's job to figure out how to convert from JSON to what we actually need.  In this case, the parser can break out what's in the JSON, figure out which author value to use, and provide an empty string on failure.

One big advantage to this is that it simplifies a lot of how our app functions.  If you handle all the dirtiness of IO on the edge of the application, nothing else has to care about the nullability and errors and so on.  They just use the data.  That prevents us from having to duplicate the same messy logic over and over again, and also makes it really easy to find where problems are happening.  If parsing is failing, check the parser.

TL;DR
- [ ] `FeedItem.Video.author` should be a non-nullable String
- [ ] Getting the author and name fields should be done by the PocketVideoParser
- [ ] We should handle the case where those fields don't exist (probably by passing an empty String)

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketViewModel.kt
```diff
@@ -34,7 +34,8 @@ class PocketViewModel(private val pocketVideoRepo: PocketVideoRepo, pocketRepoCa
             val title: String,
             val url: String,
             val thumbnailURL: String,
-            val popularitySortId: Int
+            val popularitySortId: Int,
+            val authors: JSONObject
```
Baron-Severin: I think this field should be a `String`.

See comment in PocketVideoAdapter

dnarcese: This should already be a string in the last commit.

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketVideoParser.kt
```diff
@@ -21,9 +21,24 @@ object PocketVideoParser {
...
     }
+
+    private fun getAuthorName(jsonObj: JSONObject): String {
+        var authorName = ""
+        try {
+            val authorsJSON = JSONObject(jsonObj.getString("authors"))
+
+            for (x in authorsJSON.keys()) {
+                val authorObject = JSONObject(authorsJSON[x].toString())
+                authorName = authorObject.get("name").toString()
```
Baron-Severin: If the JSON contains multiple authors, this will always return the last one.  Is that intended?  If so, we should rename the field on `FeedItem.Video` from `authors` -> `author`.  If not, `authorName` should probably be a list (and will need to be renamed).

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketVideoParser.kt
```diff
@@ -21,9 +21,24 @@ object PocketVideoParser {
...
+            }
```
Baron-Severin: I had forgotten about this, but the requirements for this story specify our fallback behavior.  If 'author' is not present, we should show "YouTube".  So we should use that as a default instead of an empty string.

We can break DEFAULT_AUTHOR out into a file level constant.

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketVideoParser.kt
```diff
@@ -21,9 +21,24 @@ object PocketVideoParser {
...
         null
     }
+
+    private fun getAuthorName(jsonObj: JSONObject): String {
+        var authorName = ""
+        try {
+            val authorsJSON = JSONObject(jsonObj.getString("authors"))
+
+            for (x in authorsJSON.keys()) {
+                val authorObject = JSONObject(authorsJSON[x].toString())
```
Baron-Severin: We can make these get().toString() calls a little more idiomatic.  `JSONObject[x].toString()` -> `JSONObject.getString(x)`

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketVideoParser.kt
```diff
@@ -21,9 +21,24 @@ object PocketVideoParser {
...
+                authorName = authorObject.get("name").toString()
+            }
+        } catch (e: JSONException) {
```
Baron-Severin: We should add a test case that verifies that JSON without an author field ends up with author == "YouTube"

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketVideoParser.kt
dnarcese: Good point. Although, I don't think we can always show YouTube since some videos are coming from other places, like BBC. The platform would have to be determined based on the video, as in previous functionality.
I'll add that back as the fallback behaviour.

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketVideoParser.kt
dnarcese: That was my intended behaviour, because I assumed there wouldn't be more than one author. But it would be more correct to handle multiple authors, I'll change that.

dnarcese: Verifying in the issue if it is possible for there to be more than one author.

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketVideoParser.kt
dnarcese: Is there a way to add this behaviour in the parser? I can't figure it out since I need the url field of the video and context to pass to `FormattedDomain#format`. Is it okay to leave it in the fragment?

Baron-Severin: Formatting the domain doesn't seem like it should be a responsibility of the parser, so we don't need to pass it in there.  It looks like it's currently being done in the RecyclerView.Adapter, which isn't great either.  If it's out of scope for this change we can put a TODO to fix it, but this is a transformation that should really be handled by the view model, imo.

app/src/test/java/org/mozilla/tv/firefox/pocket/PocketEndpointTest.kt
```diff
@@ -36,7 +36,7 @@ class PocketEndpointTest {
                 url = "https://www.youtube.com/watch?v=GHZ7-kq3GDQ",
                 thumbnailURL = "https://img-getpocket.cdn.mozilla.net/direct?url=http%3A%2F%2Fimg.youtube.com%2Fvi%2FGHZ7-kq3GDQ%2Fmaxresdefault.jpg&resize=w450",
                 popularitySortId = 17,
-                authors = "Seeker"
+                authors = "youtube"
```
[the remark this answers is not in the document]
dnarcese: The domain field is formatted as `"domain": "youtube.com"`. So yes the raw data will always be lowercase (but then we display it as uppercase).

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketVideoParser.kt
```diff
@@ -28,17 +28,17 @@ object PocketVideoParser {
...
+            var authors = mutableListOf<String?>()
             val authorsJSON = JSONObject(jsonObj.getString("authors"))
-            // This assumes there is only one author per video
+            // TODO: verify if multiple authors are possible and what the format of authors object is
             for (x in authorsJSON.keys()) {
                 val authorObject = JSONObject(authorsJSON[x].toString())
-                authorName = authorObject.getString("name")
+                authors.add(authorObject.getString("name"))
             }
+            authors.joinToString(",")
```
Baron-Severin: This is going to result in "YOUTUBE,BBC" when I assume we want "YOUTUBE, BBC".  The default behavior will give us the second value, suggest calling the method without args.

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketVideoParser.kt
```diff
@@ -28,17 +28,17 @@ object PocketVideoParser {
     }
 
     private fun getAuthorName(jsonObj: JSONObject): String {
-        var authorName = ""
-        try {
+        return try {
+            var authors = mutableListOf<String?>()
```
Baron-Severin: authors can be a val now

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketVideoParser.kt
```diff
@@ -28,17 +28,17 @@ object PocketVideoParser {
...
             val authorsJSON = JSONObject(jsonObj.getString("authors"))
-            // This assumes there is only one author per video
+            // TODO: verify if multiple authors are possible and what the format of authors object is
             for (x in authorsJSON.keys()) {
                 val authorObject = JSONObject(authorsJSON[x].toString())
-                authorName = authorObject.getString("name")
+                authors.add(authorObject.getString("name"))
```
Baron-Severin: `authors += authorObject.getString("name")` would be more idiomatic

PR #1481, opened 2018-11-07
Fixes #1408: pocket VM not instantiated crash

### LINKED ISSUE #1408: On a fresh install, Pocket megatile is skipped when scrolling down

Baron-Severin:
RCA
- BrowserNavigationOverlay.setVisibility could, with certain arguments, call a method that required PocketViewModel
- These arguments were not passed during initial instantiation (which is why this crash is not normally encountered on startup)
- PocketViewModel was being created several lines down from a call to BrowserNavigationOverlay.setVisibility

### Pull Request checklist

Trivial change


app/src/main/java/org/mozilla/tv/firefox/webrender/WebRenderFragment.kt
```diff
@@ -238,6 +238,7 @@ class WebRenderFragment : EngineViewLifecycleFragment(), Session.Observer {
             // FIXME: Need [WebRenderFragment] as lifeCycle owner until NavOverlayFragment breakout
             pinnedTileViewModel = this@WebRenderFragment.pinnedTileViewModel
             lifeCycleOwner = this@WebRenderFragment.viewLifecycleOwner
+            observeForMegaTile(this@WebRenderFragment)
             initTiles()
```
liuche: Nit: can we rename this to `initPinnedTiles`? It's a little confusing because we're observing on the Megatile before the init for the tiles, but in reality they're not related tiles.

app/src/main/java/org/mozilla/tv/firefox/webrender/WebRenderFragment.kt
```diff
@@ -256,8 +257,6 @@ class WebRenderFragment : EngineViewLifecycleFragment(), Session.Observer {
             }
 
             registerForContextMenu(browserOverlay.tileContainer)
```
liuche: Maybe some of these later lines that rely on the view having been created should be moved into `onViewCreated`, so we don't have any more assumptions on whether the views are there or not.

app/src/main/java/org/mozilla/tv/firefox/webrender/WebRenderFragment.kt
Baron-Severin: Good call.  It's safe to move it one line higher too, which makes a lot more sense.  I should have done that in the first place.

app/src/main/java/org/mozilla/tv/firefox/webrender/WebRenderFragment.kt
Baron-Severin: To be fair the views are available at this point, the problem behind this crash was the VM not being available.  It's probably a good policy to set up VMs before modifying view state, but I think we're safe keeping them in the same method.  I don't feel strongly about it though, I can move them if you'd prefer.

liuche: Hmm, I guess that's a good point but the VMs do back our views - I guess technically `onViewCreated` is more of an Android lifecycle thing. But, still a slight preference to move them to `onViewCreated` so we can clearly separate modifying view state from setting it up.

PR #1494, opened 2018-11-09
No issue: add fullscreen pinch string for FFES.

mcomella:
Localization for Echo Show is not set up yet so we still have to land
new strings in the Fire TV repository. :(

@athomasmoz When do you think we'd cut and release Firefox for Echo Show v1.2? This string would be used to introduce pinch-to-exit-fullscreen (as a workaround for the video not providing controls) but if we can't translate the string in time, I can removed the introduction, include the feature, and we can re-add the introduction in v1.3.

No tests or CL: just a string in a different product.

### Pull Request checklist

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -261,4 +261,9 @@ be temporary, and you can try again later.</li>
     %s will be replaced with the icon that matches the button in the YouTube app for this feature.
     Clicking on that button in the YouTube app will show YouTube's instructions on how to cast. -->
     <string name="youtube_casting_onboarding_toast">Cast YouTube videos from your phone or tablet. Open YouTube on your device and tap %s to get started.</string>
+
+    <!-- Text displayed in a toast when entering fullscreen mode: "pinch" refers
+         to the two-finger zoom out pinch gesture. This string will *only* be used on
+         Firefox for Echo Show where there is a touchscreen. -->
+    <string name="fullscreen_toast_pinch_to_exit">Pinch to exit fullscreen mode</string>
 </resources>


liuche:
@Delphine what do you think about this string? It's clearly Echo Show-only, and I'm just wondering if it will be confusing to localizers - or if we can get an l10n repo set up for the other project.

liuche:
@mcomella what's the date that you'll be using this string by?

Delphine:
@liuche: I have to understand what the timeline for Echo Show looks like, I'll be replying to mcomella's email today about that. If you need this string right away, we'll have to put it under Fire TV repo for now. 
I think the localization comment is clear enough, and I'll have to kick-off comms with localizers about Echo Show soon anyways (as soon as we have a better idea of the details surrounding it).

Also, CC @Pike and @flod so they are aware that we'll be adding Echo Show to Pontoon soon. We were using strings from Fire TV until now.

Pike:
I recommend that we keep using a single repository and a single strings.xml for now. Assuming that Echo Show is just a build variant of Fx TV, that is.

Once we're moving towards using the Pontoon native Android support, we can decide to use a dedicated strings-show.xml or so, for strings that are only used on Echo Show. I wouldn't want to confuse android2po with that right now.

Delphine:
@Pike How would we localize though, if this doesn't get into Pontoon?
From what I understand, this string is needed before we have strings.xml support in Pontoon - but waiting to hear back on timeline just in case

Pike:
If it goes into strings.xml, it goes into pontoon through the usual process via android2po, I assume.

Once we change that process, we can talk about different files for Echo vs TV.

Delphine:
This means we have to set up the project in Pontoon. I'll have to coordinate with Flod and see when I/we have enough bandwidth to do this. Waiting to hear back on timeline for the string in the meantime. We can decide how to proceed once we know that

liuche:
@Pike Echo Show originally was a fork of Firefox TV but now it isn't a build flavor of Firefox TV - we've been manually copying strings from one repo to the other, so ideally we can get another project set up for Echo Show.

For now, I'll merge this in and @mcomella can manually copy the strings to the Echo Show repo once they get localized.

liuche:
@mcomella string export is here: https://github.com/mozilla-l10n/firefoxtv-l10n/pull/17
I'll merge imports as @MickeyMoz makes them, and you can manually pull the strings out when they're merged.

Ideally we can get rid of those strings, but I'll wait until we have the new l10n repo for FFES set up @Pike .

mcomella:
On Mon, Nov 12, 2018 at 1:15 PM liuche <notifications@github.com> wrote:

> @mcomella <https://github.com/mcomella> what's the date that you'll be
> using this string by?
>

Maybe the week of Dec. 10th: this feature missed the latest release (v1.2)
and I'll be more-or-less OOO for three weeks (that includes one week for
all hands) so it's unlikely we'll create another release before then.

PR #1514, opened 2018-11-15
1392 toolbar mvvm

Baron-Severin:
### Pull Request checklist

Changes are not user facing


Baron-Severin:
@dnarcese I'm not sure I follow, could you provide repro steps?

dnarcese:
STR
1) Visit a site
2) Clear data in settings
3) Visit site 1
4) Visit site 2
Back button in toolbar should be enabled, but it's not.

Baron-Severin:
@dnarcese good catch!  I added a regression test for this, working on fixing it.

Baron-Severin:
Leaving some notes regarding the regression spotted by @dnarcese.

WebView.WebBackForwardList#getCurrentIndex is behaving differently before and after clearing data.

#### STR:
- Navigate to Google Search tile
- Navigate to Pinterest tile

#### If from fresh startup:
BackForwardList will look like ["www.google.com", "www.pinterest.com", null]
BackForwardList currentIndex will be 2

#### If data has been cleared:
BackForwardList will look like ["www.google.com", "www.pinterest.com", null]
BackForwardList currentIndex will be 1

I'm working on figuring out why the index initially points to the null value.

app/src/main/java/org/mozilla/tv/firefox/toolbar/ToolbarViewModel.kt
```diff
@@ -0,0 +1,183 @@
...
+        sendOverlayClickTelemetry(NavigationEvent.PIN_ACTION, pinChecked = !pinChecked)
+
+        if (pinChecked) {
+            pinnedTileRepo.removePinnedTile(url)
+            _events.value = Consumable.from(BrowserNavigationOverlay.Action.ShowTopToast(R.string.notification_unpinned_site))
+        } else {
+            pinnedTileRepo.addPinnedTile(url, sessionRepo.currentURLScreenshot())
+            _events.value = Consumable.from(BrowserNavigationOverlay.Action.ShowTopToast(R.string.notification_pinned_site))
+        }
+        setOverlayVisible(false)
```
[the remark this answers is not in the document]
aminalhazwani: Yeah, that is correct @psymoon . But I see where your doubt is coming from. Unfortunately we can only make an assumption here but it would be interesting to see how many users stay on the website after pinning and how many re-open the overlay. And base on that either change or leave the behavior as it is.

app/src/main/java/org/mozilla/tv/firefox/utils/TurboMode.kt
```diff
@@ -4,30 +4,41 @@
...
-    fun setEnabled(context: Context, enabled: Boolean) {
-        val settings = Settings.getInstance(context)
+    fun setEnabled(enabled: Boolean)
+}
+
+/**
+ * Implementation of [TurboMode] that sets and queries for Turbo state using an
+ * [Application] instance.
+ */
+class ProdTurboMode(private val app: Application) : TurboMode {
```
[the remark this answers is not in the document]
Baron-Severin: Yeah, do you think I should rename it, or just add that note to the docs?

app/src/main/java/org/mozilla/tv/firefox/toolbar/ToolbarViewModel.kt
```diff
@@ -0,0 +1,183 @@
...
+        sessionRepo.state.value?.currentUrl?.let { if (!it.isEqualToHomepage()) setOverlayVisible(false) }
+    }
+
+    @UiThread
+    fun desktopModeButtonClicked() {
+        val previouslyChecked = state.value?.desktopModeChecked ?: return
+
+        sendOverlayClickTelemetry(NavigationEvent.DESKTOP_MODE, desktopModeChecked = !previouslyChecked)
+
+        sessionRepo.setDesktopMode(!previouslyChecked)
```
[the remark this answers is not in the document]
Baron-Severin: Hitting the button toggles the state (so `!previouslyChecked`), but I wanted to hold on to that value for the telemetry.

app/src/main/java/org/mozilla/tv/firefox/toolbar/ToolbarViewModel.kt
```diff
@@ -0,0 +1,183 @@
...
+import org.mozilla.tv.firefox.ext.LiveDataHelper
+import org.mozilla.tv.firefox.ext.toUri
+import org.mozilla.tv.firefox.navigationoverlay.BrowserNavigationOverlay
+import org.mozilla.tv.firefox.navigationoverlay.NavigationEvent
+import org.mozilla.tv.firefox.telemetry.TelemetryIntegration
+import org.mozilla.tv.firefox.utils.AppConstants
+import org.mozilla.tv.firefox.utils.TurboMode
+import org.mozilla.tv.firefox.utils.UrlUtils
+
+open class ToolbarViewModel(
```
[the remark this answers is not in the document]
Baron-Severin: Not really necessary in the current implementation.  I'll remove it.

app/src/main/java/org/mozilla/tv/firefox/ViewModelFactory.kt
```diff
@@ -22,8 +24,11 @@ import org.mozilla.tv.firefox.utils.ServiceLocator
  * val myViewModel = ViewModelProviders.of(this, factory).get(ExampleViewModel::class.java)
  * ```
  */
-class ViewModelFactory(private val serviceLocator: ServiceLocator, private val app: Application) :
-    ViewModelProvider.Factory {
+class ViewModelFactory(
+    private val serviceLocator: ServiceLocator,
+    private val app: Application,
+    private val turboMode: TurboMode
```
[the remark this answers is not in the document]
Baron-Severin: > It appears that ServiceLocator already holds a reference of TurboMode. Why is this necessary?

Oversight on my part, I'll remove it.

> IMO, TurboMode should be inside SessionRepo.

I think that's a fair point.  Yeah, I'll rewrite the relevent code.

app/src/main/java/org/mozilla/tv/firefox/session/SessionRepo.kt
```diff
@@ -0,0 +1,79 @@
...
+import android.support.annotation.AnyThread
+import mozilla.components.browser.session.Session
+import mozilla.components.browser.session.SessionManager
+import mozilla.components.feature.session.SessionUseCases
+import org.mozilla.tv.firefox.ext.postIfNew
+
+/**
+ * Repository that is responsible for storing state related to the browser.
+ */
+class SessionRepo(private val sessionManager: SessionManager, private val sessionUseCases: SessionUseCases) {
```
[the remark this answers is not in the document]
Baron-Severin: It mostly has, but `WebRenderFragment` still uses those callbacks to update focus.  Once we have a focus repo, we can pull them out.

There is also some Pocket logic in `WebRenderFragment#onUrlChanged` that should have been removed in the Pocket refactor.  I will open a separate issue to pull it out and verify that the behavior is still correct.

app/src/main/java/org/mozilla/tv/firefox/toolbar/ToolbarViewModel.kt
```diff
@@ -0,0 +1,183 @@
...
+
+    init {
+        disableDesktopModeWhenHostChanges()
+        showOverlayWhenUrlIsHome()
+    }
+
+    @UiThread
+    fun backButtonClicked() {
+        sessionRepo.exitFullScreenIfPossibleAndBack()
+        setOverlayVisible(false)
```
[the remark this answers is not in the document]
Baron-Severin: That's a fair point.  If it's alright with you, I'll keep the current implementation (to at least pull it out of the fragment), but add TODOs to move this into the overlay VM.

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/BrowserNavigationOverlay.kt
```diff
@@ -155,7 +149,6 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
 
         initMegaTile()
         setupUrlInput()
```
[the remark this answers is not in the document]
Baron-Severin: I think overlay focus will probably be more easily managed in one centralized place, but other UrlBar state could definitely live here.  I'm inclined to merge what we have to unblock and add an issue for this, but it sounds like a good change.

app/src/main/java/org/mozilla/tv/firefox/session/SessionRepo.kt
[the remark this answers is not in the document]
psymoon: Sounds good

app/src/main/java/org/mozilla/tv/firefox/session/SessionObservationManager.kt
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    // Updates the repo whenever the observed session changes
+    private val sessionObserver = object : Session.Observer {
+        override fun onUrlChanged(session: Session, url: String) {
+            sessionRepo.update()
+        }
+
+        override fun onDesktopModeChanged(session: Session, enabled: Boolean) {
+            sessionRepo.update()
```
liuche: I like this, it's simple instead of trying to be extremely efficient and error-prone.

app/src/main/java/org/mozilla/tv/firefox/session/SessionObservationManager.kt
```diff
@@ -0,0 +1,56 @@
...
+        override fun onNavigationStateChanged(session: Session, canGoBack: Boolean, canGoForward: Boolean) {
+            sessionRepo.update()
+        }
+    }
+
+    // Attaches and removes the session observer whenever sessions are added
+    // and removed
+    val sessionManagerObserver = object : SessionManager.Observer {
+        override fun onSessionSelected(session: Session) {
+            session.register(sessionObserver)
```
liuche: It's unclear to me exactly whether a session can a) not be the selectedSession but b) not removed - will we be observing multiple sessions? I guess that's fine bc they don't receive updates (and for us right now, we only have the one session).

app/src/main/java/org/mozilla/tv/firefox/session/SessionObservationManager.kt
```diff
@@ -0,0 +1,56 @@
...
+            sessionRepo.update()
+        }
+
+        override fun onSessionRemoved(session: Session) {
+            session.unregister(sessionObserver)
```
liuche: More assumptions here, that there will be a session created and we won't be caught in some limbo where we have the state of the old session but we're not currently observing a session, but I think that's probably fine.

app/src/main/java/org/mozilla/tv/firefox/session/SessionRepo.kt
```diff
@@ -49,4 +50,11 @@ class SessionRepo(private val sessionManager: SessionManager, private val sessio
     fun setDesktopMode(active: Boolean) = sessionManager.selectedSession?.let { it.desktopMode = active }
 
     fun reload() = sessionUseCases.reload.invoke()
+
+    fun exitFullScreenIfPossibleAndBack() {
```
liuche: This feels a little like a UI interaction, but I guess "fullscreen" is kind of a state as well.

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/BrowserNavigationOverlay.kt
```diff
@@ -297,6 +303,16 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
                 navUrlInput.setText(it.urlBarText)
             }
         })
+
+        toolbarViewModel.events.observe(lifeCycleOwner, Observer {
```
liuche: Also events related to pinning from the tile itself, not just the pin button. I'm thinking that we might want this to eventually belong to the overlay, but need to think more about the hints/notifications that Amin designed into the 4.0 overlay (per channel).

app/src/main/java/org/mozilla/tv/firefox/toolbar/ToolbarViewModel.kt
```diff
@@ -70,15 +72,27 @@ open class ToolbarViewModel(
...
+    fun backButtonClicked() {
+        sessionRepo.exitFullScreenIfPossibleAndBack()
+        closeOverlay()
+    }
 
-    fun forwardButtonClicked() = sessionRepo.goForward()
+    @UiThread
+    fun forwardButtonClicked() {
+        sessionRepo.goForward()
+        closeOverlay()
```
liuche: hm, if every action closes the overlay, I'm wondering if it's necessary to have it in each one? But maybe this is just easier and it's not worth adding another code path to generalize that "close on click" behavior, so I'm fine with it. Just concerned that we might forget to add this with new functionality.

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/BrowserNavigationOverlay.kt
```diff
@@ -267,10 +260,67 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
         })
     }
 
+    fun observeForToolbar(fragment: Fragment) {
+        // TODO remove this
+        // See comment in fun observeForMegaTile
+        val factory = fragment.context!!.serviceLocator.viewModelFactory
```
mcomella: nit: use view context which can't be null?

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/BrowserNavigationOverlay.kt
```diff
@@ -267,10 +260,67 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
...
+        toolbarViewModel = ViewModelProviders.of(fragment, factory).get(ToolbarViewModel::class.java)
+
+        fun updateOverlayButtonState(isEnabled: Boolean, overlayButton: ImageButton) {
+            overlayButton.isEnabled = isEnabled
+            overlayButton.isFocusable = isEnabled
+            overlayButton.alpha =
```
mcomella: (Follow-up?) Should the alpha be defined as part of the "enabled/disabled" state instead?

app/src/main/java/org/mozilla/tv/firefox/session/SessionRepo.kt
```diff
@@ -0,0 +1,89 @@
...
+     * Causes [state] to emit its most recently pushed value. This can be used
+     * to reset UI that has been adjusted by the user (e.g., EditText text)
+     */
+    fun pushCurrentValue() = _state.postValue(_state.value)
+
+    fun loadURL(url: Uri) = session?.let { sessionManager.getEngineSession(it)?.loadUrl(url.toString()) }
+
+    fun setTurboModeEnabled(enabled: Boolean) = turboMode.setEnabled(enabled)
+
+    private val session: Session? get() = sessionManager.selectedSession
```
mcomella: iirc, `get()` definitions should be on a separate line, according to the style guides.

app/src/main/java/org/mozilla/tv/firefox/settings/SettingsFragment.kt
```diff
@@ -55,14 +55,20 @@ class SettingsFragment : Fragment() {
...
             ) { dialog, _ -> with(requireContext()) {
-                    requireWebRenderComponents.engine.deleteData(this)
-                    requireWebRenderComponents.sessionManager.removeAll()
+                    val components = requireWebRenderComponents
+                    components.engine.deleteData(this)
+                    components.sessionManager.removeAll()
                     dialog.cancel()
+                    serviceLocator.webViewCache.doNotPersist(doAfterRecreate = {
+                        // Sessions must be removed prior to MainActivity being recreated, in order for initialization
+                        // logic to be called. However, the URL is set before the new WebView is created, causing it to
```
mcomella: "The URL is set before the new WebView is created, causing it be out of sync with the session."

I don't understand, can you elaborate? e.g. Where is the URL set? Who sets it? What causes it to be out of sync?

app/src/main/java/org/mozilla/tv/firefox/webrender/WebRenderFragment.kt
```diff
@@ -70,7 +65,7 @@ class WebRenderFragment : EngineViewLifecycleFragment(), Session.Observer {
 
     private val mediaSessionHolder get() = activity as MediaSessionHolder? // null when not attached.
 
-    val isUrlEqualToHomepage: Boolean get() = session.url == APP_URL_HOME
+    private val isUrlEqualToHomepage: Boolean get() = session.url == AppConstants.APP_URL_HOME
```
mcomella: This function is redefined elsewhere: can we define it somewhere generic and reduce duplication?

app/src/main/java/org/mozilla/tv/firefox/webrender/WebViewCache.kt
```diff
@@ -64,10 +67,15 @@ class WebViewCache : LifecycleObserver {
...
-            }.also { cachedView = it }
+            }.also {
+                cachedView = it
+                doAfterRecreate?.invoke()
+                doAfterRecreate = null
+            }
         }
 
         cachedView?.removeFromParentIfAble()
+        sessionRepo.backForwardIndexProvider = { cachedView?.currentBackForwardIndex() ?: -1 }
```
mcomella: What does this do? Why is it necessary? Add a comment, probably on the definition in the SessionRepo.

app/src/test/java/org/mozilla/tv/firefox/ext/LiveDataPostIfNewTest.kt
```diff
@@ -0,0 +1,65 @@
...
+import org.junit.Rule
+import org.junit.Test
+import org.mockito.ArgumentMatchers.any
+import org.mockito.Mockito.spy
+import org.mockito.Mockito.times
+import org.mockito.Mockito.verify
+import org.mozilla.tv.firefox.utils.PreventLiveDataMainLooperCrashRule
+
+@Suppress("TestFunctionName")
+class LiveDataPostIfNewTest {
```
mcomella: nit: unit tests are usually named after the class or file name: why is this named after a function?

That being said, in general more encapsulation for unit tests seems reasonable (though it duplicates the fact that the method is called in the test function anyway) but I wonder if this is the way to do it?

app/src/test/java/org/mozilla/tv/firefox/ext/LiveDataPostIfNewTest.kt
```diff
@@ -0,0 +1,65 @@
...
+    fun `GIVEN old value was posted AND a different value was posted WHEN the original value is posted again THEN three total values should have been emitted`() {
+        source.postIfNew(1)
+        source.postIfNew(2)
+        source.postIfNew(1)
+        verify(observerSpy, times(3)).onChanged(any())
+    }
+
+    @Test
+    fun `GIVEN previous state was null WHEN new state is pushed THEN new state should be emitted`() {
+        source.postIfNew(null)
```
mcomella: Can we set the (pre-condition) previous state without the fun under test? That would be a better test.

app/src/test/java/org/mozilla/tv/firefox/ext/LiveDataPostIfNewTest.kt
```diff
@@ -0,0 +1,65 @@
...
+    fun `GIVEN old value was posted WHEN the same value is posted again THEN only one should be emitted`() {
+        source.postIfNew(1)
+        source.postIfNew(1)
+        verify(observerSpy, times(1)).onChanged(any())
+    }
+
+    @Test
+    fun `GIVEN old value is posted WHEN a different value is posted THEN both should be emitted`() {
+        source.postIfNew(1)
+        source.postIfNew(2)
```
mcomella: I feel like we should verify these values too, at least in one test where one value is set and then another is set (we test one value in the final test).

app/src/test/java/org/mozilla/tv/firefox/helpers/TestTurboMode.kt
```diff
@@ -0,0 +1,24 @@
...
+
+import android.arch.lifecycle.LiveData
+import android.arch.lifecycle.MutableLiveData
+import org.mozilla.tv.firefox.utils.TurboMode
+
+/**
+ * An implementation of [TurboMode] that does not persist state between
+ * instances.
+ */
+class TestTurboMode(private var isEnabled: Boolean) : TurboMode {
```
mcomella: This appears to be unused. If so, do we still need the interface and prod implementation distinction?

liuche:
Rebased this in #1573 - so when that lands this should close.

@Baron-Severin just a summary: some nits, and I think @mcomella wanted you to look at his comments and address them, but we decided to land and then fix later.

app/src/main/java/org/mozilla/tv/firefox/utils/ServiceLocator.kt
```diff
@@ -53,12 +55,14 @@ open class ServiceLocator(val app: Application) {
 
     val turboMode: TurboMode by lazy { ProdTurboMode(app) } // TODO make private before merging
     val pocketRepoCache by lazy { PocketRepoCache(pocketRepo).apply { unfreeze() } }
-    val viewModelFactory by lazy { ViewModelFactory(this, app) }
+    val viewModelFactory by lazy { ViewModelFactory(this, app, turboMode) }
```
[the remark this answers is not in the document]
Baron-Severin: Personally, I think a VM depending on two repos is fine (I worry more about keeping repos from depending upon each other).  `ToolbarVM` depends on both `SessionRepo` and `PinnedTileRepo`, for example.

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/BrowserNavigationOverlay.kt
```diff
@@ -195,7 +196,7 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
         pinnedTileViewModel.getTileList().observe(lifeCycleOwner, Observer {
             if (it != null) {
                 tileAdapter.setTiles(it)
-                updateOverlayForCurrentState()
+                updateFocusAndURL()
```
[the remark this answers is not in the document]
Baron-Severin: Agreed.  I'm looking forward to removing this method when we have our new approach to focus handling.

app/src/main/java/org/mozilla/tv/firefox/session/SessionObservationManager.kt
Baron-Severin: My understanding is that we will always have 0 or 1 sessions, but never more.  If that changes in the future I assume a lot of weird edge cases will be dredged up, but you're probably right that it's something we should think about sooner rather than later.

app/src/main/java/org/mozilla/tv/firefox/session/SessionObservationManager.kt
Baron-Severin: 🤔

app/src/main/java/org/mozilla/tv/firefox/session/SessionRepo.kt
Baron-Severin: That's a good point.  Where is the proper place to manage "fullscreen" staus?  I'm honestly not sure.

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/BrowserNavigationOverlay.kt
Baron-Severin: Added a note to move these when possible.

app/src/main/java/org/mozilla/tv/firefox/toolbar/ToolbarViewModel.kt
Baron-Severin: Overlay closing should also be moved to the overlay VM when possible, so this is fairly temporary.  As it is, it should be easy to remove.

app/src/test/java/org/mozilla/tv/firefox/helpers/TestTurboMode.kt
Baron-Severin: I swear this had a point when I wrote it....  Think I removed the class that required it.  I'll pull it out.

app/src/main/java/org/mozilla/tv/firefox/toolbar/ToolbarViewModel.kt
```diff
@@ -0,0 +1,172 @@
...
+        val pinChecked: Boolean,
+        val turboChecked: Boolean,
+        val desktopModeEnabled: Boolean,
+        val desktopModeChecked: Boolean,
+        val urlBarText: String
+    )
+
+    private var previousURLHost: String? = null
+
+    // Values should be pushed to _events using setValue. Two values are set in
```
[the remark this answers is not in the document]
Baron-Severin: I don't like that solution, but it's an improvement over mine.  Done.

app/src/main/java/org/mozilla/tv/firefox/session/SessionRepo.kt
```diff
@@ -0,0 +1,89 @@
...
+
+    fun reload() = sessionUseCases.reload.invoke()
+
+    fun setDesktopMode(active: Boolean) = session?.let { it.desktopMode = active }
+
+    /**
+     * Causes [state] to emit its most recently pushed value. This can be used
+     * to reset UI that has been adjusted by the user (e.g., EditText text)
+     */
+    fun pushCurrentValue() = _state.postValue(_state.value)
```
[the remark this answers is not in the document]
Baron-Severin: I disagree.  This will probably normally be used to reinitialize the UI, but the repo knows nothing about that.

app/src/main/java/org/mozilla/tv/firefox/toolbar/ToolbarViewModel.kt
```diff
@@ -0,0 +1,172 @@
...
+        setOverlayVisible(false)
+    }
+
+    @UiThread
+    fun turboButtonClicked() {
+        val currentUrl = sessionRepo.state.value?.currentUrl
+        val turboModeActive = sessionRepo.state.value?.turboModeActive ?: true
+
+        sessionRepo.setTurboModeEnabled(!turboModeActive)
+        sessionRepo.reload()
```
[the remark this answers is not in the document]
Baron-Severin: I don't know if we ever will, but it doesn't seem unreasonable that we someday might want to.  They don't seem intrinsically connected to me.

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/BrowserNavigationOverlay.kt
Baron-Severin: `Fragment#getView` is nullable though, unless I'm misunderstanding your suggestion.

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/BrowserNavigationOverlay.kt
Baron-Severin: That does seem cleaner, but I'm also wary of hiding styles behind more cascading XML files.  I could go either way on this one.

app/src/main/java/org/mozilla/tv/firefox/session/SessionRepo.kt
Baron-Severin: JetBrains conventions say that single line is acceptable for simple properties, although more complex gets should be multiline.
https://kotlinlang.org/docs/reference/coding-conventions.html#property-formatting

app/src/main/java/org/mozilla/tv/firefox/webrender/WebRenderFragment.kt
Baron-Severin: The other call site with this logic performs the check on a session, so the amount of code sharing is really minimal.  It would also change `session.url == URLs.APP_URL_HOME` to `URLs.isHome(session.url)` (or similar), which I don't think is more readable.

app/src/test/java/org/mozilla/tv/firefox/ext/LiveDataPostIfNewTest.kt
Baron-Severin: Having the helper methods live in one file seems easiest from a discoverability perspective, but each one needs a pretty substantial amount of testing.  Those tests also don't really share a lot of code, so keeping them together would only serve to make them more difficult to read.  This seemed appropriate at the time, but I'm open to alternate suggestions as well.

app/src/main/java/org/mozilla/tv/firefox/settings/SettingsFragment.kt
Baron-Severin: I want to provide context into why this happens, but I'm also concerned about going into so much detail that it will more quickly bitrot.

I'll see if I can strike a better balance on this.

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/BrowserNavigationOverlay.kt
mcomella: `observeForToolbar` gets called from `createView`, which inflates a view. This view has a context you can pass in instead of using a nullable context from the fragment.

PR #1516, opened 2018-11-15
Add UA with version for Pocket API call

liuche:
No changelog change, updates tests w/ new version - setting the header is basic okhttp functionality so writing a test to set and then test doesn't seem worth it.
### Pull Request checklist

### DIFF
--- a/app/src/main/java/org/mozilla/tv/firefox/pocket/PocketEndpoint.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/pocket/PocketEndpoint.kt
@@ -34,12 +34,12 @@ private val GLOBAL_VIDEO_ENDPOINT = Uri.parse("https://getpocket.cdn.mozilla.net
  * The methods of this class call the endpoint directly and does not cache results or rate limit,
  * outside of the network layer (e.g. with OkHttp).
  */
-open class PocketEndpoint {
+open class PocketEndpoint(private val appVersion: String) {
 
     /** @return The global video recommendations or null on error; the list will never be empty. */
     @AnyThread // via PocketEndpointRaw.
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/mozilla/tv/firefox/pocket/PocketEndpoint.kt
```diff
@@ -34,12 +34,13 @@ private val GLOBAL_VIDEO_ENDPOINT = Uri.parse("https://getpocket.cdn.mozilla.net
  * The methods of this class call the endpoint directly and does not cache results or rate limit,
  * outside of the network layer (e.g. with OkHttp).
  */
-open class PocketEndpoint {
+open class PocketEndpoint(version: String) {
+    val appVersion = version
```
Baron-Severin: Nit: ` open class PocketEndpoint(private val appVersion: String)`

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketEndpoint.kt
```diff
@@ -61,9 +62,10 @@ private object PocketEndpointRaw {
 
     /** @return The global video recommendations as a raw JSON str or null on error. */
     @AnyThread // executeAndAwait hands off the request to the OkHttp dispatcher.
-    suspend fun getGlobalVideoRecommendations(): String? {
+    suspend fun getGlobalVideoRecommendations(version: String): String? {
         val req = Request.Builder()
                 .url(GLOBAL_VIDEO_ENDPOINT.toString())
+                .header("User-Agent", "Firefox TV $version")
```
Baron-Severin: Nit: this will probably be easier for them to parse without spaces.  Can we replace them with dashes?

PR #1539, opened 2018-11-21
Convert to Settings MVVM

liuche:
### Pull Request checklist


liuche:
hmm, it seems that somewhere along the way I broke the telemetry UI toggle, adding a wip label to fix it.

app/src/main/java/org/mozilla/tv/firefox/settings/SettingsRepo.kt
```diff
@@ -0,0 +1,41 @@
...
+
+    private fun loadSettingsFromPreferences() {
+        // The first access to shared preferences will require a disk read.
+        val threadPolicy = StrictMode.allowThreadDiskReads()
+        try {
+            val resources = applicationContext.resources
+            _dataCollectionEnabled.value = _sharedPreferences.getBoolean(resources.getString(PREF_KEY_TELEMETRY), true)
+        } finally {
+            StrictMode.setThreadPolicy(threadPolicy)
+        }
```
Baron-Severin: Nit: we have a nice ext function that makes this construct cleaner.  Suggest replacing this with `StrictMode.ThreadPolicy.resetAfter`

app/src/main/java/org/mozilla/tv/firefox/settings/SettingsRepo.kt
```diff
@@ -0,0 +1,41 @@
...
+
+    init {
+        loadSettingsFromPreferences()
+    }
+
+    private fun loadSettingsFromPreferences() {
+        // The first access to shared preferences will require a disk read.
+        val threadPolicy = StrictMode.allowThreadDiskReads()
+        try {
+            val resources = applicationContext.resources
```
Baron-Severin: Nit: looks like we already have resources captured as a member variable.

app/src/main/java/org/mozilla/tv/firefox/settings/SettingsRepo.kt
```diff
@@ -0,0 +1,41 @@
...
+import android.content.SharedPreferences
+import android.os.StrictMode
+import android.preference.PreferenceManager
+import org.mozilla.tv.firefox.R
+
+private const val PREF_KEY_TELEMETRY = R.string.pref_key_telemetry
+
+class SettingsRepo(private val applicationContext: Application) {
+    private val _sharedPreferences: SharedPreferences = PreferenceManager.getDefaultSharedPreferences(applicationContext)
+    private val _resources = applicationContext.resources
```
Baron-Severin: Nit: normally starting underscores are used for backing properties, so suggest removing them here.
https://kotlinlang.org/docs/reference/coding-conventions.html#names-for-backing-properties

app/src/main/java/org/mozilla/tv/firefox/settings/SettingsRepo.kt
```diff
@@ -0,0 +1,41 @@
...
+
+    private val _dataCollectionEnabled = MutableLiveData<Boolean>()
+    val dataCollectionEnabled: LiveData<Boolean> = _dataCollectionEnabled
+
+    init {
+        loadSettingsFromPreferences()
+    }
```
Baron-Severin: This could be problematic if we want to test this repo in the future, since during unit tests we won't want to actually pull from disk.  Suggest moving this to a public setup function that gets called by the service locator.

app/src/main/java/org/mozilla/tv/firefox/settings/SettingsFragment.kt
```diff
@@ -34,11 +35,15 @@ class SettingsFragment : Fragment() {
         val context = view.context
 
         ic_lock.setImageResource(R.drawable.mozac_ic_lock)
-        telemetryButton.isChecked = DataUploadPreference.isEnabled(context)
+        val factory = context.serviceLocator.viewModelFactory
+        val settingsViewModel = ViewModelProviders.of(this, factory).get(SettingsViewModel::class.java)
+        settingsViewModel.dataCollectionEnabled.observe(viewLifecycleOwner, Observer<Boolean> { state ->
+            if (state != null) {
```
Baron-Severin: Optional nit: I personally prefer shorting using `state ?: return` because it flattens out some nesting, but I don't feel strongly about it.

app/src/main/java/org/mozilla/tv/firefox/telemetry/TelemetryFactory.kt
```diff
@@ -39,8 +38,12 @@ object TelemetryFactory {
                         TelemetrySettingsProvider.APP_ID
                 )
                 .setSettingsProvider(TelemetrySettingsProvider(context))
-                .setCollectionEnabled(telemetryEnabled)
-                .setUploadEnabled(telemetryEnabled)
+
+        context.serviceLocator.settingsRepo.dataCollectionEnabled.observeForever { collectionEnabled ->
+            if (collectionEnabled != null) {
+                configuration.setUploadEnabled(collectionEnabled).isCollectionEnabled = collectionEnabled
```
Baron-Severin: Optional nit: this might just be me, but combining the set and property syntax here briefly confused me.  Suggest breaking this into two lines of property syntax.

app/src/main/java/org/mozilla/tv/firefox/session/SessionRepo.kt
```diff
@@ -108,4 +111,15 @@ class SessionRepo(
...
     private val session: Session? get() = sessionManager.selectedSession
+
+    fun clearBrowsingData(context: Context, webViewCache: WebViewCache) {
+        sessionManager.engine.deleteData(context)
+        sessionManager.removeAll()
+        webViewCache.doNotPersist(doAfterRecreate = {
+            // Sessions must be removed prior to MainActivity being recreated, in order for initialization
+            // logic to be called. However, the URL is set before the new WebView is created, causing it to
+            // be out of sync with the Session. Removing all sessions puts them back into sync
+            sessionManager.removeAll()
```
Baron-Severin: This was the old comment that didn't do a great job of explaining what was going on.  Suggest replacing this with the one currently in `SettingsFragment`.  (this one can be deleted)

app/src/main/java/org/mozilla/tv/firefox/settings/SettingsFragment.kt
```diff
@@ -42,6 +45,28 @@ class SettingsFragment : Fragment() {
...
+                        // - For initialization logic to be started, all sessions must already be removed
+                        // - URL is set to app home by default when a new session is created
+                        // - Activity#recreate is posted to the event loop, and executed concurrently
+                        //
+                        // The end result is that, in a naive solution, sessions are removed, URL is set to home, the
+                        // WebView is (eventually) destroyed, and then the new WebView instance has no BackForwardHistory.
+                        // This causes the back button to be disabled until two sites have been visited. After many
+                        // attempted solutions, clearing the session both before and after recreate was the most workable.
+                        //
+                        // This is Bad Code. If you find a way to remove it, please do so.
```
Baron-Severin: IMO, this comment belongs in the `doNotPersist` block in `SessionRepo`.  Calling `Activity#recreate` when session is cleared seems reasonable to me, it's the `SessionRepo` that's doing the strange stuff that needs to be explained / apologized for.

app/src/main/java/org/mozilla/tv/firefox/settings/SettingsFragment.kt
```diff
@@ -89,24 +89,19 @@ class SettingsFragment : Fragment() {
...
+                    .setCancelable(true)
+                    .setPositiveButton(getString(R.string.action_ok)) { dialog, _ ->
+                        with(requireContext()) {
+                            settingsViewModel.clearBrowsingData(this, serviceLocator.webViewCache)
+                            dialog.cancel()
+                        }
+                    }
+                    .setNegativeButton(
+                            getString(R.string.action_cancel)) { dialog, _ -> dialog.cancel() }
+                    .create().show()
```
Baron-Severin: I like this change

app/src/main/java/org/mozilla/tv/firefox/settings/SettingsFragment.kt
```diff
@@ -89,24 +89,19 @@ class SettingsFragment : Fragment() {
...
-            builder1.setNegativeButton(
-                    getString(R.string.action_cancel)) { dialog2, _ -> dialog2.cancel() }
-
-            val alert11 = builder1.create()
-            alert11.show()
+            AlertDialog.Builder(activity)
+                    .setTitle(R.string.settings_cookies_dialog_title)
+                    .setMessage(R.string.settings_cookies_dialog_content2)
+                    .setCancelable(true)
+                    .setPositiveButton(getString(R.string.action_ok)) { dialog, _ ->
```
Baron-Severin: Nit: I think this whole block is indented once more than it should be.

liuche:
Wrote some tests, but having trouble running the test suite, so adding the WIP label until I figure it out.

app/src/main/java/org/mozilla/tv/firefox/settings/SettingsRepo.kt
liuche: Hm, this is also writing to SharedPreferences, so I don't think we can avoid that here. Could we just mock the SharedPreferences from the context to fix this?

Baron-Severin: Good point.  This is using `StrictMode` too, which is another Android class.  I think the cleanest way to get this testable would be to inject a 'persistence manager' interface, that just has a boolean getter and setter.  In prod it would be backed by `SharedPreferences`, in tests it could be an anonymous object.  Unless I'm mistaken that would let us get rid of references to `Application`, `StrictMode`, `Resources`, and `SharedPreferences`, meaning we could run the tests without Robolectric.

From a practical standpoint, that means the repo would be about ten lines long and wouldn't really even require testing.  If we don't expect Settings to grow in the future, it may be simpler to just leave it as is.

app/src/androidTest/java/org/mozilla/tv/firefox/ui/ToggleDataCollectionTest.kt
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    @Test
+    fun WHEN_data_collection_button_is_toggled_THEN_data_collection_matches_button_state() {
+        navigationOverlay {
+        }.openSettings {
+            assertDataCollectionButtonState(IS_TELEMETRY_ENABLED_DEFAULT)
+
+            toggleDataCollectionButton()
+            assert(settingsRepo.dataCollectionEnabled.value != null)
```
Baron-Severin: Nit: JUnit includes `assertNotNull` and `assertNull` methods that could replace these

Issue #1387, opened 2018-10-25
[1364] #11: Pocket tile + screen loading/error state


mcomella:
It's hard for QA to manually test that all the error states of Pocket are not working correctly so we should write an automated test with dummy data to test for us. In particular, we should verify:
- Loading state appears
- Loading state transitions to error (or success) correctly
- Error states look correct
- Try again button works correctly

Notes:
- We may want to wait until Pocket moves to channels to handle this.
- This may already be covered well enough by the screenshot tests

mcomella:
@Baron-Severin Should this be duped to #1045 or is there going to be more covered in this issue?

Baron-Severin:
This looks more in depth, so I think they should stay separate.  The other one is a smoke test that just makes sure that the happy path (data loads properly, clicking on it takes you to the fragment, selecting a video loads the URL) works properly.

dnarcese:
The best way to do this would be to mock the pocket endpoint and make it so that on first hit it returns an error state and on the next hit it returns a success state. This would require the serviceLocator to be substituted per individual test, so that each test can change the dependencies it needs.

Baron-Severin:
@dnarcese I would normally agree, but we previously set up some code to support this kind of testing.  Any Espresso test should be able to push out mocked `PocketRepo` state via `FirefoxTestApplication#pushPocketRepoState`, and it'll immediately flow out through the app.  So that part of the test should be pretty easy.

dnarcese:
@Baron-Severin I tried it that way, but I think the problem was that Try Again did not function properly. `pushPocketRepoState` created another repo state that Try Again could not update. So clicking Try Again was not updating the UI.

Baron-Severin:
@dnarcese Sounds like the code I wrote to set up tests isn't as flexible as it should be.  Is there a WIP branch with these changes where I can try to improve on it?

dnarcese:
@Baron-Severin https://github.com/dnarcese/firefox-tv/commit/4a8fcdec1271e1c4162c7bf1c327eb67554d658b

Baron-Severin:
@dnarcese just an update, I'm still working on that change.  There are a few things that make it difficult, I opened up https://github.com/mozilla-mobile/firefox-tv/issues/1651 and included more details there.

athomasmoz:
@mcomella is this still relevant with the new way we load pocket videos?

mcomella:
Nope, thanks for catching this.

Issue #1111, opened 2018-09-13
Enable link to expose Pocket megatile error screen


dnarcese:
The screen added in #889 needs a simple UI test.

dnarcese:
A link needs to be added to expose this screen so that it can be tested.
eg navigating to 'firefox:error:pocketConnection' will show the error screen

npark-mozilla:
renaming the title, will create a separate bug for UI test

AaronMT:
On master I attempted to go to _firefox:error:pocketconnection_ (according to  `${APP_URL_PREFIX}error:pocketconnection` but that yielded a WebView error. What's the link?

PR #1548, opened 2018-11-23
Closes #1547: Hide pocket error page on Try Again

### LINKED ISSUE #1369: firefox:error:pocketconnection is only visible once
Visit Firefox:error:pocketconnection, see try again, tap that and attempt to visit Firefox:error:pocketconnection again. Expected to see an error but get a white screen.

This is causing issues with testing.

### LINKED ISSUE #1547: Pocket megatile error page is never hidden
Once the Pocket megatile error page is shown (if no network connection or if Pocket fails to retrieve any videos), it is never hidden again even on clicking Try Again.

dnarcese:
Also fixes #1369.
### Pull Request checklist

The test turned out to be more complicated and take more investigation then initially thought. Going to keep it separate from this PR in issue #1387 

### DIFF
--- a/app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/BrowserNavigationOverlay.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/BrowserNavigationOverlay.kt
@@ -232,7 +232,6 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
                 resources.getString(R.string.pocket_brand_name)) + " " + resources.getString(R.string.pocket_video_feed_reload_button)
 
         megaTileTryAgainButton.setOnClickListener { _ ->
-            hideMegaTileError()
             pocketViewModel.update()
             initMegaTile()
             updateOverlayForCurrentState()
@@ -267,6 +266,7 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
                 is PocketViewModel.State.Feed -> {
                     pocketVideoMegaTileView.visibility = View.VISIBLE
                     pocketVideoMegaTileView.setContent(state.feed)
+                    hideMegaTileError()
                 }
                 is PocketViewModel.State.NotDisplayed -> pocketVideoMegaTileView.visibility = View.GONE
                 null -> return@Observer
--- a/app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/BrowserNavigationOverlay.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/BrowserNavigationOverlay.kt
@@ -234,14 +234,15 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
         megaTileTryAgainButton.setOnClickListener { _ ->
             pocketViewModel.update()
             initMegaTile()
-            updateOverlayForCurrentState()
+            pocketVideoMegaTileView.requestFocus() // TODO remove after Toolbar refactor
         }
+        updateOverlayForCurrentState()
     }
 
     private fun hideMegaTileError() {
         pocketVideosContainer.visibility = View.VISIBLE
         pocketErrorContainer.visibility = View.GONE
-        pocketVideoMegaTileView.requestFocus()
+        updateOverlayForCurrentState()
     }
 
     private fun initMegaTile() {
--- a/app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/BrowserNavigationOverlay.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/BrowserNavigationOverlay.kt
@@ -225,17 +225,24 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
     fun showMegaTileError() {
         pocketVideosContainer.visibility = View.GONE
         pocketErrorContainer.visibility = View.VISIBLE
+
         pocketMegaTileLoadError.text = resources.getString(R.string.pocket_video_feed_failed_to_load,
                 resources.getString(R.string.pocket_brand_name))
         megaTileTryAgainButton.contentDescription = resources.getString(R.string.pocket_video_feed_failed_to_load,
                 resources.getString(R.string.pocket_brand_name)) + " " + resources.getString(R.string.pocket_video_feed_reload_button)
 
         megaTileTryAgainButton.setOnClickListener { _ ->
+            hideMegaTileError()
             pocketViewModel.update()
             initMegaTile()
             updateOverlayForCurrentState()
-            pocketVideoMegaTileView.requestFocus()
         }
+    }
+
+    private fun hideMegaTileError() {
+        pocketVideosContainer.visibility = View.VISIBLE
+        pocketErrorContainer.visibility = View.GONE
+        pocketVideoMegaTileView.requestFocus()
     }
 
     private fun initMegaTile() {
--- a/app/src/main/java/org/mozilla/tv/firefox/webrender/WebRenderFragment.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/webrender/WebRenderFragment.kt
@@ -110,7 +110,7 @@ class WebRenderFragment : EngineViewLifecycleFragment(), Session.Observer {
             APP_URL_HOME -> browserOverlay?.visibility = View.VISIBLE
             APP_URL_POCKET_ERROR -> {
                 browserOverlay?.showMegaTileError()
-                browserOverlay?.visibility = View.VISIBLE
+                setOverlayVisible(true)
             }
             else -> Unit
         }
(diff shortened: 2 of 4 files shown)


app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/BrowserNavigationOverlay.kt
```diff
@@ -225,19 +225,26 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
     fun showMegaTileError() {
```
Baron-Severin: "This PR includes thorough tests or an explanation of why it does not" is checked, but this PR doesn't include any tests or, as far as I can tell, an explanation of why it doesn't.

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/BrowserNavigationOverlay.kt
```diff
@@ -225,19 +225,26 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
...
             initMegaTile()
             updateOverlayForCurrentState()
-            pocketVideoMegaTileView.requestFocus()
         }
     }
 
+    private fun hideMegaTileError() {
+        pocketVideosContainer.visibility = View.VISIBLE
+        pocketErrorContainer.visibility = View.GONE
+        pocketVideoMegaTileView.requestFocus()
```
Baron-Severin: This isn't an ideal solution, because we're duplicating visibility state between the VM and the fragment.  With this change, state in these two places could go out of sync, leading to difficult to find bugs.  Also this puts very similar code into two different places, complicating the process of making further changes.

A better solution would be to move this into the `pocketViewModel.state` observer in  `BrowserNavigationOverlay#observeForMegaTile`, so that our displayed UI would be modeled directly by our VM state.

app/src/main/java/org/mozilla/tv/firefox/webrender/WebRenderFragment.kt
```diff
@@ -131,6 +127,11 @@ class WebRenderFragment : EngineViewLifecycleFragment(), Session.Observer {
     }
 
     override fun onLoadingStateChanged(session: Session, loading: Boolean) {
+        if (!loading && session.url == APP_URL_POCKET_ERROR) {
+            browserOverlay?.showMegaTileError()
+            setOverlayVisible(true)
```
Baron-Severin: This should live in the PocketViewModel, not the view layer.

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/BrowserNavigationOverlay.kt
dnarcese: Oops, #1387 would be the test for this.

Baron-Severin: Once this logic is moved into the VM, it should be testable.  It would definitely be preferable to write tests for this now rather than pushing it off to a new issue, and in general we should try to have at least a few tests in each PR.  Not always possible, but a good goal.  LMK if you need a hand testing this, and I can assist.

app/src/main/java/org/mozilla/tv/firefox/webrender/WebRenderFragment.kt
dnarcese: I'm about confused about this- how would this fit into the VM?

psymoon: @Baron-Severin TODO: overlay refactoring perhaps? not sure about exposing presentation logic in VM

Baron-Severin: If you look at `BrowserNavigationOverlay#observeForMegaTile`, whenever it receives a `PocketViewModel.State.Error` event it displays an error megatile.  If we handle this logic in the view layer, our state will get out of sync.  (e.g., the VM might be in state `Feed`, but we've changed it by hand in the view layer to display an error).

Since we've moved Pocket over to MVVM, we want our VM to match our display state as closely as possible.  This means that, instead of putting this logic in the view layer, we should find a way to make the VM emit an `Error` object when one needs to be displayed.

Right now, that will mean adding a temporary method to the PocketVM that sends it the current URL.  Once the `SessionRepo` is in master, it can get it from there and we can remove the temporary method.

Baron-Severin: @psymoon I'm not sure I follow?  Presentation logic is the VM's job.  It's supposed to mirror display state as closely as possible, and in this case that means that `PocketVM.state` should be `Error` whenever an error is displayed.

psymoon: @Baron-Severin yes, I agree 100% on how vs. what to display logic :)

Baron-Severin: Actually, are we sure we need this code at all? As far as I'm aware, the only thing that can put Pocket into an error state is a failed network call, and the repo already handles that case.  Am I missing an edge case?

dnarcese: This part is just for testing- QA wanted a link that would show the error screen for screenshot tests (#1111).

Baron-Severin: We have a way to push mocked state to the `PocketVideoRepo` directly in Espresso tests now (`FirefoxTestApplication#pushPocketRepoState`).  If @npark-mozilla and @AaronMT are fine with it, I would like to remove the requirement to show a Pocket error on a specific URL, and use that method instead.

AaronMT: Sure. PocketErrorTest.java would need an accompanying update

psymoon:
Digging a little more deeper into this issue, it looks like `Pocket` should utilize the usage of [ConnectivityManager](https://developer.android.com/reference/android/net/ConnectivityManager). Only try pocket request if there's internet connection available. `Try Again` button also seems redundant since we are constantly trying to check for feed anyway (unless I'm mistaken). Calling `requestFocus()` within LiveData observer seems highly unstable and prone to race condition (especially with pocket request retries). This could be addressed further with `FocusRepo` implementation but for now I think we can force all focus to default if focus is lost. We could also look into different polling algorithms. What do you think, @Baron-Severin @dnarcese?

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/BrowserNavigationOverlay.kt
```diff
@@ -225,6 +225,7 @@ class BrowserNavigationOverlay @JvmOverloads constructor(
     fun showMegaTileError() {
         pocketVideosContainer.visibility = View.GONE
         pocketErrorContainer.visibility = View.VISIBLE
+
```
Baron-Severin: (Copying @psymoon's comment into a review so we can have a discussion thread)

> Digging a little more deeper into this issue, it looks like Pocket should utilize the usage of ConnectivityManager. Only try pocket request if there's internet connection available. Try Again button also seems redundant since we are constantly trying to check for feed anyway (unless I'm mistaken). Calling requestFocus() within LiveData observer seems highly unstable and prone to race condition (especially with pocket request retries). This could be addressed further with FocusRepo implementation but for now I think we can force all focus to default if focus is lost. We could also look into different polling algorithms. What do you think, @Baron-Severin @dnarcese?

Baron-Severin: > Pocket should utilize the usage of ConnectivityManager. Only try pocket request if there's internet connection available.

I don't have particularly strong feelings on this.  If the user has no internet connection, the requests instantly fail and otherwise have no impact on the user experience.  I would vote against the additional code complexity, but again, not a strong opinion.

> Try Again button also seems redundant since we are constantly trying to check for feed anyway (unless I'm mistaken).

It's definitely a narrow use case.  The backoff will eventually wait for a decent length of time between retries (the wait doubles between each attempt), it could be possible for a user to mess with their connection and then want to retry immediately.  The decision has UX ramifications as well, @aminalhazwani do you have any thoughts here?

> Calling requestFocus() within LiveData observer seems highly unstable and prone to race condition (especially with pocket request retries).

TBF, the requestFocus is within a click listener, which is a lot less concerning.

> We could also look into different polling algorithms.

For sure.  Also, the retry wait starts at 1 second, which is probably a little bit too eager.  There's an idea in A-C about breaking out a generic HTTP API client that handles this kind of stuff, so it would be good to refine our approach before writing that.  https://github.com/mozilla-mobile/android-components/issues/1508

PR #1561, opened 2018-11-30
1552/pocket null key

Baron-Severin:
### Pull Request checklist

### DIFF
--- a/app/src/main/java/org/mozilla/tv/firefox/pocket/PocketEndpoint.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/pocket/PocketEndpoint.kt
@@ -14,7 +14,6 @@ import org.json.JSONObject
 import org.mozilla.tv.firefox.BuildConfig
 import org.mozilla.tv.firefox.ext.executeAndAwait
 import org.mozilla.tv.firefox.ext.flatMapObj
-import org.mozilla.tv.firefox.utils.BuildConfigDerivables
 import org.mozilla.tv.firefox.utils.OkHttpWrapper
 import java.io.IOException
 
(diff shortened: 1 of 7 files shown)


app/src/main/java/org/mozilla/tv/firefox/pocket/PocketEndpoint.kt
```diff
@@ -61,9 +53,9 @@ private object PocketEndpointRaw {
 
     /** @return The global video recommendations as a raw JSON str or null on error. */
     @AnyThread // executeAndAwait hands off the request to the OkHttp dispatcher.
-    suspend fun getGlobalVideoRecommendations(version: String): String? {
+    suspend fun getGlobalVideoRecommendations(version: String, endpoint: Uri): String? {
```
psymoon: What's the different between `endpoint` and `pocketEndpoint`?

app/src/main/java/org/mozilla/tv/firefox/utils/BuildConfigDerivables.kt
```diff
@@ -4,18 +4,35 @@
...
-    @Suppress("SENSELESS_COMPARISON") // Values of BuildConfig can change but the compiler doesn't know that
+    @Suppress("UselessCallOnNotNull") // Values of BuildConfig can change but the compiler doesn't know that
     val initialPocketRepoState = when {
-        BuildConfig.POCKET_KEY == null -> PocketVideoRepo.FeedState.NoAPIKey
+        BuildConfig.POCKET_KEY.isNullOrEmpty() -> PocketVideoRepo.FeedState.NoAPIKey
         else -> PocketVideoRepo.FeedState.Loading
     }
+
+    // Pocket key can be null if it was not included in the build.  In this case we
+    // know that any calls to Pocket will fail, and so we do not provide the
```
psymoon: So essentially we are preventing the app from executing Pocket endpoint requests. But this also means we are not stopping async request calls in general. Fire stick devices already have perf issues due to its hardware spec

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketEndpoint.kt
Baron-Severin: This should probably be pocketEndpoint too, I'll change that.

app/src/main/java/org/mozilla/tv/firefox/utils/BuildConfigDerivables.kt
Baron-Severin: That's a good point.  I wanted to cut this off at the narrowed point possible, rather than just turning off updates, because I didn't want us to accidentally write code that bypassed those checks.  This is part of what caused our earlier problem (our initial update makes sure we have a valid key, but our background updates don't).  Killing this at the `Uri` seemed like a good way to avoid that.  

Still, it would probably be good to take a look at this and make sure the performance impact is not significant.  My guess is that it wouldn't be.  It's basically waiting for a few minutes, seeing that there's a null, and then shorting with this code.  But it would be good to verify that I'm not missing anything that could hurt performance.

PR #1562, opened 2018-11-30
Closes #1553: prevents pocket requests when locale is not english

### LINKED ISSUE #1553: Don't make any Pocket requests from non-English locales
Currently, Pocket is only shown to users in English speaking locales.  However, in the interest of separating concerns, the repo is currently unaware of locales.  Requests are made regardless, and only displayed if the locale is correct.

Non-English locales make up a substantial proportion of our user base, so it would be good to minimize these wasted calls.

### What / Requirements
Pocket requests should not be made from non-English locales.

Baron-Severin:
### Pull Request checklist

### DIFF
--- a/app/src/main/java/org/mozilla/tv/firefox/pocket/PocketFeedStateMachine.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/pocket/PocketFeedStateMachine.kt
@@ -32,6 +32,7 @@ import org.mozilla.tv.firefox.pocket.PocketVideoRepo.FeedState.Inactive
  * ### Valid Starting Configurations
  * - [Loading]
  * - [NoAPIKey]
+ * - [Inactive]
  */
 class PocketFeedStateMachine {
     fun computeNewState(
--- a/app/src/main/java/org/mozilla/tv/firefox/pocket/PocketFeedStateMachine.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/pocket/PocketFeedStateMachine.kt
@@ -40,10 +40,10 @@ class PocketFeedStateMachine {
         cacheState: PocketVideoRepo.FeedState?
     ): PocketVideoRepo.FeedState {
         return when {
-            repoState is Inactive -> repoState
-            repoState is LoadComplete -> return repoState
-            repoState === Loading && cacheState === FetchFailed -> return repoState
-            repoState === FetchFailed && cacheState === Loading -> return repoState
+            repoState === Inactive || cacheState === Inactive -> Inactive
+            repoState is LoadComplete -> repoState
+            repoState === Loading && cacheState === FetchFailed -> repoState
+            repoState === FetchFailed && cacheState === Loading -> repoState
             cacheState == null -> repoState
             else -> cacheState
         }
--- a/app/src/main/java/org/mozilla/tv/firefox/pocket/PocketFeedStateMachine.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/pocket/PocketFeedStateMachine.kt
@@ -8,6 +8,7 @@ import org.mozilla.tv.firefox.pocket.PocketVideoRepo.FeedState.LoadComplete
 import org.mozilla.tv.firefox.pocket.PocketVideoRepo.FeedState.Loading
 import org.mozilla.tv.firefox.pocket.PocketVideoRepo.FeedState.NoAPIKey
 import org.mozilla.tv.firefox.pocket.PocketVideoRepo.FeedState.FetchFailed
+import org.mozilla.tv.firefox.pocket.PocketVideoRepo.FeedState.Inactive
 
 /**
  * Manages all logic for updating cached [PocketVideoRepo.FeedState] in a testable way.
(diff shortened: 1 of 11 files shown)


app/src/main/java/org/mozilla/tv/firefox/utils/BuildConfigDerivables.kt
```diff
@@ -12,9 +12,10 @@ import org.mozilla.tv.firefox.pocket.PocketVideoRepo
  *
  * This logic is often simple but noisy, so pulling it out of client code improves readability.
  */
-class BuildConfigDerivables {
+class BuildConfigDerivables(private val localeIsEnglish: () -> Boolean) {
     @Suppress("SENSELESS_COMPARISON") // Values of BuildConfig can change but the compiler doesn't know that
     val initialPocketRepoState = when {
+        !localeIsEnglish.invoke() -> PocketVideoRepo.FeedState.Inactive
```
liuche: Good, setting initial state

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketViewModel.kt
```diff
@@ -55,13 +54,8 @@ class PocketViewModel(
                 is PocketVideoRepo.FeedState.LoadComplete -> State.Feed(repoState.videos)
                 is PocketVideoRepo.FeedState.NoAPIKey -> State.Feed(noKeyPlaceholders)
                 is PocketVideoRepo.FeedState.FetchFailed -> State.Error
+                is PocketVideoRepo.FeedState.Inactive -> State.NotDisplayed
             }
-        }.map {
```
liuche: +1 for this moving "not displayed" to an actual state.

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketVideoRepo.kt
```diff
@@ -72,6 +74,11 @@ open class PocketVideoRepo(
 
     private suspend fun updateInner() {
 
+        if (!localeIsEnglish.invoke()) {
+            _feedState.postValue(FeedState.Inactive)
```
liuche: One nit is that we hit this check all the time on the original timer, even though changing a locale is a very rare situation. Maybe we add a retryDelay here, to way less frequently for locale change? The flip side is that users changing their locale will not automatically get new content, but they'll see the "Try again" button and can force a refresh.

app/src/test/java/org/mozilla/tv/firefox/pocket/PocketVideoRepoTest.kt
```diff
@@ -0,0 +1,49 @@
...
+
+        pocketVideoRepo = PocketVideoRepo(
+            pocketEndpoint = pocketEndpoint,
+            localeIsEnglish = isEnglish,
+            pocketFeedStateMachine = PocketFeedStateMachine(),
+            buildConfigDerivables = BuildConfigDerivables(isEnglish)
+        )
+    }
+
+    @Test
```
liuche: I wonder if we should add a test for changing locales - it's not a common case, but it's the most likely place where something could break.

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketFeedStateMachine.kt
```diff
@@ -21,9 +22,9 @@ import org.mozilla.tv.firefox.pocket.PocketVideoRepo.FeedState.FetchFailed
  *
  * ### Diagram
  * [Loading] ------------------
- *    |  ^                    |     ---
- *    V  |                    V     | V
- * [FetchFailed] ---------> [LoadComplete]
+ *    |  ^                    |           ---
+ *    V  |                    V           | V
+ * [FetchFailed] ---------> [LoadComplete] / [Inactive]
```
liuche: Actually `Inactive` goes to `Loading` and could end up in `FetchFailed` if it's never been loaded before, not necessarily directly to `LoadComplete`.

Baron-Severin: Although would be possible for the repo to attempt to load, the state machine will never emit anything new if its cached state is `Inactive`.

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketVideoRepo.kt
Baron-Severin: This seems to be pretty cheap (in a quick test on the Cube, `localeIsEnglish.invoke` took about a quarter of a MS), so my inclination would be to keep the code as simple as possible.  That said, let me know if you feel strongly about it and I can cache the results.

app/src/test/java/org/mozilla/tv/firefox/pocket/PocketVideoRepoTest.kt
Baron-Severin: Hm, I guess we need to decide how we're handling those cases.  To be honest, I've been assuming that users change locales very infrequently, so I've essentially been dropping them on the floor.  With this code, if a non-English user switches to English they won't see Pocket until they restart the app.

@athomasmoz how would you like us to handle this case?

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketFeedStateMachine.kt
Baron-Severin: To correct myself, as currently implemented it can only go from `Inactive` to `Inactive` or `LoadComplete`.  This should probably change though, I'm going to put some thought into it.

PR #1574, opened 2018-12-14
Closes #1572: Close app on back button press

### LINKED ISSUE #1572: Can't exit app with back-button press
## Steps to reproduce
Open app.
Click back button on remote.

### Expected behavior
Exits app.

### Actual behavior
Doesn't exit app, no feedback.

dnarcese:
### Pull Request checklist

### DIFF
--- a/app/src/main/java/org/mozilla/tv/firefox/webrender/WebRenderFragment.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/webrender/WebRenderFragment.kt
@@ -355,20 +355,8 @@ class WebRenderFragment : EngineViewLifecycleFragment(), Session.Observer {
             }
             else -> {
                 context!!.webRenderComponents.sessionManager.remove()
-
-                // We can get into this state in two situations:
-                // - (1) The user is on the "home page" and the overlay is visible. In this situation we just want to
-                //       let the activity handle the back press (and leave the app).
-                // - (2) The overlay is not visible and we are on a website and can't go back further. In this case
-                //       "back" should show the overlay (the next back press closes the app). In this situation we
-                //       cannot look at the URL. We just removed the session and the new session may not be loaded yet.
-                if (!browserOverlay.isVisible) {
-                    // If the browser overlay is not visible then we show it now immediately. Depending on the loading
-                    // state of the page we m
-                    setOverlayVisible(true)
-                } else {
-                    return false
-                }
+                context!!.serviceLocator.webViewCache.doNotPersist()
+                return false
             }
         }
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/mozilla/tv/firefox/webrender/WebRenderFragment.kt
```diff
@@ -347,8 +347,6 @@ class WebRenderFragment : EngineViewLifecycleFragment(), Session.Observer {
                 TelemetryIntegration.INSTANCE.browserBackControllerEvent()
             }
             else -> {
-                context!!.webRenderComponents.sessionManager.remove()
-
                 // We can get into this state in two situations:
                 // - (1) The user is on the "home page" and the overlay is visible. In this situation we just want to
```
liuche: I don't think that we want this (2) case anymore - looking at the history this was written a long time ago, before we had the combined overlay/home tiles. That's what's probably actually breaking this (because it's returns true) - and the true was a result of fixing an auto-kotlin-convert: https://github.com/mozilla-mobile/firefox-tv/blame/8c61df2e2f4fe2678418314accb9e31a79d4a00b/app/src/main/java/org/mozilla/tv/firefox/webrender/WebRenderFragment.kt#L366
If the overlay is not visible and we're on a website, we should just exit, not open the overlay. It actually looks like we should just return false for this whole `else` statement of the when/else, and remove the `if !browserOverlay.isVisible` check. That should fix it because it'll allow the app to handle the back button press.

app/src/main/java/org/mozilla/tv/firefox/MainActivity.kt
```diff
@@ -193,7 +193,7 @@ class MainActivity : LocaleAwareAppCompatActivity(), OnUrlEnteredListener, Media
                 // The Browser fragment handles back presses on its own because it might just go back
                 // in the browsing history.
                 return
-            }
+            } else finish()
```
liuche: Looking at the below code, if we remove the
```
if (!browserOverlay.isVisible) {
    // If the browser overlay is not visible then we show it now immediately. Depending on the loading
    // state of the page we m
    setOverlayVisible(true)
```
is this still necessary, or does the Activity handle the back behavior correctly?

app/src/main/java/org/mozilla/tv/firefox/webrender/WebRenderFragment.kt
dnarcese: Oh yup, this is a much simpler fix!

Issue #1420, opened 2018-10-31
[1396] #7: Implement test for YouTube


Baron-Severin:
Part of #1396 

YouTube makes up a significant portion of our usage, so it would be good to have an automated test that ensures that we do not push any changes that impact its functionality.  This is relatively important, but these tests will likely be somewhat flakey.

Tests should verify that:
- YouTube loads
- D-button presses move focus around page
- Selecting a video causes it to play
- Pause, fast forward, and rewind buttons work on videos
- Full-screen works
- Back removes full screen
- Back exits video
- Back from YouTube home closes YouTube

PR #1576, opened 2018-12-14
Closes #1542: Can't go back out of YouTube

### LINKED ISSUE #1542: Can't go back out of YouTube (with remote back)
## Steps to reproduce
1. Click Google search tile
2. Open overlay and click YouTube tile
3. Click back button on remote several times to try and get back to Google
### Expected behavior
Can get back to Google
### Actual behavior
Cannot go back to Google from the remote back button. Focus stays on the 'Home' icon of YouTube.
### Device information

* Fire TV device: 4k
* Firefox version: 3.1

dnarcese:
### Pull Request checklist

In the interest of time, this will be covered in #1420.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Changed
 
 ### Fixed
+- Issue where users were unable to exit YouTube with remote back button (#1542)
 
 ## [3.1.2] - 2018-12-11
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,11 +4,9 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-## [3.1.3] - 2018-12-17
+## [3.1.3] - 2018-12-20
 ### Added
 - Ability to exit Firefox with remote back button
-
-### Changed
 
 ### Fixed
 - Issue where users were unable to exit YouTube with remote back button (#1542)
--- a/app/src/main/java/org/mozilla/tv/firefox/ext/WebBackForwardList.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/ext/WebBackForwardList.kt
@@ -1,0 +1,16 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.tv.firefox.ext
+
+import android.webkit.WebBackForwardList
+
+/**
+ * Convert WebBackForwardList to List<String> so it is easier to use.
+ */
+fun WebBackForwardList.toStringList(): List<String> {
+    return List(size) {
+        getItemAtIndex(it).originalUrl
+    }
+}
--- a/app/src/main/java/org/mozilla/tv/firefox/ext/WebBackForwardList.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/ext/WebBackForwardList.kt
@@ -5,12 +5,13 @@
 package org.mozilla.tv.firefox.ext
 
 import android.webkit.WebBackForwardList
+import android.webkit.WebHistoryItem
 
 /**
- * Convert WebBackForwardList to List<String> so it is easier to use.
+ * Convert WebBackForwardList to List<WebHistoryItem> so it is easier to use.
  */
-fun WebBackForwardList.toStringList(): List<String> {
+fun WebBackForwardList.toList(): List<WebHistoryItem> {
     return List(size) {
-        getItemAtIndex(it).originalUrl
+        getItemAtIndex(it)
     }
 }
--- a/app/src/main/java/org/mozilla/tv/firefox/webrender/WebRenderFragment.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/webrender/WebRenderFragment.kt
@@ -39,7 +39,7 @@ import org.mozilla.tv.firefox.ext.toUri
 import org.mozilla.tv.firefox.ext.isYoutubeTV
 import org.mozilla.tv.firefox.ext.focusedDOMElement
 import org.mozilla.tv.firefox.ext.serviceLocator
-import org.mozilla.tv.firefox.ext.toStringList
+import org.mozilla.tv.firefox.ext.toList
 import org.mozilla.tv.firefox.navigationoverlay.BrowserNavigationOverlay
 import org.mozilla.tv.firefox.navigationoverlay.NavigationEvent
 import org.mozilla.tv.firefox.pinnedtile.PinnedTileAdapter
@@ -64,6 +64,7 @@ class WebRenderFragment : EngineViewLifecycleFragment(), Session.Observer {
         const val APP_URL_PREFIX = "firefox:"
         const val APP_URL_HOME = "${APP_URL_PREFIX}home"
         const val APP_URL_POCKET_ERROR = "${APP_URL_PREFIX}error:pocketconnection"
+        const val YOUTUBE_TILE_URL = "https://ftv.cdn.mozilla.net/ytht"
 
         @JvmStatic
         fun createForSession(session: Session) = WebRenderFragment().apply {
@@ -413,29 +414,40 @@ class WebRenderFragment : EngineViewLifecycleFragment(), Session.Observer {
         }
 
         if (!browserOverlay.isVisible && session.isYoutubeTV &&
-                event.keyCode == KeyEvent.KEYCODE_BACK &&
-                event.action == KeyEvent.ACTION_DOWN) {
+                event.keyCode == KeyEvent.KEYCODE_BACK) {
+            val escKeyEvent = KeyEvent(event.action, KeyEvent.KEYCODE_ESCAPE)
+
+            /**
+             * If YouTube guide-list element is focused:
+             * - Suppress the DOWN,BACK key event
+             * - Change the UP,BACK key event to go back in history before YouTube
+             * Else:
+             * - Dispatch ESC key event
+             */
+            val jsCallback = ValueCallback<String> {
+                if (it == "true") {
+                    if (event.action == KeyEvent.ACTION_DOWN) Unit
+                    else goBackBeforeYouTube()
+                } else (activity as MainActivity).dispatchKeyEvent(escKeyEvent)
+            }
+            // This will return true if the currently focused YouTube element is a button in the left sidebar, false otherwise
             webView?.evalJS("""
                 (function () {
                     return document.activeElement.parentElement.parentElement.id === 'guide-list';
                 })();
                 """,
-                    ValueCallback {
-                        val keyCode = if (it == "true") {
-                            val wv = (webView as ViewGroup).getChildAt(0) as WebView
-                            val backForward = wv.copyBackForwardList().toStringList()
-                            val youtubeIndex = backForward.lastIndexOf("https://ftv.cdn.mozilla.net/ytht")
-                            val goBackSteps = backForward.size - youtubeIndex
-                            wv.goBackOrForward(-goBackSteps)
-                            KeyEvent.KEYCODE_BACK
-                        } else KeyEvent.KEYCODE_ESCAPE
-
-                        val newKeyEvent = KeyEvent(KeyEvent.ACTION_UP, keyCode)
-                        (activity as MainActivity).dispatchKeyEvent(newKeyEvent)
-                    })
+                    jsCallback)
             return true
         }
         return false
+    }
+
+    private fun goBackBeforeYouTube() {
+        val webView = (webView as ViewGroup).getChildAt(0) as WebView
+        val backForwardUrlList = webView.copyBackForwardList().toList().map { it.originalUrl }
+        val youtubeIndex = backForwardUrlList.lastIndexOf(YOUTUBE_TILE_URL)
+        val goBackSteps = backForwardUrlList.size - youtubeIndex
+        webView.goBackOrForward(-goBackSteps)
     }
 
     /**
--- a/app/src/main/java/org/mozilla/tv/firefox/webrender/WebRenderFragment.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/webrender/WebRenderFragment.kt
@@ -16,7 +16,6 @@ import android.view.View
 import android.view.ViewGroup
 import android.view.WindowManager
 import android.webkit.ValueCallback
-import android.webkit.WebBackForwardList
 import android.webkit.WebView
 import kotlinx.android.synthetic.main.browser_overlay.*
 import kotlinx.android.synthetic.main.browser_overlay.view.*
@@ -40,6 +39,7 @@ import org.mozilla.tv.firefox.ext.toUri
 import org.mozilla.tv.firefox.ext.isYoutubeTV
 import org.mozilla.tv.firefox.ext.focusedDOMElement
 import org.mozilla.tv.firefox.ext.serviceLocator
+import org.mozilla.tv.firefox.ext.toStringList
 import org.mozilla.tv.firefox.navigationoverlay.BrowserNavigationOverlay
 import org.mozilla.tv.firefox.navigationoverlay.NavigationEvent
 import org.mozilla.tv.firefox.pinnedtile.PinnedTileAdapter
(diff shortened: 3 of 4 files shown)


app/src/main/java/org/mozilla/tv/firefox/webrender/WebRenderFragment.kt
```diff
@@ -284,6 +289,8 @@ class WebRenderFragment : EngineViewLifecycleFragment(), Session.Observer {
         if (session.url == APP_URL_HOME) {
             browserOverlay?.visibility = View.VISIBLE
         }
+
+        webView.addJavascriptInterface(YouTubeBackHandler(), "youtubeBackHandler")
```
mcomella: `youtubeBackHandler` gets written directly into the DOM so you probably want to name it something:
- identifiable by website owners wondering what this random object is
- identifiable by us to know we added it
- that won't get clobbered by site content

[For media session, we use `_firefoxTV_playbackStateObserverJava`](https://github.com/mozilla-mobile/firefox-tv/blob/23a1288990e22a3fc06a2975bac0c307775a9e8d/app/src/main/java/org/mozilla/tv/firefox/webrender/VideoVoiceCommandMediaSession.kt#L52).

By the way, see `VideoVoiceCommandMediaSession` for an example of how you can break a JavascriptInterface dependent class into a separate file.

app/src/main/java/org/mozilla/tv/firefox/webrender/WebRenderFragment.kt
```diff
@@ -421,9 +428,10 @@ class WebRenderFragment : EngineViewLifecycleFragment(), Session.Observer {
         }
 
         if (!browserOverlay.isVisible && session.isYoutubeTV &&
-                event.keyCode == KeyEvent.KEYCODE_BACK) {
-            val escKeyEvent = KeyEvent(event.action, KeyEvent.KEYCODE_ESCAPE)
-            (activity as MainActivity).dispatchKeyEvent(escKeyEvent)
+                event.keyCode == KeyEvent.KEYCODE_BACK &&
+                !youtubeOverride) {
+            youtubeKeyEvents.add(event)
+            webView?.evalJS("youtubeBackHandler.youtubeBack(document.activeElement.parentElement.parentElement.id === 'guide-list');")
```
mcomella: Did you verify this is the best way to identify if we're at the top level in YouTube's back stack? I did this rather quickly.

Also, we should clarify what `document.active.parent.parent` actually means and why it works, preferably in code but in comments if that fails.

app/src/main/java/org/mozilla/tv/firefox/webrender/WebRenderFragment.kt
```diff
@@ -447,6 +455,44 @@ class WebRenderFragment : EngineViewLifecycleFragment(), Session.Observer {
...
+    /**
+     * JavascriptInterface for handling when to map back click to ESC vs perform back in history.
+     * Clicking remote back will remap to ESC until the active DOM element is an item in YouTube's guidelist-
+     * then it will go back in the history stack.
+     */
+    inner class YouTubeBackHandler {
+        @JavascriptInterface
+        fun youtubeBack(sendBack: Boolean) {
+            if (sendBack) {
+                Handler(Looper.getMainLooper()).post {
```
mcomella: Creating a new Handler might be expensive so we wouldn't want to do it each time this is called. Maybe `launch(UI)`?

app/src/main/java/org/mozilla/tv/firefox/webrender/WebRenderFragment.kt
```diff
@@ -447,6 +455,44 @@ class WebRenderFragment : EngineViewLifecycleFragment(), Session.Observer {
...
+                    val wv = (webView as ViewGroup).getChildAt(0) as WebView
+                    val backForward = wv.copyBackForwardList().toList()
+                    val youtubeIndex = backForward.lastIndexOf("https://ftv.cdn.mozilla.net/ytht")
```
mcomella: This isn't always correct: we need to find this or `youtube.com/tv`, e.g. if the user navigates there directly.

app/src/main/java/org/mozilla/tv/firefox/webrender/WebRenderFragment.kt
```diff
@@ -447,6 +455,44 @@ class WebRenderFragment : EngineViewLifecycleFragment(), Session.Observer {
...
+                    wv.goBackOrForward(-goBackSteps)
+                }
+                return
+            }
+            val keyCode = if (sendBack) KeyEvent.KEYCODE_BACK else KeyEvent.KEYCODE_ESCAPE
+
+            youtubeOverride = true
+            youtubeKeyEvents.forEach {
+                val newKeyEvent = KeyEvent(it.action, keyCode)
+                Handler(Looper.getMainLooper()).post {
```
mcomella: Again, new handler created.

app/src/main/java/org/mozilla/tv/firefox/webrender/WebRenderFragment.kt
```diff
@@ -447,6 +455,44 @@ class WebRenderFragment : EngineViewLifecycleFragment(), Session.Observer {
...
+                    val backForward = wv.copyBackForwardList().toList()
+                    val youtubeIndex = backForward.lastIndexOf("https://ftv.cdn.mozilla.net/ytht")
+                    val goBackSteps = backForward.size - youtubeIndex
+                    wv.goBackOrForward(-goBackSteps)
+                }
+                return
+            }
+            val keyCode = if (sendBack) KeyEvent.KEYCODE_BACK else KeyEvent.KEYCODE_ESCAPE
+
+            youtubeOverride = true
```
mcomella: I wonder if we can do this without the youtubeOverride? I did it because simulating the `ESC` presses called this method again so I needed a flag to prevent nested calls but I wonder if there is a better way.

CHANGELOG.md
```diff
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Changed
 
 ### Fixed
-- Some web fonts were blocked which caused icons to be missing
```
mcomella: I don't think we should delete this line, right?

CHANGELOG.md
```diff
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
...
+- Unable to back out of YouTube with remote back button
```
mcomella: "An issue where users were unable to leave YouTube with remote back button (#...)"

Include a bug number and perhaps elaborate more, like ^

app/src/main/java/org/mozilla/tv/firefox/webrender/WebRenderFragment.kt
```diff
@@ -421,9 +428,10 @@ class WebRenderFragment : EngineViewLifecycleFragment(), Session.Observer {
         }
 
         if (!browserOverlay.isVisible && session.isYoutubeTV &&
-                event.keyCode == KeyEvent.KEYCODE_BACK) {
-            val escKeyEvent = KeyEvent(event.action, KeyEvent.KEYCODE_ESCAPE)
-            (activity as MainActivity).dispatchKeyEvent(escKeyEvent)
+                event.keyCode == KeyEvent.KEYCODE_BACK &&
```
mcomella: Did you try this with the browser navigation overlay back button too?

mcomella:
@dnarcese Looking at my comments, I think it seems reasonable to _attempt_ to address everything except trying to abstract the code away to a one file, which can take a while given the uncertainty of design. Since the code works, if something seems like it'll take too long and we'll miss our deadline, **feel free to skip it** and make a note!

To summarize the issues to address, in order of priority:
- Pressing back from Pocket can exit YouTube (not sure if this already works; note that it does not need to go back to Pocket, just needs to exit YouTube)
- Try to have a better system than boolean mutable flags and key event lists
- Ensure we're using an optimal way to ensure we're at the top level in YouTube's ESC back stack
- Don't create new handlers on each button press (cache a handler or use coroutines)
- Disambiguate the object injected into the DOM with a better name
- Address changelog nit

We should ask Ashley if these use cases are important and if so, we should fix them too (probably at the highest priority):
- Overlay back button can also leave YouTube
- We can also back out if the user visits youtube.com/tv by typing the URL (this might be the same as the Pocket use case)

Let me know if you have questions!

dnarcese:
@mcomella I've addressed everything from above except for 

- Pressing back from Pocket can exit YouTube
- We can also back out if the user visits youtube.com/tv by typing the URL

Those 2 issues are trickier to deal with and it was decided that since more users click the YouTube home tile than the Pocket megatile, this issue is lower priority and can be skipped for now.

liuche:
@dnarcese can you file a bug for the follow-ups, and document the problems with the Pocket youtube redirect back stack, and optionally ideas for how to address them?

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Changed
 
 ### Fixed
+- Issue where users were unable to exit YouTube with remote back button (#1542)
```
mcomella: Not your fault: the release date for 3.1.3 is inaccurate and the Changed header should only be present if there are items under it

app/src/main/java/org/mozilla/tv/firefox/webrender/WebRenderFragment.kt
```diff
@@ -409,9 +413,26 @@ class WebRenderFragment : EngineViewLifecycleFragment(), Session.Observer {
...
+                event.action == KeyEvent.ACTION_DOWN) {
+            webView?.evalJS("""
+                (function () {
+                    return document.activeElement.parentElement.parentElement.id === 'guide-list';
```
mcomella: nit: comment to explain what is happening here. Probably outside of the JS because it doesn't make sense to inject JS comments.

app/src/main/java/org/mozilla/tv/firefox/webrender/WebRenderFragment.kt
```diff
@@ -409,9 +413,26 @@ class WebRenderFragment : EngineViewLifecycleFragment(), Session.Observer {
...
+                })();
+                """,
+                    ValueCallback {
+                        val keyCode = if (it == "true") {
+                            val wv = (webView as ViewGroup).getChildAt(0) as WebView
```
mcomella: nit: This is a lot of nesting: can we define the `ValueCallback` elsewhere so that it isn't as heavily nested?

mcomella: nit: We avoid one or two letter variable names. -> `webView`?

app/src/main/java/org/mozilla/tv/firefox/webrender/WebRenderFragment.kt
```diff
@@ -409,9 +413,26 @@ class WebRenderFragment : EngineViewLifecycleFragment(), Session.Observer {
...
+                            val backForward = wv.copyBackForwardList().toStringList()
+                            val youtubeIndex = backForward.lastIndexOf("https://ftv.cdn.mozilla.net/ytht")
+                            val goBackSteps = backForward.size - youtubeIndex
+                            wv.goBackOrForward(-goBackSteps)
+                            KeyEvent.KEYCODE_BACK
+                        } else KeyEvent.KEYCODE_ESCAPE
+
+                        val newKeyEvent = KeyEvent(KeyEvent.ACTION_UP, keyCode)
+                        (activity as MainActivity).dispatchKeyEvent(newKeyEvent)
```
mcomella: iiuc, I see a few problems with this:
- We're not dispatching the down event. That might work in this case but I'm concerned what side effects there might be
- We're dispatching the up event twice.
- We're not converting the BACK, UP physical key press into an ESC, UP physical key press

fwiw, I know we're pressed for time but I think it's think it's these types of edge cases that get uncovered while writing tests for code like this.

To be explicit, I think the logic we want here is:
```
if (youtube.com/tv) {
    if (!atYouTubeTopLevel) {
        dispatchHomeEvent(event.action)
    } else {
        goBackBeforeYouTube()
    }
} else {
    dispatchKeyEventAsNormal()
}
```

app/src/main/java/org/mozilla/tv/firefox/webrender/WebRenderFragment.kt
```diff
@@ -409,9 +413,26 @@ class WebRenderFragment : EngineViewLifecycleFragment(), Session.Observer {
...
+            webView?.evalJS("""
+                (function () {
+                    return document.activeElement.parentElement.parentElement.id === 'guide-list';
+                })();
+                """,
+                    ValueCallback {
+                        val keyCode = if (it == "true") {
+                            val wv = (webView as ViewGroup).getChildAt(0) as WebView
+                            val backForward = wv.copyBackForwardList().toStringList()
+                            val youtubeIndex = backForward.lastIndexOf("https://ftv.cdn.mozilla.net/ytht")
```
mcomella: nit: Is this url stored in a constant somewhere? If not, can we add it as one?

app/src/main/java/org/mozilla/tv/firefox/webrender/WebRenderFragment.kt
```diff
@@ -409,9 +413,26 @@ class WebRenderFragment : EngineViewLifecycleFragment(), Session.Observer {
...
+                event.action == KeyEvent.ACTION_DOWN) {
+            webView?.evalJS("""
+                (function () {
+                    return document.activeElement.parentElement.parentElement.id === 'guide-list';
+                })();
+                """,
+                    ValueCallback {
+                        val keyCode = if (it == "true") {
+                            val wv = (webView as ViewGroup).getChildAt(0) as WebView
+                            val backForward = wv.copyBackForwardList().toStringList()
```
mcomella: nit: `backForwardUrls` or similar

app/src/main/java/org/mozilla/tv/firefox/ext/WebBackForwardList.kt
```diff
@@ -0,0 +1,16 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.tv.firefox.ext
+
+import android.webkit.WebBackForwardList
+
+/**
+ * Convert WebBackForwardList to List<String> so it is easier to use.
+ */
+fun WebBackForwardList.toStringList(): List<String> {
```
mcomella: nit: `toUrlStringList` or similar: we want to be explicit that we're extracting the URLs.

However, I think I'd prefer to leave it as, `WebBackForwardList.toList()` and the callers can extract the string if they want: it's more flexible this way.

app/src/main/java/org/mozilla/tv/firefox/webrender/WebRenderFragment.kt
```diff
@@ -409,9 +413,26 @@ class WebRenderFragment : EngineViewLifecycleFragment(), Session.Observer {
...
+                            val youtubeIndex = backForward.lastIndexOf("https://ftv.cdn.mozilla.net/ytht")
+                            val goBackSteps = backForward.size - youtubeIndex
+                            wv.goBackOrForward(-goBackSteps)
+                            KeyEvent.KEYCODE_BACK
```
mcomella: Do we need to dispatch this back press at all if we're already going back before YouTube? I don't think so but I could be mistaken.

mcomella:
If we're pressed for time, everything but the logic correction can slip.

app/src/main/java/org/mozilla/tv/firefox/webrender/WebRenderFragment.kt
```diff
@@ -411,12 +416,40 @@ class WebRenderFragment : EngineViewLifecycleFragment(), Session.Observer {
         if (!browserOverlay.isVisible && session.isYoutubeTV &&
                 event.keyCode == KeyEvent.KEYCODE_BACK) {
             val escKeyEvent = KeyEvent(event.action, KeyEvent.KEYCODE_ESCAPE)
-            (activity as MainActivity).dispatchKeyEvent(escKeyEvent)
+
+            /**
+             * If YouTube guide-list element is focused:
```
mcomella: nit: the comments in here don't explain what problem we're trying to solve and, because our solution is complex, the solution at a high level. If I came across this code for the first time, I might be confused at why this BACK->ESC override is necessary (i.e. YouTube doesn't handle their back stack correctly) and why the override is only necessary _sometimes_.

For example, I might comment, "youtube/tv doesn't handle their back stack correctly: a user can navigate back by pressing ESC but going back in the browser back stack does nothing. To fix this, we convert BACK presses on youtube/tv to ESC. However, when we get to the top of YouTube's back stack, we back all the way out of YouTube."

app/src/main/java/org/mozilla/tv/firefox/webrender/WebRenderFragment.kt
```diff
@@ -411,12 +416,40 @@ class WebRenderFragment : EngineViewLifecycleFragment(), Session.Observer {
...
+                    jsCallback)
             return true
         }
         return false
     }
 
+    private fun goBackBeforeYouTube() {
+        val webView = (webView as ViewGroup).getChildAt(0) as WebView
+        val backForwardUrlList = webView.copyBackForwardList().toList().map { it.originalUrl }
+        val youtubeIndex = backForwardUrlList.lastIndexOf(YOUTUBE_TILE_URL)
```
mcomella: Finding the index and going back the appropriate number of steps would be a good candidate for writing tests, e.g.:
- How does this handle youtube.com/tv URLs?
- What happens if the CDN url is not in the format we expect (no trailing slash)?
- What happens if the CDN is at the end of the url list?
- What happens if the CDN url is not in the list?
- What if the CDN url is first?
- What if we remove firefox:home from the pages loaded by the application?

You get the idea. :P

PR #1578, opened 2018-12-15
Closes #1528: Fix PocketOnboardTest

### LINKED ISSUE #1528: PocketOnboardTest is failing (Espresso)
It looks like using rule `MainActivityTestRule` with argument `skipOnboarding = false` is not showing an onboarding screen.  

Removing the rule and replacing it with the former `new ActivityTestRule<MainActivity>(MainActivity.class, false, false);` does not solve the problem either.

### Device information

* Fire TV device: Cube
* Firefox version: af3ea79c44c930b353f45e527f15760d7c6c30cb

psymoon:
### Pull Request checklist

### DIFF
--- a/app/src/main/java/org/mozilla/tv/firefox/MainActivity.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/MainActivity.kt
@@ -5,7 +5,6 @@
 
 package org.mozilla.tv.firefox
 
-import android.app.Activity
 import android.content.Context
 import android.content.Intent
 import android.os.Bundle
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/mozilla/tv/firefox/MainActivity.kt
```diff
@@ -111,26 +109,15 @@ class MainActivity : LocaleAwareAppCompatActivity(), OnUrlEnteredListener, Media
             )
         }
 
-        if (Settings.getInstance(this@MainActivity).shouldShowTurboModeOnboarding()) {
-            val onboardingIntent = Intent(this@MainActivity, OnboardingActivity::class.java)
-            startActivityForResult(onboardingIntent, ONBOARDING_REQ_CODE)
+        if (Settings.getInstance(this@MainActivity).shouldShowPocketOnboarding()) {
```
Baron-Severin: I'm not sure if this changes anything that matters, but it looks like we're changing the order of execution here (the Pocket onboarding block is now executed before the Turbo one).

I did some brief testing and everything _seems_ to be working as expected, so I'm not sure if that has any impact at all.

psymoon: Activities open in a stack so unless we want the pocket onboarding to show first, this is the correct way

PR #1587, opened 2018-12-19
Closes #1557: add telemetry for app ID

### LINKED ISSUE #1557: Separate out Mozilla services components from code base
Currently we include all services endpoints, like telemetry, crash reporting, and Pocket in the repo codebase, but any fork of this repo will also send requests and telemetry. We assume only our own apps are using this code, which, being open source, isn't always a good assumption.

A few things we might want to do here:
- Protect endpoints by adding them at build time (through Taskcluster, like we do with API keys)
- Include an app-id especially for telemetry, so we know for sure that the telemetry we're collecting is from our devices

Baron-Severin:
@liuche 

- How many instances of this codebase are our own, and how many are forked

- We have run into problems with forked code in the past, and need to be aware of how many instances are our own users

- No viable alternatives were found

- No telemetry currently exists to answer these questions.

| Measurement Description                                      | Data Collection Category | Tracking Bug #                                               |
| ------------------------------------------------------------ | ------------------------ | ------------------------------------------------------------ |
| App ID       | 1                        | [Issue 1557](https://github.com/mozilla-mobile/firefox-tv/issues/1557) |

- This will be collected indefinitely

- All users

- There is an option in settings to stop sending usage data.

- The data will be analyzed through Redash queries and dashboards.

- On Redash and with mobile teams.

### DIFF
--- a/docs/telemetry.md
+++ b/docs/telemetry.md
@@ -18,6 +18,9 @@ As part of the event ping the most recent state of the user's setting is sent (d
 | Turbo mode currently enabled?      | tracking_protection_enabled     | **true**/false
 | Total home tile count    | total_home_tile_count           | `<int>`
 | Custom home tile count   | custom_home_tile_count          | `<int>`
+| App ID                   | app_id                          | `<string>`*
+
+(*) This is the public, published app ID, and does not contain any user information
 
 
 ## Events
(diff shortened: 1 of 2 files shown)


docs/telemetry.md
```diff
@@ -18,6 +18,7 @@ As part of the event ping the most recent state of the user's setting is sent (d
 | Turbo mode currently enabled?      | tracking_protection_enabled     | **true**/false
 | Total home tile count    | total_home_tile_count           | `<int>`
 | Custom home tile count   | custom_home_tile_count          | `<int>`
+| App ID                   | app_id                          | `<string>`
```
liuche: Nit: can you provide an example of this, or clarify that this is a public, published app-id so it's clear that isn't tied to the user?

PR #1599, opened 2018-12-19
Cherry-pick firefox:home history fix

liuche:
This cherry-picks the fix for a missing `firefox:home` after clearing sessions, which we now do after exiting the app on "back".


liuche:
Simpler fix in #1600

Issue #1620, opened 2018-12-31
Track Clicks on YouTube Tile


athomasmoz:
### Why/User Benefit/User Problem

Determine how many people & sessions use Firefox for things outside YouTube

### What / Requirements

* YouTube Tile clicks are tracked

### Acceptance Criteria (how do I know when I’m done?)

* We can pull data to determine the count & percent of sessions that include non-YouTube tile clicks
* We can pull data to determine the count & percent of users that click non-YouTube tiles

### Anti-vision

* Aside from YouTube, we do not track which tiles are clicked on or which URLs they go to

dnarcese:
@athomasmoz Should a YouTube tile click still count for a click on a `bundled` home tile? 
ie. in our Telemetry we have an entry for `bundled` `home_tile` clicks and there will be a new entry for `youtube` `home_tile` clicks. When a user clicks the YouTube tile, should the number for both entries increase, or just the youtube entry?

athomasmoz:
Yes, I don't want to change the behaviour there. We should still have it count for a home tile click. 

I'm glad you called it out though, sounds like something that should be documented. Do we have documentation on the different events tracked?

Issue #879, opened 2018-05-01
HIDE MENU telemetry intended to be shown when user hides menu without action; is used when loading URLs


mcomella:
### Why/User Benefit/User Problem
We want to know when the user closes the overlay without an action so we have a heuristic for when the user hasn't found the overlay to be useful for them.

This probe is broken: it's used when the user loads a URL or loads a tile in addition to when they take no action (via hitting back or menu).

@bbinto How important is this to fix? Should we fix it for v3.0?

I'll update the docs to indicate this problem.

bbinto:
@mcomella - what's the effort to fix it?

Sdaswani:
I suggest if we make changes then we change the probe name.

mcomella:
[triage] v3.0, we don't want to have bad telemetry.

If this broke in the v3.0 refactor, just fix it. If it broke before and we already have bad data, consider a new probe.

Baron-Severin:
@bbinto @Sdaswani Could you help me clarify what it is that we intend to capture?

We want to know when the user opens the menu then closes it without performing an action.  Does changing focus (i.e., moving the selection around) count as an action?  One option that occurred to me is that we could send an event if the user opens and closes the menu without performing a click.

Sdaswani:
@Baron-Severin I think 'not useful' could mean I opened it 1) by mistake, so it's not useful or 2) intentionally, and then didn't find anything to help get where I need to go. So I _think_ your option ('send an event if the user opens and closes the menu without performing a click') covers both those cases? If so, that's a great option IMHO. cc @bbinto

Baron-Severin:
I'm going to move forward with that condition.  @bbinto please let me know if you would like me to take a different approach.

Baron-Severin:
Looks like the telemetry events were broken in v2.0.1 on April 18th, so we'll have to rename them.  I've fixed the menu open / close events, and added a new one as discussed above.  The new event addresses an edge case in the earlier approach (user opens menu, user pins tile, user closes menu, our data shows that as a confused user).

I'm going to discuss event names with @liuche, then the code will be ready to land.
https://github.com/mozilla-mobile/firefox-tv/pull/1056

bbinto:
@Baron-Severin and @Sdaswani sorry for the delay - agreed and OK with that change

PR #1642, opened 2019-01-07
Issue #1620: Track Clicks on YouTube Tile

dnarcese:
### Pull Request checklist

Just a telemetry change - testing telemetry is not set up yet

### DIFF
--- a/app/src/main/java/org/mozilla/tv/firefox/telemetry/TelemetryIntegration.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/telemetry/TelemetryIntegration.kt
@@ -27,6 +27,7 @@ import java.util.Collections
 
 private const val SHARED_PREFS_KEY = "telemetryLib" // Don't call it TelemetryWrapper to avoid accidental IDE rename.
 private const val KEY_CLICKED_HOME_TILE_IDS_PER_SESSION = "clickedHomeTileIDsPerSession"
+private const val YOUTUBE_TILE_ID = "youtube"
 
 @Suppress(
         // Yes, this a large class with a lot of functions. But it's very simple and still easy to read.
@@ -90,6 +91,7 @@ open class TelemetryIntegration protected constructor(
         val TILE_BUNDLED = "bundled"
         val TILE_CUSTOM = "custom"
         val POCKET_VIDEO_MEGATILE = "pocket_video_tile"
+        val YOUTUBE_TILE = "youtube_tile"
     }
 
     private object Extra {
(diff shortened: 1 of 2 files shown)


dnarcese:

- How many users & sessions use Firefox for Fire TV for things outside of YouTube?

* Determine the count & percent of sessions that include non-YouTube tile clicks.
* Determine the count & percent of users that click non-YouTube tiles.

- This is the most accurate way to determine this information.

- No telemetry currently exists to answer these questions.

| Measurement Description                                      | Data Collection Category | Tracking Bug #                                               |
| ------------------------------------------------------------ | ------------------------ | ------------------------------------------------------------ |
| When the YouTube tile is clicked                             | 2                        | [Issue 1620](https://github.com/mozilla-mobile/firefox-tv/issues/1620) |
| How many sessions do not contain a click on the YouTube tile | 2                        | [Issue 1620](https://github.com/mozilla-mobile/firefox-tv/issues/1620) |

6) How long will this data be collected?
* This data will be permanently monitored by @athomasmoz

* All users with telemetry enabled, on release channel.

- There is an option in settings to stop sending usage data.

- The data will be analyzed through Redash queries and dashboards.

- On Redash and with mobile teams.

docs/telemetry.md
```diff
@@ -54,20 +54,17 @@ The event ping contains a list of events ([see event format on readthedocs.io](h
...
-| Unique tiles clicked per session**     | aggregate| click                 | home_tile  | `<int>` | |
+| Unique tiles clicked per session     | aggregate| click                 | home_tile  | `<int>` | |
+| Sessions that **do not** contain a click on  YouTube tile | aggregate | click | home_tile | session_without_youtube | |
 | Menu shown by user              | action   | user_show       | menu         | ||
 | Menu hidden by user              | action   | user_hide       | menu         | ||
 | No menu action taken              | aggregate   | no_action_taken       | menu         | ||
 
 (*)When the pin site switch is clicked, the state (on/off) of the desktop mode switch is also sent.
 
-(**)This is broken: see [#879](https://github.com/mozilla-mobile/firefox-tv/issues/879).** For overlay hidden, we only capture cases where the user closes the overlay themselves:
```
liuche: I actually think this `**` is actually supposed to be on the "Menu hidden by user" telemetry probe, so don't remove it. The formatting of the `(**)` is also broken, I think.

app/src/main/java/org/mozilla/tv/firefox/telemetry/TelemetryIntegration.kt
```diff
@@ -92,6 +92,7 @@ open class TelemetryIntegration protected constructor(
         val TILE_CUSTOM = "custom"
         val POCKET_VIDEO_MEGATILE = "pocket_video_tile"
         val YOUTUBE_TILE = "youtube_tile"
+        val SESSION_WITHOUT_YOUTUBE = "session_without_youtube"
```
liuche: I talked to Su (who is doing metrics for this) and I think that this doesn't need to be collected - he said it's easy enough to group by sessions that don't have a youtube_tile event, so can we remove this commit? (and the telemetry docs for it)

docs/telemetry.md
dnarcese: Issue #879 is closed, so I figured this isn't actually broken anymore- which is why I removed it.

docs/telemetry.md
```diff
@@ -54,20 +54,16 @@ The event ping contains a list of events ([see event format on readthedocs.io](h
...
-- Controller menu button is pressed
-- Controller back button is pressed
-
-e.g. we don't log an event when the overlay is closed automatically after "Home" is selected.
+(**)YouTube tile clicks are a sub-category of bundled tile clicks- each click will count towards both values.
```
liuche: One more nit: (**) doesn't show up properly in the docs in markup (see the docs/Telemetry.md) - even though it shows up fine in the comments. I think the fix is to wrap it in single quotes so: (`**`) (you can look at the raw text of this by editing it). I tested this by editing the docs/Telemetry.md and looking at preview (even though it's not possible to commit).

Issue #334, opened 2018-01-17
Localizations for 1.1


Sdaswani:
We need localizations in ten languages () for the following strings: Simplified Chinese, English, French, German, Hindi, Japanese, Korean, Russian, Spanish, Italian

### Title
Turbo Mode

### [Hed + Body Copy]
Turbo Mode helps you browse faster
Turbo Mode automatically blocks trackers and ads to help get you where you're going faster. If there's a speed bump – like a site that isn't behaving the way you expect – just turn Turbo Mode off.

### [Buttons]
Keep Turbo Mode Enabled
Turn Turbo Mode Off

lime124:
We also just use the string "Turbo Mode" in the browse lane. 

What are thoughts are on translating turbo mode vs keeping it in English?

Delphine:
We usually translate new features. Especially given that here we are dealing with a number of different scripts for these languages, so users would be lost.
I didn't realize there would be a new feature name introduced in these strings. Machine translation may not do well on that, but we can go for it. 
Anyways, once we have the l10n repos set up localizers will be able to go in and correct these strings.

Sdaswani:
@Delphine it is _sort_ of a new feature - it's really just Focus' tracking protection, renamed to 'Turbo Mode'. And @lime124 is suggesting perhaps we keep it 'Turbo Mode' in all languages, so maybe that helps a little?

> Machine translation may not do well on that, but we can go for it.
👍 🙏

Delphine:
After giving this more thought, I'd like to insist on translating "Turbo mode". Another example to add to what I've already mentioned above is that "Mode" in German means "Fashion" so they'd be getting "Turbo Fashion" the way it is.
"Turbo" is usually understood universally, it's especially the "mode" that I'm worried about :)

Delphine:
Here are the translations: 
https://public.etherpad-mozilla.org/p/l10n.fireTV

Sdaswani:
Thanks @Delphine - one of the engineers will get these integrated soon.

npark-mozilla:
@Delphine : just noticed the machine translation of Korean is really awkward (really difficult to understand too, thanks to the embedded sentence), is it possible for me to make a suggestion for it?  (I swear I won't put any nasty texts :) )

Delphine:
@npark-mozilla : Of course! Be my guest :) 
The only ones that were not machine translated are French, Italian and Simplified Chinese.
[...]

PR #1643, opened 2019-01-07
No issue: Remove robotranslations/ because not used.

liuche:
According to :Pike, this isn't used anywhere in the toolchain, so we
should remove it - https://bugzilla.mozilla.org/show_bug.cgi?id=1518294

Removing unused code (also not in Focus - it looks like this was used for machine translation but we don't need that anymore).

### Pull Request checklist


liuche:
Originally added in https://github.com/mozilla-mobile/firefox-tv/issues/334

mcomella:
@liuche fwiw, to save duplicated effort, this should have been the same reviewer as FFES.

Please remove .gitignore too! See https://github.com/mozilla-mobile/firefox-echo-show/pull/166#pullrequestreview-190044619

PR #1654, opened 2019-01-10
1651 espresso swap service locator

Baron-Severin:
### Pull Request checklist

Not a user facing change


app/src/androidTest/java/org/mozilla/tv/firefox/FirefoxTestApplication.kt
```diff
@@ -11,8 +11,23 @@ import org.mozilla.tv.firefox.utils.ServiceLocator
...
+ * [FirefoxTestApplication.createServiceLocator] should check here for any
+ * dependencies that must vary between tests, then return super if they are found
+ * to be null. This allows individual tests to set these properties, thus
+ * substituting their own fakes into tests.
+ *
+ * Note that Application#onCreate is called by Espresso before @Before blocks,
+ * so to use this class dependencies must be fulfilled from an init block in a test
+ */
+object TestProvider {
+    var pocketVideoRepo : PocketVideoRepo? = null
```
mcomella: This is fragile. Since the test `init` block is called after `Application.onCreate`, in order for this to work, it relies on:
- `pocketRepo` to be loaded lazily
- `pocketRepo` is not used in Application.onCreate (which would trigger lazy loading with super)

The fragility is that this will stop working for devs if their new dependencies don't meet the criteria or they change existing dependencies to not meet this criteria. I'd like to do better, if possible.

How crazy would it be to mutate the serviceLocator in test init, after `onCreate`, to use the new dependencies? It fixes the issues above but unfortunately may result in confusing situations where one serviceLocator is used for init and one is used for the test.

---

More involved option 3: I wonder if we can override the serviceLocator from within the test runner. I played around a bit and it looks like we can override `FirefoxTestRunner.onCreate`, parse the test arguments from the bundle (fragile!), use that to reach into the test class with reflection, and modify the application under test (maybe in `FirefoxTestRunner.callApplicationOnCreate`?).

A few things I tried that won't work:
- Adding a JUnit `RunListener`: these callbacks are called after Application.onCreate

Some things we can try:
- Overriding the classLoader via gradle testInstrumentationRunnerArgs (unclear if called before application on create)

app/src/androidTest/java/org/mozilla/tv/firefox/FirefoxTestApplication.kt
```diff
@@ -11,8 +11,23 @@ import org.mozilla.tv.firefox.utils.ServiceLocator
 class FirefoxTestApplication : FirefoxApplication() {
 
     override fun createServiceLocator() = object : ServiceLocator(this) {
-        override val pocketRepo = PocketRepoFaker.fakedPocketRepo
+        override val pocketRepo by lazy { TestProvider.pocketVideoRepo ?: super.pocketRepo }
```
mcomella: Why are we mutating individual items in the serviceLocator? It feels more straightforward, and less boilerplate, to swap in an entirely new ServiceLocator where `EachTestServiceLocator : ServiceLocator`.

app/src/androidTest/java/org/mozilla/tv/firefox/FirefoxTestApplication.kt
Baron-Severin: I think substituting a `ServiceLocator` after `onCreate` via reflection would cause more problems than it would solve, especially with things like Pocket that initialize content on app start.  I really like the idea about hooking into `FirefoxTestRunner#onCreate`, though.  I'm going to do some quick work on that and push up a POC.

Baron-Severin: Pushed some new commits with those changes.  Commits prior to Jan 14 have not changed and don't need to be re-reviewed.

app/src/androidTest/java/org/mozilla/tv/firefox/FirefoxTestRunner.kt
```diff
@@ -6,11 +6,46 @@ package org.mozilla.tv.firefox
...
+        return super.newApplication(cl, FirefoxTestApplication::class.java.name, context).also {
+            app = it
+        }
+    }
+
+    override fun onCreate(arguments: Bundle?) {
+        when (arguments.extractClass()) {
+            // Set up test class specific ServiceLocator
+            PocketBasicUserFlowTest::class.java.name,
+            PocketErrorTest::class.java.name -> {
```
mcomella: I don't like that we explicitly need to define every test class using subbed deps here. What if we:
```kotlin
class TestRunner {
    val testClass = arguments.extractClass()
    if (testClass is TestDependencyProvider) {
        app.overrideServiceLocator(testClass.testServiceLocator)
    }
}

class WhateverTest : TestDependencyProvider {
    override val testServiceLocator = ...
}
```

That way:
- All the test code is defined (at least initially) within its own file
- We don't have to maintain a giant list of explicit classes separate from the test itself

Baron-Severin: I like that change a lot.  I'll set it up.

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketVideoRepo.kt
```diff
@@ -40,9 +40,9 @@ open class PocketVideoRepo(
     }
 
     private val _feedState = MutableLiveData<FeedState>().apply {
-        // _feedState.value should always be initialized at the top of init,
-        // because we treat it as non-null
-        value = buildConfigDerivables.initialPocketRepoState
+        // setValue must be called from the main thread, and this repo is instantiated
+        // off of the main thread in Espresso tests, so postValue is used instead
```
mcomella: No need to add a comment: why is the repo initialized off the main thread? I don't actually know how threading works in tests (e.g. is the main thread different from the thread running the junit code?).

fwiw, I don't love that the testing harness is impacting our production code: an immediate set is much simpler to handle than the post, e.g. the posted value may not be received by the calling application if another value is posted quickly: is that a problem?. I'm not sure what a solution would be here though without better understanding the test threading.

app/src/androidTest/java/org/mozilla/tv/firefox/helpers/FakePocketVideoRepoProvider.kt
```diff
@@ -0,0 +1,41 @@
...
+import org.mozilla.tv.firefox.pocket.PocketVideoRepo
+import org.mozilla.tv.firefox.pocket.PocketViewModel
+import org.mozilla.tv.firefox.utils.BuildConfigDerivables
+
+/**
+ * Provides a fake [PocketVideoRepo] implementation for testing purposes.
+ *
+ * Any values pushed to [fakedPocketRepoState] will be immediately emitted.
+ */
+object FakePocketVideoRepoProvider {
```
mcomella: You should make this a class and instantiate it for each test: if the test runner doesn't the class state between tests, this will break and it's also harder to think about.

mcomella: nit: -> `CustomPocketFeedStateProvider` or similar.

I think `FakePocketVideoRepoProvider` is unclear about what specifically this class is doing: it's too generic. While generic is good for reusability, it's more complex. We can rename it later if the functionality expands.

mcomella: Out of scope of this PR but I don't love that this provider is not created with composable parts: i.e. both tests rely on this code but if they needed to change it (e.g. adding functionality), this class would need to change so both tests are affected by those changes (they're coupled together).

I haven't given enough thought about how to fix this but we can do it later as we write more tests, or run into this problem for real.

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketVideoRepo.kt
Baron-Severin: Test code can be run on the main thread using `ActivityTestRule#runOnUiThread`, but by default it runs on a thread created by our `AndroidJunitTestRunner`.  In this case, we need to create the object before `Application` is initialized, so the `ActivityTestRule` thread isn't yet available.

I agree with your discomfort here, and will give it more thought.  I don't think it should block the merge, but hopefully we can come up with a better solution in the future.

PR #1692, opened 2019-01-16
Closes #1690: add new intent telemetry

### LINKED ISSUE #1690: Add Telemetry Probe for new intents

As a Product Manager, I want tracking on new intents, so that I can see how many times Firefox is opened from another app (ie. youtube.com or YouTube Fling feature in future)

### What / Requirements

* Add probe for new intent

* We can see a count of the number of times a new intent is made

dnarcese:
### Pull Request checklist

Telemetry change

### DIFF
--- a/app/src/main/java/org/mozilla/tv/firefox/IntentValidator.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/IntentValidator.kt
@@ -65,7 +65,7 @@ object IntentValidator {
                 if (TextUtils.isEmpty(dataString)) {
                     return null // We can't create a session from an Intent without a URL.
                 }
-
+                TelemetryIntegration.INSTANCE.viewIntentEvent()
                 return ValidatedIntentData(dataString, Session.Source.ACTION_VIEW)
             }
             Intent.ACTION_SEND -> {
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/mozilla/tv/firefox/IntentValidator.kt
```diff
@@ -65,7 +65,7 @@ object IntentValidator {
                 if (TextUtils.isEmpty(dataString)) {
                     return null // We can't create a session from an Intent without a URL.
                 }
-
+                TelemetryIntegration.INSTANCE.viewIntentEvent()
```
Baron-Severin: I feel like we should be logging all intents (rather than just View ones), along with what their type is.  

The story specifies that YouTube fling events should be logged, for example, and if I understand correctly those are `ACTION_MAIN` intents.

dnarcese: We already log YouTube fling events. What other type is there besides view?

Baron-Severin: > We already log YouTube fling events.

I'm blind.  Thanks.

> What other type is there besides view?

I don't know if there are any now, I'd rather log all intents rather than risk missing an update that adds a new one.  This way feels a little brittle.  Then again, we don't want to update the YouTube fling event on short notice, and we don't want to double up on them either.  So I suppose we should leave it like this for now.

Issue #1693, opened 2019-01-16
Develop consistent assertion strategy


Baron-Severin:
### Why/User Benefit/User Problem

We currently have several different assertion strategies, including throwing exceptions on unexpected state, logging it to Sentry, and ignoring it completely.  These respectively lead to unnecessary production crashes, confusing Sentry reports that look like crashes, and developers being unaware of strange states in the app.  Improving this would help us to find problems and write more robust code.

### What / Requirements

Investigate how we should be handling assertions, write a small, reusable solution for making these more consistent.

~Develop an assertion class with a simple API, which will crash the app on a failed assertion *if in debug mode*, and log it to Sentry if in release.  These logs should clearly show that they represent a failed assertion, rather than a crash.~

mcomella:
I agree that we should investigate and agree upon a consistent assertion strategy. However...

> Develop an assertion class with a simple API, which will crash the app on a failed assertion if in debug mode, and log it to Sentry if in release.

This "what" section includes a solution and I think we should avoid prescribing one at the outset of this ticket.

## My opinion
Whether or not you should turn off asserts at runtime is a hotly contested issue. For the sake of discussion, here's my opinion.

I disagree with disabling assertions at runtime. Assertions are intended to solve the problem of letting the application go in to states the developer did not intend and notifying the developer when it does. Turning them off means the app can go into these states which can be bad user experiences such as:
- Data loss
- Nonfunctional application (e.g. its very challenging to restart apps on Fire TV so users can get stuck, unable to move)
- Very confusing or otherwise low quality experience that degrades our application and brand in a way we can't observe

The counterargument is that users can continue using the app instead of crashing but, as per my examples, that isn't always the case and sometimes it's worse.

Furthermore, with assertions disabled, the application we're developing in debug mode will be taking different branches in release mode: users in the wild may be getting different experiences than what we're developing with, especially if the assertions have unintentional side effects or offset a race condition just enough to not cause issues in debug mode.

Lastly, users may never notice an error if the application doesn't crash. This means we won't get user bug reports with steps to reproduce which means we may never have an STR. Without an STR, we may never fix the issue. Users may never notice this error in isolation - so not a big deal - but it may manifest itself to the user in combination with other unintended states - potentially a big deal. An error state that results from multiple unintended states is even more difficult to track down and probably results in complex code to manage when, if we crashed, we'd be able to fix the errors one-by-one.

### When I overturn my opinion
[...]

PR #1694, opened 2019-01-16
Closes #1691: fixes inconsistent state assertion

### LINKED ISSUE #1691: ScreenController Inconsistent state assertion on 3.2
ScreenController Inconsistent state, expected NAVIGATION_OVERLAY, got WEB_RENDER 

https://sentry.prod.mozaws.net/share/issue/074800e7d1f7460ea3be3412edc9194e/

Baron-Severin:
RCA:
When clearing data, the displayed screen was changed to NavigationOverlayFragment, however the ScreenController did not actually update its currentActiveScreen value. This caused a disconnect between expected and actual current state. At this time, no user impact has been noticed, but there could be strange transaction bugs related to this.

Fix:
Setting up fragments for a new session is guaranteed to display the NavigationOverlayFragment, so instead of asserting that state, the method now sets it.

Regressions:
To avoid regressions, we should work out a system for assertions that crash on debug and report to Sentry silently on release, then periodically assert app state throughout ScreenController to guarantee that no similar situations exist in the app.  Opened https://github.com/mozilla-mobile/firefox-tv/issues/1693

### Pull Request checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Fixed
 - Fixed Pocket Crash issue for non-EN locales (#1684)
 - Crash when using unrecognized remote control (#1685)
+- Fixed mismatch between internal and display state after clearing data (#1691)
 
 ## [3.2.2] - 2019-01-15
 ### Fixed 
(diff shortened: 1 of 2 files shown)


Baron-Severin:
@psymoon it's happening because `ScreenController` doesn't update its current screen when we clear data.  It's unrelated to whether or not we clear using `Activity.recreate`.

Another potential solution would be to create a new `ServiceLocator` any time `MainActivity` is created.  I think that might be a good idea, to avoid leaking information that we're trying to clear, but doing so could introduce bugs.  So that's something to think about later.

Issue #1683, opened 2019-01-15
Capture variant from split tests in Telemetry data


athomasmoz:
### Why/User Benefit/User Problem

As a Product Manager, I want to see & compare metrics from each variate of my split test so that I can determine which is better for my user

### What / Requirements

* All telemetry pings from user actions include the test variant

### Acceptance Criteria (how do I know when I’m done?)

* Dashboards can be filtered on a particular variant of a split test

PR #1701, opened 2019-01-21
Closes #1676: Add fretboard

### LINKED ISSUE #1676: Add Fretboard

As a PM, I want Fretboard set up on Firefox for Fire TV to run split testing, so that I can measure the impact of new features using A/B tests

### What / Requirements

* Fretboard is added to Firefox for Fire TV

* Fretboard package has been added to FFTV
* Determine package structure

dnarcese:
Waiting on https://bugzilla.mozilla.org/show_bug.cgi?id=1521161 before merging
### Pull Request checklist

### DIFF
--- a/app/src/main/java/org/mozilla/tv/firefox/FirefoxApplication.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/FirefoxApplication.kt
@@ -42,7 +42,11 @@ open class FirefoxApplication : LocaleAwareApplication() {
         PreferenceManager.setDefaultValues(this, R.xml.settings, false)
 
         serviceLocator = createServiceLocator()
-        serviceLocator.fretboardProvider.loadExperiments()
+
+        with(serviceLocator.fretboardProvider) {
+            updateExperiments()
+            loadExperiments()
+        }
 
         enableStrictMode()
 
@@ -51,8 +55,6 @@ open class FirefoxApplication : LocaleAwareApplication() {
         visibilityLifeCycleCallback = VisibilityLifeCycleCallback(this).also {
             registerActivityLifecycleCallbacks(it)
         }
-
-        serviceLocator.fretboardProvider.updateExperiments()
     }
 
     // ServiceLocator needs to be created in onCreate in order to accept Application
--- a/app/src/main/java/org/mozilla/tv/firefox/FirefoxApplication.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/FirefoxApplication.kt
@@ -43,14 +43,14 @@ open class FirefoxApplication : LocaleAwareApplication() {
 
         serviceLocator = createServiceLocator()
 
+        TelemetryIntegration.INSTANCE.init(this)
+
         with(serviceLocator.fretboardProvider) {
             updateExperiments()
             loadExperiments()
         }
 
         enableStrictMode()
-
-        TelemetryIntegration.INSTANCE.init(this)
 
         visibilityLifeCycleCallback = VisibilityLifeCycleCallback(this).also {
             registerActivityLifecycleCallbacks(it)
--- a/app/src/main/java/org/mozilla/tv/firefox/FirefoxApplication.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/FirefoxApplication.kt
@@ -42,6 +42,7 @@ open class FirefoxApplication : LocaleAwareApplication() {
         PreferenceManager.setDefaultValues(this, R.xml.settings, false)
 
         serviceLocator = createServiceLocator()
+        serviceLocator.fretboardProvider.loadExperiments()
 
         enableStrictMode()
 
@@ -50,6 +51,8 @@ open class FirefoxApplication : LocaleAwareApplication() {
         visibilityLifeCycleCallback = VisibilityLifeCycleCallback(this).also {
             registerActivityLifecycleCallbacks(it)
         }
+
+        serviceLocator.fretboardProvider.updateExperiments()
     }
 
     // ServiceLocator needs to be created in onCreate in order to accept Application
--- a/app/src/main/java/org/mozilla/tv/firefox/FirefoxApplication.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/FirefoxApplication.kt
@@ -46,8 +46,8 @@ open class FirefoxApplication : LocaleAwareApplication() {
         TelemetryIntegration.INSTANCE.init(this)
 
         with(serviceLocator.fretboardProvider) {
-            updateExperiments()
             loadExperiments()
+            updateExperiments()
         }
 
         enableStrictMode()
--- a/app/src/main/java/org/mozilla/tv/firefox/experiments/FretboardProvider.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/experiments/FretboardProvider.kt
@@ -1,0 +1,44 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.tv.firefox.experiments
+
+import android.content.Context
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.launch
+import mozilla.components.service.fretboard.Fretboard
+import mozilla.components.service.fretboard.ValuesProvider
+import mozilla.components.service.fretboard.source.kinto.KintoExperimentSource
+import mozilla.components.service.fretboard.storage.flatfile.FlatFileExperimentStorage
+import org.mozilla.tv.firefox.telemetry.TelemetryIntegration
+import java.io.File
+import kotlin.coroutines.CoroutineContext
+
+const val EXPERIMENTS_JSON_FILENAME = "experiments.json"
+const val EXPERIMENTS_BASE_URL = "https://settings.prod.mozaws.net/v1"
+const val EXPERIMENTS_BUCKET_NAME = "main"
+const val EXPERIMENTS_COLLECTION_NAME = "fftv-experiments"
+
+class FretboardProvider(private val applicationContext: Context) : CoroutineScope {
+    private var job = Job()
+    override val coroutineContext: CoroutineContext
+        get() = job + Dispatchers.Main
+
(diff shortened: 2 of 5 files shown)


app/src/main/java/org/mozilla/tv/firefox/FirefoxApplication.kt
```diff
@@ -50,6 +51,8 @@ open class FirefoxApplication : LocaleAwareApplication() {
         visibilityLifeCycleCallback = VisibilityLifeCycleCallback(this).also {
             registerActivityLifecycleCallbacks(it)
         }
+
+        serviceLocator.fretboardProvider.updateExperiments()
```
Baron-Severin: Both `loadExperiments` and `updateExperiments` are called in `onCreate`, is there any reason why they're separated like this?

app/src/main/java/org/mozilla/tv/firefox/experiments/FretboardProvider.kt
```diff
@@ -0,0 +1,44 @@
...
+    val fretboard: Fretboard = Fretboard(experimentSource, FlatFileExperimentStorage(experimentsFile),
+            object : ValuesProvider() {
+                override fun getClientId(context: Context): String {
+                    return TelemetryIntegration.INSTANCE.clientId
+                }
+            })
+
+    fun updateExperiments() = launch(Dispatchers.IO) { fretboard.updateExperiments() }
+
+    fun loadExperiments() = launch(Dispatchers.IO) { fretboard.loadExperiments() }
```
Baron-Severin: It's not immediately apparent how these act differently.  A small kdoc on each would be helpful, so we don't have to jump into Fetboard source just to find that out.

app/src/main/java/org/mozilla/tv/firefox/FirefoxApplication.kt
Baron-Severin: The chances of our network request finishing before our local storage retrieval are pretty minimal, but I still don't love that this is nondeterministic.  I think I would prefer if `updateExperiments` and `loadExperiments` were suspend funs, and our `onCreate` calls looked something like this

```kotlin
launch(Dispatchers.IO) {
  with(serviceLocator.fretboardProvider) {
    loadExperiments()
    updateExperiments()
  }
}
```

That way `loadExperiments` would block `updateExperiments`, and they would always complete in the same order.

app/src/main/java/org/mozilla/tv/firefox/experiments/FretboardProvider.kt
```diff
@@ -0,0 +1,44 @@
...
+            EXPERIMENTS_BASE_URL, EXPERIMENTS_BUCKET_NAME, EXPERIMENTS_COLLECTION_NAME
+    )
+    val fretboard: Fretboard = Fretboard(experimentSource, FlatFileExperimentStorage(experimentsFile),
+            object : ValuesProvider() {
+                override fun getClientId(context: Context): String {
+                    return TelemetryIntegration.INSTANCE.clientId
+                }
+            })
+
+    fun updateExperiments() = launch(Dispatchers.IO) { fretboard.updateExperiments() }
```
Baron-Severin: This should be a pretty quick call, can we measure how long it takes and see if we can block startup on it?  It would simplify our implementations down the line if we could always assume that our experiments have already been loaded.

Down the line we can parallelize this with other startup work (e.g., WebView init), so the startup time impact will drop from 'very small' to '0'.

dnarcese: I thought we decided that it was okay to use the currently loaded experiments, and then wait until the next app startup to use the updated ones? Or are you referring to something different

Baron-Severin: Apologies, I meant to leave this comment on `loadExperiments`.

app/src/main/java/org/mozilla/tv/firefox/experiments/FretboardProvider.kt
```diff
@@ -0,0 +1,51 @@
...
+    )
+    val fretboard: Fretboard = Fretboard(experimentSource, FlatFileExperimentStorage(experimentsFile),
+            object : ValuesProvider() {
+                override fun getClientId(context: Context): String {
+                    return TelemetryIntegration.INSTANCE.clientId
+                }
+            })
+
+    /**
+     * Requests new experiments from the server and
```
Baron-Severin: Nit: "Asynchronously requests new experiments..."

app/src/main/java/org/mozilla/tv/firefox/experiments/FretboardProvider.kt
```diff
@@ -0,0 +1,51 @@
...
+     * saves them to local storage
+     */
+    fun updateExperiments() = launch(Dispatchers.IO) { fretboard.updateExperiments() }
+
+    /**
+     * Loads experiments from local storage
```
Baron-Severin: Nit: "Synchronously loads experiments from local storage.  This is completed quickly, so we are comfortable blocking in order to reduce complexity by making experiments always available."

app/src/main/java/org/mozilla/tv/firefox/experiments/FretboardProvider.kt
```diff
@@ -0,0 +1,44 @@
...
+        get() = job + Dispatchers.Main
+
+    private val experimentsFile = File(applicationContext.filesDir, EXPERIMENTS_JSON_FILENAME)
+    private val experimentSource = KintoExperimentSource(
+            EXPERIMENTS_BASE_URL, EXPERIMENTS_BUCKET_NAME, EXPERIMENTS_COLLECTION_NAME
+    )
+    val fretboard: Fretboard = Fretboard(experimentSource, FlatFileExperimentStorage(experimentsFile),
+            object : ValuesProvider() {
+                override fun getClientId(context: Context): String {
+                    return TelemetryIntegration.INSTANCE.clientId
```
liuche: Why are we using Telemetry clientId here? I feel uncomfortable using the same telemetry id, because it adds an unnecessary dependency on the Telemetry client id.
@dnarcese can we just generate an id just for Fretboard instead?

psymoon: @liuche https://github.com/mozilla-mobile/firefox-tv/issues/1683 ?

dnarcese: I think we need it to be the same id so that we can link user actions with which experiments they are in. That telemetry will be collected in #1683

liuche: We should be able to use `Telemetry.recordActiveExperiments` with `Fretboard.getActiveExperiments`, which means we don't need to send the telemetry id and it decouples that dependency (and just explicitly sends the active experiments).
However, let me double check:
@pocmo is there a reason why Focus overrides the Experiment clientId to use the TelemetryClientId rather than using the default behavior of generating a random experiment id?

app/src/main/java/org/mozilla/tv/firefox/FirefoxApplication.kt
pocmo: Please file this in AC. Happy to make those suspension methods (or add separate ones that suspend).

Other than that: Yes, you'll absolutely want to load before updating since the update call will only request a diff from the server to minimize traffic.

app/src/main/java/org/mozilla/tv/firefox/experiments/FretboardProvider.kt
pocmo: First of all, there are two options:
* Do not provide anything and fretboard will generate a UUID; no need to do that yourself
* Provide a different ID in case you want to attach it to some other system; like telemetry

We added this option here:
* https://github.com/mozilla-mobile/android-components/issues/906
* Related Focus issue: https://github.com/mozilla-mobile/focus-android/issues/3489

I think the primary reason was that this allowed the data team to verify and reproduce the bucketing on the server side. I agree that for evaluating experiments we provide everything by other means:
* You can record the list of active experiments (in core ping) -> `recordActiveExperiments`
* You can record a map of **all** experiments the client knows about and whether they are active or not (in events ping; recording all experiments allows you to distinguish between experiments the client is not part of and experiments the client just doesn't know about yet). -> `recordExperiments`
* You can attach additional data like the "bucket" via the `settings` map in the events ping

So, yeah, I agree that the telemetry ID should not be needed. But please consult with the data team since they requested this for the A/A test originally. In any case the ID is only needed for determining the bucket locally and is never send to a server by fretboard. For the same reason it's good to bring that up now since changing the ID after release will shuffle buckets around and may skew your data.

app/src/main/java/org/mozilla/tv/firefox/FirefoxApplication.kt
Baron-Severin: Done!

https://github.com/mozilla-mobile/android-components/issues/1855

app/src/main/java/org/mozilla/tv/firefox/experiments/FretboardProvider.kt
liuche: Pinged @fbertsch and @mmccorks in #telemetry-mobile to ask about this.

The context behind this: Data wants to be able to reproduce buckets server-side in order to verify that clients are being bucketed correctly. That seems a reasonable ask. I'd like to hear back from them about whether it's acceptable to send another persistent, randomly generated id. However, the risk of reusing the clientid for Fretboard bucketing could be pretty low, because a) fretboard/telemetry ids go through Mozilla servers (telemetry and kinto) and b) by themselves, telemetry clientid + bucket don't leak personally identifiable information.

@dnarcese let's make a decision on Monday, and whichever case we decide we should document it in-code (either here where we reuse the telemetry clientid, or wherever we upload the randomly generated id).

georgf: For possible context for a decision:
My team is taking ownership of providing an Android experimentation library by Q2 (targeted at Fenix first as the priority project, [planning doc](https://docs.google.com/document/d/1gpqCdq5qbG7_yLMjeM3MPySsoibYytn6lOCrrPq5bNU/))
I'll track this as an item / use-case to consider when going over the design and APIs.

liuche: We discussed this during our standup, and we decided to land this as-is - if we decide that we need to change this from Telemetry ID, we still have time to add a new id before we start experiments, but we don't want to block on that since it's also reversible until we start actually running experiments.

PR #1709, opened 2019-01-24
Bump version.

liuche:
Version bump & changelog
### Pull Request checklist

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -25,7 +25,7 @@ android {
         targetSdkVersion 28
         versionCode 11 // This versionCode is "frozen" for local builds. For "release" builds we
         // override this with a generated versionCode at build time.
-        versionName "3.2.2"
+        versionName "3.2.6.1"
         testInstrumentationRunner "org.mozilla.tv.firefox.FirefoxTestRunner"
         testInstrumentationRunnerArguments clearPackageData: 'true'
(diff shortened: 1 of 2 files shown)


app/build.gradle
```diff
@@ -25,7 +25,7 @@ android {
         targetSdkVersion 28
         versionCode 11 // This versionCode is "frozen" for local builds. For "release" builds we
         // override this with a generated versionCode at build time.
-        versionName "3.2.2"
+        versionName "3.2.6.1"
```
liuche: I'm not sure if we should just bump per major/minor version? Or have a commit for all releases. In any case, when we do the 3.3 release, I'll update this version too, but we need the CHANGELOG changes.

Baron-Severin: Our releases tab is getting pretty hard to read.  Suggest updating this to include pre-release metadata as per https://semver.org/#spec-item-9 (e.g., `3.2.6-LAT.1`).

Baron-Severin: Not sure I'm understanding you correctly, I do think we need to tag everything that goes out to users differently though.  There will be a lot of headings under our releases tab, but I think it is important to keep that record.

liuche:
Closing this for now, I'll make a new PR with the suggested changes.

PR #1713, opened 2019-01-24
Closes #1362: fixes StrictMode violations preventing ui tests

### LINKED ISSUE #1362: [1364] Run UI/integration tests on phone/tablet emulators
Right now, we can't run our tests on CI. They get run at least once a day off of master. However, this means that the feedback cycle is delayed, it's unclear whose change broke the tests, and no one feels ownership over fixing them: they don't get fix. When the tests break, it is a sign that the user experience is broken (assuming the tests are reliable, and they should be!) so the bustage needs to be investigated!

We can run our tests on CI on phone/tablet emulators. However, because we intend to run on TV emulators, there are issues to solve.

### What / Requirements
- Figure out how to open the overlay on phone/tablet emulators (menu button on these devices opens recent apps list)
- Identify and fix any other issues with UI/integration tests on these emulators
- Talk to QA about getting these tests run on CI

Baron-Severin:
### Pull Request checklist

Test only change

### DIFF
--- a/app/src/androidTest/java/org/mozilla/tv/firefox/FirefoxTestApplication.kt
+++ b/app/src/androidTest/java/org/mozilla/tv/firefox/FirefoxTestApplication.kt
@@ -6,6 +6,7 @@ package org.mozilla.tv.firefox
 
 import android.arch.lifecycle.LiveData
 import android.arch.lifecycle.MutableLiveData
+import android.os.StrictMode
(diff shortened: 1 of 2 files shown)


Baron-Severin:
@liuche StrictMode is disabled entirely in production, so I think we're OK there.  This is causing failures on all tests (but only on Remote TestKit), so unfortunately we can't disable per test.

Issue #1700, opened 2019-01-19
Fix ToggleDataCollectionTest#WHEN_data_collection_button_is_toggled_THEN_button_state_persists_when_returning_to_settings


Baron-Severin:
I noticed that this is broken on master.  This needs to be fixed (or at the very least ignored) before we can get ui tests on CI.

liuche:
I'll pick up this bug - I wonder if I broke this while resolving conflicts during the rebase.

liuche:
No need to QA, this should just pass now in the UI tests.

AaronMT:
Awesome, thanks!

Issue #1664, opened 2019-01-11
Refactor overlay into split


athomasmoz:
Refactor overlay into split

    This will involve adjusting fragment transactions so that WebRenderFragment is visible

mcomella:
I haven't made progress beyond the initial PR that landed: feel free to take this (but if it's untaken, I can tackle it again tomorrow).

liuche:
I'm also sending @brampitoyo a build so I can get some feedback on various sizing of the toolbar, window space, etc, but since this is landing turned off I can file the polish in a later bug.

PR #1746, opened 2019-01-29
Issue #1664: Pre-work: move toolbar code into ToolbarUiController

mcomella:
This will make NavOverlayFragment a little less unwieldy. I didn't move the code for the bottom part of the navigation overlay (after the split) because that'll get refactored to channels and my changes may not make sense.

### Pull Request checklist

  - No functionality changes
- [x] The **UI tests** are passing after this PR
  - #1700 fails
  - OnboardingLaunchTest fails but also fails on master. Filed #1748

### DIFF
--- a/app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/NavigationOverlayFragment.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/NavigationOverlayFragment.kt
@@ -149,8 +149,7 @@ class NavigationOverlayFragment : Fragment() {
         ToolbarUiController(
             toolbarViewModel,
             ::exitFirefox,
-            { currFocus },
-            ::updateFocusableViews,
+            { updateFocusableViews() },
             onNavigationEvent
         ).onCreateView(view, viewLifecycleOwner, fragmentManager!!)
 
--- a/app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/NavigationOverlayFragment.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/NavigationOverlayFragment.kt
@@ -74,7 +74,6 @@ enum class NavigationEvent {
     }
 }
 
-@Suppress("LargeClass") // TODO remove this. See https://github.com/mozilla-mobile/firefox-tv/issues/1187
 class NavigationOverlayFragment : Fragment() {
     companion object {
         const val FRAGMENT_TAG = "overlay"
--- a/app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/NavigationOverlayFragment.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/NavigationOverlayFragment.kt
@@ -32,7 +32,6 @@ import kotlinx.coroutines.Job
 import org.mozilla.tv.firefox.MainActivity
 import org.mozilla.tv.firefox.R
 import org.mozilla.tv.firefox.experiments.ExperimentConfig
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/ToolbarUiController.kt
```diff
@@ -0,0 +1,158 @@
...
+import org.mozilla.tv.firefox.widget.IgnoreFocusMovementMethod
+import org.mozilla.tv.firefox.widget.InlineAutocompleteEditText
+
+private const val NAVIGATION_BUTTON_ENABLED_ALPHA = 1.0f
+private const val NAVIGATION_BUTTON_DISABLED_ALPHA = 0.3f
+
+/**
+ * An encapsulation of the toolbar to set up and respond to UI operations.
+ */
+class ToolbarUiController(
```
psymoon: This is an interesting concept and I agree with benefits you've listed; I do question if adding another layer of encapsulation adds more complications to our current architecture. What do you think about creating a custom `ToolbarView` instead? I think it's also worth looking into dependency injection in the near future. @Baron-Severin might also have some good inputs on this as well

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/NavigationOverlayFragment.kt
```diff
@@ -84,7 +74,7 @@ enum class NavigationEvent {
 }
 
 @Suppress("LargeClass") // TODO remove this. See https://github.com/mozilla-mobile/firefox-tv/issues/1187
```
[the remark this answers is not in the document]
mcomella: Yes: done.

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/ToolbarUiController.kt
mcomella: > I do question if adding another layer of encapsulation adds more complications to our current architecture.

I don't consider this to be an extension of our architecture – it's a grouping of some of the parts of a single architectural component – the NavigationOverlayFragment – necessitated by  the file growing too large. The ToolbarUiController is intended to have same responsibility in the architecture as the Fragment – it's just focused on a specific part of the fragment.

fwiw, the CursorController _kind of_ acts in a similar way: it's a grouping of Cursor functionality that is an extension of WebRenderFragment.

I can see how one might think this is another part of the architecture to understand – especially because it's called "Controller" – and the problems that might entail. I think this could be fixed with a better name but I couldn't come up with one though (e.g. `ToolbarSubFragment` isn't quite right because it isn't really a fragment). I used `Controller` because it evokes a similar responsibility to the Controller in MVC: it (more or less) hooks up the view to the model. Let me know if you can think of a better name (`ToolbarFragmentExtension`?).

> What do you think about creating a custom ToolbarView instead?

I think this would have a larger impact on our architecture: the view would now take on some responsibilities that the fragment current encapsulates and generally our views are pretty dumb.

> I think it's also worth looking into dependency injection in the near future.

This seems unrelated so we can talk about that elsewhere.

mcomella: We can follow-up in a new bug if this discussion makes us decide to change the code.

mcomella:
I didn't realize my IDE had indented some lines that were generally unrelated to my change: I'm going to indent them back, fold those commits into the appropriate place, and force push over the content. To save time, I'm not going to go through re-review.

@psymoon ^ If you notice these types of changes, please notify the assignee! We don't want to break blame on unnecessary formatting changes if possible.

Issue #1670, opened 2019-01-13
Unable to disable "send usage data"


mcl21014:
Ticking off the box results in no change. Running Firefox for Fire TV 3.2 on a Fire Stick 4k with the latest updates installed.

AaronMT:
Oof, yeah I see this too.

Sdaswani:
@psymoon @dnarcese can you tag team on this ASAP please?

PR #1750, opened 2019-01-30
Closes #1700 - Fix broken Data Toggle test and a11y

### LINKED ISSUE #1700: Fix ToggleDataCollectionTest#WHEN_data_collection_button_is_toggled_THEN_button_state_persists_when_returning_to_settings

liuche:
This code is annoyingly complex because we have two code paths for using VoiceView and not.

The fix in #1670 actually broke data toggling when VoiceView was enabled (because that is controlled directly through the checkbox, so it would never allow the checkbox to be changed) and the previous fix worked because it pushed a checkbox change through the SessionRepo. The fix here toggles the checkbox when the container being clicked.

### Pull Request checklist

- [X] The **UI tests** are passing after this PR

### DIFF
--- a/app/src/androidTest/java/org/mozilla/tv/firefox/ui/ToggleDataCollectionTest.kt
+++ b/app/src/androidTest/java/org/mozilla/tv/firefox/ui/ToggleDataCollectionTest.kt
@@ -26,6 +26,41 @@ class ToggleDataCollectionTest {
         settingsRepo = activityTestRule.activity.application.serviceLocator.settingsRepo
     }
 
+    /**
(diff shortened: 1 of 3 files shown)


liuche:
To summarize, it seems like this patch isn't working on the Fire OS version 6.2.3.8 4k pendant, but works on 6.2.3.5 on Cube and 4k Pendant. A few things we're trying:
- Adding a test "version" change to make sure the build is updating correctly
- Testing on another device
We're wondering if there is something changed with VoiceView on the the new Fire OS?

Issue #1628, opened 2019-01-01
Investigate potential memory leak in state restoration (i.e. don't keep activities)


mcomella:
See also FFES: https://github.com/mozilla-mobile/firefox-echo-show/issues/153

## Steps to reproduce
- On Android tablet emulator, Enable don't keep activities
- Open app
- Press home
- Open app
- Open profiler
- Force GC
- Take heap dump
- Open heap dump, search for MainActivity

### Expected behavior
One instance of MainActivity

### Actual behavior
Multiple instances of MainActivity

I don't know if this is a real memory leak or a problem with the Android profiler's representation of our memory state. We can consider adding LeakCanary, which should make it immediately obvious if there's a leak.

### Device information
* Fire TV device: Android tablet emulator
* Firefox version: local build, master

mcomella:
User consequences: there are a few cases:
- Users with long running sessions (including opening/closing the app) may have Firefox crash on them when we run out of memory
- Users with long running sessions may have Firefox be killed in the background, making it take longer to start up
- Other apps may be starved for memory so will be killed and take longer to start up
- Firefox (and other apps) may become slower and slower as it becomes more challenging to allocate necessary memory

This may be causing all of our out-of-memory errors.

mcomella:
I notice that `AccessibilityManager.addTouchExplorationStateChangeListener` adds listeners with strong references:
- http://androidxref.com/9.0.0_r3/xref/frameworks/base/core/java/android/view/accessibility/AccessibilityManager.java#732
- http://androidxref.com/9.0.0_r3/xref/frameworks/base/core/java/android/view/accessibility/AccessibilityManager.java#142

In Echo Show, I do not remove these listeners in `BrowserAppBarLayoutController`, which could be causing the memory leak.

mcomella:
I'm still seeing this issue.

mcomella:
I have found at least one source:
https://github.com/mozilla-mobile/firefox-tv/blob/master/app/src/main/java/org/mozilla/tv/firefox/telemetry/SentryIntegration.kt#L32-L44

This anonymous class instance is getting bound to the Activity context on `observeForever`. Another problem with this code is that a new listener will be added each time the Activity is recreated.

mcomella:
After fixing this leak, I'm still getting memory leaks:
![screenshot_1548876985](https://user-images.githubusercontent.com/759372/52007439-4cd73a00-2483-11e9-9b8b-b3155390efbc.png)

[...]

PR #1762, opened 2019-01-30
Closes #750, Issue #1628: Move Sentry init to application

### LINKED ISSUE #750: Sentry: consider moving init to Application
e.g. we're missing crash reporting for OnboardingActivity

The docs put it in an Activity but that would require multiple inits and it seems fine to put it in application since it uses the applicationContext.

mcomella:
We verified with the Sentry team that this is an acceptable way to
initialize Sentry.

This change is beneficial because:
- It fixes a memory leak (not completely though: there are still things to follow-up on)
- It allows all Activites to get crash reporting
- We get crash reporting earlier in the init process

It's concerning because:
- we're modifying crash reporting code
- Sentry is slow to start up so this may further affect startup time

This change could be better if:
- We could initialize Sentry before the ServiceLocator is initialized
and before the PreferenceManager default values get set

I verified this change was working correctly by adding code to crash the
app (throw when M key is pressed) and ensuring the events were received
by the server. I also tested events are not received by the server when
data collection is disabled.

### Pull Request checklist

  - No real issues fixed for users
  - #1700 is a known failure

### DIFF
--- a/app/src/main/java/org/mozilla/tv/firefox/telemetry/SentryIntegration.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/telemetry/SentryIntegration.kt
@@ -5,10 +5,12 @@
 package org.mozilla.tv.firefox.telemetry
 
 import android.content.Context
+import android.support.annotation.VisibleForTesting
+import android.support.annotation.VisibleForTesting.NONE
 import io.sentry.Sentry
 import io.sentry.android.AndroidSentryClientFactory
 import org.mozilla.tv.firefox.BuildConfig
-import org.mozilla.tv.firefox.ext.serviceLocator
+import org.mozilla.tv.firefox.utils.ServiceLocator
 
 /**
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/mozilla/tv/firefox/telemetry/SentryIntegration.kt
```diff
@@ -28,8 +30,20 @@ import org.mozilla.tv.firefox.ext.serviceLocator
...
-        context.serviceLocator.settingsRepo.dataCollectionEnabled.observeForever { isEnabled ->
+
+    @VisibleForTesting(otherwise = NONE)
+    var isInit = false
+        private set
+
+    /**
+     * Initializes Sentry. This method should only be called once.
+     */
+    fun init(appContext: Context, serviceLocator: ServiceLocator) {
```
Baron-Severin: If we only need the locator for the `SettingsRepo`, could we pass that instead of the `ServiceLocator`?

PR #1768, opened 2019-01-31
Closes #1744: EngineView to remove its child webview prior to onDestroy

### LINKED ISSUE #1744: Remove webview from its parent view hiearchy on webview.destroy()
We call `Engineview#onDestroy` (which calls `webview.destroy()`) in `WebViewCache#onDestroy`

https://developer.android.com/reference/android/webkit/WebView.html#destroy() shows that
> Destroys the internal state of this WebView. This method should be called after this WebView has been removed from the view system. No other methods may be called on this WebView after destroy. 

We should remove webview from its parent view hiearchy before destroying to ensure there is no memory leak

psymoon:
### Pull Request checklist

### DIFF
--- a/app/src/main/java/org/mozilla/tv/firefox/webrender/WebViewCache.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/webrender/WebViewCache.kt
@@ -93,6 +93,8 @@ class WebViewCache(private val sessionRepo: SessionRepo) : LifecycleObserver {
 
     private fun clear() {
         sessionRepo.canGoBackTwice = null
+        cachedView?.removeAllViews()
+        cachedView?.onPause()
         cachedView?.onStop()
         cachedView?.onDestroy()
         cachedView = null


app/src/main/java/org/mozilla/tv/firefox/webrender/WebViewCache.kt
```diff
@@ -93,6 +93,8 @@ class WebViewCache(private val sessionRepo: SessionRepo) : LifecycleObserver {
 
     private fun clear() {
         sessionRepo.canGoBackTwice = null
+        cachedView?.removeAllViews()
+        cachedView?.onPause()
         cachedView?.onStop()
         cachedView?.onDestroy()
         cachedView = null
```
Baron-Severin: Nit: might be time to reformat this a bit.  Suggest:

```kotlin
cachedView?.apply {
  removeAllViews()
  onPause()
  onStop()
  onDestroy()
}
cachedView = null
```

psymoon:
@csadilek is updating `SystemEngineView#onDestroy` to do this, so no need to merge;

PR #1777, opened 2019-02-01
Closes #1761, Issue #1628: Remove and replace sessions when Activity …

### LINKED ISSUE #1761: WebView stops working with Don't Keep Activities enabled, caused by components WebView memory leak
## Steps to reproduce
- Enable Don't Keep Activities
- Open app
- Open page from home tiles
- Return to home screen (leaving app open)
- Open app again
- Try to navigate to new page

### Expected behavior
WebView goes to new page

### Actual behavior
WebView does not respond to new events

### Device information

* Fire TV device: Android TV emulator API 26; have not tried on device
* Firefox version: master

Could be indicative of problems in our WebView state restoration code.

mcomella:
…recreates.

This address a memory leak in components:
https://github.com/mozilla-mobile/android-components/issues/1915

I have tested that regular browsing, opening and closing the app, and
clear data work with:
- Don't keep activities on a TV emulator
- On 4K device
- Don't keep activities on tablet emulator + LeakCanary does not display
this leak

### Pull Request checklist

  - We haven't finished fixing all the memory leaks yet so no need for changelog entry

### DIFF
--- a/app/src/main/java/org/mozilla/tv/firefox/MainActivity.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/MainActivity.kt
@@ -61,7 +61,10 @@ class MainActivity : LocaleAwareAppCompatActivity(), OnUrlEnteredListener, Media
 
         val intentData = IntentValidator.validateOnCreate(this, intent.toSafeIntent(), savedInstanceState)
 
+        // This restores the SessionManager sessions so must be called before getOrCreateSession.
+        webRenderComponents.applicationLifecycleSessionCache.onCreateMaybeRestoreSessions()
         val session = getOrCreateSession(intentData)
+
         val screenController = serviceLocator.screenController
         screenController.setUpFragmentsForNewSession(supportFragmentManager, session)
 
@@ -88,6 +91,13 @@ class MainActivity : LocaleAwareAppCompatActivity(), OnUrlEnteredListener, Media
         }
 
         serviceLocator.intentLiveData.value = Consumable.from(intentData)
+    }
+
+    override fun onDestroy() {
+        super.onDestroy()
+
+        // This removes the Sessions from the SessionManager so should be called last.
+        webRenderComponents.applicationLifecycleSessionCache.onDestroyCacheSessions()
     }
 
     @SuppressLint("MissingSuperCall")
--- a/app/src/main/java/org/mozilla/tv/firefox/components/ApplicationLifecycleSessionCache.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/components/ApplicationLifecycleSessionCache.kt
@@ -1,0 +1,57 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.tv.firefox.components
+
+import android.support.annotation.VisibleForTesting
+import android.support.annotation.VisibleForTesting.PRIVATE
+import mozilla.components.browser.session.Session
+import mozilla.components.browser.session.SessionManager
+
+/**
+ * An object that caches [Session]s in place of the [SessionManager] after the
+ * [mozilla.components.concept.engine.EngineView] owner's lifecycle has ended and restores the sessions when the
+ * EngineView is reattached. This is necessary because Android Components has a memory leak where [Session]s will keep a
+ * reference to their system WebView. They null the WebView reference when [Session]s are removed from the
+ * [SessionManager] so, as a workaround, we manually remove the [Session]s at the end of the lifecycle and re-add them
+ * when a new lifecycle begins.
+ *
+ * See more details in the issue mentioned below:
+ * TODO: implement the components solution in android-components#1915, remove this class and its calls
+ *
+ * We don't use lifecycle observers in this class because each method must be called at a precise time (see each
+ * lifecycle method's kdoc for details).
+ */
+class ApplicationLifecycleSessionCache(
+    private val sessionManager: SessionManager
+) {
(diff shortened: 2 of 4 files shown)


app/src/main/java/org/mozilla/tv/firefox/MainActivity.kt
```diff
@@ -90,6 +93,13 @@ class MainActivity : LocaleAwareAppCompatActivity(), OnUrlEnteredListener, Media
         serviceLocator.intentLiveData.value = Consumable.from(intentData)
     }
 
+    override fun onDestroy() {
+        super.onDestroy()
+
+        // This removes the Sessions from the SessionManager so should be called last.
+        webRenderComponents.applicationLifecycleSessionCache.onDestroyCacheSessions()
+    }
```
Baron-Severin: Are there any potential edge cases here if the `Activity` is destroyed without hitting `onDestroy`?

app/src/main/java/org/mozilla/tv/firefox/components/ApplicationLifecycleSessionCache.kt
```diff
@@ -0,0 +1,63 @@
...
+
+package org.mozilla.tv.firefox.components
+
+import android.support.annotation.VisibleForTesting
+import android.support.annotation.VisibleForTesting.PRIVATE
+import mozilla.components.browser.session.Session
+import mozilla.components.browser.session.SessionManager
+
+/**
+ * A object that caches [Session]s in place of the [SessionManager] after the
```
Baron-Severin: Nit: 'A object' -> 'An object'

app/src/main/java/org/mozilla/tv/firefox/components/ApplicationLifecycleSessionCache.kt
```diff
@@ -0,0 +1,63 @@
...
+ * See more details in the issue mentioned below:
+ * TODO: implement the components solution in android-components#1915
+ *
+ * We don't use lifecycle observers in this class because each method must be called at a precise time (see each
+ * lifecycle method's kdoc for details).
+ *
+ * TODO
+ * - test clear data, on device.
+ * - Check session repo for correctness.
+ * - Will all our onDestroy methods work if the WebView is hidden in onStop?
```
Baron-Severin: Is this TODO still valid?

app/src/main/java/org/mozilla/tv/firefox/components/ApplicationLifecycleSessionCache.kt
```diff
@@ -0,0 +1,63 @@
...
+/**
+ * A object that caches [Session]s in place of the [SessionManager] after the
+ * [mozilla.components.concept.engine.EngineView] owner's lifecycle has ended and restores the sessions when the
+ * EngineView is reattached. This is necessary because Android Components has a memory leak where [Session]s will keep a
+ * reference to their system WebView. They null the WebView reference when [Session]s are removed from the
+ * [SessionManager] so, as a workaround, we manually remove the [Session]s at the end of the lifecycle and re-add them
+ * when a new lifecycle begins.
+ *
+ * See more details in the issue mentioned below:
+ * TODO: implement the components solution in android-components#1915
```
Baron-Severin: Will this class be removed when A-C fixes this?  Please specify either way.

app/src/test/java/org/mozilla/tv/firefox/components/ApplicationLifecycleSessionCacheTest.kt
```diff
@@ -0,0 +1,79 @@
...
+
+        cache.onCreateMaybeRestoreSessions()
+
+        verifySessionManagerAdd(expectedTimes = expected.size)
+    }
+
+    private fun addSessionsToCache(): List<Session> {
+        return listOf(mock(Session::class.java), mock(Session::class.java)).also { sessions ->
+            sessions.forEach { cache.cachedSessions.add(it) }
+        }
```
Baron-Severin: Nit: personally, I would remove the side effect from this method for clarity, and load the cache in the test itself (or a separate helper method exclusively devoted to the side effect).

app/src/main/java/org/mozilla/tv/firefox/components/ApplicationLifecycleSessionCache.kt
mcomella: No: I forgot to commit these deletions apparently.

app/src/main/java/org/mozilla/tv/firefox/MainActivity.kt
mcomella: If the Activity is destroyed without onDestroy getting called but the process is not destroyed then yes, the memory leak persists. After doing some reading, I was under the impression that this can't happen but the Android docs aren't explicit enough on it that I'd trust them 100%.

The original implementation was in `onStart/onStop` and also `onCreate` (in the case the activity is recreated, we need to add it before we try to use it) but it was fairly complex (e.g. what happens if someone tries to access the session after onStop?) so I prefer the simplicity of this but I understand it might not always be correct.

I'll see if we can add telemetry to detect the case where onCreate is called but the leak still exists.

app/src/test/java/org/mozilla/tv/firefox/components/ApplicationLifecycleSessionCacheTest.kt
mcomella: I removed the method entirely: the code was not used in more than one place.

app/src/main/java/org/mozilla/tv/firefox/MainActivity.kt
mcomella: @Baron-Severin I implemented a Sentry error when this happens: please review the new commit.

Issue #1780, opened 2019-02-04
Test Orchestrator does not reset state within a file


npark-mozilla:
ToggleDataCollectionTest is failing on cloud (Bitbar, Firebase) only, because it does not reset the app state between tests that are residing in a same file.  The workaround is to set the flags back at the end of each test.

npark-mozilla:
Both cloud test frameworks now support resetting states

PR #1781, opened 2019-02-04
Reset the app state for running in Cloud Test

npark-mozilla:
Addressing #1780

### DIFF
--- a/app/src/androidTest/java/org/mozilla/tv/firefox/ui/ToggleDataCollectionTest.kt
+++ b/app/src/androidTest/java/org/mozilla/tv/firefox/ui/ToggleDataCollectionTest.kt
@@ -58,6 +58,7 @@ class ToggleDataCollectionTest {
         }.exitToOverlay {
         }.openSettings {
             assertDataCollectionButtonState(cachedDataCollectionButtonIsChecked)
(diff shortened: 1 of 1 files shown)


app/src/androidTest/java/org/mozilla/tv/firefox/ui/ToggleDataCollectionTest.kt
```diff
@@ -58,6 +58,7 @@ class ToggleDataCollectionTest {
         }.exitToOverlay {
         }.openSettings {
             assertDataCollectionButtonState(cachedDataCollectionButtonIsChecked)
+            toggleDataCollectionButton() // For resetting the data on Cloud Test
```
liuche: I see, so this is for manually "undoing" the data collection click?

Can you make the comment more specific, like // Manually undo data collection button click until Issue #XXX to correctly clear data between tests with Test Orchestrator is resolved

Baron-Severin:
We should try to reset this state in an @After block if possible.  That kind of block is frequently used for data cleanup, and it will be easier to understand what's going on.  I'm running into a 1:1 now, I'll get a PR up shortly after that finishes.

npark-mozilla:
Turns out both platforms support reset, so removing this PR.

PR #1792, opened 2019-02-05
WIP: Enabling Firebase UI test on PR

npark-mozilla:

### DIFF
--- a/tools/taskcluster/execute-firebase-test.sh
+++ b/tools/taskcluster/execute-firebase-test.sh
@@ -1,0 +1,61 @@
+#!/usr/bin/env bash
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+# This script installs google cloud sdk, logs into google firebase, executes tests,
+# and collects test artifacts into the test_artifacts folder
+
+# If a command fails then do not proceed and fail this script too.
+set -ex
+#########################
+# The command line help #
+#########################
+display_help() {
+    echo "Usage: $0 Build_Variant Apk_Name Device_Config [Device_Config...]"
+    echo
+    echo "Do not put .apk prefix in Apk_Name"
+    echo "Device_Config should be in landscape mode (e.g. model=taimen,version=26,orientation=landscape)"
+    echo "Example: execute-firebase-test.sh debug app-debug model=taimen,version=26,orientation=landscape"
+    echo
+}
+
+# Basic parameter check
+if [[ $# -lt 3 ]]; then
+    echo "Your command line contains $# arguments"
+    display_help
+    exit 1
+fi
+
+# Get test configuration
+config_count=`expr $# - 2`
+config_param="--device $3"
+for ((i=4; i <= $#; i++))
+{
+    config_param+=" --device ${!i}"
+}
+
+# From now on disable exiting on error. If the tests fail we want to continue
+# and try to download the artifacts. We will exit with the actual error code later.
+set +e
+
+# Execute test set
+./google-cloud-sdk/bin/gcloud --format="json" firebase test android run \
+--type instrumentation \
+--app ./app/build/outputs/apk/$1/$2.apk \
+--test ./app/build/outputs/apk/androidTest/$1/$2-androidTest.apk \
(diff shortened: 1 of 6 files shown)


mcomella:
Unfortunately, I have no idea how TC and TC scopes/secrets work. If Johan doesn't get back to you, please let me know No-Jun! fwiw, liuche and Sebastian know a lot more about TC if you wanted to ask them.

JohanLorenzo:
> also pinging @JohanLorenzo about this too, I also wonder whether we should be using `project/taskcluster/tc-queue/production` for pull requests? (that one has access to secrets)

Chatted over IRC with No-Jun. I noticed there were 2 sets of secrets: `project/firefox-tv/token` and `project/mobile/firefox-tv/token`. The latter contained a subset of the former. These the exact same tokens. So I suggested to No-Jun to copy the remainings token to the latter. I prefer using the latter so the android projects are all under the namespace `mobile`. 

I just created the role [`repo:github.com/mozilla-mobile/firefox-tv:pull-request`](https://tools.taskcluster.net/auth/roles/repo%3Agithub.com%2Fmozilla-mobile%2Ffirefox-tv%3Apull-request) and it now contains `secrets:get:project/mobile/firefox-tv/tokens`. If you edit the PR to reference this secret, then taskcluster shouldn't complain anymore.

Regarding `project/taskcluster/tc-queue/production`, that's a misleading message. Basically, `project/taskcluster/tc-queue/production` is the client that reads `.tasklcuster.yml` and spins the task(s). This client finds `repo:github.com/mozilla-mobile/firefox-tv:pull-request` (and other scopes) to know what kind of task it's allowed to spin. Please let me know if you need more information.

npark-mozilla:
@liuche So we need your confirmation on:
- deleting project/firefox-tv/tokens and use only mobile/project/firefox-tv/tokens
- Granting same access to mobile/project/firefox-tv/tokens so team members can modify it when needed

I wonder, why were there two secrets in TC (and mobile/project one was being unused?) ? I feel like I'm missing something important

mcomella:
I love your commit messages: "stupid dash missing" lol

npark-mozilla:
Also addresses #1800

npark-mozilla:
> I love your commit messages: "stupid dash missing" lol

I run out of  things to write for those 20+ commit PRs :)

tools/taskcluster/execute-firebase-test.sh
```diff
@@ -0,0 +1,60 @@
...
+    echo "Do not put .apk prefix in Apk_Name"
+    echo "Device_Config should be in landscape mode (e.g. model=taimen,version=26,orientation=landscape)"
+    echo
+}
+
+# Basic parameter check
+if [[ $# -lt 3 ]]; then
+    echo "Your command line contains $# arguments"
+    display_help
+    exit 1
```
Baron-Severin: This error message would be pretty confusing out of context.  Could we print how many it should have, and why?  Maybe even include "See execute-firebase-test.sh for details"

tools/taskcluster/execute-firebase-test.sh
```diff
@@ -0,0 +1,60 @@
...
+# Execute test set
+./google-cloud-sdk/bin/gcloud --format="json" firebase test android run \
+--type instrumentation \
+--app ./app/build/outputs/apk/$1/$2.apk \
+--test ./app/build/outputs/apk/androidTest/$1/$2-androidTest.apk \
+--use-orchestrator \
+--environment-variables clearPackageData=true \
+--results-bucket firefox-tv_test_artifacts \
+--timeout 30m \
+--no-auto-google-login $config_param
```
Baron-Severin: This block seems like something we will probably want to share in both Firebase and other tests.  I'm not terribly familiar with Shell scripting, is there any way to break this out and reuse it?

tools/taskcluster/execute-firebase-test.sh
npark-mozilla: Actually it calls display_help which prints out the usage - I just added example use case to be more clear.

tools/taskcluster/execute-firebase-test.sh
npark-mozilla: This is google gcloud sdk cli command, and I have thought about creating a wrapper method for it, but for each project, most parameters would be slightly different - so it would be more clear to the user if the actual gcloud test request is readily visible.

PR #1809, opened 2019-02-07
Closes #1776: Regular Builds for Performance update

### LINKED ISSUE #1776: Update and merge performance build variant PR to master
https://github.com/mozilla-mobile/firefox-tv/pull/1243 is fairly out of date and the performance build variant migration should't break anything on the SystemEngineView build variant.

psymoon:
### Pull Request checklist

- [x] The **UI tests** are passing after this PR (`./gradlew connectedSystemDebugAndroidTest`)


app/build.gradle
```diff
@@ -226,9 +246,9 @@ def addSecretToBuildConfig(variant, fieldName, fileBaseName) {
     // We separate the debug files from the release files so that we don't
     // accidentally ship a release using the debug key or vice versa.
     def fileSuffix
-    if (variantName.contains("debug")) {
+    if (variantName.contains("Debug")) {
         fileSuffix = "debug"
-    } else if (variantName.contains("release")) {
+    } else if (variantName.contains("Release")) {
```
Baron-Severin: Optional nit: a potentially less brittle way of handling this is `variantName.toLowerCase().contains("release")`

buildSrc/src/main/java/Gecko.kt
```diff
@@ -0,0 +1,11 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+private object GeckoVersions {
+    const val nightly_version = "67.0.20190206095431"
```
Baron-Severin: Possible that I'm confused here, but if we're declaring this version statically, it's not really nightly, is it?  Should this be renamed?

app/src/gecko/java/org/mozilla/tv/firefox/ext/EngineView.kt
```diff
@@ -29,15 +26,15 @@ fun EngineView.setupForApp() {
...
-    }
+//    if (BuildConstants.isDevBuild) {
+//        GeckoView.setWebContentsDebuggingEnabled(true)
+//    }
 
-    // WebView can be null temporarily after clearData(); however, activity.recreate() would
-    // instantiate a new WebView instance
-    webView?.setOnFocusChangeListener { _, hasFocus ->
+    // GeckoView can be null temporarily after clearData(); however, activity.recreate() would
+    // instantiate a new GeckoView instance
```
Baron-Severin: Is this comment accurate for GeckoView?

buildSrc/src/main/java/Gecko.kt
psymoon: Nightly just means a release channel. Dynamic version updates (especially for dependencies) are generally frowned upon

app/src/gecko/java/org/mozilla/tv/firefox/ext/EngineView.kt
psymoon: No, I actually haven't changed comments in gv/EngineView.kt. Will update!

buildSrc/src/main/java/Gecko.kt
Baron-Severin: Agreed that we should avoid dynamic versioning, but I do find the name confusing.

On a side note, is there a standard for how to style code shared by two languages?  It's weird to see snake casing in a Kotlin file, but I suppose it's only really going to be used in Gradle.

buildSrc/src/main/java/Gecko.kt
```diff
@@ -0,0 +1,11 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+private object GeckoVersions {
```
Baron-Severin: We may want to consider renaming this to `EngineVersions` and keeping this version of the file in the Gecko source set.

psymoon: @Baron-Severin Hmm, how come? We'd only maintain gecko since we have no control over system webview versioning (I can see us adding different CPU and channel variants in the future though)

Baron-Severin: We made pretty heavy use of source sets at Macy's (the Bloomingdale's app was just a different build flavor), and we found that the complexity involved tended to compound really quickly.

Gecko is a source set, and to me that means that Gecko related code should never be accessible to System code.  If we wind up with a lot of conditional checks to execute System vs Gecko code, things get messy and complicated very fast.  I strongly feel that that kind of problem should be handled using source sets as often as possible.

In this case, the System version of the class might just query a `WebView` and gets its version, or something similar.  But I think we should strictly enforce a policy where Main code knows about Main concerns and nothing else.

psymoon: @Baron-Severin Just to clarify, this is a `/buildSrc/` that works specifically with declarative portion of `build.gradle`, in other words it's not an actual source set? Main or build variant source codes has no direct exposure to this

liuche:
On a quick first pass this looks fine to me, but I'll take a closer look too - but no need to block on my review.

buildSrc/src/main/java/Gecko.kt
Baron-Severin: I totally missed that, sorry.  Ignore me!

PR #1825, opened 2019-02-12
On merge to master, run on-device UI Test on Bitbar Cloud

AaronMT:
This PR integrates a Gradle plugin for supporting Bitbar Cloud UI testing on master merge. Configuration has to be defined in a Gradle task. Currently it uses a preset “Firefox for Fire TV” project defined on Bitbar using a predefined group of devices under a group name “Amazon”. We can customize the group of device selection through the front-end. Test run names are just a build + date (we can probably improve on that, just temporarily  using this approach).

Edit: I'll squash before merge.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -3,6 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 import org.mozilla.gradle.tasks.ValidateAndroidAppReleaseConfiguration
+import java.text.SimpleDateFormat
 
 apply plugin: 'com.android.application'
 apply plugin: 'kotlin-android'
@@ -12,6 +13,7 @@ apply plugin: 'jacoco'
 apply plugin: 'pmd'
 apply plugin: 'checkstyle'
 apply plugin: 'kotlin-android-extensions'
+apply plugin: 'testdroid'
 
 apply from: "$project.rootDir/tools/gradle/versionCode.gradle"
 
@@ -465,4 +467,35 @@ if (project.hasProperty("coverage")) {
 
 afterEvaluate {
     check.dependsOn 'findbugs', 'pmd', 'checkstyle'
+}
+
+// -------------------------------------------------------------------------------------------------
+// Bitbar Cloud Testing: testDroidUpload
+// Upload both the application and test APK and conduct a UI test-run
+// Utilizes a preset group of Amazon Fire TV devices on a preset project and framework
+// -------------------------------------------------------------------------------------------------
+
+testdroid {
+
+    try {
+        def tokenFile = new File("${rootDir}/.bitbar_token.json")
+        def parsedFile = new groovy.json.JsonSlurper().parseText(tokenFile.text)
+
+        apiKey parsedFile.secret.bitbarToken.api_key
+        cloudUrl parsedFile.secret.bitbarToken.cloud_url
+
+        println "(Added from .bitbar_token file)"
+    } catch (FileNotFoundException ignored) {
+        println("X_X")
+    }
+
+    authorization "APIKEY"
+    deviceGroup "Amazon"
+    projectName "Firefox for Fire TV"
+    testRunName "Build: " + android.defaultConfig.versionName + " - " + new SimpleDateFormat("yyyy-MM-dd HH:mm:ss").format(new Date())
+    frameworkId 24
+    fullRunConfig {
+        instrumentationRunner = android.defaultConfig.testInstrumentationRunner
+        withAnnotation = "org.mozilla.tv.firefox.ui"
+    }
 }
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -480,6 +480,8 @@ testdroid {
         def tokenFile = new File("${rootDir}/.bitbar_token.json")
         def parsedFile = new groovy.json.JsonSlurper().parseText(tokenFile.text)
 
+        println parsedFile
+
         apiKey parsedFile.api_key
         cloudUrl parsedFile.cloud_url
 
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -476,13 +476,12 @@ afterEvaluate {
 // -------------------------------------------------------------------------------------------------
 
 testdroid {
-
     try {
         def tokenFile = new File("${rootDir}/.bitbar_token.json")
         def parsedFile = new groovy.json.JsonSlurper().parseText(tokenFile.text)
 
-        apiKey parsedFile.secret.bitbarToken.api_key
-        cloudUrl parsedFile.secret.bitbarToken.cloud_url
+        apiKey parsedFile.bitbarToken.api_key
+        cloudUrl parsedFile.bitbarToken.cloud_url
 
         println "(Added from .bitbar_token file)"
     } catch (FileNotFoundException ignored) {
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -493,6 +493,7 @@ testdroid {
     projectName "Firefox for Fire TV"
     testRunName "Build: " + android.defaultConfig.versionName + " - " + new SimpleDateFormat("yyyy-MM-dd HH:mm:ss").format(new Date())
     frameworkId 24
+    scheduler "PARALLEL"
     fullRunConfig {
         instrumentationRunner = android.defaultConfig.testInstrumentationRunner
         withAnnotation = "org.mozilla.tv.firefox.ui"
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -480,8 +480,8 @@ testdroid {
         def tokenFile = new File("${rootDir}/.bitbar_token.json")
         def parsedFile = new groovy.json.JsonSlurper().parseText(tokenFile.text)
 
-        apiKey parsedFile.bitbarToken.api_key
-        cloudUrl parsedFile.bitbarToken.cloud_url
+        apiKey parsedFile.api_key
+        cloudUrl parsedFile.cloud_url
 
         println "(Added from .bitbar_token file)"
     } catch (FileNotFoundException ignored) {
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -480,8 +480,6 @@ testdroid {
         def tokenFile = new File("${rootDir}/.bitbar_token.json")
         def parsedFile = new groovy.json.JsonSlurper().parseText(tokenFile.text)
 
-        println parsedFile
-
         apiKey parsedFile.api_key
         cloudUrl parsedFile.cloud_url
 
--- a/tools/taskcluster/get-bitbar-token.py
+++ b/tools/taskcluster/get-bitbar-token.py
@@ -1,0 +1,22 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+"""
+This script talks to the taskcluster secrets service to obtain the
+Bitbar token and write it to the .bitbar_token file in the root
+directory.
+"""
+
+import os
+import taskcluster
+import json
+
+# Get JSON data from taskcluster secrets service
+secrets = taskcluster.Secrets({'baseUrl': 'http://taskcluster/secrets/v1'})
+data = secrets.get('project/mobile/firefox-tv/tokens')
+
+with open(os.path.join(os.path.dirname(__file__), '../../.bitbar_token.json'), 'w') as file:
+    json.dump(data['secret']['bitbarToken'], file)
+
+print("Imported bitbar token from secrets service")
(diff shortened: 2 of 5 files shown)


AaronMT:
Test-run notifications are currently supported in Slack. 

Wish-list help on integrating with Github. Super wish-list, device selection at submission time.

app/build.gradle
```diff
@@ -466,3 +468,34 @@ if (project.hasProperty("coverage")) {
...
+// Bitbar Cloud Testing: testDroidUpload
+// Upload both the application and test APK and conduct a UI test-run
+// Utilizes a preset group of Amazon Fire TV devices on a preset project and framework
+// -------------------------------------------------------------------------------------------------
+
+testdroid {
+
+    try {
+        def tokenFile = new File("${rootDir}/.bitbar_token.json")
+        def parsedFile = new groovy.json.JsonSlurper().parseText(tokenFile.text)
```
Baron-Severin: Nit: it would be cleaner if we imported this class instead of using a fully qualified reference.

app/build.gradle
```diff
@@ -466,3 +468,34 @@ if (project.hasProperty("coverage")) {
...
+
+        apiKey parsedFile.secret.bitbarToken.api_key
+        cloudUrl parsedFile.secret.bitbarToken.cloud_url
+
+        println "(Added from .bitbar_token file)"
+    } catch (FileNotFoundException ignored) {
+        println("X_X")
```
Baron-Severin: Is it possible for this task to succeed if the file is not found?  I feel like we should probably throw an exception and cancel the build if not.  

If it is possible for this to succeed, we should flesh out this log so that someone who is debugging this problem can easily trace the log back to this task.  Maybe something like "BitBar credentials not found, canceling remote UI test".

app/build.gradle
```diff
@@ -466,3 +468,34 @@ if (project.hasProperty("coverage")) {
 afterEvaluate {
     check.dependsOn 'findbugs', 'pmd', 'checkstyle'
 }
+
+// -------------------------------------------------------------------------------------------------
+// Bitbar Cloud Testing: testDroidUpload
+// Upload both the application and test APK and conduct a UI test-run
```
Baron-Severin: Nit: `a UI test-run` -> `a UI test-run on BitBar`

app/build.gradle
```diff
@@ -466,3 +468,34 @@ if (project.hasProperty("coverage")) {
...
+testdroid {
+
+    try {
+        def tokenFile = new File("${rootDir}/.bitbar_token.json")
+        def parsedFile = new groovy.json.JsonSlurper().parseText(tokenFile.text)
+
+        apiKey parsedFile.secret.bitbarToken.api_key
+        cloudUrl parsedFile.secret.bitbarToken.cloud_url
+
+        println "(Added from .bitbar_token file)"
```
Baron-Severin: Nit: what did we add from that file?  Can we include that in the log?

app/build.gradle
```diff
@@ -466,3 +468,34 @@ if (project.hasProperty("coverage")) {
...
+        println("X_X")
+    }
+
+    authorization "APIKEY"
+    deviceGroup "Amazon"
+    projectName "Firefox for Fire TV"
+    testRunName "Build: " + android.defaultConfig.versionName + " - " + new SimpleDateFormat("yyyy-MM-dd HH:mm:ss").format(new Date())
+    frameworkId 24
+    fullRunConfig {
+        instrumentationRunner = android.defaultConfig.testInstrumentationRunner
```
Baron-Severin: Some of our tests will fail if run without using `org.mozilla.tv.firefox.FirefoxTestRunner`.

tools/taskcluster/get-bitbar-token.py
```diff
@@ -0,0 +1,22 @@
...
+Bitbar token and write it to the .bitbar_token file in the root
+directory.
+"""
+
+import os
+import taskcluster
+import json
+
+# Get JSON data from taskcluster secrets service
+secrets = taskcluster.Secrets({'baseUrl': 'http://taskcluster/secrets/v1'})
```
Baron-Severin: What will happen if this network request fails?  Will it cancel the build?

app/build.gradle
Baron-Severin: E.g, `Added test credentials from .bitbar_token file`

app/build.gradle
AaronMT: This file is necessary (as it contains the API key and cloud URL), otherwise a build won't be uploaded. I changed `ignored` to `e` and clarified the message. What else is needed?

app/build.gradle
AaronMT: Cool. Done.

app/build.gradle
AaronMT: I thought that is what is defined by defaultConfig in line 33 of build.gradle ?
`        testInstrumentationRunner "org.mozilla.tv.firefox.FirefoxTestRunner"`

If not, I can change that.

tools/taskcluster/get-bitbar-token.py
AaronMT: No idea actually. How is this handled in other projects? @npark-mozilla ?

npark-mozilla: The build will fail, and that's good, we want to know when this service fails, since we rely on it.  We do the same for the pocket/sentry key, accessing the same URL.  I think this is good.

app/build.gradle
Baron-Severin: I think failing to load it should be considered a build failure, so in the `catch` block I would add `throw(e)`.  I assume that a task throwing an exception should cancel the rest of the build?

app/build.gradle
Baron-Severin: It is, but what does this line do?  It looks like it's redeclaring the same information.  We saw earlier that BitBar and Firebase were ignoring some of what we declared in Gradle, so it's not implausible.

tools/taskcluster/get-bitbar-token.py
Baron-Severin: Agreed, I just wanted to make sure it wouldn't silently fail.

app/build.gradle
AaronMT: The line just specifies a custom instrumentation runner framework (Default: `android.test.InstrumentationTestRunner`. It doesn't seem to be required, but I didn't see any problems leaving it in. If the default is fine, I can remove it?

http://docs.bitbar.com/testing/api/tests/index.html?highlight=instrumentationrunner

app/build.gradle
AaronMT: Should be yeah; added.

Baron-Severin: It's possible that I'm being overly nitpicky here, but I think the ideal situation would be to have this task only run on PRs.  When we're building for a PR, it seems like this running successfully **_should_** be a hard requirement (please correct me if I'm wrong there, you understand the process much better than I do).  But I agree that blocking local builds on this would be unnecessary.

What do you think?

app/build.gradle
Baron-Severin: I'm not honestly clear on what it's doing, but if everything passes without it that suggests to me that we should delete it.

app/build.gradle
AaronMT: It's up to the team really. Originally I had it on PRs, I think that's where it makes sense to be. @npark-mozilla thoughts?

npark-mozilla: I probably am not understanding the full conversation thread, but just to clarify, the bitbar testing will only run on merge to master, and not on PR.  And, in case the file is not found, it will fail almost instantly on the next line when it tries to log in, with invalid credential.  I don't think throw() would be too useful here, and for simplicity, I think having a detailed println statement as above should be sufficient.

app/build.gradle
npark-mozilla: Does the test run on bitbar without 498? then `android.test.InstrumentationTestRunner` as default value is working, and we can remove line 498.  If not, then we should keep it.  Sometimes the cloud test frameworks needs to re-declare ones that are already defined in gradle, because they are disregarding most of it.

app/build.gradle
AaronMT: Agreed. So to clarify once more

- Currently, it runs on merge to master (via task cluster.yml)
- If for whatever reason credentials can't be fetched, we'll see the printed line at least
- A green build will still happen (which is fine for now).
- Results can be checked in Slack. In the case of failure, I suppose you'll notice no results.

Does this sound good?

npark-mozilla: So only thing I'm kinda worried about is that it returns green build after failure.  Devs probably won't give a 2nd look to check for the test result if it is marked green.

 If it was a TC shell script, then you can make it return nonzero int exit value (`exit 1`)  to signal failure on TC.  But because we're running it as a gradle command, it'll be difficult to do that.  
If we can somehow write a file that shows whether the UI test run is successful or not, then run a shell script as a part of TC run that checks this file to return zero/nonzero int, then we can fail the TC run appropriately.  If that is difficult, we can leave it as is, and create a followup issue where we either use their rest api or something else where we can at least know whether there was a failure.

Baron-Severin: Yeah, a green build on failure is definitely a big problem.

app/build.gradle
AaronMT: If `instrumentationRunner` is empty the primary Android test runner class `android.test.InstrumentationTestRunner` is used. Yes, I can remove it.

app/build.gradle
AaronMT: The only form of direct callback AFAIK is the call back URL *after* a test run has finished. Bitbar testing sends a POST message to a defined URL when test run is finished.

E.g, `testRunId=%s&projectId=%s&status=FINISHED`

The URL has to be reachable online.

We can continue to explore that if we want?

http://docs.bitbar.com/testing/cloud-integrations/jenkins-plugin/jenkins-ui/index.html?highlight=hook

npark-mozilla: Actually, also if the testdroid returns any pass/fail status, we can try to fail the gradle run by raising gradleexception as well. https://stackoverflow.com/questions/10312259/recommended-way-to-stop-a-gradle-build

npark-mozilla: Or, much easier way would be to call the rest api method with the testrunid we got from above to get the test result.
 We can do that on a separate script and return nonzero if it contains failure.

AaronMT: After investigation on Sunday it looks like `Status` (from the POST parameter value: http://ptsv2.com/t/xuzfr-1550774903/d/30001)  will only provide us details about a test run’s overall state; in the case of failure it still finishes, hence `FINISHED`. Regardless of failures it will always value as `FINISHED` (or maybe `Waiting` if devices are offline). 

If we want test failure count, we need the test `runID`. We have a set `projectID` we can re-use. 

We can still use the hook callback URL but ignore that state and look at the other parameters (http://ptsv2.com/t/xuzfr-1550774903/d/30001). We would need some kind of service to wait for, poll, and read POST data after a test-run to a server online. From there we get their generated `runID `. With the `runID` and `projectID` that is also available, we can get JSON data back (e.g, Python requests) that will have `"failedTestCaseCount”` and any non zero integer s a failure. From there we can do whatever. But suddenly you introduce a lot of wait (~15 min) and a lot of complication.

npark-mozilla: I put a question to bitbar team on the slack channel about getting the test result - I think it is sufficient to get whether tests have all passed or not.  Anything more detailed than that they can check in on bitbar.  Currently the problem is that bitbar just says `Finished` even when there is a failure.  I have an idea that we can try later.  IF that wait is the wait during the actual test execution, I think that is okay.  Other TC test runs don't finish until tests are completed anyways.  If this doesn't work, we can ask Levente to check the slack channel for failure, since slack notification tells whether there are failures or not.

npark-mozilla:
Got the response from bitbar, and they showed a python API that 'polls' for the test result once submitted. So we need to use it to obtain the test result from their server.  So the flow would be like:

- Install python api in the docker image (or during the taskcluster script execution)
- Submit tests via python api
- Poll for test execution result using python api, determine pass/fail status, and generate appropriate exit value depending on it
- Possibly provide the bitbar URL for test result.

Which means that a lot of the work might have to be redone in python. @AaronMT can elaborate/correct on above points.

AaronMT:
> Got the response from bitbar, and they showed a python API that 'polls' for the test result once submitted. So we need to use it to obtain the test result from their server. So the flow would be like:
> 
> * Install python api in the docker image (or during the taskcluster script execution)
> * Submit tests via python api
> * Poll for test execution result using python api, determine pass/fail status, and generate appropriate exit value depending on it
> * Possibly provide the bitbar URL for test result.
> 
> Which means that a lot of the work might have to be redone in python. @AaronMT can elaborate/correct on above points.

Yes, this seems like the way to go. This has spun out into different needs and testing and I'll probably close this PR and open a new one with the Python integration.

npark-mozilla:
We now have a Docker related PR (#1895) that awaits review before this PR can proceed.  We ran locally, and it looks good.

AaronMT:
Closing in favor of a Python integration

Issue #1832, opened 2019-02-13
Write YouTube navigation UI test


Baron-Severin:
YouTube is core functionality for our product, and it's important that it consistently functions.  It is, however, difficult to test, and has therefore regressed several times in the past (most recently in #1830).

Since we cannot unit test YouTube, or mock out its implementation, an Espresso test will have to be written that makes actual network calls.  This will be slow and potentially flakey, but given how important the site is and the potential for regression, it's probably worth doing.

Test should verify that:

- [x] ~~Clicking on the YouTube tile, opening the overlay, waiting for loading to end, then closing the overlay does not disrupt navigation~~
This is currently not possible with the `SessionLoadedIdlingResource`, as it will not allow the overlay to open until the page has finished loading. Opened #1892
- [x] D-pad navigation works
- [x] Hardware back cycles focus until the sidebar is focused, and then opens the overlay

Baron-Severin:
FYI, be aware that the first bullet point will fail until #1830 is closed.

PR #1834, opened 2019-02-13
Closes #1830: fixes YouTube focus when overlay is open

### LINKED ISSUE #1830: Overlay can break YouTube navigation
## Steps to reproduce
- From home screen (does not have to be a fresh install)
- Click YouTube tile
- Immediately reopen overlay with menu button
- Wait for YouTube to finish loading
- Close overlay

### Expected behavior
- YouTube should operate as normal

### Actual behavior
- Focus movement, clicks, and back presses do nothing
- Menu button still works
- Reloading page fixes this

### Device information
* Fire TV device: Cube
* Latest affected Firefox version: 65ad45e6be65136aee2f2885c5e0556504759f8c

Baron-Severin:
RCA:
There is a WebView bug that, when the WV loses and regains focus, causes the incorrect DOM element to receive focus (https://bugs.chromium.org/p/chromium/issues/detail?id=826577). Our workaround for this caches the currently focused view on overlay open, then focuses it on overlay close. When we open the overlay during a loading screen, we cache the body element. The overlay is then closed, and the body element focused, rather than the view on which YouTube has programmatically requested focus.

Fix:
Instead of caching focus on overlay open, this now caches it just before overlay close.  This allows us to use the most recently focused DOM element.

### Pull Request checklist

Testing this will involve writing a complex Espresso test.  This is out of scope for this story, but https://github.com/mozilla-mobile/firefox-tv/issues/1832 was opened

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed links on settings About page that did not load (#1731)
 - A memory leak (#1628)
 - Fixed bug that could cause YouTube to display vertically offset from where it should be (#1719)
+- Bug that could cause YouTube to lose focus if the overlay was opened during loading (#1830)
 
 ## [3.3.1] - 2019-02-04
 *Released to Fire TV 4K.*
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed links on settings About page that did not load (#1731)
 - A memory leak (#1628)
 - Fixed bug that could cause YouTube to display vertically offset from where it should be (#1719)
-- Bug that could cause YouTube to lose focus if the overlay was opened during loading (#1830)
+- Bug that could cause YouTube to be unresponsive if the overlay was opened during loading (#1830)
 
 ## [3.3.1] - 2019-02-04
 *Released to Fire TV 4K.*
--- a/app/src/main/java/org/mozilla/tv/firefox/ScreenController.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/ScreenController.kt
@@ -174,6 +174,7 @@ class ScreenController {
                 _currentActiveScreen.value = ActiveScreen.NAVIGATION_OVERLAY
             }
             Transition.REMOVE_OVERLAY -> {
+                fragmentManager.webRenderFragment().onShowEngineView()
                 showNavigationOverlay(fragmentManager, false)
                 _currentActiveScreen.value = ActiveScreen.WEB_RENDER
             }
@@ -212,6 +213,7 @@ class ScreenController {
                     .hide(fragmentManager.pocketFragment())
                     .hide(fragmentManager.settingsFragment())
                     .commitNow()
+                fragmentManager.webRenderFragment().onShowEngineView()
                 _currentActiveScreen.value = ActiveScreen.WEB_RENDER
             }
             Transition.EXIT_APP -> { return false }
--- a/app/src/main/java/org/mozilla/tv/firefox/ScreenController.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/ScreenController.kt
@@ -30,7 +30,9 @@ class ScreenController {
     private val _currentActiveScreen = MutableLiveData<ActiveScreen>().apply {
         value = ActiveScreen.NAVIGATION_OVERLAY
     }
-    // This is updated just before the fragment transaction is committed
+    /**
+     * Observers will be notified just before the fragment transaction is committed
+     */
     val currentActiveScreen: LiveData<ActiveScreen> = _currentActiveScreen
 
     /**
@@ -171,6 +173,7 @@ class ScreenController {
         // Call show() before hide() so that focus moves correctly to the shown fragment once others are hidden
         when (transition) {
             Transition.ADD_OVERLAY -> {
+                // We always update the currentActiveScreen value before beginning the fragment transaction
                 _currentActiveScreen.value = ActiveScreen.NAVIGATION_OVERLAY
                 fragmentManagerShowNavigationOverlay(fragmentManager, true)
             }
--- a/app/src/main/java/org/mozilla/tv/firefox/ScreenController.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/ScreenController.kt
@@ -27,9 +27,10 @@ import org.mozilla.tv.firefox.widget.InlineAutocompleteEditText
 
 class ScreenController {
 
-    private var _currentActiveScreen = MutableLiveData<ActiveScreen>().apply {
+    private val _currentActiveScreen = MutableLiveData<ActiveScreen>().apply {
         value = ActiveScreen.NAVIGATION_OVERLAY
     }
+    // This is updated just before the fragment transaction is committed
     val currentActiveScreen: LiveData<ActiveScreen> = _currentActiveScreen
 
     /**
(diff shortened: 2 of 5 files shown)


app/src/main/java/org/mozilla/tv/firefox/ScreenController.kt
```diff
@@ -212,6 +213,7 @@ class ScreenController {
                     .hide(fragmentManager.pocketFragment())
                     .hide(fragmentManager.settingsFragment())
                     .commitNow()
+                fragmentManager.webRenderFragment().onShowEngineView()
```
mcomella: We're adding `onShowEngineView` each time currentActiveScreen is assigned to WEB_RENDER: this approach is fragile because if the code changes and we have a new assignment, we may forget to add this line.

What if we were reactive and did it as an observer instead? Do the LiveData observers get called before the View focus changes? If not, could we write a small wrapper around the currentActiveScreen update so that our observer can get called first before the LiveData changes?

app/src/main/java/org/mozilla/tv/firefox/webrender/WebRenderFragment.kt
```diff
@@ -211,4 +212,11 @@ class WebRenderFragment : EngineViewLifecycleFragment(), Session.Observer {
         }
         return false
     }
+
+    /**
+     * Called before the [EngineView] is shown
```
mcomella: -> Must be called...

app/src/system/java/org/mozilla/tv/firefox/ext/EngineView.kt
```diff
@@ -39,14 +39,7 @@ fun EngineView.setupForApp() {
...
-        if (!hasFocus) {
-            // For why we're modifying the focusedDOMElement, see FocusedDOMElementCacheInterface.
-            //
-            // Any views (like BrowserNavigationOverlay) that may clear the cache, e.g. by
-            // reloading the page, are required to handle their own caching. Here we'll handle
-            // cases where the page cache isn't cleared.
-            focusedDOMElement.cache()
-        } else {
+        if (hasFocus) {
             // Trying to restore immediately doesn't work - perhaps the WebView hasn't actually
```
mcomella: You should preface this comment with, "// For why we're modifying the focusedDOMElement, see FocusedDOMElementCacheInterface."

Also, we may want to mention that cache the element is done elsewhere.

app/src/main/java/org/mozilla/tv/firefox/webrender/WebRenderFragment.kt
```diff
@@ -211,4 +212,11 @@ class WebRenderFragment : EngineViewLifecycleFragment(), Session.Observer {
...
+     */
+    fun onShowEngineView() {
+        engineView?.focusedDOMElement?.cache()
```
mcomella: We need to explain why we cache here and specifically here. Basically, the gist of your commit summary (cache the element at the last possible second) should really be in the code base as comments instead.

app/src/main/java/org/mozilla/tv/firefox/ScreenController.kt
Baron-Severin: That's a good point.  The LiveData observers are called synchronously when the value is set, and asynchronously when post is used.  Changing these to set the value before kicking off the transaction would be trivial, but I'll have to double check that it won't have negative effects.

Good catch.

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed links on settings About page that did not load (#1731)
 - A memory leak (#1628)
 - Fixed bug that could cause YouTube to display vertically offset from where it should be (#1719)
+- Bug that could cause YouTube to lose focus if the overlay was opened during loading (#1830)
```
mcomella: Has this been seen by users yet? If not, we don't need a changelog item for it.

mcomella: nit: -> "YouTube to be unresponsive"

Users don't know what losing focus means. :)

app/src/main/java/org/mozilla/tv/firefox/ScreenController.kt
```diff
@@ -170,49 +171,49 @@ class ScreenController {
         // Call show() before hide() so that focus moves correctly to the shown fragment once others are hidden
         when (transition) {
             Transition.ADD_OVERLAY -> {
-                fragmentManagerShowNavigationOverlay(fragmentManager, true)
                 _currentActiveScreen.value = ActiveScreen.NAVIGATION_OVERLAY
```
mcomella: Add a comment (maybe above `when`, next to first currentActiveScreen assignment, or wrap currentActiveScreen assignment and comment in that method) that we need to update the active screen before we perform the fragment transactions so this behavior that we rely on doesn't regress.

Can we add a test for this?

app/src/main/java/org/mozilla/tv/firefox/ScreenController.kt
```diff
@@ -27,9 +27,10 @@ import org.mozilla.tv.firefox.widget.InlineAutocompleteEditText
 
 class ScreenController {
 
-    private var _currentActiveScreen = MutableLiveData<ActiveScreen>().apply {
+    private val _currentActiveScreen = MutableLiveData<ActiveScreen>().apply {
         value = ActiveScreen.NAVIGATION_OVERLAY
     }
+    // This is updated just before the fragment transaction is committed
```
mcomella: nit: use property kdoc rather than "//": that indicates the comment is for consumers rather than file private comments.

Also, the comment should be more explicit, "Observers will be notified just before the fragment transaction is committed."

app/src/main/java/org/mozilla/tv/firefox/webrender/WebRenderFragment.kt
```diff
@@ -135,8 +135,13 @@ class WebRenderFragment : EngineViewLifecycleFragment(), Session.Observer {
          *  memory leak while clearing data. See [WebViewCache.clear] as well as #1720
          */
         serviceLocator?.screenController?.currentActiveScreen?.observe(viewLifecycleOwner, Observer {
-            if (it != ScreenControllerStateMachine.ActiveScreen.WEB_RENDER) {
-                engineView.pauseAllVideoPlaybacks()
+            when (it) {
+                // Cache focused DOM element just before WebView gains focus. See comment in
+                // FocusedDOMElementCacheInterface for details
+                ScreenControllerStateMachine.ActiveScreen.WEB_RENDER -> engineView.focusedDOMElement.cache()
```
mcomella: nit: -> `ActiveScreen.WEB_RENDER` to make this slightly more readable.

app/src/main/java/org/mozilla/tv/firefox/webrender/WebRenderFragment.kt
```diff
@@ -135,8 +135,13 @@ class WebRenderFragment : EngineViewLifecycleFragment(), Session.Observer {
          *  memory leak while clearing data. See [WebViewCache.clear] as well as #1720
          */
         serviceLocator?.screenController?.currentActiveScreen?.observe(viewLifecycleOwner, Observer {
-            if (it != ScreenControllerStateMachine.ActiveScreen.WEB_RENDER) {
-                engineView.pauseAllVideoPlaybacks()
+            when (it) {
```
mcomella: nit: I think this is more readable as an if statement because the comparison (`it == WEB_RENDER`) is on a single line.

CHANGELOG.md
mcomella: nit: commit message: what changelog is being updated?

app/src/main/java/org/mozilla/tv/firefox/webrender/FocusedDOMElementCacheInterface.kt
```diff
@@ -17,6 +17,10 @@ import android.view.View
  * to restore the focused element, dpad navigation is entirely broken (#393): that is
```
mcomella: nit: commit message: which comments are improved?

app/src/main/java/org/mozilla/tv/firefox/ScreenController.kt
Baron-Severin: This does seem like a good thing to test, but it's not clear to me at the moment how.  Do you have any suggestions?

liuche:
@Baron-Severin can you open a PR for this against the 3.4.1-LAT2 branch? Then we can send out a LAT and if everything seems fixed, push a release to Amazon.

app/src/main/java/org/mozilla/tv/firefox/ScreenController.kt
mcomella: First thought: if somehow you could stub `setCurrentActiveScreen` and `fragmentManagerShowNavigationOverlay`, you can have each toggle a boolean flag and check that the other boolean flag is in the appropriate state. It seems obnoxious to decouple these invocations so heavily, however.

PR #1839, opened 2019-02-13
Optimize and re-build docker image.

pocmo:
This PR updates the Docker image to 2.1. I've built and uploaded this image to Docker Hub today - so that it includes the latest cached dependencies.

In addition to that I ported some Dockerfile optimizations that we did in AC over to the FFFTV app:
- Use "--depth=1" to only checkout the latest state. No need to have the whole git history inside the image.
- Run "gradle clean" after the build. No need to keep outdated build artifacts inside the image.
- Do not download anything with "sdkmanager". The following gradle task will download what it needs anyways. There's a higher chance that we pick the wrong things or just too much.
- Do not run "apt-gen clean". Because of the way Docker images are layered this doesn't actually save any space.

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -28,7 +28,7 @@ tasks:
     payload:
       maxRunTime: 3600
       deadline: "{{ '2 hours' | $fromNow }}"
-      image: 'mozillamobile/firefox-tv:2.0'
+      image: 'mozillamobile/firefox-tv:2.2'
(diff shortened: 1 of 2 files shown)


pocmo:
>    raise exceptions.TaskclusterFailure('baseUrl option is no longer allowed')

Aye, I need to pin the taskcluster client version like in the other images.

PR #1852, opened 2019-02-15
Closes #1584: Add ability to exit YT from Pocket video

### LINKED ISSUE #1584: Can't go back out of YouTube from a Pocket video with remote back
Similar to #1542, but specifically for the case where you got to YouTube from clicking a video in the Pocket feed.
## Steps to reproduce
1. Visit any site
2. Open overlay and click Pocket
3. Click video
4. Try to exit YouTube with remote back button
### Expected behavior
Ability to go to the previously visited site.
### Actual behavior
Stuck on YouTube side navigation bar.
### Device information

* Fire TV device: 4k
* Firefox version: 3.1.2

dnarcese:
### Pull Request checklist

Writing in a separate PR
- [x] The **UI tests** are passing after this PR (`./gradlew connectedSystemDebugAndroidTest`)


app/src/androidTest/java/org/mozilla/tv/firefox/ui/YouTubeNavigationTest.kt
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    /**
+     * Clicks remote back until the active element is in the YouTube sidebar.
+     * Then, clicks remote back one more time to go to the previous site.
+     */
+    private fun backOutOfYouTube() {
+        browser {
+            var activeElementParentId = executeJS(
+                "return document.activeElement.parentElement.parentElement.id"
```
dnarcese: These tests pass on device, but are failing on emulator at this line with:
`Cannot read property 'id' of null`

@Baron-Severin Do you of any differences between device and emulator that could be causing this?

Baron-Severin: I've run this branch a few times now on both device and emulator and haven't encountered that failure, so there seems to be some flakiness here.  

If the grandparent of the focused element is null, it means the focused element is `body` (the root element is `html`, which is one level higher than `body`).  You can verify that by logging the active element.  I know that `body` is the element focused by default until YouTube has finished loading, after which I assume they programmatically request focus on the first video.  My guess is that our test is sometimes trying to move on before YouTube is fully loaded.

dnarcese: So do you think the best thing would be to 
a) change the test to check `noElementFocused || bodyElementFocused || sidebarFocused` like in the implementation
b) change the test to make sure it waits for YT to be loaded (not entirely sure how to accomplish this since it's already using `SessionLoadedIdlingResource` which should be waiting)

Baron-Severin: B, although I'm not sure how to accomplish it.  The `SessionLoadedIdlingResource` knows whether or not the page load is complete, but maybe the focus change itself happens on page load complete?  So it becomes a race condition between the two?  I'm not sure, it'll take some investigation.

I think one option here would be to break the Espresso test into a separate PR so we can avoid blocking the other story.

dnarcese:
TC is failing because for some reason, sometimes when entering a new url, its not clearing the old one. It passes when I run on fire tv, though

app/src/main/java/org/mozilla/tv/firefox/webrender/YouTubeBackHandler.kt
```diff
@@ -59,7 +61,34 @@ class YouTubeBackHandler(
...
+        } else if (!url.isUriYouTubeTV) goingToYouTube = false
     }
+
+    private fun getIndexToGoBackTo(): Int {
+        return if (urlIndexList.isNotEmpty())
+            urlIndexList.removeAt(urlIndexList.lastIndex)
+        else 0
+    }
+
+    private fun goBackBeforeYouTube() = engineView!!.handleYoutubeBack(getIndexToGoBackTo())
```
Baron-Severin: This logic has gotten a little hard to follow IMO.  I'm seeing if I can work out any other way we can format this that might be easier to read, but it's a hard problem.

Baron-Severin:
> TC is failing because for some reason, sometimes when entering a new url, its not clearing the old one. It passes when I run on fire tv, though

Do you think this is a test problem, or a regression that we hadn't caught yet?

dnarcese:
> Do you think this is a test problem, or a regression that we hadn't caught yet?
Test problem. `enterUrlAndEnterToBrowser` seems to be flaky.

app/src/main/java/org/mozilla/tv/firefox/webrender/YouTubeBackHandler.kt
Baron-Severin: This is the same logic as your code, but I took some time to try to improve its readability.  A lot of this is subjective, so feel free to take some / all / or none of this advice.

```kotlin
private val preYouTubeIndexHistory: MutableList<Int> = mutableListOf()
private var currentPreYouTubeIndex: Int? = null

fun onUrlChanged(url: String) {
    if (!url.isUriYouTubeTV) currentPreYouTubeIndex = null
    else if (currentPreYouTubeIndex == null) {
        // Store the current (pre-YouTube) backForwardIndex when the URL first changes to YouTube
        currentPreYouTubeIndex = engineView!!.backForwardList.currentIndex
    }
}

/**
 * We only want to add indexes to the list that are a forward navigation from non-YouTube site
 * to YouTube. If you hit back from a non-YouTube site to YouTube, we do not want to save the index
 * of that non-YouTube site. The web history list is only updated after loading has completed,
 * which is why this function is called from onLoadingStateChanged callback.
 */
fun onLoadComplete() {
    // Don't store anything if we aren't navigating to YouTube
    val preYouTubeIndex = currentPreYouTubeIndex ?: return

    val navigationWasForward = preYouTubeIndex < engineView!!.backForwardList.currentIndex
    if (navigationWasForward) {
        preYouTubeIndexHistory += preYouTubeIndex
    }
}

private fun getIndexToGoBackTo(): Int {
    return if (preYouTubeIndexHistory.isEmpty()) 0
    else preYouTubeIndexHistory.removeAt(preYouTubeIndexHistory.lastIndex)
}

private fun goBackBeforeYouTube() = engineView!!.handleYoutubeBack(getIndexToGoBackTo())
```

Changes made:
- Made `backForwardUrlList` a local variable (as we were updating it before reading anyway)
- Renamed `urlIndexList` and `indexToAdd` to `preYouTubeIndexHistory` and `currentPreYouTubeIndex` to try and communicate more about what they represented
- Merged `indexToAdd` and `goingToYouTube` concepts into a nullable Int
- Renamed `updateUrlIndexList` and `saveIndex` to show that they are callbacks. This keeps us from having to read the other file to understand when they are called
- Moved `saveIndex` above `updateUrlIndexList` (it's called first, so the file can now be read more top to bottom)
- Restructured some conditional statements to complete their simpler checks first, to make them easier to follow

Again, a lot of this is subjective.  This would be my advice, but I'm wrong a lot!

PR #1854, opened 2019-02-16
1554 test pocket backoff

Baron-Severin:
### Pull Request checklist

Non-functional change


app/src/main/java/org/mozilla/tv/firefox/pocket/PocketEndpoint.kt
```diff
@@ -8,24 +8,47 @@ import android.net.Uri
...
+    override fun request(): Single<Response<PocketData>> {
+        fun requestAndWrap(): Single<Response<PocketData>> {
+            val videos = runBlocking { getRecommendedVideos() }
+            return when {
+                videos == null || videos.isEmpty() -> Single.just(Response.Failure())
+                else -> Single.just(Response.Success(videos))
+            }
+        }
+        return requestAndWrap()
+            .subscribeOn(Schedulers.io())
```
sashei: I find it obfuscating to put my scheduling inside of the observable source instead of as an operator when subscribing; personal preferences on this vary, but IMO it should be a project-level decision one way or the other.

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketRepoCache.kt
```diff
@@ -24,23 +23,40 @@ import android.arch.lifecycle.MutableLiveData
...
-            // Freezing the cache should prevent new videos from overwriting old ones, but we don't want to (for
-            // example) prevent new videos from overwriting a loading state. cachedValueIsBad means that the cached
-            // value is not a successful load
-            val cachedValueIsBad = _feedState.value !is PocketVideoRepo.FeedState.LoadComplete
-
-            if (!frozen || cachedValueIsBad) {
-                _feedState.postValue(it)
-            }
+    open val feedState : Observable<PocketVideoRepo.FeedState> =
+        repo.feedState.scan { cachedValue: PocketVideoRepo.FeedState, currentValue: PocketVideoRepo.FeedState ->
```
sashei: great use of `scan`!

another clarity note that we find useful in Lockbox is to use `\n` to help with readability when applying operators. so this would read:
```kotlin
repo.feedState // <- get your observable
    .scan { // <- each operator gets its own `\n`
        ...
    }
    .distinctUntilChanged()
    ...
```

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketVideoRepo.kt
```diff
@@ -39,102 +29,40 @@ open class PocketVideoRepo(
         object Inactive : FeedState()
     }
 
-    private val _feedState = MutableLiveData<FeedState>().apply {
-        // setValue must be called from the main thread, and this repo is instantiated
-        // off of the main thread in Espresso tests, so postValue is used instead
-        postValue(buildConfigDerivables.initialPocketRepoState)
-    }
-    open val feedState: LiveData<FeedState> = _feedState
+    private val _feedState = BehaviorSubject.createDefault(initialState)
```
sashei: one nice way to avoid the `_`-prefixed variables is:

```kotlin
open val feedState: Observable<PocketVideoRepo.FeedState> = BehaviorSubject.createDefault(initialState)
```

Then, when you need to push to the `feedState`, you can cast to `as Subject` below.

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketVideoRepo.kt
```diff
@@ -39,102 +29,40 @@ open class PocketVideoRepo(
...
-    @Volatile private var lastUpdateAttemptMillis = -1L
-    @Volatile private var retryDelayMillis = BASE_RETRY_TIME
-    @Volatile private var lastAttemptWasSuccessful = false
+    private val periodicRequester = PeriodicRequester(pocketEndpoint)
+    private val compositeDisposable = CompositeDisposable()
 
     fun update() {
-        retryDelayMillis = BASE_RETRY_TIME
-        GlobalScope.launch { updateInner() }
+        compositeDisposable += pocketEndpoint.request().subscribe { data -> postUpdate(data) }
```
sashei: For readability, I would rewrite this:
```kotlin
pocketEndpoint.request()
    .subscribe(this::postUpdate)
    .addTo(compositeDisposable)
```

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketVideoRepo.kt
```diff
@@ -39,102 +29,40 @@ open class PocketVideoRepo(
...
     }
 
     @UiThread // update backgroundUpdates.
     fun startBackgroundUpdates() {
-        backgroundUpdates?.cancel() // Cancelling unexpectedly active Pocket update job to ensure only one is running
-        backgroundUpdates = startBackgroundUpdatesInner()
+        compositeDisposable.clear()
+        compositeDisposable += periodicRequester.start().subscribe { postUpdate(it) }
```
sashei: similarly:
```kotlin
periodicRequester.start()
    .subscribe(this::postUpdate)
    .addTo(compositeDisposable)
```

app/src/main/java/org/mozilla/tv/firefox/utils/PeriodicRequester.kt
```diff
@@ -0,0 +1,91 @@
...
+ * Failed requests should be represented as [Maybe.empty].
+ *
+ * Should operate on a background thread.
+ */
+interface Endpoint<T> {
+    fun request(): Single<Response<T>>
+}
+
+sealed class Response<T> {
+    data class Success<T>(val data: T) : Response<T>()
```
sashei: is there a reason you wouldn't use the `onSuccess` and `onError` cases of a `Single` instead of using this custom `Response` type?

sashei: Oh, I see, you want to use this in an `Observable` sometimes....

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketEndpoint.kt
Baron-Severin: That's a good point, I'll move that.

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketVideoRepo.kt
Baron-Severin: My only concern with that approach is that downstream objects would be able to cast it to a subject as well.

app/src/main/java/org/mozilla/tv/firefox/utils/PeriodicRequester.kt
Baron-Severin: Yeah, and the repo also needs to be aware of failure responses for some of its logic.  The observable between the requester and the repo needs to stay open even after failure responses, so we can't use `onError` there.

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketVideoRepo.kt
psymoon: > My only concern with that approach is that downstream objects would be able to cast it to a subject as well.

@Baron-Severin what's the downside of that?

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketVideoRepo.kt
```diff
@@ -39,102 +31,46 @@ open class PocketVideoRepo(
...
+            .subscribe(this::postUpdate)
+            .addTo(compositeDisposable)
     }
 
     // When we the app is not in use, we don't want to hit the network for no reason, so we cancel updates
     @UiThread // stop updating backgroundUpdates.
     fun stopBackgroundUpdates() {
-        backgroundUpdates?.cancel()
-        backgroundUpdates = null
+        compositeDisposable.clear()
```
psymoon: Is this asynchronous?

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketVideoFragment.kt
```diff
@@ -40,33 +40,45 @@ class PocketVideoFragment : Fragment() {
...
 
+    private val compositeDisposable = CompositeDisposable()
+
     override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View {
-        val layout = inflater.inflate(R.layout.fragment_pocket_video, container, false)
+        return inflater.inflate(R.layout.fragment_pocket_video, container, false)
+    }
+
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        pocketHelpButton.setOnClickListener { _ ->
```
psymoon: Have we always had pocket help button?! 😮

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketVideoRepo.kt
```diff
@@ -39,102 +31,46 @@ open class PocketVideoRepo(
...
-                // Exponential backoff on failure
-                false -> {
-                    val doubled = retryDelayMillis * 2
-                    // Do nothing on overflow
-                    if (doubled > retryDelayMillis) retryDelayMillis = doubled
-                }
-            }
-        }
+    private fun postUpdate(response: Response<PocketData>) {
+        fun Response<PocketData>.toRepoState(): FeedState =
```
psymoon: This really simplifies our post logic! 👍

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketViewModel.kt
```diff
@@ -57,6 +57,8 @@ class PocketViewModel(
                 is PocketVideoRepo.FeedState.Inactive -> State.NotDisplayed
             }
         }
+        .replay(1)
+        .autoConnect(0)
```
psymoon: Do we have to specify this every time we don't want observable to autoconnect?

app/src/main/java/org/mozilla/tv/firefox/utils/PeriodicRequester.kt
psymoon: So this would be more on the data abstraction layer? Data transformation feed?

app/src/main/java/org/mozilla/tv/firefox/utils/PeriodicRequester.kt
```diff
@@ -0,0 +1,92 @@
...
+import java.util.concurrent.TimeUnit
+import kotlin.math.pow
+
+private const val CACHE_UPDATE_FREQUENCY_MINUTES = 45L
+private val CACHE_UPDATE_FREQUENCY_SECONDS = TimeUnit.MINUTES.toSeconds(CACHE_UPDATE_FREQUENCY_MINUTES)
+
+/**
+ * Simple interface for making a network request.
+ *
+ * This may operate on any thread, and consumers should take care to schedule
```
psymoon: This is a bold claim, can we use annotations?

app/src/main/java/org/mozilla/tv/firefox/utils/PeriodicRequester.kt
```diff
@@ -0,0 +1,92 @@
...
+     * Follow up requests will be made every 45 minutes, regardless of previous
+     * success or failure.
+     *
+     * Repeated calls to this method will create separate [Observable]s,
+     * potentially duplicating calls.
+     */
+    fun start(): Observable<Response<T>> = normalTimer()
+        .flatMap {
+            Observable.concat(requestAsynchronously().toObservable(), backoffRequests())
+                .takeUntil { it is Response.Success }
```
psymoon: So essentially we request forever until the 200 response (with backoff logic)? Maybe I am not understanding your backoff logic - is the backoff interval exponentially incremental? So we could get to a point `Integer.MAX_VALUE` ^ 2?

This should be okay, but generally this is a good place where we can manipulate tcp headers (i.e. etags) and/or http status codes. Maybe something we can work with Pocket API on

app/src/test/java/org/mozilla/tv/firefox/RxTest.kt
```diff
@@ -0,0 +1,56 @@
...
+
+        subject.onNext(3)
+
+        val observable2 = source.test()
+
+        observable1.assertValues(1, 2, 3)
+        observable2.assertValues(3)
+
+        observable1.dispose()
+        observable2.dispose()
```
psymoon: Ah this is cool, you can dispose observable subscription but the subject emitted data is not lost :).

app/src/test/java/org/mozilla/tv/firefox/RxTest.kt
```diff
@@ -0,0 +1,56 @@
...
+import io.reactivex.subjects.PublishSubject
+import org.junit.Test
+
+/**
+ * This class is for testing combinations of operators to ensure that they work
+ * as we expect.
+ *
+ * Rx is hard.  Don't wing it; test.
+ */
+class RxTest {
```
psymoon: Can we add a lifecycle test here as well? That is important considering we are looking to replace `LiveData` eventually.

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketVideoRepo.kt
Baron-Severin: Subjects can be observed, but they can also be pushed to (they implement both Observable and Observer interfaces).  This means they're a lot like mutable properties: sometimes necessary, but always dangerous, and should be kept as narrowly scoped as possible.

app/src/androidTest/java/org/mozilla/tv/firefox/helpers/CustomPocketFeedStateProvider.kt
```diff
@@ -20,22 +21,22 @@ import org.mozilla.tv.firefox.utils.BuildConfigDerivables
  */
 class CustomPocketFeedStateProvider {
 
-    private val pocketEndpoint = object : PocketEndpoint("VERSION", "www.mock.com".toUri()) {
+    private val localeIsEnglish: () -> Boolean = { true }
```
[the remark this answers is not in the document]
Baron-Severin: We have a unit test for non-EN locales, and it's simple enough logic that I don't think we need an Espresso test for it.  This is just a dependency for a few of these objects, so I wanted to declare it in one place instead of two.

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/NavigationOverlayFragment.kt
```diff
@@ -179,10 +185,42 @@ class NavigationOverlayFragment : Fragment() {
         updateFocusableViews()
     }
 
+    override fun onStart() {
+        super.onStart()
+        observePocketState()
```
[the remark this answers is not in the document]
Baron-Severin: Content will be set before the first frame is rendered anyway, I don't think we need to worry too much about this one.

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketVideoRepo.kt
Baron-Severin: Nope, it's synchronous.  Most Rx is synchronous by default, although there are a few exceptions.

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketVideoFragment.kt
Baron-Severin: Yeah, but it's pretty hidden....

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketViewModel.kt
Baron-Severin: I'm planning on giving a presentation soon on some of this, but here's the abridged version.

There are hot and cold observables.  Most `Observables` are cold, and they will start producing emissions whenever you subscribe to them.  Hot `Observables` are ones where the `Observer` can't control the source (e.g., with an `Observable<TouchEvent>`, you can't ask the user to not touch the screen until you've subscribed).  `autoConnect(n)` means "take this hot `Observable`, and tell it that it can start emitting values as soon as n `Observers` have subscribed to it".  So in this case, "start emitting results immediately, even if no one is listening".  `replay(n)` means it will cache the n most recently values it saw.  The end result is that we will always immediately have the most recently seen value cached, and it'll be emitted as soon as someone subscribes.

app/src/main/java/org/mozilla/tv/firefox/utils/PeriodicRequester.kt
Baron-Severin: Sounds like a good idea.  What would be the correct annotation for that?  I'm not sure.

app/src/main/java/org/mozilla/tv/firefox/utils/PeriodicRequester.kt
Baron-Severin: Yeah, the backoff is exponentially incremental.  It will cap out at `cache_update_frequency / 2` though, so we'll never actually reach `Integer.MAX_VALUE` ^ 2`.

Maybe we could talk later about how we can accomplish this with networking flags?  I'm not super knowledgable on that front.

app/src/test/java/org/mozilla/tv/firefox/RxTest.kt
Baron-Severin: I'm disposing manually in the `Fragment`, so this should be identical to how we're using it in practice.  I suppose we could add a test to the `Fragment`, and assert that dispose is called?

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketViewModel.kt
Baron-Severin: To follow up, most `Observables` you work with will be cold, so you won't need `autoconnect` or any similar operators.  Whenever working with a hot `Observable`, you will need to tell it when to start emitting data.

PR #1882, opened 2019-02-25
Pass Intent arguments through to GeckoView in developer builds.

### LINKED ISSUE #1881: Pass Intent arguments through to GeckoView in developer builds

I'm automating performance testing of the system WebView-based and the GeckoView-based versions, and I need to start a lot of sessions.  A lot.  Those sessions need to configure Turbo Mode and can't have the first run experience get in the way.  @npark-mozilla has a UI automator solution to this but it's not particularly flexible, and it doesn't smoothly extend to general GeckoView debugging.  For better or for worse, the way to achieve that is to pass through Intent arguments to the GeckoView engine creation process.

### What / Requirements

Allow to set Turbo Mode via Intent arguments; pass through arguments to GeckoView.

When I can automate [browsertime](https://github.com/sitespeedio/browsertime) in the way that I want!  Patch inbound.

ncalexan:
Closes #1881

### Pull Request checklist

I have no idea if tests are needed for this.  These are developer features; I expect bustage to be witnessed pretty quickly.

I don't really think this needs an entry, since it's not really user-facing.

Shouldn't CI do this for me?

I don't think this needs QA, other than other developers not having broken apps in the next few days.

### DIFF
--- a/app/src/gecko/java/org/mozilla/tv/firefox/webrender/WebRenderComponents.kt
+++ b/app/src/gecko/java/org/mozilla/tv/firefox/webrender/WebRenderComponents.kt
@@ -15,6 +15,7 @@ import org.mozilla.geckoview.GeckoRuntimeSettings
 import org.mozilla.tv.firefox.R
 import org.mozilla.tv.firefox.ext.serviceLocator
 import org.mozilla.tv.firefox.utils.BuildConstants
+import org.mozilla.tv.firefox.utils.SafeIntent
 import org.mozilla.tv.firefox.utils.Settings
 
 /**
@@ -22,6 +23,19 @@ import org.mozilla.tv.firefox.utils.Settings
  * application.
  */
 class WebRenderComponents(applicationContext: Context, systemUserAgent: String) {
+    // The first intent the App was launched with.  Used to pass configuration through to Gecko.
+    lateinit var launchSafeIntent: SafeIntent
+
+    fun notifyLaunchWithSafeIntent(safeIntent: SafeIntent): Boolean {
+        // We can't access the property reference outside of our own lexical scope,
+        // so this helper must be in this class.
+        if (!this::launchSafeIntent.isInitialized) {
+            launchSafeIntent = safeIntent
+            return true
+        }
+        return false
+    }
+
     val engine: Engine by lazy {
         fun getUserAgent(): String = UserAgent.buildUserAgentString(
                 applicationContext,
@@ -32,7 +46,7 @@ class WebRenderComponents(applicationContext: Context, systemUserAgent: String) 
         if (BuildConstants.isDevBuild) {
             // In debug builds, allow to invoke via an Intent that has extras customizing Gecko.
             // In particular, this allows to add command line arguments for custom profiles, etc.
-            val extras = applicationContext.serviceLocator.launchSafeIntent.extras
+            val extras = launchSafeIntent.extras
             if (extras != null) {
                 runtimeSettingsBuilder.extras(extras)
             }
--- a/app/src/gecko/java/org/mozilla/tv/firefox/webrender/WebRenderComponents.kt
+++ b/app/src/gecko/java/org/mozilla/tv/firefox/webrender/WebRenderComponents.kt
@@ -13,7 +13,6 @@ import mozilla.components.feature.session.SessionUseCases
 import org.mozilla.geckoview.GeckoRuntime
 import org.mozilla.geckoview.GeckoRuntimeSettings
 import org.mozilla.tv.firefox.R
-import org.mozilla.tv.firefox.ext.serviceLocator
 import org.mozilla.tv.firefox.utils.BuildConstants
 import org.mozilla.tv.firefox.utils.SafeIntent
 import org.mozilla.tv.firefox.utils.Settings
--- a/app/src/gecko/java/org/mozilla/tv/firefox/webrender/WebRenderComponents.kt
+++ b/app/src/gecko/java/org/mozilla/tv/firefox/webrender/WebRenderComponents.kt
@@ -10,7 +10,10 @@ import mozilla.components.browser.session.SessionManager
 import mozilla.components.concept.engine.DefaultSettings
 import mozilla.components.concept.engine.Engine
 import mozilla.components.feature.session.SessionUseCases
+import org.mozilla.geckoview.GeckoRuntime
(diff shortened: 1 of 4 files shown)


psymoon:
@ncalexan Looks like TC's failing due to `ktlint`. Make sure to run the pre-push hooks locally :)

ncalexan:
> @ncalexan Looks like TC's failing due to `ktlint`. Make sure to run the pre-push hooks locally :)

Thanks, @psymoon.  I've done this and it's green in CI: can I get re-review?

app/src/main/java/org/mozilla/tv/firefox/MainActivity.kt
```diff
@@ -76,15 +83,21 @@ class MainActivity : LocaleAwareAppCompatActivity(), OnUrlEnteredListener, Media
             }
         })
 
-        if (Settings.getInstance(this@MainActivity).shouldShowPocketOnboarding()) {
-            val onboardingIntents =
-                    Intent(this@MainActivity, PocketOnboardingActivity::class.java)
-            startActivity(onboardingIntents)
-        }
+        if (!intent.hasExtra("TURBO_MODE")) {
```
psymoon: So you want to be able to inject "TURBO_MODE" check through intent? If so, this isn't the best place to do it. If the Activity is never destroyed (i.e. backgrounded), onCreate() would never get called again. Please see `onNewIntent()`

app/src/gecko/java/org/mozilla/tv/firefox/webrender/WebRenderComponents.kt
```diff
@@ -25,6 +28,19 @@ class WebRenderComponents(applicationContext: Context, systemUserAgent: String)
                 systemUserAgent = systemUserAgent,
                 appName = applicationContext.resources.getString(R.string.useragent_appname))
 
+        val runtimeSettingsBuilder = GeckoRuntimeSettings.Builder()
+        if (BuildConstants.isDevBuild) {
+            // In debug builds, allow to invoke via an Intent that has extras customizing Gecko.
+            // In particular, this allows to add command line arguments for custom profiles, etc.
+            val extras = applicationContext.serviceLocator.launchSafeIntent.extras
```
psymoon: Is this copy by reference? If so, shouldn't this be synchronized? But I guess lazy{} by default grants thread safe. This seems flaky since `Engine` Initialization happens prior to `Activity.onCreate()`. Also closing the app does not guarantee the Activity to be destroyed

app/src/main/java/org/mozilla/tv/firefox/utils/ServiceLocator.kt
```diff
@@ -85,4 +85,17 @@ open class ServiceLocator(val app: Application) {
     }
     open val sessionRepo by lazy { SessionRepo(sessionManager, sessionUseCases, turboMode).apply { observeSources() } }
     open val settingsRepo by lazy { SettingsRepo(app) }
+
+    // The first intent the App was launched with.  Used to pass configuration through to Gecko.
+    lateinit var launchSafeIntent: SafeIntent
+
+    fun notifyLaunchWithSafeIntent(safeIntent: SafeIntent): Boolean {
```
psymoon: I'd prefer this to be in `WebRenderComponents` than in `ServiceLocator`. Alternatively you can create companion class in `SafeIntent`

app/src/main/java/org/mozilla/tv/firefox/MainActivity.kt
ncalexan: I think you're correct, this should use `onNewIntent`: it makes sense to "re-launch" with a different turbo mode setting.  I'll do that.  I'm going to keep it here too, though, to disable the first-run experience in automation.

app/src/gecko/java/org/mozilla/tv/firefox/webrender/WebRenderComponents.kt
ncalexan: I'm not sure what the ownership model is here.  The `launchSafeIntent` itself is synchronized by the Swift runtime, and I expect all access to `extras` to be read-only, but that's (probably?) not guaranteed.

Re: flakiness and initialization order, that's the devil we must dance with.  Gecko is initialized at most once *per process*, and it's initialized very early.  I am not aware of a way to get the launching Intent extras (if there are any) before the first `Activity.onCreate()`.  So yes, this necessarily races Gecko startup.  @psymoon, do you know of such a way to get the extras in another place?  It can't be in `Application.onCreate` since a service could start the App before any Activity starts GeckoView...

app/src/main/java/org/mozilla/tv/firefox/utils/ServiceLocator.kt
ncalexan: I will try to make this so.

app/src/gecko/java/org/mozilla/tv/firefox/webrender/WebRenderComponents.kt
ncalexan: Also, it cannot be true that `Engine` initialization happens prior to `Activity.onCreate`, because locally the flags I send in as Intent extras make it to the GV creation.

app/src/main/java/org/mozilla/tv/firefox/MainActivity.kt
psymoon: Sounds good

ncalexan:
@psymoon okay, I've rebased this and pushed the changes as fixup commits.  I'll squash for landing if you're happy with them.  I have plans to improve the Gecko arguments stuff in a way that won't require GV consumers to pass Intent arguments around at all, but that's not for today.

app/src/gecko/java/org/mozilla/tv/firefox/webrender/WebRenderComponents.kt
psymoon: > Also, it cannot be true that `Engine` initialization happens prior to `Activity.onCreate`, because locally the flags I send in as Intent extras make it to the GV creation.

I think you're right technically. If it's not, we are doing something wrong...

> @psymoon, do you know of such a way to get the extras in another place? It can't be in Application.onCreate since a service could start the App before any Activity starts GeckoView...

Activity.onCreate is the correct way, I think since we are checking for null anyway it may be just fine

PR #1922, opened 2019-03-07
Issue #1921: Use `substitute-local-geckoview.gradle` from Bug 1533465…

### LINKED ISSUE #1921: Use `substitute-local-geckoview.gradle` from Bug 1533465 for local GeckoView substitution.
This is a development ergonomics change only. Over in [Bug 1533465](https://bugzilla.mozilla.org/show_bug.cgi?id=1533465) I'm trying to unify how downstream GeckoView consumers (at least, Mozilla consumers) substitute local GeckoView versions. (This is all leading up to Gradle composite builds, someday, but that day is not this day.)

This is an example of consuming that ticket in a downstream consumer.

ncalexan:
… for local GeckoView substitution.

See discussion in https://bugzilla.mozilla.org/show_bug.cgi?id=1533465.

This can land before Bug 1533465, since there's no existing support for this.  We may want to build more support for this in, i.e., parse `local.properties` to find substitution paths rather than requiring changes to `build.gradle`.  But that can be follow-up: it would be best to do that with a shared plugin, rather than with a script like this.

Closes #1921

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -147,7 +147,7 @@ dependencies {
     // Required to override android.support lib in GeckoView dependency (uses palette-v7:26.1.0)
     geckoImplementation "androidx.palette:palette:$androidx_version"
     geckoImplementation "org.mozilla.components:browser-engine-gecko-nightly:$moz_components_version"
-    geckoImplementation Gecko.geckoview_nightly_arm
+    geckoImplementation "${Gecko.geckoview_nightly_arm}@aar"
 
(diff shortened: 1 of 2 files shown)


ncalexan:
> @ncalexan Is there a follow-up for the shared plugin somewhere? Maybe in https://github.com/mozilla-mobile/android-automation-tools/tree/master/gradle-plugin perhaps?

I think, in fact, that I'm going to do this in the Gradle plugin rather than this way at all.  Will update when that's done.  Sorry for the noise!

PR #1928, opened 2019-03-07
1916 fix back from webpage exit

### LINKED ISSUE #1916: Back to exit doesn't work from webpage
Looks like caused in a recent commit with #277 

## Steps to reproduce
Launch Firefox
Load website
Click remote Back to return to about:home/overlay
Click remote Back again

### Expected behavior
On back from about:home/overlay, app exits

### Actual behavior
Overlay toggles, doesn't exit

### Device information
* Fire TV device: Emulator 26
* Latest affected Firefox version: master 785501a9eb6c68386b84d0f1104fc5b2f19ed9cb
* Earliest affected Firefox version: master 785501a9eb6c68386b84d0f1104fc5b2f19ed9cb

Baron-Severin:
Closes #1916 

### Pull Request checklist

### DIFF
--- a/app/src/androidTest/java/org/mozilla/tv/firefox/ui/BasicNavigationTest.kt
+++ b/app/src/androidTest/java/org/mozilla/tv/firefox/ui/BasicNavigationTest.kt
@@ -128,6 +128,7 @@ class BasicNavigationTest {
             assertNotEquals(preReloadTitle, getPageTitle())
         }
     }
+
     @Test
     fun exitAppByBackButtonTest() {
         navigationOverlay {
@@ -135,6 +136,24 @@ class BasicNavigationTest {
             assertActivityFinishing(activityTestRule)
         }
     }
+
+    @Test
+    fun navigateThenExitAppByBackButtonTest() {
+        val server = MockWebServer().apply {
+            setDispatcher(AndroidAssetDispatcher())
+            start()
+        }
+
+        val pages = TestAssetHelper.getGenericAssets(server)
+
+        navigationOverlay {
+        }.enterUrlAndEnterToBrowser(pages.first().url) {
+        }.backButtonToOverlay {
+            remoteBack()
+            assertActivityFinishing(activityTestRule)
+        }
+    }
+
     @Test
     fun screenNavigationFocusTest() {
         navigationOverlay {
--- a/app/src/main/java/org/mozilla/tv/firefox/ScreenController.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/ScreenController.kt
@@ -18,6 +18,7 @@ import org.mozilla.tv.firefox.ScreenControllerStateMachine.Transition
 import org.mozilla.tv.firefox.ext.serviceLocator
 import org.mozilla.tv.firefox.navigationoverlay.NavigationOverlayFragment
 import org.mozilla.tv.firefox.pocket.PocketVideoFragment
+import org.mozilla.tv.firefox.session.SessionRepo
 import org.mozilla.tv.firefox.settings.SettingsFragment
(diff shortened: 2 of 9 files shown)


app/src/androidTest/java/org/mozilla/tv/firefox/ui/BasicNavigationTest.kt
```diff
@@ -128,13 +128,32 @@ class BasicNavigationTest {
...
     fun exitAppByBackButtonTest() {
         navigationOverlay {
             remoteBack()
             assertActivityFinishing(activityTestRule)
         }
     }
+
+    @Test
+    fun navigateThenExitAppByBackButtonTest() {
+        val server = MockWebServer().apply {
```
liuche: Since we might be using this for multiple tests in this file, I wonder if there are any savings/clarity by setting this up in one place, but a) not every test uses it, and b) since we clear app state in between test runs, I guess it doesn't matter because the server object can't persist?

app/src/main/java/org/mozilla/tv/firefox/ScreenController.kt
```diff
@@ -169,19 +170,21 @@ class ScreenController {
...
 
-    private fun isOnHomeUrl(fragmentManager: FragmentManager): Boolean {
-        // TODO: Would be more correct to get this from the model rather than the Fragment.
-        val webRenderFragment = fragmentManager.webRenderFragment()
-        return webRenderFragment.session.url == URLs.APP_URL_HOME
+    private fun canGoBack(): Boolean {
+        return sessionRepo.state.value?.backEnabled == true
+    }
+
+    private fun isOnHomeUrl(): Boolean {
```
liuche: Good on pulling this out.

app/src/main/java/org/mozilla/tv/firefox/ScreenController.kt
```diff
@@ -169,19 +170,21 @@ class ScreenController {
...
+        val transition = ScreenControllerStateMachine.getNewStateMenuPress(_currentActiveScreen.value!!, isOnHomeUrl())
         return handleTransitionAndUpdateActiveScreen(fragmentManager, transition)
     }
 
-    private fun isOnHomeUrl(fragmentManager: FragmentManager): Boolean {
-        // TODO: Would be more correct to get this from the model rather than the Fragment.
-        val webRenderFragment = fragmentManager.webRenderFragment()
-        return webRenderFragment.session.url == URLs.APP_URL_HOME
+    private fun canGoBack(): Boolean {
+        return sessionRepo.state.value?.backEnabled == true
```
liuche: It's good to gate this on back stack - one thing this does do is add another layer on top of the `canGoBackTwice` which still bothers me, but that's a bigger fix (and depends on https://github.com/mozilla-mobile/android-components/issues/1993)

app/src/androidTest/java/org/mozilla/tv/firefox/ui/BasicNavigationTest.kt
Baron-Severin: I think we'd ideally just have a really nice helper that sets things up, but we've tried that and, to be perfectly honest, I think we've landed on a state where it's just confusing.  The requirements are a little bit different between each test.  I haven't really figured out how I want this to work, but I share your concerns.

app/src/main/java/org/mozilla/tv/firefox/ScreenController.kt
Baron-Severin: I think it should be roughly independent of `canGoBackTwice`, right?  Regardless of how we calculate when the back button is enabled, once it's not we should exit the app.  To me that seems like a relatively generic seam.

PR #1941, opened 2019-03-12
Closes #1865 - Workaround for grey screen on Youtube resume.

### LINKED ISSUE #1865: Grey screen on YouTube home page after Home button interruption
## Steps to reproduce

1. Start up the browser for the first time
2. Access the YouTube tile
3. When on the YouTube home page, make a Home button interruption
4. Return back to the browser

### Expected behavior

YouTube home page loads correctly

### Actual behavior

Grey screen appears after re-entering the browser. I need to tap on the Back button or D-Pad and the Home page reappears

The issues is not 100% reproducible. From 15 tries I managed to reproduce it 4-5 times

### Device information
* Fire TV device: 4kPendant (OS 6.2.5.8), Basic Stick (OS 5.2.6.6)
* Latest affected Firefox version: 3.4.1-LAT2
* Earliest affected Firefox version: 3.4.1-LAT2

https://www.youtube.com/watch?v=oY2awf32zMI&feature=youtu.be

## Alternative steps to repro from #1559
- Open YouTube (not reproducible on any other site)
- Press menu button to open overlay
- Press home button
- Reopen FFTV
- Press menu button to close overlay

liuche:
This is a workaround for getting Youtube to not grey-screen that sends keypress events. Unfortunately, other no-op keypresses don't seem to work - maybe it's something that requires a repaint?
Could not repro this with 30 tries.

This requires a SuppressLint because of an Android bug: https://issuetracker.google.com/issues/37130193#comment21
### Pull Request checklist

### DIFF
--- a/app/src/main/java/org/mozilla/tv/firefox/webrender/WebRenderFragment.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/webrender/WebRenderFragment.kt
@@ -4,6 +4,7 @@
 
 package org.mozilla.tv.firefox.webrender
 
+import android.annotation.SuppressLint
 import android.app.Activity
 import androidx.lifecycle.Observer
 import androidx.lifecycle.ViewModelProviders
@@ -15,6 +16,10 @@ import android.view.View
 import android.view.ViewGroup
 import android.view.WindowManager
 import kotlinx.android.synthetic.main.fragment_browser.view.*
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.GlobalScope
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.launch
 import mozilla.components.browser.session.Session
 import mozilla.components.concept.engine.EngineView
 import mozilla.components.concept.engine.permission.Permission
@@ -73,6 +78,19 @@ class WebRenderFragment : EngineViewLifecycleFragment(), Session.Observer {
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
         initSession()
+    }
+
+    @SuppressLint("RestrictedApi")
+    override fun onResume() {
+        super.onResume()
+        if (session.isYoutubeTV) {
+            // Send key events on the UI thread to clear webview grey screen, see #1865
+            GlobalScope.launch(Dispatchers.Main) {
+                delay(50)
+                activity?.dispatchKeyEvent(KeyEvent(KeyEvent.ACTION_DOWN, KeyEvent.KEYCODE_DPAD_RIGHT))
+                activity?.dispatchKeyEvent(KeyEvent(KeyEvent.ACTION_DOWN, KeyEvent.KEYCODE_DPAD_LEFT))
+            }
+        }
     }
 
     private fun initSession() {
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/mozilla/tv/firefox/webrender/WebRenderFragment.kt
```diff
@@ -75,6 +80,19 @@ class WebRenderFragment : EngineViewLifecycleFragment(), Session.Observer {
...
 
+    @SuppressLint("RestrictedApi")
+    override fun onResume() {
+        super.onResume()
+        if (session.isYoutubeTV) {
+            // Send key events on the UI thread to clear webview grey screen, see #1865
+            GlobalScope.launch(Dispatchers.Main) {
+                delay(50)
+                activity?.dispatchKeyEvent(KeyEvent(KeyEvent.ACTION_DOWN, KeyEvent.KEYCODE_DPAD_RIGHT))
+                activity?.dispatchKeyEvent(KeyEvent(KeyEvent.ACTION_DOWN, KeyEvent.KEYCODE_DPAD_LEFT))
```
Baron-Severin: Nit: I think we could more clearly communicate what this is doing by breaking it into a new class (something like `YouTubeGreyScreenWorkaround#invoke`).  The code is pretty trivial, but it would 1) let us add a kdoc with some relevant context, and 2) clearly mark it is a workaround.

app/src/main/java/org/mozilla/tv/firefox/webrender/WebRenderFragment.kt
```diff
@@ -75,6 +80,19 @@ class WebRenderFragment : EngineViewLifecycleFragment(), Session.Observer {
...
     }
 
+    @SuppressLint("RestrictedApi")
+    override fun onResume() {
+        super.onResume()
+        if (session.isYoutubeTV) {
+            // Send key events on the UI thread to clear webview grey screen, see #1865
+            GlobalScope.launch(Dispatchers.Main) {
+                delay(50)
+                activity?.dispatchKeyEvent(KeyEvent(KeyEvent.ACTION_DOWN, KeyEvent.KEYCODE_DPAD_RIGHT))
```
Baron-Severin: Nit: I think it would be good to include a comment with a few of the things you tried to get this to work.  I know the information is in the comment, but a brief list helps to contextualize why we're doing this.

app/src/main/java/org/mozilla/tv/firefox/webrender/WebRenderFragment.kt
```diff
@@ -75,6 +80,19 @@ class WebRenderFragment : EngineViewLifecycleFragment(), Session.Observer {
         initSession()
     }
 
+    @SuppressLint("RestrictedApi")
+    override fun onResume() {
+        super.onResume()
+        if (session.isYoutubeTV) {
+            // Send key events on the UI thread to clear webview grey screen, see #1865
+            GlobalScope.launch(Dispatchers.Main) {
+                delay(50)
```
Baron-Severin: Nit: I know we're only blocking for 50 MS here, but I'm a little concerned about blocking the main thread.  It seems like it would be better to delay on a background thread then jump back (or post) to dispatch the events.

liuche: Actually `delay` is suspending, not blocking: 
https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/delay.html

Baron-Severin: Huh, so this launches on the main loop, suspends on a background thread, returns to main, then does its work?  Well, no changes needed then.  Thanks!

Issue #1939, opened 2019-03-11
YouTube Back Broken


Baron-Severin:
## Steps to reproduce
- Launch app
- Open YouTube tile
- Press back repeatedly

### Expected behavior
- App should return to home screen after focus moves to YouTube side bar

### Actual behavior
- App never returns to home screen

### Device information
* Fire TV device: Cube, emulator
* Latest affected Firefox version: ac9c612b5e4c44af98a25d6906f962668b0f9e6c (head of master)
* Earliest affected Firefox version: c9f19ecb131be8f313b2c8efc918b794c106b8e1

### Debugging Notes
- Commenting out the following line in `MainActivity#onCreate` fixes the problem:
```kotlin
webRenderComponents.sessionManager.getOrCreateEngineSession().resetView(this@MainActivity)
```
- The proximate cause of this bug is that, when we attempt to back from YouTube, "app:home" is not in the backstack (we expect that it would be)
  - Note that this means that the following steps still work as expected
```
* Launch app
* Open IMDB tile
* Open Menu
* Open YouTube Tile
* Press back repeatedly
(success)
```

Baron-Severin:
Replacing `webRenderComponents.sessionManager.getOrCreateEngineSession().resetView(this@MainActivity)`
with 
`webRenderComponents.sessionManager.getOrCreateEngineSession()`
 fixes the bug.  So this is specifically related to resetting the view, and not just a change in initialization order.

liuche:
When this gets fixed, please cherry-pick this onto the 3.5-LAT2 branch.

levisacal:
wasn't able to reproduce anymore in 3.5-LAt on 4k pendant

PR #1943, opened 2019-03-12
1927 fix youtube menu back

### LINKED ISSUE #1927: Back/forward buttons in overlay do not work on YouTube TV
## Steps to reproduce
- Open app
- Click Youtube tile
- open video
- open overlay
- hit back

### Expected behavior
Youtube goes back to homepage

### Actual behavior
Youtube stays on the current video

### Device information
* Fire TV device: Stick gen 2
* Latest affected Firefox version: latest, v3.4.1-lat1?

I cannot reproduce this issue when using the hardware back button.

Baron-Severin:
Closes #1927 
Depends on #1939 

### Pull Request checklist

I tried to get this under test for hours, but ran into issues on the edge of A-C.  I'm going to have to think about SessionRepo for a while and come back to it. See https://github.com/Baron-Severin/firefox-tv/pull/3

### DIFF
--- a/app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/ToolbarViewModel.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/ToolbarViewModel.kt
@@ -70,7 +70,7 @@ class ToolbarViewModel(
 
     @UiThread
     fun backButtonClicked() {
-        sessionRepo.exitFullScreenIfPossibleAndBack()
+        sessionRepo.attemptBack(forceYouTubeExit = true)
         hideOverlay()
     }
 
--- a/app/src/main/java/org/mozilla/tv/firefox/session/SessionRepo.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/session/SessionRepo.kt
@@ -4,16 +4,20 @@
 
 package org.mozilla.tv.firefox.session
 
-import androidx.lifecycle.LiveData
-import androidx.lifecycle.MutableLiveData
 import android.graphics.Bitmap
 import android.net.Uri
 import androidx.annotation.AnyThread
+import androidx.lifecycle.LiveData
+import androidx.lifecycle.MutableLiveData
+import io.reactivex.subjects.PublishSubject
+import io.reactivex.subjects.Subject
 import mozilla.components.browser.session.Session
 import mozilla.components.browser.session.SessionManager
 import mozilla.components.feature.session.SessionUseCases
+import org.mozilla.tv.firefox.ext.isYoutubeTV
 import org.mozilla.tv.firefox.ext.postIfNew
 import org.mozilla.tv.firefox.ext.toUri
+import org.mozilla.tv.firefox.telemetry.TelemetryIntegration
 import org.mozilla.tv.firefox.utils.TurboMode
 import org.mozilla.tv.firefox.webrender.EngineViewCache
 
@@ -34,8 +38,15 @@ class SessionRepo(
         val currentUrl: String
     )
 
+    enum class Event {
+        YouTubeBack, ExitYouTube
+    }
+
     private val _state = MutableLiveData<State>()
     val state: LiveData<State> = _state
+
+    private val _events: Subject<Event> = PublishSubject.create()
+    val events = _events.hide()
 
     var canGoBackTwice: (() -> Boolean?)? = null
     private var previousURLHost: String? = null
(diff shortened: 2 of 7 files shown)


Baron-Severin:
Merging this is blocked on https://github.com/mozilla-mobile/firefox-tv/issues/1939, but the code is ready for review.

app/src/main/java/org/mozilla/tv/firefox/session/SessionRepo.kt
```diff
@@ -34,9 +38,16 @@ class SessionRepo(
...
 
+    enum class Event {
+        YouTubeBack, ExitYouTube
+    }
+
     private val _state = MutableLiveData<State>()
     val state: LiveData<State> = _state
 
+    private val _events: Subject<Event> = PublishSubject.create()
+    val events = _events.hide()
```
liuche: Okay, so you `hide()` here to prevent others from calling operations like `next()` on `_events`? What's the scope of that restriction, e.g. this class, and is it redundant because `_events` is private to the class anyway? (I guess it seems like good hygiene in any case.)
And no one can just call `events.onNext` because it's just an Observable, not an ObservableSource.

app/src/main/java/org/mozilla/tv/firefox/webrender/WebRenderFragment.kt
```diff
@@ -155,13 +161,21 @@ class WebRenderFragment : EngineViewLifecycleFragment(), Session.Observer {
...
+            }.forceExhaustive
+        }.addTo(compositeDisposable)
     }
 
     override fun onStop() {
         super.onStop()
 
         serviceLocator!!.sessionRepo.exitFullScreenIfPossible()
         sessionFeature?.stop()
+        compositeDisposable.clear()
```
liuche: Is it better to clear the compositeDisposable in bulk (like here in `onStop`) vs as we receive each event?

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/ToolbarViewModel.kt
```diff
@@ -70,7 +70,7 @@ class ToolbarViewModel(
 
     @UiThread
     fun backButtonClicked() {
-        sessionRepo.exitFullScreenIfPossibleAndBack()
+        sessionRepo.attemptBack(forceYouTubeExit = true)
```
liuche: What if the user wants to go back through the YT history stack? e.g. they have watched a series of videos. (Though maybe this is already the behavior now...)
But this is a weird case - e.g. we want to handle web browsing history, but not trigger things like the "back" cursor movement.
What do you think?

liuche:
(Sorry for the delay and maybe the over-detailed Rx review/questions - it seemed like a good opportunity to get some Rx learning in so I took it!)

app/src/main/java/org/mozilla/tv/firefox/webrender/YouTubeBackHandler.kt
```diff
@@ -30,15 +30,14 @@ class YouTubeBackHandler(private val engineView: EngineView?, private val activi
     private val preYouTubeIndexHistory: MutableList<Int> = mutableListOf()
     private var currentPreYouTubeIndex: Int? = null
 
-    fun handleBackClick(event: KeyEvent) {
+    fun handleBackClick() {
```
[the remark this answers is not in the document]
Baron-Severin: This comes from `Activity#onBackPressed`.  Do you think I should rename it to `onBackPressed` to better communicate that?

app/src/main/java/org/mozilla/tv/firefox/webrender/YouTubeBackHandler.kt
```diff
@@ -68,10 +67,10 @@ class YouTubeBackHandler(private val engineView: EngineView?, private val activi
         }
     }
 
+    fun goBackBeforeYouTube() = engineView!!.handleYoutubeBack(getIndexToGoBackTo())
```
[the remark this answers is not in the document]
Baron-Severin: The param is nullable, but looking more closely it probably shouldn't be, since the whole class won't function without it.  I'm going to give some thought to how to fix that.

app/src/main/java/org/mozilla/tv/firefox/webrender/WebRenderFragment.kt
```diff
@@ -155,13 +161,21 @@ class WebRenderFragment : EngineViewLifecycleFragment(), Session.Observer {
         super.onStart()
 
         sessionFeature?.start()
+        serviceLocator!!.sessionRepo.events.subscribe {
+            when (it) {
+                SessionRepo.Event.YouTubeBack -> youtubeBackHandler.handleBackClick()
+                SessionRepo.Event.ExitYouTube -> youtubeBackHandler.goBackBeforeYouTube()
+                null -> return@subscribe
```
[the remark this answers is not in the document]
Baron-Severin: It can't, but the compiler doesn't believe me (frustratingly, nullability annotations are inconsistently applied in RxJava).  Maybe I should add a comment about that.

app/src/main/java/org/mozilla/tv/firefox/session/SessionRepo.kt
Baron-Severin: Code local to the class can still push values through `_events`, but anything else only has access to the read-only `events`.  This is similar to what we were doing before (with `val events: LiveData<T> = _events`, see `SessionRepo.events`), but with the upside that `Observables` obtained using `.hide()` cannott be cast back to subjects (`LiveData` can be cast back and forth freely at any time).

app/src/main/java/org/mozilla/tv/firefox/webrender/WebRenderFragment.kt
Baron-Severin: `Disposable` is equivelent to a coroutine `Job`, so it doesn't hold references to passed events.  Instead, `compositeDisposable.clear()` is basically `List<Job>.forEach { it.cancel() }`

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/ToolbarViewModel.kt
Baron-Severin: I _think_ that's handled by YouTube using the ESC key, but I'll verify that before merging.

Baron-Severin: > I think that's handled by YouTube using the ESC key, but I'll verify that before merging.

It looks like this is true.  It is interesting to note that, if you've watched a bunch of videos, back will still just take you to YouTube TV home.  But that's really Google's decision.

Issue #1919, opened 2019-03-07
Show build variant when in debug mode


psymoon:
### Why/User Benefit/User Problem

As a Mozilla team member running tests, I'd like an easy way to see which build variant I'm currently on, so that I don't run my testing against the wrong variant

### What / Requirements

* Display build variant when in debug mode
* Display engine type (GV or WV) and version

### Acceptance Criteria (how do I know when I’m done?)

* Technical & non-technical mozillians can see which build they are running when in debug mode
* Technical & non-technical mozillians can see whether they're using GV version or WV version when in debug mode
* There are no end user facing changes
* There are 2 related issues. They should be closed off or reprioritized once this work is completed #824 #1785 

- [x] debug overlay for developers
- [ ] about page for QA, LAT and end users (#1785 )

mcomella:
I think production Focus builds provide the GeckoView build number. We should:
- Do at least what they do
- Share their code (can this be a component?)

mcomella:
See also adding commit to builds – #824 – and I notice that Fenix provides a, "Built on Aug 4th @ 2am" too.

psymoon:
Related: https://github.com/mozilla-mobile/firefox-tv/issues/1785

PR #1952, opened 2019-03-13
Show build variant when in debug mode

psymoon:
Connected to #1919 

### Pull Request checklist

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -276,6 +276,14 @@ android.applicationVariants.all { variant ->
 }
 
 // -------------------------------------------------------------------------------------------------
+// Gecko: Put GeckoVersion into BuildConfig
+// -------------------------------------------------------------------------------------------------
+
+android.applicationVariants.all {
+    buildConfigField 'String', 'GECKOVIEW_VERSION', "\"${GeckoVersions.nightly_version}\""
+}
+
+// -------------------------------------------------------------------------------------------------
 // L10N: Generate list of locales
 // Firefox TV provides its own (Android independent) locale switcher. That switcher requires a list
 // of locale codes. We generate that list here to avoid having to manually maintain a list of locales:
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -279,8 +279,10 @@ android.applicationVariants.all { variant ->
 // Gecko: Put GeckoVersion into BuildConfig
 // -------------------------------------------------------------------------------------------------
 
-android.applicationVariants.all {
-    buildConfigField 'String', 'GECKOVIEW_VERSION', "\"${GeckoVersions.nightly_version}\""
+android.applicationVariants.all { variant ->
+    if (variant.flavorName == "gecko") {
+        buildConfigField 'String', 'GECKOVIEW_VERSION', "\"${GeckoVersions.nightly_version}\""
+    }
 }
 
 // -------------------------------------------------------------------------------------------------
--- a/app/src/main/java/org/mozilla/tv/firefox/MainActivity.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/MainActivity.kt
@@ -16,6 +16,7 @@ import android.view.KeyEvent
 import android.view.View
 import io.sentry.Sentry
 import kotlinx.android.synthetic.main.activity_main.*
+import kotlinx.android.synthetic.main.overlay_debug.*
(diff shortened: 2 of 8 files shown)


app/build.gradle
```diff
@@ -275,6 +275,14 @@ android.applicationVariants.all { variant ->
     addSecretToBuildConfig(variant, 'POCKET_KEY', ".pocket_key")
 }
 
+// -------------------------------------------------------------------------------------------------
+// Gecko: Put GeckoVersion into BuildConfig
+// -------------------------------------------------------------------------------------------------
+
+android.applicationVariants.all {
```
Baron-Severin: Could we filter this to Gecko variants only?

app/src/main/res/layout/overlay_debug.xml
```diff
@@ -0,0 +1,13 @@
...
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
+<TextView
+    xmlns:android="http://schemas.android.com/apk/res/android"
+    android:id="@+id/debugLog"
+    android:layout_width="wrap_content"
+    android:layout_height="wrap_content"
+    android:layout_gravity="end"
+    android:padding="@dimen/lb_action_icon_margin"
+    android:background="@color/lb_basic_card_bg_color"
+    android:textSize="@dimen/lb_action_text_size" />
```
Baron-Severin: IMO, visibility should default to `GONE` here.

app/src/main/java/org/mozilla/tv/firefox/MainActivity.kt
```diff
@@ -100,6 +101,16 @@ class MainActivity : LocaleAwareAppCompatActivity(), OnUrlEnteredListener, Media
...
         serviceLocator.intentLiveData.value = Consumable.from(intentData)
+
+        if (BuildConfig.DEBUG) {
+            debugLog.visibility = View.VISIBLE
+            val debugLogStr = "TYPE: ${BuildConfig.BUILD_TYPE} / " +
+                    "FLAVOR: ${BuildConfig.FLAVOR} / VERSION: ${BuildConfig.VERSION_NAME}"
+
+            debugLog.text = debugLogStr
+        } else {
+            debugLog.visibility = View.GONE
```
Baron-Severin: I'm of the opinion that test code shouldn't ship with our APK, so this kind of check worries me a little bit.  Would you be opposed to handling this via source set instead?  We could have a `BuildDisplayer` class with debug and release implementations.  That way the code to set this view up wouldn't even ship to our users.

app/src/main/java/org/mozilla/tv/firefox/utils/BuildConstants.kt
```diff
@@ -13,10 +13,14 @@ import org.mozilla.tv.firefox.BuildConfig
...
+    private const val PRODUCT_FLAVOR_GECKO = "gecko"
 
     val isDevBuild: Boolean
         get() = BUILD_TYPE_DEBUG == BuildConfig.BUILD_TYPE
 
     val isReleaseBuild: Boolean
         get() = BUILD_TYPE_RELEASE == BuildConfig.BUILD_TYPE
+
+    val isGeckoBuild: Boolean
+        get() = PRODUCT_FLAVOR_GECKO == BuildConfig.FLAVOR
```
Baron-Severin: Same thing as before.  This might be me being overly inflexible, so feel free to push back on this, but I don't really like the existence of these methods.  It seems like a code smell to me.  I would much rather handle this via source set.

app/build.gradle
psymoon: @Baron-Severin then we would have conditional BuildConfig :/. all checks should be split into variants

app/src/main/java/org/mozilla/tv/firefox/MainActivity.kt
psymoon: @Baron-Severin  We have test codes in other places (i.e. enable remoteDebugging, show PocketError, etc)

app/src/main/java/org/mozilla/tv/firefox/utils/BuildConstants.kt
psymoon: @Baron-Severin you do raise a good concern though, i'll create a follow-up story for all debug/test related code?

PR #1954, opened 2019-03-13
Closes #1824: Fix internal url obstruction in gv

### LINKED ISSUE #1824: Gecko build "about:home" is broken on launch
## Steps to reproduce
- Select "geckoDebug" build variant
- Run and open the FFTV app

### Expected behavior
- Overlay opens with `about:home` loaded in the url bar

### Actual behavior
- Overlay opens with `data:text/html,<html></html>` in the url bar

psymoon:
### Pull Request checklist

### DIFF
--- a/app/src/main/java/org/mozilla/tv/firefox/utils/UrlUtils.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/utils/UrlUtils.kt
@@ -158,8 +158,11 @@ object UrlUtils {
      */
     @JvmStatic
     fun toUrlBarDisplay(url: String): String {
-        return when (url) {
-            URLs.APP_URL_HOME -> "" // Empty string forces the EditText to show hint text
+        return when {
+            url == URLs.APP_URL_HOME -> "" // Empty string forces the EditText to show hint text
+            // GeckoView loads [data: url] when RequestInterceptor.InterceptionResponse.Content()
+            // due to lack of internal page api (bug #1535131)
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mozilla/tv/firefox/utils/UrlUtils.kt
```diff
@@ -158,8 +158,11 @@ object UrlUtils {
      */
     @JvmStatic
     fun toUrlBarDisplay(url: String): String {
-        return when (url) {
-            URLs.APP_URL_HOME -> "" // Empty string forces the EditText to show hint text
+        return when {
+            url == URLs.APP_URL_HOME -> "" // Empty string forces the EditText to show hint text
+            // GeckoView loads [data: url] when RequestInterceptor.InterceptionResponse.Content()
+            // due to lack of internal page api (bug #1535131)
+            url.startsWith("data:", false) -> ""
```
liuche: Hm, it looks like from bug 1535131, this means that the url will actually just show the hint, but not any sort of "firefox:about" url? I guess that's fine for this workaround.
From the discussion though, it looks like this isn't high enough priority for GeckoView to fix because it'll be a non-trivial amount of work, so our "about:" pages in GV Firefox will just show the hint - heads up @athomasmoz is that fine?

athomasmoz: Agreed. This sounds fine to me

PR #1968, opened 2019-03-18
Closes #1967, add l10n.toml 🔄

### LINKED ISSUE #1967: Add l10n.toml configuration to add Fire TV to android-l10n
Kicking off the l10n-drivers side to migrate FFTV to the new l10n ecosystem.

This is a sibling to https://github.com/mozilla-mobile/firefox-echo-show/pull/261, note the different Spanish variant here compared to there.

Which according to @Delphine is how it should be. CC @athomasmoz

Pike:
This is only a configuration change, to be used by the upcoming l10n ecosystem. Thus not tests or changelogs or QA tags on this one.

### Pull Request checklist

### DIFF
--- a/l10n.toml
+++ b/l10n.toml
@@ -1,0 +1,27 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+basepath = "."
+
+locales = [
+    "de",
+    "es-ES",
+    "fr",
+    "it",
+    "ja",
+    "pt-BR",
+    "zh-CN",
+]
+
+# Expose the following branches to localization
+# Changes to this list should be announced to the l10n team ahead of time.
+branches = [
+    "master",
+]
+
+[env]
+
+[[paths]]
+  reference = "app/src/main/res/values/strings.xml"
+  l10n = "app/src/main/res/values-{android_locale}/strings.xml"


liuche:
I'm not sure why the TC failed, but the error seems weird and I'm just going to restart it:
```
> Transform artifact okhttp-urlconnection.jar (com.squareup.okhttp3:okhttp-urlconnection:3.13.1) with JetifyTransform
ERROR: [TAG] Failed to resolve variable '${animal.sniffer.version}'
```

Pike:
You use this to check for l10n problems. You'd install compare-locales,

```sh
pip install compare-locales
```

and then run

```sh
compare-locales --validate l10n.toml .
```

to check for errors in en-US. You can give that a try by adding a duplicate message, for example. Or by adding a `localizable="false"` attribute. It basically checks for stuff that throws off the l10n toolchain.

With

```sh
compare-locales l10n.toml .
```

you check the localizations for problems. You'll find that there are tons here, mostly due to `android-po` adding useless `"` characters all over the place, which we don't allow in our native Android support. This is gonna be a tad interesting when converting from `android-po` to our native support, but the quotes are not in the pontoon database, so we'll use those strings in the conversion, that should be good for us, and fine for FFTV.

In case you wonder, the `.` is required to denote that the localizations are in the FFTV repo. That's mostly used for desktop l10n, where the l10n repos are "wherever" the developer or automation put it. We're using the same compare-locales tool there, too.

Pike:
... weird, travis shows as pending on the PR, but is green when I click on details.

liuche:
Thank you! This is very helpful, I'll add it to our l10n wiki page.

PR #1970, opened 2019-03-18
Send better logs for ExperimentsProvider fallback

### LINKED ISSUE #1969: Use better logging for ExperimentsProvider fallbacks
`checkBranchVariants()` would technically never return a `null` value so users not in any experiments is currently not captured properly via sentry.

https://sentry.prod.mozaws.net/operations/firefox-tv/issues/5201045/?query=is:unresolved

psymoon:
Closes #1969

### Pull Request checklist

### DIFF
--- a/app/src/main/java/org/mozilla/tv/firefox/experiments/ExperimentExceptions.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/experiments/ExperimentExceptions.kt
@@ -11,3 +11,11 @@ package org.mozilla.tv.firefox.experiments
  * considered an illegal experiment state
  */
 open class ExperimentIllegalStateException(message: String) : IllegalStateException(message)
+
+/**
+ * [NotInExperimentException] is for logging users that are currently not part of an
+ * Experiment
+ *
+ * @param name: Name of the experiment
+ */
+open class NotInExperimentException(name: String) : Exception(name)
--- a/app/src/main/java/org/mozilla/tv/firefox/experiments/ExperimentExceptions.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/experiments/ExperimentExceptions.kt
@@ -16,6 +16,11 @@ open class ExperimentIllegalStateException(message: String) : IllegalStateExcept
  * [NotInExperimentException] is for logging users that are currently not part of an
  * Experiment
  *
+ * Our current implementation for [FretboardProvider.loadExperiments] and
+ * [FretboardProvider.updateExperiments] are not synchronized. This is causing some of the
+ * new users to be not part of an experiment (even though they should be) in the initial
+ * launch (restarting the app would have you properly sync'd up) - #1978
+ *
  * @param name: Name of the experiment
  */
 open class NotInExperimentException(name: String) : Exception(name)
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/mozilla/tv/firefox/experiments/ExperimentsProvider.kt
```diff
@@ -51,13 +51,17 @@ class ExperimentsProvider(private val fretboard: Fretboard, private val context:
...
-            if (fretboard.isInExperiment(context, expDescriptor)) break
+            if (fretboard.isInExperiment(context, expDescriptor)) {
+                // Correct experiment variant is found
+                return expDescriptor
+            }
         }
 
-        return expDescriptor
+        // No matching experiment, so return null
+        return null
```
Baron-Severin: Nit: It looks like `expDescriptor` doesn't need to be mutable anymore.  Suggest updating to:

```kotlin
private fun checkBranchVariants(expConfig: ExperimentConfig): ExperimentDescriptor? {
        for (suffix in ExperimentSuffix.values()) {
            val expDescriptor = ExperimentDescriptor(expConfig.value + ":" + suffix.name)
            if (fretboard.isInExperiment(context, expDescriptor)) {
                // Correct experiment variant is found
                return expDescriptor
            }
        }
        return null
}

```

app/src/main/java/org/mozilla/tv/firefox/experiments/ExperimentExceptions.kt
```diff
@@ -11,3 +11,11 @@ package org.mozilla.tv.firefox.experiments
  * considered an illegal experiment state
  */
 open class ExperimentIllegalStateException(message: String) : IllegalStateException(message)
+
+/**
+ * [NotInExperimentException] is for logging users that are currently not part of an
+ * Experiment
+ *
+ * @param name: Name of the experiment
+ */
```
Baron-Severin: Nit: might be good to put a note here that this can happen on the first launch after an experiment has been created, since it uses persisted values.

psymoon: @Baron-Severin Hm, not sure if that is _the_ use case (I personally like to keep it as general as possible). We should be looking to optimize that in the future when we are looking at reducing launch time fwiw. What do you think?

Baron-Severin: I haven't spent a lot of time on Fretboard, so correct me if I'm wrong here.  I didn't think it waited for a response on startup?  My understanding was that it would just use the previously persisted value, and in the background make a request and store a new one.  So we would pretty much guarantee that we were using the Fretboard data pulled during our previous session, but also that it wouldn't have any significant impact on launch time.

Agreed that we shouldn't describe that as the only thing that can cause this error.  Other problems could definitely occur as well.  But it's a known issue that isn't immediately obvious, and anyone investigating this crash will probably look in this file first.  I think it would be good to mention it as a possibility.

psymoon: @Baron-Severin You're right, we are still blocking the UI thread with internal data storage load though. I still think there can be a reactive solution to this while finding a way to synchronize experiments load with fetch.

Having said that, I think it's a feasible possibility that an user is not part of any experiments :)

psymoon: I'll add a follow-up story and a reference comment :)

Baron-Severin: For sure.  I was thinking something like "This can happen when a user is not part of any experiment, or they have not yet fetched updated data from Fretboard."

PR #1977, opened 2019-03-19
Closes #1935: Add main process check in FirefoxApplication

### LINKED ISSUE #1935: Don't do stuff in `FirefoxApplication` in Gecko processes
Right now, `FirefoxApplication` runs code in every process in the app. Unfortunately, this includes processes that are private to Gecko, and can result in some very difficult to debug problems. Apps should not run any of their own code in processes owned by Gecko at all. It seems like we may need some help from a-c to make this easier on apps.

psymoon:
### Pull Request checklist


psymoon:
as @liuche mentioned, ac-47 should have this already - closing

PR #2004, opened 2019-03-30
Settings channel 🔄!!

liuche:
### Pull Request checklist


liuche:
@npark-mozilla @AaronMT I'm running into one problem here with the tests - in the Screenshots, specifically SettingsTest, I'm not sure how the `@ClassRule` localeTestRule is being used. This test doesn't actually run unless I comment out the `localeTestRule` companion object (now that I've converted it into kotlin). It seems like something stupid I'm doing so I thought I'd check :)

AaronMT:
I don't see any difference in their documentation, but yeah the rule is declared to handle automatic switching of locales.

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/NavigationOverlayFragment.kt
```diff
@@ -72,14 +71,13 @@ private const val COL_COUNT = 5
 private val uiHandler = Handler(Looper.getMainLooper())
 
 enum class NavigationEvent {
-    SETTINGS, BACK, FORWARD, RELOAD, LOAD_URL, LOAD_TILE, TURBO, PIN_ACTION, POCKET, DESKTOP_MODE, EXIT_FIREFOX;
+    BACK, FORWARD, RELOAD, LOAD_URL, LOAD_TILE, TURBO, PIN_ACTION, POCKET, DESKTOP_MODE, EXIT_FIREFOX;
```
Baron-Severin: Hm, when we hit 'send usage data' or 'clear all data' it opens a new screen.  The screen is very simple (it can only go back to the overlay), but is this something we should consider a navigation event?

This might be something that we add in when it becomes a problem, but I'm wondering if this makes our navigation code less consistent and therefore harder to follow.

app/src/main/java/org/mozilla/tv/firefox/ScreenController.kt
```diff
@@ -17,6 +17,7 @@ import org.mozilla.tv.firefox.ScreenControllerStateMachine.ActiveScreen
 import org.mozilla.tv.firefox.ScreenControllerStateMachine.Transition
```
Baron-Severin: > Remove unnecessary initial Settings fragment

👍

app/src/main/java/org/mozilla/tv/firefox/settings/SettingsFragment.kt
```diff
@@ -13,40 +13,70 @@ import android.view.ViewGroup
...
-            serviceLocator?.screenController?.showBrowserScreenForUrl(fragmentManager!!, URLs.URL_ABOUT)
-        }
-        view.privacyNoticeButton.setOnClickListener {
-            serviceLocator?.screenController?.showBrowserScreenForUrl(fragmentManager!!, URLs.PRIVACY_NOTICE_URL)
+        val settingsVM = FirefoxViewModelProviders.of(this@SettingsFragment).get(SettingsViewModel::class.java)
+        val type: SettingsType = SettingsType.valueOf(arguments!!.getString(KEY_SETTINGS_TYPE)!!)
+        return when (type) {
+            SettingsType.DATA_COLLECTION -> setupDataCollectionScreen(inflater, container, settingsVM)
+            SettingsType.CLEAR_COOKIES -> setupClearCookiesScreen(inflater, container, settingsVM)
+            else -> return container!!
```
Baron-Severin: This would be a good place to log an inconsistent state assertion to Sentry.

app/src/main/java/org/mozilla/tv/firefox/settings/SettingsFragment.kt
```diff
@@ -123,7 +139,12 @@ class SettingsFragment : Fragment() {
     companion object {
         const val FRAGMENT_TAG = "settings"
 
-        @JvmStatic
-        fun create() = SettingsFragment()
+        fun newInstance(type: SettingsType): SettingsFragment {
```
Baron-Severin: > This would be a good place to log an inconsistent state assertion to Sentry.

Same as above.  Although I wonder if we could enforce this with the type system, instead.  It seems like our `SettingsType` has two separate purposes: display on the overlay, and show some settings screen.  This approach is similar to one I took with Pocket, which in retrospect I think was a mistake.

What if we had an interface `SettingsScreen` or similar, that only some `SettingsType`s implemented?  Then we could enforce that a `SettingsScreen` instance is passed here?

app/src/main/res/color/preferences_switch_selector.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
+   - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
+<selector xmlns:android="http://schemas.android.com/apk/res/android">
+    <item android:color="@color/photonBlue40" android:state_checked="true" />
+    <item android:color="@color/photonGrey10" android:state_checked="false" />
+</selector>
```
Baron-Severin: Nit: New line at end of file.

app/src/main/res/values/styles.xml
```diff
@@ -78,4 +78,22 @@
         <item name="android:textColor">@color/photonGrey10_a80p</item>
     </style>
 
+    <style name="ConfirmButton">
+        <item name="android:background">@drawable/confirm_button_background</item>
+        <item name="android:textColor">#F9F9F9</item>
```
Baron-Severin: Do we have a Photon color for this?

app/src/main/res/layout/settings_tile.xml
```diff
@@ -3,32 +3,36 @@
...
-</androidx.cardview.widget.CardView>
\ No newline at end of file
+    <TextView
+        android:id="@+id/settings_title"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:gravity="center"
+        android:padding="12dp"
+        android:singleLine="true" />
+</LinearLayout>
```
Baron-Severin: Nit: extra line at end of file.

app/src/main/res/drawable/settings_item_selector.xml
```diff
@@ -2,24 +2,13 @@
...
+<shape
+    xmlns:android="http://schemas.android.com/apk/res/android"
+    android:shape="rectangle">
+    <size
+        android:height="40dp"
+        android:width="40dp" />
+    <solid android:color="@color/overlay_button_selector" />
+    <stroke android:color="@color/overlay_button_border"
+        android:width="4dp" />
+</shape>
```
Baron-Severin: Nit: line at end of file

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/NavigationOverlayFragment.kt
```diff
@@ -30,9 +30,7 @@ import androidx.recyclerview.widget.RecyclerView
 import io.reactivex.disposables.CompositeDisposable
 import io.reactivex.disposables.Disposable
 import io.reactivex.rxkotlin.addTo
-import kotlinx.android.synthetic.main.fragment_navigation_overlay_orig.navUrlInput
-import kotlinx.android.synthetic.main.fragment_navigation_overlay_orig.pocketVideoMegaTileView
-import kotlinx.android.synthetic.main.fragment_navigation_overlay_orig.tileContainer
+import kotlinx.android.synthetic.main.fragment_navigation_overlay_orig.*
```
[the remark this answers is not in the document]
liuche: I think "no star imports" - this was an oversight!

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/NavigationOverlayFragment.kt
liuche: Makes sense - it's a little awkward now because the SettingsType now gets converted into a NavigationEvent, and then back to a SettingsType for ScreenController, but putting it into the main flow of NavigationEvent means there aren't multiple places to look for navigation, which is good - and we can also clean it up later.

liuche:
Hmm, @npark-mozilla I'm running into problems here on the Data Collection Toggle test, which is related, but I fixed it locally and it runs green on my emulator.
I'll try it on a physical TV too, but I can't figure out what else is going wrong - maybe the test is not waiting for the next screen to show up, and I should add some kind of wait? Are the tests extra slow on bitbar?
https://console.firebase.google.com/project/moz-firefox-tv/testlab/histories/bh.3c34c7cf235a156d/matrices/6389914415755510748/executions/bs.3e8fda704ef057eb

npark-mozilla:
@liuche So basically, the screen changed, but the DOM didn't at this point. 
On 3:15 mark of the video, it shows the settings page: https://console.firebase.google.com/u/0/project/moz-firefox-tv/testlab/histories/bh.3c34c7cf235a156d/matrices/6389914415755510748/executions/bs.3e8fda704ef057eb/videos

But the test fails, because it fails to see the id:/toggle element, and spits out the current DOM here:
https://console.firebase.google.com/u/0/project/moz-firefox-tv/testlab/histories/bh.3c34c7cf235a156d/matrices/6389914415755510748/executions/bs.3e8fda704ef057eb/testcases/22
It appears the DOM isn't updated yet, because `InlineAutocompleteEditText`is still enabled in the DOM, which should be only in the main view.

The test assumes that after entering settings, it will move to the next line to execute when the page is open.  But often this is not the case. When a large UI change is involved, screen might have updated but the hierarchy hasn't been refreshed yet.  Most people think setting the idlingresource will fix this, but to my experience I haven't seen it completely eliminated. So what we used to do, (as well in iOS) is to wait for the element that should be visible when you open the settings view, using the UIAutomator syntax, kinda like I did here: https://github.com/mozilla-mobile/firefox-tv/commit/5ca3e3662920c15a779dd8be414e751b298917b1  

I think many of intermittent failures are of this kind, and we should deal with them as we go along.

liuche:
Gah @npark-mozilla it looks like it's still failing on that even with the test change you suggested, and the video isn't present for this for some reason. https://console.firebase.google.com/project/moz-firefox-tv/testlab/histories/bh.3c34c7cf235a156d/matrices/8004553002719159317

I'll retry this again, see if I can get a video and hopefully see if something else is going wrong...

npark-mozilla:
So here's the video, and the failing test starts at 3:24 mark : https://console.firebase.google.com/u/0/project/moz-firefox-tv/testlab/histories/bh.3c34c7cf235a156d/matrices/8004553002719159317/executions/bs.1ec191851c889c9/videos

When I ran this test locally, it also failed on simulator. It didn't click the "Send Usage Data" button reliably - perhaps it's because there was a popup text overlay? it seems like the button lost its focus quite fast.  So the usage data dialog never got opened, and failed. I guess we need to identify why when tapping down key. I suspect we press the return key too fast after tapping the down button

liuche:
Thanks for all your help! I **think** I have it fixed, and it passes on the Pixel emulator locally now too. And that's a helpful tip, that you can't watch the video from the individual test, but have to watch the whole thing -_-;;

npark-mozilla:
@liuche np! and it seems like it's doing well on bitbar after merge, so all good. You can also manually upload tests to bitbar directly beforehand if you wish.

PR #2013, opened 2019-04-04
Closes #1908: Add tooltips to nav bar

### LINKED ISSUE #1908: Helper Text for Nav Bar Controls

As a user, I'm not sure what the icons in the navigation bar mean

### What / Requirements

Add helper text when an icon in the navigation bar is focused

* Helper text is added for the navigation bar icons:
  * "Back"
  * "Forward"
  * "Refresh Page"
  * "Search or enter address"
  * "Turbo Mode ON" / "Turbo Mode OFF"
  * "Request Desktop Site ON" / "Request Desktop Site OFF"
  * "Settings"
  * "Quit Firefox"
* Text is localized
* Transitions...

dnarcese:
### Pull Request checklist

### DIFF
--- a/app/src/androidTest/java/org/mozilla/tv/firefox/ui/TooltipTest.kt
+++ b/app/src/androidTest/java/org/mozilla/tv/firefox/ui/TooltipTest.kt
@@ -1,0 +1,83 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.tv.firefox.ui
+
+import org.junit.Rule
+import org.junit.Test
+import org.mozilla.tv.firefox.helpers.MainActivityTestRule
+import org.mozilla.tv.firefox.ui.robots.navigationOverlay
+
+private const val TURBO_MODE_TEXT = "Turbo Mode"
+private const val BACK_TEXT = "Navigate back"
+private const val FORWARD_TEXT = "Navigate forward"
+private const val RELOAD_TEXT = "Reload website"
+private const val PIN_TEXT = "Pin to homescreen"
+private const val DESKTOP_MODE_TEXT = "Request desktop version of this site"
+private const val EXIT_TEXT = "Exit Firefox"
+private const val SETTINGS_TEXT = "Settings"
+
+/**
+ * A test for the nav bar tooltips including:
+ * - Each tooltip shows the correct string
+ * - The string dynamically updates the ON/OFF state (only applicable to Turbo Mode)
+ */
+class TooltipTest {
+
+    @get:Rule val activityTestRule = MainActivityTestRule()
+
+    /* ktlint-disable no-blank-line-before-rbrace */ // This imposes unreadable grouping.
+    @Test
+    fun tooltipTest() {
+        navigationOverlay {
+            // Url bar is focused on startup and next focus up will be turbo mode button
+            remoteUp()
+            assertTooltipText(TURBO_MODE_TEXT)
+            remoteCenter()
+            assertTooltipText(TURBO_MODE_TEXT)
+        }.openTileToBrowser(1) {
+        }.openOverlay {
+        }.openTileToBrowser(2) {
+        }.openOverlay {
+            // Focus back button
+            remoteUp()
+            assertTooltipText(BACK_TEXT)
+        }.goBack {
+        }.openOverlay {
+            // Focus forward button
+            remoteUp()
+            assertTooltipText(FORWARD_TEXT)
+            // Focus reload button
+            remoteRight()
+            assertTooltipText(RELOAD_TEXT)
+            // Focus pin button
+            remoteRight()
+            assertTooltipText(PIN_TEXT)
+        }.unpinSite {
+        }.openOverlay {
+            // This sequence focuses the pin button from the url bar
+            remoteUp()
+            remoteRight(2)
+            assertTooltipText(PIN_TEXT)
+            // Focus turbo mode button
+            remoteRight()
+            assertTooltipText(TURBO_MODE_TEXT)
+            // Focus desktop mode button
+            remoteRight()
+            assertTooltipText(DESKTOP_MODE_TEXT)
+        }.turnDesktopModeOn {
+        }.openOverlay {
+            // This sequence focuses the desktop mode button from the url bar
+            remoteUp()
+            remoteRight(4)
+            assertTooltipText(DESKTOP_MODE_TEXT)
+            // Focus exit button
+            remoteRight()
+            assertTooltipText(EXIT_TEXT)
+            // Focus settings button
+            remoteRight()
+            assertTooltipText(SETTINGS_TEXT)
+        }
+    }
+}
--- a/app/src/androidTest/java/org/mozilla/tv/firefox/ui/robots/NavigationOverlayRobot.kt
+++ b/app/src/androidTest/java/org/mozilla/tv/firefox/ui/robots/NavigationOverlayRobot.kt
@@ -12,6 +12,7 @@ import androidx.test.espresso.action.ViewActions.pressImeActionButton
 import androidx.test.espresso.action.ViewActions.typeText
 import androidx.test.espresso.assertion.ViewAssertions.matches
 import androidx.test.espresso.contrib.RecyclerViewActions
+import androidx.test.espresso.matcher.RootMatchers.isPlatformPopup
 import androidx.test.espresso.matcher.ViewMatchers.hasDescendant
 import androidx.test.espresso.matcher.ViewMatchers.withHint
 import androidx.test.espresso.matcher.ViewMatchers.withId
@@ -39,10 +40,14 @@ class NavigationOverlayRobot {
     private val device = UiDevice.getInstance(InstrumentationRegistry.getInstrumentation())
 
     fun goBack() = backButton().click()
-    fun remoteBack() = device.pressBack()
     fun goForward() = forwardButton().click()
     fun reload() = reloadButton().click()
     fun toggleTurbo() = turboButton().click()
+
+    fun remoteBack() = device.pressBack()
+    fun remoteUp() = device.pressDPadUp()
+    fun remoteRight(x: Int = 1) = repeat(x) { device.pressDPadRight() }
+    fun remoteCenter() = device.pressDPadCenter()
 
     // The implementation of this method is arbitrary. We could run this check
     // against any of its views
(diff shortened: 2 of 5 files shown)


dnarcese:
@brampitoyo Here a couple of screenshots of the tooltips, let me know if they look okay, or if you would like screenshots or a build to checkout all of them.

A couple things to note:
- I did not add a tooltip to the urlbar, since we already have the default hint there. It is also a little trickier to implement. But if you think it is still needed despite the hint, I can add it in.
- The desktop site button has a long accessibility string (image 2). If we want it to just say "Request desktop site on/off" we would need to add and localize a new string.

image

image

athomasmoz:
@dnarcese looks great! I suggest adding this in for release and we can make updates in follow-up tickets as needed. It doesn't deviate from the original design and the strings can be updated after if needed.

liuche:
@dnarcese change requested by L10N that we should [hard-code all the on/off labels](https://github.com/mozilla-l10n/firefoxtv-l10n/pull/21#issuecomment-481026441) (the whole button description, not just the string on/off) so that the adjective is sure to match the button name in other languages.

Could you open a separate PR for that? (since I already pre-landed the "on/off" strings, but those can't be localized properly)

app/src/androidTest/java/org/mozilla/tv/firefox/ui/robots/NavigationOverlayRobot.kt
```diff
@@ -97,6 +102,23 @@ class NavigationOverlayRobot {
     fun disableSessionIdling(activityTestRule: MainActivityTestRule) { activityTestRule.loadingIdlingResource.ignoreLoading = true }
     fun enableSessionIdling(activityTestRule: MainActivityTestRule) { activityTestRule.loadingIdlingResource.ignoreLoading = false }
 
+    // Focus the pin button using keypresses, starting from urlbar assuming back is disabled, forward is enabled
+    fun navigateToPinButton() {
```
npark-mozilla: Perhaps we should change the title of this method for clarity, since it assumes it's always starting from URLbar.
navigateToPinBtnWhenURLBarSelected() would give less room for confusion.  Either, we could create a method that taps left or right key x times, and comment in the main test script that x taps to right would give navigatedToPinBtn selected.

app/src/androidTest/java/org/mozilla/tv/firefox/ui/robots/NavigationOverlayRobot.kt
```diff
@@ -97,6 +102,23 @@ class NavigationOverlayRobot {
...
+        device.apply {
+            pressDPadUp()
+            repeat(2) { pressDPadRight() }
+        }
+    }
+    // Focus the desktop site button using keypresses, starting from urlbar assuming back is disabled, forward is enabled
+    fun navigateToDesktopSiteButton() {
```
npark-mozilla: same comment as above

app/src/androidTest/java/org/mozilla/tv/firefox/ui/TooltipTest.kt
```diff
@@ -0,0 +1,72 @@
...
+ */
+class TooltipTest {
+
+    @get:Rule val activityTestRule = MainActivityTestRule()
+
+    /* ktlint-disable no-blank-line-before-rbrace */ // This imposes unreadable grouping.
+    @Test
+    fun tooltipTest() {
+        navigationOverlay {
+            remoteUp()
```
npark-mozilla: Let's add a comment that by default URL bar is selected when overlay is opened, and pressing UP will select Turbo Mode. If this behavior changes in the future, then we would know why we did it this way

app/src/androidTest/java/org/mozilla/tv/firefox/ui/robots/NavigationOverlayRobot.kt
```diff
@@ -97,6 +102,23 @@ class NavigationOverlayRobot {
     fun disableSessionIdling(activityTestRule: MainActivityTestRule) { activityTestRule.loadingIdlingResource.ignoreLoading = true }
     fun enableSessionIdling(activityTestRule: MainActivityTestRule) { activityTestRule.loadingIdlingResource.ignoreLoading = false }
 
+    // Focus the pin button using keypresses, starting from urlbar assuming back is disabled, forward is enabled
+    fun navigateToPinButton() {
+        device.apply {
+            pressDPadUp()
+            repeat(2) { pressDPadRight() }
```
npark-mozilla: hmm, isn't this same as remoteRight()?  wouldn't it be better to change to remoteRight(int x) so it can repeat right button tap x times?  (Or as I said above, use pressDPadUp() and remoteRight(2) in the main test script with comments

dnarcese:
@brampitoyo Do you think it is okay if we omit the on/off state of the button in the tooltips? This will add a lot of new strings that will not be localized in time for this release (it also complicates implementation). Without the button state, we can re-use our content description strings :)

brampitoyo:
@dnarcese wrote:
> Do you think it is okay if we omit the on/off state of the button in the tooltips?

Yes. Totally okay. So then, the strings would be “Turbo mode” and “Request desktop site”.

brampitoyo:
@dnarcese Everything looks great other than a small typography tweak, as shown in this spec:

https://brampitoyo.github.io/people/bpitoyo/firefox-for-fire-tv/new-homescreen/i03/

Note the height of the tooltip box, as well as the 4dp top/bottom paddings and 8dp left-right paddings for the tip text inside it.

The final tooltip will also be set using the Ember Display font?

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/ToolbarUiController.kt
```diff
@@ -36,17 +45,40 @@ class ToolbarUiController(
...
         initToolbar(layout, viewLifecycleOwner, fragmentManager)
     }
 
+    private fun showTooltip(navBarButton: View) {
+        tooltip.contentView.tooltip.text = navBarButton.contentDescription
+        tooltip.isClippingEnabled = false
+
+        // Since the height and width are WRAP_CONTENT, we need to force the layout measure to happen before we can
+        // access the true dimensions
+        tooltip.contentView.measure(tooltip.contentView.width, tooltip.contentView.height)
```
liuche: This should probably use the WRAP_CONTENT constants, rather than fetching them from the tooltip.

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/ToolbarUiController.kt
```diff
@@ -36,17 +45,40 @@ class ToolbarUiController(
...
         setupUrlInput(layout)
         initToolbar(layout, viewLifecycleOwner, fragmentManager)
     }
 
+    private fun showTooltip(navBarButton: View) {
+        tooltip.contentView.tooltip.text = navBarButton.contentDescription
+        tooltip.isClippingEnabled = false
+
+        // Since the height and width are WRAP_CONTENT, we need to force the layout measure to happen before we can
+        // access the true dimensions
```
liuche: Glad you found a solution, and this makes sense why this wasn't working before.

I looked into this more, because using `measure()` directly is usually a red flag:
By calling `measure` here means it will get called twice - once here, and once during the normal layout pass that the View does (wherever this would actually get measured). Measuring is usually expensive (because it has to recurse through all the children) but in this case, we *know* that it's only a single TextView, so it's not *terrible* - if this view got more complicated though, we'd want to change this to do it correctly.

The correct way to do this would be to hook into the normal layout pass by setting an OnGlobalLayoutListener, which will get called during the normal layout pass, and then call `tooltip.update` with the new offset, because the height will be guaranteed to have been set at that time. This way, we only call `measure` once.

A few options here:
a) Do it the correct way, tradeoff is that this way is "okay" and that will add a little more dev time, but it makes our code base cleaner/correct (e.g., in the case someone new comes in, they wouldn't be able to copy this antipattern somewhere else)
b) Add a comment explaining why this is okay in this one case (single TextView), and optionally include an explanation of the correct way. Say this is *NOT* how it should be done elsewhere.

app/src/main/res/layout/tooltip.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
+   - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
+<RelativeLayout
```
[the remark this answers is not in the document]
liuche: Actually, can you just use a TextView here, or do you need the parent layout?

app/src/androidTest/java/org/mozilla/tv/firefox/ui/TooltipTest.kt
```diff
@@ -0,0 +1,72 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.tv.firefox.ui
+
+import org.junit.Rule
+import org.junit.Test
+import org.mozilla.tv.firefox.helpers.MainActivityTestRule
+import org.mozilla.tv.firefox.ui.robots.navigationOverlay
+
+private const val TURBO_MODE_TEXT = "Turbo Mode "
```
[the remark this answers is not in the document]
npark-mozilla: Actually that was my feedback - the standard method of writing UI test involves not getting the comparison values from the code. If we rely on the context.resources, the verification kinda becomes circular, i.e. if the resource file is incorrect, then the test will give false negative.  This also means that the test will run on EN locale only, but from my experience, that seems acceptable .

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/ToolbarUiController.kt
```diff
@@ -36,17 +45,50 @@ class ToolbarUiController(
...
+
+        // The measurement of the popup happens in onGlobalLayout. We need to update the position of the
+        // popup after this measurement has happened. Beforehand, the measuredWidth will be 0.
+        tooltipView.viewTreeObserver.addOnGlobalLayoutListener(
+            object : ViewTreeObserver.OnGlobalLayoutListener {
+                override fun onGlobalLayout() {
+                    tooltip.update(
+                        navBarButton,
+                        0 - (tooltip.contentView.measuredWidth - navBarButton.width) / 2,
+                        10, -1, -1
```
liuche: Nit: break these last 2 args into separate lines too, I think.

dnarcese: Ok, I wasn't sure about that. They looked so empty all on separate lines.

Issue #689, opened 2018-03-16
[1304] Update Settings


bbinto:
**Why/User Benefit**
- Enable users to access settings in less clicks
- Make it easier for users to understand settings and choices

**What**
- Settings is a channel on the homepage
- All individual settings are a pin on the homepage and open up to a detail view
- Detail view includes:
-- header area with a title and a back button
-- content area that supports:
--- text only (eg about page)
--- text + buttons (eg clear cookies example above)
--- text + toggles (eg send usage and data)
--- list + toggles (eg homescreen customization)
-- The back button uses the same icon as the Fire TV remote so that we build on top of what users already are familiar with.

**Acceptance Criteria (how do I know when I’m done?)**
- Probes on all toggles
- Probe to enter settings

**KPI/Data**
- How many of these people change their settings (toggle)
- How many of the default settings are (not) changed by the user

bbinto:
@aminalhazwani -- would you mind providing some ideas here so eng can estimate. (no mocks needed right now) -- we think it doesn't have to be a big change, just some clean-up

aminalhazwani:
@bbinto I am happy that you asked! The current settings page includes:

- Send usage data, which is a control/checkbox
- About, which opens a subview
- Privacy notice, which is a link to a subview/webview
- Clear all cookies and site data, which opens a modal

We could maintain the same layout and navigation, bringing in also some of learning from the Pocket feed, and invest more into the interaction and overall experience.

Every setting could be a little card that could grow in size and change the background if the user selects it. The same card could expand to a larger view if the users presses it. The expanded area could be used to suit texts, list controls (checkboxes/radios/switches), and all the components necessary to let users customize their browsing experience. The flexibility of this pattern would allow the the settings page to be easily extended in the future.

In the image below you can see the main settings view with the first card - send usage and stats - highlighted.  If selected it grows to a bigger area where you can read more and toggle on and off the switches.

<img width="512" alt="screen shot 2018-04-23 at 6 41 29 pm" src="https://user-images.githubusercontent.com/1721875/39140836-13c394e2-4726-11e8-8a6e-e9844f7e4269.png">

The other example - about firefox - can displays the same text that we currently feature in the app. As mentioned the same pattern could be extended to parental control, privacy notice, cookies and site data and more to come.

bbinto:
#1189 should/can be considered to be fixed with this update revamp

aminalhazwani:
[...]

PR #2028, opened 2019-04-09
Make telemetry checkbox consistent in VoiceView on or off.

eeejay:
This removes a lot of the special accessibility code.

When the actual checkbox is non-interactive (not focusable or
clickable), its state is uttered when the cursor is on an ancestor node.

### Pull Request checklist

- [ ] ~This PR includes thorough **tests** or an explanation of why it does not~
- [ ] ~I have considered adding **QA labels** on the associated issue (not this PR; `qa-ready` or `qa-denied`)~


eeejay:
@mcomella I think this was your original stuff, and I wasn't informed enough to suggest this solution. I also faintly recall a similar convo we had about a checkbox in Echo, and I wonder if this kind of solution could help.

liuche:
This is a great idea, @eeejay, I didn't realize we could do this 😅 
We're actually changing this settings code entirely though, and removing that fake checkbox, in PR #2004 / issue #689.

PR #2050, opened 2019-04-11
Closes #1165: Use Ember fonts

### LINKED ISSUE #1165: Use Ember fonts

Matches OS system fonts as we also do on other platforms

### What / Requirements

Set text and type to Amazon Ember

As above.

dnarcese:
### Pull Request checklist


dnarcese:
@brampitoyo Here are screenshots with the new font:
image
image
image

The only areas that do not have the font (because I haven't figured out how to yet/ are more complicated to change) are the context menu when removing a pinned tile, and the dialog when clearing data:
image
image

To not block this PR, I will create another issue to investigate changing those fonts as well!

app/src/main/java/org/mozilla/tv/firefox/utils/ViewUtils.java
```diff
@@ -114,7 +116,13 @@ public static void showCenteredBottomToast(Context context, int resId) {
     }
 
     private static void showToast(Context context, int resId, String toastLocation) {
-        Toast toast = Toast.makeText(context, resId, Toast.LENGTH_SHORT);
+        LayoutInflater inflater = (LayoutInflater) context.getSystemService(Context.LAYOUT_INFLATER_SERVICE);
```
psymoon: Inflating layouts are generally costly. Can we define a `WeakReference` to persist the toast layout (like a ViewHolder pattern)?

brampitoyo:
@dnarcese Looks amazing! Thanks heaps.

app/src/main/java/org/mozilla/tv/firefox/utils/ViewUtils.java
dnarcese: After trying this, the toast visibly translate horizontally when a new message is shown, instead of always being centred. This doesn't look the greatest when a new toast is shown before the previous one has disappeared. Going to keep the current implementation for this reason.

Issue #2031, opened 2019-04-09
Youtube TV stuck at loading screen


psymoon:
## Steps to reproduce
- open FFTV
- go to Youtube TV tile
- close the app by going home
- reopen the app
- go to Youtube TV tile again

### Expected behavior
- Youtube TV loads correctly

### Actual behavior
- Youtube TV stuck at loading screen (happens intermittently)

### Device information
* Fire TV device: ?
* Latest affected Firefox version: ?
* Earliest affected Firefox version: ?

Baron-Severin:
Looks like this is reproducible on emulator as well.

FFTV version: f48803168d87fa91115e946b925bc27d512af883
Emulator: Android TV, API 25

cpearce:
Is this when using GeckoView, or with the SystemView? If GeckoView, this could be issue #2040. You could verify that by pressing a button on the remote once YouTube has started loading, and then you should see playback start.

psymoon:
@cpearce this is for SystemWebView

Baron-Severin:
Some additional notes.

100% reproducible STR
- Using emulator
- Turn on "don't keep activities"
- Open FF
- Open YouTube Tile (will load)
- Press the home button (FF will be killed in background)
- Reopen FF
- Load YouTube (will be stuck on loading screen)

Notes:
- If you force kill the app when YouTube is failing to load, on the next launch YT will load as expected
- Refreshing YT does not fix the problem
- Navigating to a new site and then returning to YT does not fix the problem
- I haven't yet noticed anything weird in the WV debugger during YT loads, but I'm still investigating

Baron-Severin:
- Removing the WebViewCache doesn't fix this bug

Baron-Severin:
- When following the above STR, most websites load as expected
  - Some sites never finish loading (progress bar remains active)
  - Such sites include www.google.com, www.mail.google.com, dive.google.com, www.flickr.com

Baron-Severin:
I'm logging page ready status on this branch here.  Build and connect with Chrome remote device inspector to view logs.  https://github.com/mozilla-mobile/firefox-tv/pull/2043

When the app is working normally, pages go through ready state: interactive, ready state: complete, then emit an onLoad event.  When following the above STR, pages stick on ready state: interactive.  This happens on every website I've tested.

[...]

PR #2051, opened 2019-04-11
Issue #2031: Move SessionFeature initialization

psymoon:
### Pull Request checklist

### DIFF
--- a/app/src/main/java/org/mozilla/tv/firefox/webrender/WebRenderFragment.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/webrender/WebRenderFragment.kt
@@ -80,7 +80,6 @@ class WebRenderFragment : EngineViewLifecycleFragment(), Session.Observer {
     var cursor: CursorController? = null
         @UiThread get set // Set from the UI thread so serial access is required for simplicity.
 
-    var sessionFeature: SessionFeature? = null
     // If YouTubeBackHandler is instantiated without an EngineView, YouTube won't
     // work properly, so we !!
     private val youtubeBackHandler by lazy { YouTubeBackHandler(engineView!!, activity as MainActivity) }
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/mozilla/tv/firefox/webrender/WebRenderFragment.kt
```diff
@@ -155,13 +154,6 @@ class WebRenderFragment : EngineViewLifecycleFragment(), Session.Observer {
     }
 
     override fun onEngineViewCreated(engineView: EngineView): Disposable? {
```
Baron-Severin: Nit: since this isn't where the `EngineView` is being created anymore, could we rename it to `onEngineViewStarted`?

Baron-Severin: I guess that's not correct either.  What should this be called?

psymoon: That's a good question, why do we need this? We can have a follow-up story to clean up all of our start-up logic, thoughts? @Baron-Severin

Baron-Severin: It's related to the old YouTube focus caching issue.  It looks like we're starting up that caching when the `EngineView` is instantiated, but I don't think we need to be nearly so strict with it (it just needs to be in place before the user has a chance to do anything).  I'm fine putting this into a follow up story, this logic could easily just live in `onStart`.

Baron-Severin: https://github.com/mozilla-mobile/firefox-tv/issues/2053

PR #2058, opened 2019-04-12
Closes #1940: GreyScreen fix

### LINKED ISSUE #1940: Investigate root cause of webview grey screen on app re-open
## Steps to reproduce
Open Firefox, load Youtube
Click home button on remote to return to Amazon home
Reopen Firefox

### Expected behavior
Youtube page shows

### Actual behavior
Grey screen

### Device information
* Fire TV device: all
* Latest affected Firefox version: master ac9c612b5e4c44af98a25d6906f962668b0f9e6c
* Earliest affected Firefox version: #1559 or more reliably with 3.2.2 w/ AC 37

Some bisection work and attempted workarounds in #1865 - landed a hacky fix in #1865 to get this out on release, so that will need to be backed out in order to investigate this.

### Bisection + workarounds
For AC32-36.1, grey screen is not a problem. The upgrade **to 37** shows the grey screen, and this is where AC switches from using the the webview from the layout to the EngineView.

Some workarounds tried:
- (didn't work) Tried calling `engineView.resume` to trigger the webview to "load" from grey screen
- (didn't work) Tried scrolling the webview in WebRenderFragment resume with `engineView.scrollByClamped` which scrolls the webview
- (works, but scrolls) Sending a dpad-right event when the fragment resumes if it's on the Youtube site

I also tried bisecting across the individual AC commits, but for one of the commits there were a lot of significant changes that I didn't pursue and did the hacky workaround.

psymoon:
This PR fixes 2 outstanding grey screen bugs - updated STR available at https://github.com/mozilla-mobile/firefox-tv/issues/1940#issuecomment-482699025

### Pull Request checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Fixed
 - Fixed Youtube loading screen stuck issue (#2031)
+- Fixed bug that could cause websites to show up as blank, grey screens when resuming Firefox (#1940)
 
 ## [3.6-RO] 2019-03-26
 *Released to Fire TV 4K, staged roll-out*
--- a/app/src/main/java/org/mozilla/tv/firefox/ScreenController.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/ScreenController.kt
@@ -54,7 +54,6 @@ class ScreenController(private val sessionRepo: SessionRepo) {
             .add(R.id.container_web_render, renderFragment, WebRenderFragment.FRAGMENT_TAG)
             // We add NavigationOverlayFragment last so that it takes focus
             .add(R.id.container_navigation_overlay, NavigationOverlayFragment(), NavigationOverlayFragment.FRAGMENT_TAG)
-            .hide(renderFragment) // TODO note that this will need to be changed in order to display WebRenderFragment under a split overlay
             .hide(pocketFragment)
             .hide(settingsFragment)
             .commitNow()
@@ -230,7 +229,6 @@ class ScreenController(private val sessionRepo: SessionRepo) {
             Transition.SHOW_BROWSER -> {
                 _currentActiveScreen.onNext(ActiveScreen.WEB_RENDER)
                 fragmentManager.beginTransaction()
-                    .show(fragmentManager.webRenderFragment())
                     .hide(fragmentManager.navigationOverlayFragment())
                     .hide(fragmentManager.pocketFragment())
                     .hide(fragmentManager.settingsFragment())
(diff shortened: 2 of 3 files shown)


app/src/main/java/org/mozilla/tv/firefox/ScreenController.kt
```diff
@@ -145,6 +145,7 @@ class ScreenController(private val sessionRepo: SessionRepo) {
             // overlayFragment.navOverlayScrollView.updateOverlayForHomescreen(isOnHomeUrl(fragmentManager))
         } else {
             transaction.hide(overlayFragment)
+            transaction.show(fragmentManager.webRenderFragment())
```
Baron-Severin: We only hide `WebRenderFragment` one place, on init.  Most everywhere else we seem to assume that it's always shown.  

I think it might be better to stop hiding it in the first place, rather than calling show here.  That would mean removing line 57 (where it's hidden on init) and 233 (where it is shown on `Transition.SHOW_BROWSER`).

I've done some brief testing and this seems to work as expected, although a QA pass over this would be appropriate.

app/src/main/java/org/mozilla/tv/firefox/webrender/WebRenderFragment.kt
```diff
@@ -159,6 +159,13 @@ class WebRenderFragment : EngineViewLifecycleFragment(), Session.Observer {
                 // Cache focused DOM element just before WebView gains focus. See comment in
                 // FocusedDOMElementCacheInterface for details
                 engineView.focusedDOMElement.cache()
+
+                // EngineView focus may be lost after waking up from sleep & screen saver.
+                // Forcibly request focus onStart(), after DOMElement cache, IFF webRenderFragment
```
Baron-Severin: Is `IFF` a typo for `if`, or an acronym I'm missing?

app/src/main/java/org/mozilla/tv/firefox/webrender/WebRenderFragment.kt
```diff
@@ -159,6 +159,13 @@ class WebRenderFragment : EngineViewLifecycleFragment(), Session.Observer {
...
+                // is the current ActiveScreen
+                // TODO: move this when focus repo is in place (#1395)
+                // TODO: this method needs to be renamed (#2053); preliminary onStart() setup
```
Baron-Severin: Good comments, very clear.

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Fixed
 - Fixed Youtube loading screen stuck issue (#2031)
+- Fixed GreyScreen issue (#1940)
```
Baron-Severin: I don't know if users will understand what that means.  What about `Fixed bug that could cause websites to show up as blank, grey screens when resuming Firefox`

app/src/main/java/org/mozilla/tv/firefox/webrender/WebRenderFragment.kt
psymoon: if and only if

PR #2067, opened 2019-04-16
1912 cursor improvements

Baron-Severin:
### Pull Request checklist


app/src/main/java/org/mozilla/tv/firefox/webrender/cursor/CursorView.kt
```diff
@@ -5,89 +5,65 @@
...
+    override fun onDraw(canvas: Canvas?) {
+        super.onDraw(canvas)
+
+        cursorController?.mutatePosition(position)
+        x = position.x - (width / 2)
+        y = position.y - (height / 2)
+
+        canvas?.drawBitmap(bitmap, 0f, 0f, paint)
+
+        invalidate()
```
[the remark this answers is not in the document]
Baron-Severin: Yeah, just haven't gotten that far yet.

mcomella:
Another thing I thought of about major refactoring of the cursor (especially without tests): if the cursor stops working, users can't use the browser (exception: YT TV). It's a significant piece of the user experience that needs to work.

app/src/main/java/org/mozilla/tv/firefox/webrender/cursor/NewCursorController.kt
```diff
@@ -0,0 +1,77 @@
...
+class NewCursorController(private val activeScreen: Observable<ScreenControllerStateMachine.ActiveScreen>) {
+
+    var screenBounds: PointF? = null
+    var lastDirectionKeyPressed: KeyEvent by Delegates.observable(KeyEvent(KeyEvent.ACTION_UP, KeyEvent.KEYCODE_DPAD_UP)) { _, _, newValue ->
+        if (newValue.action == KeyEvent.ACTION_UP) {
+            accelerationDisposable?.dispose()
+            accelerationDisposable = null
+            currentAcceleration = 0f
+        } else if (newValue.action == KeyEvent.ACTION_DOWN && accelerationDisposable == null) {
+            accelerationDisposable = Observable.interval(10, TimeUnit.MILLISECONDS)
```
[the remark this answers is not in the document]
Baron-Severin: I wanted to be able to tweak translation frequency and acceleration independently from one another, at least while I was working out what their ratios should be.  This'll be rewritten before merging.

app/src/main/java/org/mozilla/tv/firefox/webrender/cursor/NewCursorController.kt
```diff
@@ -0,0 +1,77 @@
...
+    fun setup() {
+        activeScreen
+                .switchMap {
+                    if (it == ScreenControllerStateMachine.ActiveScreen.WEB_RENDER) {
+                        Observable.interval(1, TimeUnit.MILLISECONDS)
+                    } else {
+                        Observable.empty()
+                    }
+                }
+                .observeOn(Schedulers.computation())
```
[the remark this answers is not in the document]
Baron-Severin: Talked about this in person

app/src/main/java/org/mozilla/tv/firefox/MainActivity.kt
```diff
@@ -248,8 +248,16 @@ class MainActivity : LocaleAwareAppCompatActivity(), OnUrlEnteredListener, Media
         // code paths consistent between the two.
         if (event.keyCode == KeyEvent.KEYCODE_BACK ||
                 event.keyCode == KeyEvent.KEYCODE_DEL) return super.dispatchKeyEvent(event)
+
         serviceLocator.cursorEventRepo.pushKeyEvent(event)
 
+        serviceLocator.cursorController.maybeToMotionEvent(event)?.let {
```
mcomella: This is called redundantly inside `handleKeyEvent`. I feel like we should return two values or a sealed class from `handleKeyEvent` so that we keep a single entry point. While having two entry points can increase clarity in some ways, I dislike that it adds complexity to the call contract where two methods need to be called instead of one.

app/src/main/java/org/mozilla/tv/firefox/utils/ServiceLocator.kt
```diff
@@ -76,7 +78,9 @@ open class ServiceLocator(val app: Application) {
     val sessionManager get() = app.webRenderComponents.sessionManager
     val sessionUseCases get() = app.webRenderComponents.sessionUseCases
     val searchEngineManager by lazy { SearchEngineManagerFactory.create(app) }
-    val cursorEventRepo by lazy { CursorEventRepo(screenController) }
+    val cursorEventRepo by lazy { CursorEventRepo(cursorController, screenController) }
+    val cursorController by lazy { CursorModel(scrollBus, screenController.currentActiveScreen, frameworkRepo, sessionRepo) }
```
mcomella: nit: -> `cursorModel`

app/src/main/java/org/mozilla/tv/firefox/utils/Direction.kt
```diff
@@ -4,6 +4,15 @@
 
 package org.mozilla.tv.firefox.utils
 
+import android.view.KeyEvent
+
+val DIRECTION_KEY_CODES = listOf(
```
mcomella: nit: As `Direction.KEY_CODES` to avoid polluting the namespace?

app/src/main/java/org/mozilla/tv/firefox/settings/SettingsFragment.kt
```diff
@@ -55,7 +55,7 @@ class SettingsFragment : Fragment() {
         parentView: ViewGroup?,
         settingsViewModel: SettingsViewModel
     ): View {
-        val view = inflater.inflate(R.layout.settings_screen_switch, parentView, false)
+    val view = inflater.inflate(R.layout.settings_screen_switch, parentView, false)
```
mcomella: nit: unnecessary indentation change.

app/src/main/java/org/mozilla/tv/firefox/webrender/WebRenderFragment.kt
```diff
@@ -135,18 +125,10 @@ class WebRenderFragment : EngineViewLifecycleFragment(), Session.Observer {
...
-            viewModel = ViewModelProviders.of(this, viewModelFactory).get(CursorViewModel::class.java)
-        ).also { (controller, disposable) ->
-            this.cursor = controller
-            onCreateViewCompositeDisposable.add(disposable)
-            lifecycle.addObserver(controller)
-            context.serviceLocator.cursorEventRepo.setCursorController(controller)
+        layout.browserFragmentRoot.addOnLayoutChangeListener { _, _, _, right, bottom, _, _, _, _ ->
+            context.serviceLocator.cursorController.screenBounds = PointF(right.toFloat(), bottom.toFloat())
         }
+        context.serviceLocator.cursorEventRepo.setWebViewCouldScrollInDirection { layout.engineView.couldScrollInDirection(it) }
```
mcomella: nit: -> `canWebViewScrollInDirectionProvider` or similar: I find it confusing to distinguish between references that are values and functions that provide those values since they often get named the same thing.

Also (with old naming convention for time), probably not that important but:
```kotlin
context.serviceLocator.cursorEventRepo.setWebViewCouldScrollInDirection(layout.engineView::couldScrollInDirection)
```

Or even:
```kotlin
context.serviceLocator.cursorEventRepo.webViewCouldScrollInDirection = layout.engineView::couldScrollInDirection
```

app/src/main/java/org/mozilla/tv/firefox/MainActivity.kt
```diff
@@ -248,8 +248,16 @@ class MainActivity : LocaleAwareAppCompatActivity(), OnUrlEnteredListener, Media
         // code paths consistent between the two.
         if (event.keyCode == KeyEvent.KEYCODE_BACK ||
                 event.keyCode == KeyEvent.KEYCODE_DEL) return super.dispatchKeyEvent(event)
+
         serviceLocator.cursorEventRepo.pushKeyEvent(event)
```
mcomella: Is there any reason not to move this call inside `cursorModel.handleKeyEvent()` so that we can have a single entry point to cursor key code?

app/src/main/java/org/mozilla/tv/firefox/webrender/WebRenderFragment.kt
```diff
@@ -218,6 +204,9 @@ class WebRenderFragment : EngineViewLifecycleFragment(), Session.Observer {
 
         HintBinder.bindHintsToView(hintViewModel, hintBarContainer, animate = true)
                 .forEach { startStopCompositeDisposable.add(it) }
+
+        cursorView.setup(context!!.serviceLocator.cursorController)
```
mcomella: The `cursorView` is owned by the `WebRenderFragment`: why aren't the key events also handled in the `webRenderFragment`?

app/src/main/java/org/mozilla/tv/firefox/webrender/cursor/CursorModel.kt
```diff
@@ -0,0 +1,303 @@
...
+            oldPos.x = oldPos.x.coerceIn(0f, screenBounds.x)
+            oldPos.y = oldPos.y.coerceIn(0f, screenBounds.y)
+        }
+
+        fun calculateAndSendScrollEvent() {
+            // This should not live inside internalMutatePositionAndReturnVelocity.
+            // Move it if you can find a better solution
+            val approxFramesPassed = (timePassed / 16).toInt()
+            val shouldScroll = getScrollDistance(velocity, oldPos, approxFramesPassed)
+            shouldScroll?.let { scrollBus.scrollRequests.onNext(it) }
```
mcomella: I think this would be acceptable, and fit our app's architecture, if the `scrollBus` Observable lived inside `CursorModel`. Do you agree?

app/src/androidTest/java/org/mozilla/tv/firefox/ui/IWebViewExecuteJavascriptTest.kt
```diff
@@ -44,7 +45,7 @@ class IWebViewExecuteJavascriptTest {
 
         // Load the mock page.
         onView(withId(R.id.navUrlInput))
-                .perform(typeText(url), pressImeActionButton())
+                .perform(typeText(url), closeSoftKeyboard(), pressImeActionButton())
```
mcomella: Are these changes related to this PR? If not, can we take them out? If so, can you explain how it's related? This PR is challenging to read already.

app/src/main/java/org/mozilla/tv/firefox/MainActivity.kt
Baron-Severin: That was the original implementation before you asked me to change it to this one.  I'm fine with either, do you have a strong preference one way or the other?

app/src/main/java/org/mozilla/tv/firefox/webrender/cursor/CursorModel.kt
Baron-Severin: That's definitely an improvement.  I still don't like it, because it's a very similar conceptual space to what `CursorEventRepo` handles, but I spent some time trying to unify them and wasn't able to find any easy solution to that.

So I'll take your advice, good call.

app/src/androidTest/java/org/mozilla/tv/firefox/ui/IWebViewExecuteJavascriptTest.kt
Baron-Severin: These tests were failing on CI.  I think we've seen some periodic failures from them in the past, so rerunning might solve it, but fixing it seemed better.  I can pull them out and throw them into their own PR.

app/src/main/java/org/mozilla/tv/firefox/MainActivity.kt
mcomella: I don't remember asking for that or why I would – I would prefer a single entry point.

app/src/main/java/org/mozilla/tv/firefox/webrender/cursor/CursorModel.kt
mcomella: tbh, I'm still confused on the distinction between `CursorEventRepo` and `CursorModel`: maybe it's this same dissonance that makes it challenging to combine the `ScrollRepo` with `CursorEventRepo`?

Given what we learned from the original implementation (having an arbitrary architecture that is even slightly unclear is hard to maintain because eventually code will be added in the wrong place then it will be even more unclear), I wonder if there's value in merging CursorEventRepo directly into CursorModel: it'd make the responsibilities much clearer between the two remaining classes (Model & View). Once we better understand the problem space and the `CursorModel` grows too large, we can pull whatever functionality makes sense into a separate class.

I'm not sure this is a good idea, just thinking aloud...

app/src/main/java/org/mozilla/tv/firefox/MainActivity.kt
Baron-Severin: Hm, looks like this forces a cyclical dependency between `CursorModel` and `CursorEventRepo`.  I don't think this change is worth that, but let me know if you disagree.

app/src/main/java/org/mozilla/tv/firefox/MainActivity.kt
Baron-Severin: `handleKeyEvent` returned a sealed class that looked something like `HandleResponse(val wasConsumed: Boolean, motionEvent: MotionEvent?)`.  It wasn't great code.  I prefer to have a single entry point as well, but both implementations have drawbacks.

app/src/main/java/org/mozilla/tv/firefox/webrender/cursor/CursorView.kt
```diff
@@ -5,89 +5,103 @@
...
-        viewWeakReference.get()
-                ?.animate()
-                ?.setDuration(HIDE_ANIMATION_DURATION_MILLIS)
-                ?.alpha(0f)
-                ?.start()
+    override fun onDraw(canvas: Canvas?) {
+        super.onDraw(canvas)
+
+        val shouldInvalidate = cursorModel?.mutatePosition(position) ?: true
+        x = position.x - (width / 2)
```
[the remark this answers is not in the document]
Baron-Severin: We do need to set `View.x` and `View.y` at some point.  I suppose we could pass the View itself into the model, instead of a `PointF`.

app/src/main/java/org/mozilla/tv/firefox/webrender/cursor/CursorModel.kt
```diff
@@ -0,0 +1,303 @@
...
+            if (pos.y == 0f && directionKeysPressed.contains(Direction.UP)) {
+                scrollVelReturnVal.y = -percentMaxVel
+            } else if (pos.y == screenBounds.y && directionKeysPressed.contains(Direction.DOWN)) {
+                scrollVelReturnVal.y = percentMaxVel
+            }
+        }
+
+        val scrollX = Math.round(scrollVelReturnVal.x * MAX_SCROLL_VELOCITY * framesPassed)
+        val scrollY = Math.round(scrollVelReturnVal.y * MAX_SCROLL_VELOCITY * framesPassed)
+        return if (scrollX != 0 || scrollY != 0) scrollX to scrollY
```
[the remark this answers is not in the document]
Baron-Severin: We need to either allocate or calculate it twice, which has less of an impact?  We could also store them to a private var, so we can at least avoid deallocation and finding available memory space costs.

app/src/main/java/org/mozilla/tv/firefox/webrender/cursor/CursorModel.kt
Baron-Severin: That's an interesting observation, and I think it might be a good suggestion.  I think I need to mull it over before making that change, though.

app/src/main/java/org/mozilla/tv/firefox/webrender/cursor/CursorView.kt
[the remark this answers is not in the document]
mcomella: My solution would do that under "...":
```kotlin
val onDrawMutablePositionCache = PointF()

override fun onDraw(canvas: Canvas?) {
    onDrawMutablePositionCache.set(x, y)
    val shouldInvalidate = cursorModel?.mutatePosition(onDrawMutablePositionCache) ?: false
    x = onDrawMutablePositionCache.x
    y = onDrawableMutablePositionCache.y
    // ...
```

app/src/main/java/org/mozilla/tv/firefox/webrender/cursor/CursorModel.kt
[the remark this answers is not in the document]
mcomella: It's better to cache in a private var, like my `onDrawMutablePositionCache` example.

app/src/main/java/org/mozilla/tv/firefox/webrender/cursor/CursorModel.kt
mcomella: It'd also address the circular dependency between CursorEventRepo and CursorModel (if we were to move EventRepo inside), though there may be other solutions for that.

app/src/main/java/org/mozilla/tv/firefox/webrender/cursor/CursorView.kt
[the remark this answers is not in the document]
Baron-Severin: This could me being dense, so apologies if I'm missing the point here, but I'm not clear on what this solves.  As it is, `position` (which seems to be equivalent to `onDrawMutablePositionCache`) is our source of truth.  We mutate it and read it to see where we should draw.  `View.x` and `View.y` are only read in `onLayout` (and looking more closely, that can be removed entirely).  `position` is our source of truth, `x` and `y` are set only for their side effects.

Again, possible that I'm just missing your point.

app/src/main/java/org/mozilla/tv/firefox/webrender/cursor/CursorEventRepo.kt
```diff
@@ -44,11 +47,13 @@ class CursorEventRepo(screenController: ScreenController) {
         data class CursorMoved(val direction: Direction) : CursorEvent()
```
mcomella: `CursorEventRepo.webRenderDirectionEvents` is duplicating existing logic: it's filtering the events to make sure they're direction-events, filtering action down, mapping them to the appropriate direction, filtering if the WebRenderFragment is active, etc. However, we already have that logic in order to translate the cursor in response to a direction key.

Technically, scroll events happen in response to Cursor move events so I think it makes more sense to push scroll events in response to these cursor move events, i.e. around the call site of `mutatePosition`. This would also remove the duplication and, I think, clarify the code. In fact, I think that means we won't need the `CursorEventRepo` at all.

app/src/main/java/org/mozilla/tv/firefox/webrender/cursor/CursorView.kt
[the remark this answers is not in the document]
mcomella: They serve a similar purpose: both cache a position in order to avoid allocation. However, the difference is about scoping: my suggestion is scope to the `onDraw` function (which is sadly not enforced by the compiler) while `position` scopes to the entire class.

My concern is that by scoping to the entire class, we're forced to duplicate a real source of truth: `View.x/y`. While I'm not always against doing that, it creates the burden of having to accurately duplicate state: you have to find all the places those values can change and override them (e.g. your code in `onLayout` would be unnecessary in my solution). If you miss a spot, you end up with bugs (e.g. a very real problem in the current implementation is `CursorModel` or `WebRenderFragment` could call `setX/Y` and break the implementation).

app/src/main/java/org/mozilla/tv/firefox/webrender/cursor/CursorView.kt
```diff
@@ -5,89 +5,103 @@
...
-    private val hideHandler = CursorHideHandler(this)
+    private val position = PointF(x, y)
 
-    @UiThread
-    fun updatePosition(x: Float, y: Float) {
-        // translation* sets the top-left corner so we offset it in order to center the asset.
-        translationX = x - width / 2
-        translationY = y - height / 2
+    private var cursorModel: CursorModel? = null
+    private var width = 0f
```
mcomella: In hindsight, I think my idea to micro-optimize this is dumb: we're already calling a bunch functions in mutatePosition so it's not that helpful. Also, it forces us to duplicate an existing source of truth (which we may have gotten wrong). Also also, it makes the class less readable by adding a handful of unnecessary lines of code (var declarations, override onSizeChanged).

app/src/main/java/org/mozilla/tv/firefox/webrender/cursor/CursorView.kt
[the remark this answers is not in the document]
Baron-Severin: That's a good point, I'll make the change.

app/src/main/java/org/mozilla/tv/firefox/webrender/cursor/CursorEventRepo.kt
Baron-Severin: It hadn't occurred to me, but you're right.  That's definitely a a pretty glaring oversight on my part.  I'll merge the two.

app/src/main/java/org/mozilla/tv/firefox/webrender/cursor/CursorView.kt
Baron-Severin: It's hard to know where to draw the line on that kind of thing, but I agree that we can do without this one.

app/src/main/java/org/mozilla/tv/firefox/webrender/cursor/CursorView.kt
```diff
@@ -84,11 +87,12 @@ class CursorView(context: Context, attrs: AttributeSet) : AppCompatImageView(con
...
         super.onDraw(canvas)
-        onDrawMutablePositionCache.set(x, y)
+        viewOffset = width / 2
+        onDrawMutablePositionCache.set(x + viewOffset, y + viewOffset)
 
         val shouldInvalidate = cursorModel?.mutatePosition(onDrawMutablePositionCache) ?: false
-        x = onDrawMutablePositionCache.x - (width / 2)
-        y = onDrawMutablePositionCache.y - (height / 2)
+        x = onDrawMutablePositionCache.x - viewOffset
+        y = onDrawMutablePositionCache.y - viewOffset
```
mcomella: This is `height / 2`, not `width / 2`. Same above.

app/src/main/java/org/mozilla/tv/firefox/webrender/cursor/CursorView.kt
```diff
@@ -5,89 +5,95 @@
...
+        return compositeDisposable
     }
 
-    private fun setMaxVisibility() {
-        animate().cancel()
-        alpha = 1f
-    }
+    override fun onDraw(canvas: Canvas?) {
+        super.onDraw(canvas)
+        viewOffset = width / 2
```
mcomella: Why isn't this a local variable?

app/src/main/java/org/mozilla/tv/firefox/webrender/WebRenderFragment.kt
```diff
@@ -258,15 +243,15 @@ class WebRenderFragment : EngineViewLifecycleFragment(), Session.Observer {
...
+        val handledCursorEvent = serviceLocator?.cursorModel?.handleKeyEvent(event)
+
+        handledCursorEvent?.simulatedTouch?.let {
+            activity?.dispatchTouchEvent(it)
+            it.recycle()
+        }
+
         return handleSpecialKeyEvent(event) ||
-                (cursor?.keyDispatcher?.dispatchKeyEvent(event) ?: false)
+                handledCursorEvent?.wasConsumed == true
```
mcomella: My understanding of this key dispatch pattern is that if one function returns true, that function handled the key press and subsequent functions should not get to handle it. However, the way this is written, `handleSpecialKeyEvent` and `cursorModel.handleKeyEvent` could both handle a key event.

I'd like to keep this pattern. I think it's helpful because it allows the developer to consider a smaller scope: "if this module (i.e. function) handles this key event, I do not need to consider any others" rather than having to check every function for whether it handles a particular key (here and back in MainActivity).

Maybe a reasonable phrasing that minimally changes behavior is:
```kotlin
fun handleCursorKeyEvent(event: KeyEvent): Boolean {
    val handledCursorEvent = serviceLocator?.cursorModel?.handleKeyEvent(event)
    handledCursorEvent?.simulatedTouch?.use {
        activity?.dispatchTouchEvent(it)
    }
    return wasConsumed == true
}

return handleSpecialKeyEvent(event) ||
    handleCursorKeyEvent(event)

File: app/src/main/java/org/mozilla/tv/firefox/webrender/WebRenderFragment.kt [hunk 20]
```diff
@@ -258,15 +243,15 @@ class WebRenderFragment : EngineViewLifecycleFragment(), Session.Observer {
...
-        /**
-         * Key handling order:
-         * - Menu to control overlay
-         * - Youtube remap of BACK to ESC
-         * - Cursor
-         * - Return false, as unhandled
-         */
+        val handledCursorEvent = serviceLocator?.cursorModel?.handleKeyEvent(event)
+
+        handledCursorEvent?.simulatedTouch?.let {
```
mcomella: nit: `use { ... }`

app/src/main/java/org/mozilla/tv/firefox/webrender/cursor/CursorModel.kt
```diff
@@ -0,0 +1,352 @@
...
+import org.mozilla.tv.firefox.session.SessionRepo
+import org.mozilla.tv.firefox.utils.Direction
+import org.mozilla.tv.firefox.utils.DirectionHelper
+
+// Constants that can be tweaked to adjust cursor behavior
+private const val BASE_SPEED = 5f
+private const val MAX_VELOCITY = 25f
+private const val MS_TO_MAX_ACCELERATION = 200
+private const val MAX_SCROLL_VELOCITY = 13
+// Other constants
```
mcomella: nit: newline above?

app/src/main/java/org/mozilla/tv/firefox/webrender/cursor/CursorModel.kt
```diff
@@ -0,0 +1,352 @@
...
+import io.reactivex.subjects.PublishSubject
+import org.mozilla.tv.firefox.ScreenControllerStateMachine
+import org.mozilla.tv.firefox.ScreenControllerStateMachine.ActiveScreen.WEB_RENDER
+import org.mozilla.tv.firefox.ext.isUriYouTubeTV
+import org.mozilla.tv.firefox.framework.FrameworkRepo
+import org.mozilla.tv.firefox.session.SessionRepo
+import org.mozilla.tv.firefox.utils.Direction
+import org.mozilla.tv.firefox.utils.DirectionHelper
+
+// Constants that can be tweaked to adjust cursor behavior
```
mcomella: nit: Constants we expect to be tweaked...

Just clarifying that you can change other things but we don't expect you to.

app/src/main/res/layout/fragment_browser.xml
```diff
@@ -47,9 +47,6 @@
         android:id="@+id/cursorView"
         android:layout_width="@dimen/remote_cursor_size"
         android:layout_height="@dimen/remote_cursor_size"
-        android:layout_gravity="top|start"
-        android:src="@drawable/cursor_full"
```
mcomella: nit: add a comment mentioning the src is set in code.

app/src/main/res/layout/fragment_browser.xml
```diff
@@ -47,9 +47,6 @@
         android:id="@+id/cursorView"
         android:layout_width="@dimen/remote_cursor_size"
         android:layout_height="@dimen/remote_cursor_size"
-        android:layout_gravity="top|start"
```
mcomella: nit: For defensive coding, it might be helpful to leave this (e.g. if Android decides to layout us somewhere other than the origin and all of our draw calls are off).

app/src/main/res/layout/fragment_browser.xml
```diff
@@ -47,9 +47,6 @@
...
-        android:src="@drawable/cursor_full"
-        android:scaleType="centerInside"
```
mcomella: nit: I can't find out what the default scaleType is: we should leave this or set it in code.

app/src/main/java/org/mozilla/tv/firefox/webrender/cursor/CursorView.kt
```diff
@@ -5,89 +5,96 @@
...
-                setImageResource(R.drawable.cursor_full_active)
-            } else if (event.action == KeyEvent.ACTION_UP) {
-                setImageResource(R.drawable.cursor_full)
+    @CheckResult(suggest = "Dispose me, please. 🥰")
+    fun setup(cursorModel: CursorModel): Disposable {
+        this.cursorModel = cursorModel
+
+        val compositeDisposable = CompositeDisposable()
+
+        cursorModel.isCursorActive.subscribe { isCursorActive ->
```
mcomella: nit: it's unclear what "active" means in this context: is it enabled? Moving? etc. Maybe, `CursorModel.isCursorEnabledForAppState`. "AppState" helps separate this from states like, "The cursor is disabled because it's not moving."

app/src/main/java/org/mozilla/tv/firefox/webrender/cursor/CursorView.kt
```diff
@@ -5,89 +5,96 @@
...
+            isVisible = isCursorActive
+            if (!isCursorActive) {
+                animate().cancel()
             }
+        }.addTo(compositeDisposable)
+
+        Observables.combineLatest(cursorModel.isCursorMoving, cursorModel.isSelectPressed) {
```
mcomella: nit: not sure this is better but maybe the model should only expose `isSelectPressed` (which is used elsewhere) and this combined value, e.g. `isAnyCursorKeyPressed`, rather than adding complexity to the View here.

app/src/main/java/org/mozilla/tv/firefox/webrender/cursor/CursorView.kt
```diff
@@ -5,89 +5,96 @@
...
+ * Renders the cursor visible on top of the browser.
+ *
+ * Most behavior is controlled by [CursorModel]
+ *
+ * Any images set must be Bitmaps, not Drawables
  */
 class CursorView(context: Context, attrs: AttributeSet) : AppCompatImageView(context, attrs) {
 
-    private val hideHandler = CursorHideHandler(this)
+    private val onDrawMutablePositionCache = PointF(x, y)
```
mcomella: nit: explain that this is only intended to be used in `onDraw` to avoid allocation and that it works because it's only ever called from the UI thread on the same method (no concurrent access)

app/src/main/java/org/mozilla/tv/firefox/webrender/cursor/CursorView.kt
```diff
@@ -5,89 +5,96 @@
...
+        // (otherwise the PointF we set would be its top left corner)
+        val xOffset = width / 2
+        val yOffset = height / 2
 
-    private fun resetCountdown() {
-        hideHandler.removeMessages(HIDE_MESSAGE_ID)
-        hideHandler.sendEmptyMessageDelayed(HIDE_MESSAGE_ID, HIDE_AFTER_MILLIS)
-    }
-}
+        onDrawMutablePositionCache.set(x + xOffset, y + yOffset)
```
mcomella: nit: rather than duplicating our offset logic (to add it and remove it), maybe we should just pass width & height to the `mutatePosition` method and let it handle the complexities:
```kotlin
fun mutatePosition(topLeftPosition: PointF, width: Int, height: Int) { ...

File: app/src/main/java/org/mozilla/tv/firefox/webrender/cursor/CursorModel.kt [hunk 30]
```diff
@@ -0,0 +1,352 @@
...
+     */
+    private fun directionKeyPress(event: KeyEvent): Boolean {
+        if (!isCursorActive.blockingFirst()) {
+            return false
+        }
+        require(DirectionHelper.KEY_CODES.contains(event.keyCode)) {
+            "Invalid key event passed to CursorController#directionKeyPress: $event"
+        }
+
+        val direction = when (event.keyCode) {
```
mcomella: nit: we were going to pull this out, right? Maybe `KeyEvent.toDirection()`?

app/src/main/java/org/mozilla/tv/firefox/webrender/cursor/CursorModel.kt
```diff
@@ -0,0 +1,352 @@
...
+        return when (event.action) {
+            KeyEvent.ACTION_UP -> {
+                _isSelectPressed.onNext(false)
+                motionEvent
+            }
+            KeyEvent.ACTION_DOWN -> {
+                _isSelectPressed.onNext(true)
+                motionEvent
+            }
+            else -> null
```
mcomella: The motionEvent doesn't get recycled in this case: we should avoid creating it in this case.

app/src/main/java/org/mozilla/tv/firefox/webrender/cursor/CursorModel.kt
```diff
@@ -0,0 +1,352 @@
...
+    @CheckResult(suggest = "Recycle MotionEvent after use")
+    private fun maybeToMotionEvent(event: KeyEvent): MotionEvent? {
+        if (!isCursorActive.blockingFirst()) {
+            return null
+        }
+        if (event.keyCode != KeyEvent.KEYCODE_DPAD_CENTER && event.keyCode != KeyEvent.KEYCODE_ENTER) {
+            return null
+        }
+
+        val motionEvent = MotionEvent.obtain(event.downTime, event.eventTime, event.action,
```
mcomella: nit: we're not very explicit about why we're transforming a KeyEvent into a MotionEvent: we should comment on `touchSimulationEvent` (and reference it here).

app/src/main/java/org/mozilla/tv/firefox/webrender/cursor/CursorModel.kt
```diff
@@ -0,0 +1,352 @@
...
+     *
+     * Architecture:
+     * Having a View#onDraw call VM updates does not mesh well with our architecture, but we found
+     * no way to achieve desired performance with any other approach. Instead, we contained this
+     * architectural aberration to this class, and as much as possible it should be left here.
+     *
+     * STRONGLY avoid accessing this state outside of this class, and think very carefully about
+     * how you do if it is unavoidable.
+     *
+     * @return whether or not the view should continue to invalidate itself
```
mcomella: nit: comment that we mutate `oldPos`

app/src/main/java/org/mozilla/tv/firefox/webrender/cursor/CursorModel.kt
```diff
@@ -0,0 +1,352 @@
...
+     */
+    fun mutatePosition(oldPos: PointF): Boolean {
```
mcomella: nit: `oldPos` is not quite correct. Maybe `oldPosAndReturnedPos`.

It might even be clearer to have two: `fun mutatePosition(oldPos: PointF, returnedPosition: PointF)`

app/src/main/java/org/mozilla/tv/firefox/webrender/cursor/CursorModel.kt
```diff
@@ -0,0 +1,352 @@
...
+) {
+    // This is set early in the class lifecycle. Most methods short if it is not available
+    var screenBounds: PointF? = null
+    var webViewCouldScrollInDirectionProvider: (Direction) -> Boolean = { false }
+
+    private val directionKeysPressed = mutableSetOf<Direction>()
+    private var lastVelocity = 0f
+    private var lastUpdatedAtMS = NOT_RECENTLY_UPDATED
+    private var lastKnownCursorPos = PointF(0f, 0f)
+    private var cursorHasBeenCentered = false
```
mcomella: nit: `isInitialCursorPositionSet`? Without "initial position", it's unclear when the cursor gets centered and why.

app/src/main/java/org/mozilla/tv/firefox/webrender/cursor/CursorModel.kt
```diff
@@ -0,0 +1,352 @@
...
+     *
+     * STRONGLY avoid accessing this state outside of this class, and think very carefully about
+     * how you do if it is unavoidable.
+     *
+     * @return whether or not the view should continue to invalidate itself
+     */
+    fun mutatePosition(oldPos: PointF): Boolean {
+        lastKnownCursorPos = oldPos
+        when {
+            !cursorHasBeenCentered && screenBounds != null -> {
```
mcomella: At a high level, my assumption is that if the initial cursor position is set, we don't want to attempt to do any mutation in this class. This code doesn't necessarily do that: we only continue when the initial cursor position is set and screen bounds is not null. I think this should be:

```kotlin
if (!cursorHasBeenCentered) {
    if (screenBounds != null) { 
        // center
        cursorHasBeenCenter = true
    }
    
    return true
}

...

File: app/src/main/java/org/mozilla/tv/firefox/webrender/cursor/CursorModel.kt [hunk 37]
```diff
@@ -0,0 +1,352 @@
...
+        return when {
+            lastKnownCursorPos.y < EDGE_OF_SCREEN_MARGIN -> Direction.UP
+            lastKnownCursorPos.y > screenBounds.y - EDGE_OF_SCREEN_MARGIN -> Direction.DOWN
+            lastKnownCursorPos.x < EDGE_OF_SCREEN_MARGIN -> Direction.LEFT
+            lastKnownCursorPos.x > screenBounds.x - EDGE_OF_SCREEN_MARGIN -> Direction.RIGHT
+            else -> null
+        }
+    }
+
+    private fun resetCursorSpeed() {
```
mcomella: nit: resetCursorState

Since this also clears the direction keys, this does more than resetCursor position (but clearing keys is only used in one place: maybe we should move that?)

app/src/main/java/org/mozilla/tv/firefox/webrender/cursor/CursorModel.kt
```diff
@@ -0,0 +1,352 @@
...
+
+        scrollX = (scrollVelReturnVal.x * MAX_SCROLL_VELOCITY * framesPassed).toInt()
+        scrollY = (scrollVelReturnVal.y * MAX_SCROLL_VELOCITY * framesPassed).toInt()
+        return if (scrollX != 0 || scrollY != 0) {
+            scrollX to scrollY
+        } else null
+    }
+
+    @SuppressLint("CheckResult") // Does not need to be disposed as this survives for the duration of the app
+    private fun resetSpeedWhenCursorIsInactive() {
```
mcomella: nit: This is attaching an Observer but I find it confusing that it sounds like an immediate conditional state change listener (if cursor is inactive, reset the state right now). Maybe, `attachResetStateObserver`? Or even `attachStateObservers()` and we can add arbitrary subscribe code to it.

app/src/main/java/org/mozilla/tv/firefox/webrender/cursor/CursorModel.kt
```diff
@@ -0,0 +1,352 @@
...
+    /**
+     * Called by CursorView in order to drive cursor movement.
+     *
+     * Performance:
+     * Previous code updated cursor VM position every 16MS (1_000 MS / 60 frames == 16.66). This
+     * led to minor jitter because we were not tied exactly to the actual framerate. This new
+     * implementation is harder to understand, but is noticeably more performant.
+     *
+     * Architecture:
+     * Having a View#onDraw call VM updates does not mesh well with our architecture, but we found
```
mcomella: nit: it's not clear from the docs that our draw/update loop is coupled and driven by `View.onDraw` (the line I'm commenting on mentions `onDraw` but doesn't really provide context) and that we're performing the cursor position update here.

Similarly, the kdoc on this class does not convey the architecture between these two classes (View & Model, how they interact and that we should look out for this method). I think we can do a better job being explicit about our touch points (class kdoc, this method, View kdoc). Sorry this isn't entirely clear but it's the end of the day and it'd be hard to convey this through comments without writing it myself. We could probably pair program to write out what is important, if you'd like.

app/src/main/java/org/mozilla/tv/firefox/webrender/cursor/CursorModel.kt
```diff
@@ -0,0 +1,352 @@
...
+                cursorHasBeenCentered = true
+                return true
+            }
+            directionKeysPressed.isEmpty() -> {
+                resetCursorSpeed()
+                return false
+            }
+            else -> {
+                val currTime = System.currentTimeMillis()
+                // If the cursor pos has not recently been updated, reset to current time.
```
mcomella: nit: I think we should be more explicit:

> If we're starting a new update loop, we don't have a previous update value. Instead, we assume the average amount of time has passed (one frame).

app/src/main/java/org/mozilla/tv/firefox/webrender/cursor/CursorModel.kt
```diff
@@ -0,0 +1,352 @@
...
+                )
+
+                calculateAndSendScrollEvent(timePassedSinceLastMutation, lastVelocity, oldPos)
+                lastUpdatedAtMS = currTime
+                return true
+            }
+        }
+    }
+
+    private fun calculateAndSendScrollEvent(timePassed: Long, velocity: Float, oldPos: PointF) {
```
mcomella: nit: this is newPos now, right?

app/src/main/java/org/mozilla/tv/firefox/webrender/cursor/CursorModel.kt
```diff
@@ -0,0 +1,352 @@
...
+
+    private fun resetCursorSpeed() {
+        lastVelocity = BASE_SPEED
+        lastUpdatedAtMS = NOT_RECENTLY_UPDATED
+        directionKeysPressed.clear()
+    }
+
+    // This is taken from older code.  Crufty, but it works
+    private fun getScrollDistance(vel: Float, pos: PointF, framesPassed: Int): Pair<Int, Int>? {
+        val scrollVelReturnVal = PointF(0f, 0f)
```
mcomella: We shouldn't be allocating here either.

app/src/main/java/org/mozilla/tv/firefox/webrender/cursor/CursorModel.kt
```diff
@@ -0,0 +1,352 @@
...
+    /**
+     * @return returns a MotionEvent if the event is consumed, null otherwise
+     */
+    @SuppressLint("Recycle")
+    @CheckResult(suggest = "Recycle MotionEvent after use")
+    private fun maybeToMotionEvent(event: KeyEvent): MotionEvent? {
+        if (!isCursorActive.blockingFirst()) {
+            return null
+        }
+        if (event.keyCode != KeyEvent.KEYCODE_DPAD_CENTER && event.keyCode != KeyEvent.KEYCODE_ENTER) {
```
mcomella: This should be an assertion (like `handleDirectionKeyEvent`), right?

mcomella:
My reviews are really nitty so this is starting to grow in scope for both me and you (and this is excluding possible regressions). I'd like to land this sooner, rather than later, to try to catch any regressions so I wonder if we shouldn't block until all of the nits are complete. maybe we can do something like:
- Address all the functionality problems, land it as is, then fix the nits (though nit fixing can cause regressions too)
- Address functionality problems then pair program the nits (though we bike shed pretty hard together), then land
- Address functionality problems, then I'll fix the nits (it's faster for me to change what I dislike than to explain why I dislike it), you can review them, then land

fwiw, even if we were unable to fix the nits, I think I'd rather merge this than not because I think it the rearchitecture makes the code much easier to change for folks. The code is self-contained enough that I think it's okay if it's a little messy (provided the high level architecture description is there).

@Baron-Severin What do you think?

app/src/main/java/org/mozilla/tv/firefox/utils/Direction.kt
```diff
@@ -4,6 +4,17 @@
 
 package org.mozilla.tv.firefox.utils
 
+import android.view.KeyEvent
+
+object DirectionHelper {
```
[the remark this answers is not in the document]
Baron-Severin: There's already a `Direction` class in this file

app/src/main/res/layout/fragment_browser.xml
Baron-Severin: We programmatically set our starting position when the view is first measured, so this shouldn't matter.  I can add a comment saying that though.

app/src/main/java/org/mozilla/tv/firefox/webrender/cursor/CursorView.kt
Baron-Severin: It winds up being considerably more complex if we take that approach, because the model and the view have different conceptions of where the view actually is.  Touch events need to be at the center of the view, so this kind of one off calculation is always going to live somewhere.  It might be simpler to do when creating the touch event (at least we only calculate it once, instead of twice), but it's a marginal difference.

app/src/main/java/org/mozilla/tv/firefox/utils/Direction.kt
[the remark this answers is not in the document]
mcomella: Right: can we combine them so they share the same namespace?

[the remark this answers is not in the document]
Baron-Severin: Right, ya.  I'll do that.

app/src/main/java/org/mozilla/tv/firefox/webrender/cursor/CursorView.kt
mcomella: > (at least we only calculate it once, instead of twice), but it's a marginal difference.

I'm concerned about this duplication (rather than the existence of this calculation): if the code changes and we change the calculation in one place, will we remember to update in the other? Without knowing what it'd look like, that biases me towards doing it in one place, but maybe it's too complex.

Baron-Severin: Personally, I'm inclined to leave it as it is for that very reason.  The model considers the point to be 0,0, and the view offsets the display of that point.  I find that to be a reasonable division of concerns.

app/src/main/java/org/mozilla/tv/firefox/webrender/cursor/CursorModel.kt
```diff
@@ -0,0 +1,352 @@
...
+import org.mozilla.tv.firefox.utils.Direction
+import org.mozilla.tv.firefox.utils.DirectionHelper
+
+// Constants that can be tweaked to adjust cursor behavior
+private const val BASE_SPEED = 5f
+private const val MAX_VELOCITY = 25f
+private const val MS_TO_MAX_ACCELERATION = 200
+private const val MAX_SCROLL_VELOCITY = 13
+// Other constants
+private const val MAX_ACCELERATION = MAX_VELOCITY - BASE_SPEED
```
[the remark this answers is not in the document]
Baron-Severin: Perhaps it's poorly named.  Our acceleration ramps up as the user holds down the button, and this is as high as it can grow.  If that is true, this equation could not be changed without changing the meaning of either `MAX_VELOCITY` or `INITIAL_VELOCITY` (formerly `BASE_SPEED`).

app/src/main/java/org/mozilla/tv/firefox/webrender/cursor/CursorModel.kt
```diff
@@ -0,0 +1,352 @@
...
+     * content in that direction.
+     */
+    data class ScrolledToEdge(val edge: Direction) : CursorEvent() // TODO how often does this happen?  Telemetry would be good
+    data class CursorMoved(val direction: Direction) : CursorEvent()
+}
+
+/**
+ * Calculates cursor movements.
+ *
+ * Remote key events are pushed here from the rest of the app. Actual cursor position changes are
```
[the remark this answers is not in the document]
mcomella: We'll address this in the other comment comment: https://github.com/mozilla-mobile/firefox-tv/pull/2067#discussion_r277483227

mcomella:
Closing in favor of #2120.

PR #2077, opened 2019-04-16
Add wait until the progress indicator disappears when website is loading for UI Test

npark-mozilla:
### Pull Request checklist


npark-mozilla:
The latest UI test failure in bitbar was caused by returning to overlay too fast after back button was pressed. When the previous website was loaded, then the URL bar entry got updated and lost focus.

The attempt made here is to wait at least for the progress indicator to disappear when the site is being loaded.

npark-mozilla:
this fix seems to make the test pass in bitbar twice in a row, good sign.

npark-mozilla:
What's interesting is that when Forward button is pressed, the progressbar appears instantly, but when Back button is pressed from overlay, it takes about half second to show the progressbar, thus two checks for the presence/absence of progress indicator was needed.

PR #2079, opened 2019-04-17
Closes #2072 - Don't emit screen changes unless distinct.

### LINKED ISSUE #2072: Youtube loses control after menu bar
## Steps to reproduce
- Open youtube, play video
- Play with controls, make it visible
- Press menu to go to overlay
- Press menu again to return to video
- Access youtube video control, or exit the video

### Expected behavior
- Control is visible, and video can be stopped with back button

### Actual behavior
- App does not respond to inputs, other than menu overlay

### Device information
* Fire TV device: Gen 2 stick

liuche:
Didn't add a changelog because this fixes a new regression created in 3.7, didn't add a test but I added another item to #1420 .

### Pull Request checklist

### DIFF
--- a/app/src/main/java/org/mozilla/tv/firefox/ScreenController.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/ScreenController.kt
@@ -35,7 +35,9 @@ class ScreenController(private val sessionRepo: SessionRepo) {
     /**
      * Observers will be notified just before the fragment transaction is committed
      */
-    val currentActiveScreen: Observable<ActiveScreen> = _currentActiveScreen.hide()
+    val currentActiveScreen: Observable<ActiveScreen> = _currentActiveScreen
+            .distinctUntilChanged()
+            .hide()
 
     /**
      * To keep things simple, we add all the fragments at start instead of creating them when needed


app/src/main/java/org/mozilla/tv/firefox/webrender/WebRenderFragment.kt
```diff
@@ -154,7 +154,7 @@ class WebRenderFragment : EngineViewLifecycleFragment(), Session.Observer {
     }
 
     override fun onEngineViewCreated(engineView: EngineView): Disposable? {
-        return serviceLocator?.screenController?.currentActiveScreen?.subscribe {
+        return serviceLocator?.screenController?.currentActiveScreen?.distinctUntilChanged()?.subscribe {
```
Baron-Severin: This is probably going to be a gotcha in other parts of the codebase, too.  I think we can reasonably assume that this observable should always use `distinctUntilChanged`, could you add this operator in `ScreenController` line 37 instead?

Baron-Severin: By convention, individual Rx operators are generally put on separate lines.  So this would be 

```kotlin
serviceLocator?.screenController?.currentActiveScreen
    ?.distinctUntilChanged()
    ?.subscribe {
```
or
```kotlin
serviceLocator?.screenController?.currentActiveScreen?.distinctUntilChanged()
    ?.subscribe {
```

liuche: Yes! Makes sense.

I didn't look into why this was emitting multiple times, which was also strange.

Issue #1912, opened 2019-03-06
Refactor Cursor to improve maintainability; fix dropped frames


devinreams:
Related to https://github.com/mozilla-mobile/firefox-tv/issues/1890

### ACCEPTANCE CRITERIA

* Perceived performance of cursor on GV+FFTV is equal or better than FFTV+AWV
* Perceived performance of cursor is equal or better than what's live today

---

_edit: We have observed the cursor has three problems:_
- The code is hard to maintain
- The cursor jitters when scrolling the page on GeckoView builds
- Every 1/8 of the screen length or so, the cursor drops a frame while moving

We will address the 1st and 3rd; bug #2102 is filed for the 2nd item.

mcomella:
See #472 for implemented performance improvements and #685 for follow-up thoughts. Feel free to pick my brain for ideas!

athomasmoz:
Blocks #1819

athomasmoz:
@psymoon A few questions
* What about the scrolling is blocking #1819 ?
* Do the recent cursor changes from #1907 resolve this?

psymoon:
@athomasmoz 
>     * What about the scrolling is blocking #1819 ?
Depends on the target audience. If end-user, yes this is required. If internal, merging https://github.com/mozilla-mobile/firefox-tv/pull/1971 would suffice IMO
>     * Do the recent cursor changes from #1907 resolve this?
No

athomasmoz:
So is this story just a general suggestion to improve how smooth the cursor is? And if so, how does that relate to GV and #1819 ?

mcomella:
> _edit: We have observed the cursor has three problems:_
> 
>     * The code is hard to maintain
> 
>     * The cursor jitters when scrolling the page on GeckoView builds
> 
>     * Every 1/8 of the screen length or so, the cursor drops a frame while moving

We're looking into all three of these (and timeboxing the refactor).

mcomella:
QA: we almost entirely rewrote the cursor code. Please ensure:
- It's working properly on each web page
- It correctly enables itself and doesn't steal key events when transitioning between screens (e.g. opening and closing the overlay, going to YouTube TV which doesn't support cursor, etc.)
- The cursor still feels good to use, is very functional
- The cursor has minimum dropped frames

mcomella:
I filed https://github.com/mozilla-mobile/firefox-tv/issues/2130 to restore the tests that were deleted during this refactor.

[...]

PR #2084, opened 2019-04-17
Closes #2083: removes validate motion event

### LINKED ISSUE #2083: Remove CursorViewModel#validateMotionEvent
After close inspection, it looks like we can remove this entire branch of code if we stop altering the event times of our touch events at all.

Baron-Severin:
### Pull Request checklist

### DIFF
--- a/app/src/main/java/org/mozilla/tv/firefox/ext/KeyEvent.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/ext/KeyEvent.kt
@@ -1,0 +1,12 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.tv.firefox.ext
+
+import android.graphics.PointF
+import android.view.KeyEvent
+import android.view.MotionEvent
+
+fun KeyEvent.toMotionEvent(pos: PointF): MotionEvent =
+    MotionEvent.obtain(this.downTime, this.eventTime, this.action, pos.x, pos.y, 0)
--- a/app/src/main/java/org/mozilla/tv/firefox/webrender/cursor/CursorController.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/webrender/cursor/CursorController.kt
@@ -62,7 +62,7 @@ class CursorController private constructor(
                 else -> Unit
             }
         }, onSelectKey = { event ->
-            viewModel.onSelectKeyEvent(event.action, legacyViewModel.pos)
+            viewModel.onSelectKeyEvent(event, legacyViewModel.pos)
             view.updateCursorPressedState(event)
         })
 
--- a/app/src/main/java/org/mozilla/tv/firefox/webrender/cursor/CursorController.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/webrender/cursor/CursorController.kt
@@ -10,12 +10,12 @@ import android.view.View
 import androidx.annotation.CheckResult
 import androidx.lifecycle.Lifecycle
 import androidx.lifecycle.LifecycleObserver
-import androidx.lifecycle.Observer
 import androidx.lifecycle.OnLifecycleEvent
 import io.reactivex.disposables.Disposable
 import kotlinx.coroutines.Job
 import org.mozilla.tv.firefox.ext.couldScrollInDirection
 import org.mozilla.tv.firefox.ext.scrollByClamped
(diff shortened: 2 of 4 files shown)


app/src/main/java/org/mozilla/tv/firefox/webrender/cursor/CursorViewModel.kt
```diff
@@ -63,63 +49,8 @@ class CursorViewModel(
...
-            else -> {
-                // Fallback Option - See [MotionEvent#Consistency Guarantees]
-                prevDownMotionEvent?.recycle()
-                prevDownMotionEvent = null
-            }
-        }.forceExhaustive
-
-        return null
+    fun onSelectKeyEvent(keyEvent: KeyEvent, pos: PointF) {
+        val event = MotionEvent.obtain(keyEvent.downTime, keyEvent.eventTime, keyEvent.action, pos.x, pos.y, 0)
```
mcomella: nit: I do kind of prefer `KeyEvent.toMotionEvent()` because it makes it declaratively clear that this is a pure conversion as opposed to having to read through each attribute.

app/src/test/java/org/mozilla/tv/firefox/webrender/cursor/CursorViewModelTest.kt
```diff
@@ -163,73 +157,6 @@ class CursorViewModelTest {
         isEnabledTestObs.assertValues(*Array(emittedEventCount) { true })
     }
 
-    @Test
-    fun `GIVEN intended Click event THEN ACTION_UP MotionEvent returns same event_time as ACTION_DOWN`() {
```
mcomella: As we discussed, I would prefer to have a test but I'm currently in the, "why not test everything?" mindset and recognize that it's impractical to test everything. Given how simple it is and how adding a test will only make this code harder to intentionally change, I think it's reasonable to not write a test for this code.

app/src/main/java/org/mozilla/tv/firefox/ext/KeyEvent.kt
```diff
@@ -0,0 +1,12 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.tv.firefox.ext
+
+import android.graphics.PointF
+import android.view.KeyEvent
+import android.view.MotionEvent
+
+fun KeyEvent.toMotionEvent(pos: PointF): MotionEvent =
```
mcomella: nit: Add `@CheckResult` for recycle.

app/src/main/java/org/mozilla/tv/firefox/webrender/cursor/CursorViewModel.kt
```diff
@@ -63,63 +50,7 @@ class CursorViewModel(
...
-            else -> {
-                // Fallback Option - See [MotionEvent#Consistency Guarantees]
-                prevDownMotionEvent?.recycle()
-                prevDownMotionEvent = null
-            }
-        }.forceExhaustive
-
-        return null
+    fun onSelectKeyEvent(keyEvent: KeyEvent, pos: PointF) {
+        _touchSimulationLiveData.value = Consumable.from(keyEvent.toMotionEvent(pos))
```
mcomella: We need to recycle this MotionEvent (should we have a test for that?).

mcomella: We can greatly simplify the code, making the recycling easier, by handling this directly in `onSelectKey` rather than unnecessarily passing it around through the ViewModel.

app/src/main/java/org/mozilla/tv/firefox/webrender/cursor/CursorController.kt
```diff
@@ -62,7 +62,9 @@ class CursorController private constructor(
                 else -> Unit
             }
         }, onSelectKey = { event ->
-            viewModel.onSelectKeyEvent(event.action, legacyViewModel.pos)
+            val motionEvent = event.toMotionEvent(legacyViewModel.pos)
```
mcomella: nit: there should be a `MotionEvent.use {}` annotation function

Baron-Severin: Neat.

Baron-Severin: Actually, this class is deleted in #1912 and I'd rather not rebase over this change, so I'm going to leave it as is.  That's good to know though.

Issue #2089, opened 2019-04-18
Convert PinnedTileRepo to Rx


psymoon:
Required for https://github.com/mozilla-mobile/firefox-tv/issues/1395

It's one of the few core remaining places where LiveData runs the show. We are incrementally migrating everything to Rx.

PR #2090, opened 2019-04-18
Issue #2089: Convert PinnedTileRepo to use Rx

psymoon:
### Pull Request checklist

### DIFF
--- a/app/src/main/java/org/mozilla/tv/firefox/pinnedtile/PinnedTileRepo.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/pinnedtile/PinnedTileRepo.kt
@@ -5,7 +5,6 @@
 package org.mozilla.tv.firefox.pinnedtile
 
 import android.app.Application
-import androidx.lifecycle.LiveData
 import android.content.Context
 import android.content.SharedPreferences
 import android.graphics.Bitmap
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/mozilla/tv/firefox/pinnedtile/PinnedTileRepo.kt
```diff
@@ -75,7 +80,7 @@ class PinnedTileRepo(private val applicationContext: Application) {
     @UiThread
     fun removePinnedTile(url: String): String? {
         val tileToRemove = _pinnedTiles.value?.remove(url) ?: return null
-        _pinnedTiles.value = _pinnedTiles.value
+        _pinnedTiles.onNext(_pinnedTiles.value!!)
```
Baron-Severin: I guess this was a problem with LD too, and just wasn't caught by the compiler, but I'm not super happy about `!!`'ing this.  `BehaviorSubject#getValue` is nullable because it can be accessed before any values have been pushed.  We're pushing values on init, so this won't crash at the moment, but I'm still wary of it.

This could be refactored to follow a more reactive approach, but I don't think it's worth the effort right now.  So no change needed here, but bear in mind this kind of danger in the future.

PR #2100, opened 2019-04-19
No-issue: fixes (hopefully!) a periodic Espresso failure

### LINKED ISSUE #1912: Refactor Cursor to improve maintainability; fix dropped frames

Baron-Severin:
Solution taken from https://stackoverflow.com/a/53430379/9307461.  This suggests that the soft keyboard was occasionally interfering with some of our actions, causing tests to occasionally fail.

### Pull Request checklist

### DIFF
--- a/app/src/androidTest/java/org/mozilla/tv/firefox/ui/IWebViewExecuteJavascriptTest.kt
+++ b/app/src/androidTest/java/org/mozilla/tv/firefox/ui/IWebViewExecuteJavascriptTest.kt
@@ -5,6 +5,7 @@
 package org.mozilla.tv.firefox.ui
 
 import androidx.test.espresso.Espresso.onView
+import androidx.test.espresso.action.ViewActions.closeSoftKeyboard
 import androidx.test.espresso.action.ViewActions.pressImeActionButton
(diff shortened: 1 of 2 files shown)


Baron-Severin:
> Seems reasonable, since if it's already closed this is a no-op. Did you see the keyboard open in the videos though?

I didn't, so this fix is pretty speculative.  But it's also cheap and no-risk, so I think it's worth trying.  With all the independent weirdness around Espresso and soft keyboards, I would not be at all surprised if there were bad interactions between the two, even when it wasn't visible to the user.

Issue #2118, opened 2019-04-26
Hint bar animation is choppy over web content


mcomella:
## Steps to reproduce
- Load a page
- Scroll the hint bar in and out of view (scroll to top of page, move cursor down, repeat)

### Expected behavior
Hint bar is smooth

### Actual behavior
Hint bar is choppy and causes dropped frames in other parts of the app (e.g. cursor movement)

### Device information
* Fire TV device: Stick
* Latest affected Firefox version: master
* Earliest affected Firefox version: ?

After a quick analysis, it looks like if we disable the alpha part of the animation, it performs as expected.

PR #2138, opened 2019-04-30
Closes #2133: fixes hint bar showing when A/B is disabled

### LINKED ISSUE #2133: Empty hint bar appears on bottom of screen
## Steps to reproduce
- Open debug build
- Loda a page

### Expected behavior
No hint bar or hint bar with content

### Actual behavior
Hint bar appears but is empty

### Device information
* Fire TV device: Cube
* Latest affected Firefox version: master

~I think this is  a regression from https://github.com/mozilla-mobile/firefox-tv/issues/2029 (which just landed).~
EDIT: regression from #2118

Baron-Severin:
Root cause:
We previously enabled hint bar visibility on any isDisplayed change, then animated it to 0f opacity if !isDisplayed.  This opacity animation was removed in #2118, however always setting visibility to true was left in.  This led to an empty hint bar being always visible.

### Pull Request checklist

Bug never reached users

### DIFF
--- a/app/src/main/java/org/mozilla/tv/firefox/hint/HintBinder.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/hint/HintBinder.kt
@@ -37,7 +37,6 @@ object HintBinder {
                                 .start()
                     }
                     .subscribe { shouldDisplay ->
-                        hintContainer.isVisible = true
                         val translationY = when (shouldDisplay) {
                             true -> 0f
                             false -> hintContainer.height.toFloat()


mcomella:
#2133

app/src/main/java/org/mozilla/tv/firefox/hint/HintBinder.kt
```diff
@@ -37,7 +37,6 @@ object HintBinder {
                                 .start()
                     }
                     .subscribe { shouldDisplay ->
-                        hintContainer.isVisible = true
```
mcomella: This isn't quite right: in the `animate` case, after we dispose once, the view will never be visible again because e set `isVisible` to false but this is the only line that sets `isVisible` to true.

mcomella:
The Netflix architecture has the fragment talk to interchangable components, each of which contains its own Presenter, View, and layout XML. Model events are sent to the presenter; user input events are sent back to the Fragment (through the presenter). The fragment purely functions on lifecycle (I have screenshots of the slides).

Let's see what we can learn from that. 🤔

mcomella:
Closing in favor of a new PR (Simon and Daniela are working on this instead, I believe).

PR #2140, opened 2019-04-30
2110 pinned tile vm to rx

Baron-Severin:
### Pull Request checklist

Part of a larger change that will have a changelog entry

### DIFF
--- a/app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/NavigationOverlayFragment.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/NavigationOverlayFragment.kt
@@ -23,7 +23,6 @@ import androidx.core.content.ContextCompat
 import androidx.core.view.updateLayoutParams
 import androidx.core.widget.NestedScrollView
 import androidx.fragment.app.Fragment
-import androidx.lifecycle.Observer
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/NavigationOverlayFragment.kt
```diff
@@ -149,7 +148,7 @@ class NavigationOverlayFragment : Fragment() {
     private lateinit var pocketViewModel: PocketViewModel
     private lateinit var hintViewModel: HintViewModel
 
-    private lateinit var tileAdapter: PinnedTileAdapter
+    private var tileAdapter: PinnedTileAdapter? = null
```
psymoon: Why did you decide to make this nullable?

Baron-Severin: I wanted to move the subscription into `onStart`, since that's where we're setting up all of our existing subscriptions.  But the adapter winds up being used before that gets called, so I needed null checks.  The alternative was to add a new `CompositeDisposable`, which seemed like too much complexity.  I tested this pretty thoroughly and didn't encounter any problems.

PR #2157, opened 2019-05-02
Closes #2156: Use geckoview.BuildConfig for gecko flavor specific version string generation

### LINKED ISSUE #2156: Expose GeckoVersion from org.mozilla.geckoview.BuildConfig
We are currently maintaining our own GeckoVersion through [app/build.config](https://github.com/mozilla-mobile/firefox-tv/blob/master/app/build.gradle#L269). This causes confusion for anyone who's using a custom gecko build (local) with FFTV since the versioning does not come from the geckoview directly.

psymoon:
### Pull Request checklist

### DIFF
--- a/app/src/gecko/java/org/mozilla/tv/firefox/utils/BuildFlavor.kt
+++ b/app/src/gecko/java/org/mozilla/tv/firefox/utils/BuildFlavor.kt
@@ -18,7 +18,7 @@ class BuildFlavor {
             return "DEBUG / " +
                     "FLAVOR: ${BuildConfig.FLAVOR} / " +
                     "VERSION: ${BuildConfig.VERSION_NAME} / " +
-                    "GECKO: ${BuildConfig.GECKOVIEW_VERSION}"
+                    "GECKO: " + generateGeckoVersionStr()
         }
 
         return null
@@ -28,7 +28,7 @@ class BuildFlavor {
      * SystemWebView requires context to acquire version
      */
     fun getEngineVersion(@Suppress("UNUSED_PARAMETER") context: Context): String {
-        return BuildConfig.GECKOVIEW_VERSION
+        return generateGeckoVersionStr()
     }
 
     /**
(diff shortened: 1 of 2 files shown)


app/src/gecko/java/org/mozilla/tv/firefox/utils/BuildFlavor.kt
```diff
@@ -42,4 +42,8 @@ class BuildFlavor {
     fun getInterceptionResponseContent(localizedContent: String): RequestInterceptor.InterceptionResponse.Content {
         return RequestInterceptor.InterceptionResponse.Content(localizedContent, "text/html", "base64")
     }
+
+    private fun generateGeckoVersionStr(): String {
+        return "${org.mozilla.geckoview.BuildConfig.MOZ_APP_VERSION}-${org.mozilla.geckoview.BuildConfig.MOZ_APP_BUILDID}"
```
mcomella: I don't know what MOZ_APP_VERSION is (like FF 68?) or MOZ_APP_BUILDID (e.g. like the random string of numbers specifying the build time) but this looks reasonable? Ideally, how their used should be specified in their kdoc but I guess that's hard because it's `BuildConfig`.

nit: in any case, I think it'd be helpful to add an inline comment above this line showing what this is supposed to look like, e.g.:
```kotlin
// Expected: 68.0a1-20190427
return "...

File: app/src/gecko/java/org/mozilla/tv/firefox/utils/BuildFlavor.kt [hunk 2]
```diff
@@ -28,7 +28,7 @@ class BuildFlavor {
      * SystemWebView requires context to acquire version
      */
     fun getEngineVersion(@Suppress("UNUSED_PARAMETER") context: Context): String {
-        return BuildConfig.GECKOVIEW_VERSION
+        return generateGeckoVersionStr()
```
mcomella: This is unused so it's probably fine to do this. 🤷‍♀

PR #2158, opened 2019-05-02
Issue #1652: Integrate Glean SDK

### LINKED ISSUE #1652: Integrate Glean SDK Trial
Glean is the path forward for mobile telemetry SDKs at Mozilla so we should use it to reduce the maintenance burden of using our own telemetry SDK owned by the product teams. This should be more reliable and more easy to test than our existing solution.

### What / Requirements

- Review PR from Glean team
- Create SLA (agreement with glean team on post-production support)

- The glean trial is set up in our codebase
- SLA with Glean team for support post-launch

---

The Glean integration is a ways off but this bug can be used to track concerns before we get there.

Instructions (Step 1 and 3):
https://github.com/mozilla-mobile/android-components/blob/master/components/service/glean/README.md#usage

See also FFES: https://github.com/mozilla-mobile/firefox-echo-show/issues/171

travis79:
Closes #1652: This adds the basic Glean SDK integration.  **Only the baseline ping will be sent at this time** and this is why no `metrics.yaml` file is needed for this PR.  Any additional application specific metrics will need to be added in a later iteration and will require a `metrics.yaml` file to be added (and require a new data-review).

This is intended to be instrumented alongside the existing telemetry in order to do a side-by-side comparison and validation between the two. 

It should also be noted that additional pipeline work may be required before any data will be collected by the back end.

I believe this will also require data review from @liuche, @boek, or another data steward to be legitimate before merging.  For their reference, Glean has been reviewed and approved for the [reference-browser](https://github.com/mozilla-mobile/reference-browser/pull/677) and [fenix](https://github.com/mozilla-mobile/fenix/issues/1190).  Both of the links refer to custom specific metrics

### Pull Request checklist

### DIFF
--- a/app/src/main/java/org/mozilla/tv/firefox/FirefoxApplication.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/FirefoxApplication.kt
@@ -51,15 +51,7 @@ open class FirefoxApplication : LocaleAwareApplication() {
             // Enable crash reporting. Don't add anything above here because if it crashes, we won't know.
             SentryIntegration.init(this, serviceLocator.settingsRepo)
 
-            serviceLocator.settingsRepo.dataCollectionEnabled.observeForever { collectionEnabled ->
-                if (collectionEnabled != null) {
-                    // This needs to be called before Glean.initialize, or we risk 1) not
-                    // sending startup data, or 2) sending even when the user has toggled
-                    // off data collection
-                }
-                    Glean.setUploadEnabled(collectionEnabled)
-            }
-            Glean.initialize(applicationContext)
+            initGlean()
 
             TelemetryIntegration.INSTANCE.init(this)
 
@@ -74,6 +66,18 @@ open class FirefoxApplication : LocaleAwareApplication() {
                 registerActivityLifecycleCallbacks(it)
             }
         }
+    }
+
+    private fun initGlean() {
+        serviceLocator.settingsRepo.dataCollectionEnabled.observeForever { collectionEnabled ->
+            if (collectionEnabled != null) {
+                // This needs to be called before Glean.initialize, or we risk 1) not
+                // sending startup data, or 2) sending even when the user has toggled
+                // off data collection
+                Glean.setUploadEnabled(collectionEnabled)
+            }
+        }
+        Glean.initialize(applicationContext)
     }
 
     // ServiceLocator needs to be created in onCreate in order to accept Application
--- a/app/src/main/java/org/mozilla/tv/firefox/FirefoxApplication.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/FirefoxApplication.kt
@@ -7,6 +7,7 @@ package org.mozilla.tv.firefox
 import android.os.StrictMode
 import androidx.annotation.VisibleForTesting
 import android.webkit.WebSettings
+import mozilla.components.service.glean.Glean
 import mozilla.components.support.ktx.android.content.runOnlyInMainProcess
 import org.mozilla.tv.firefox.components.locale.LocaleAwareApplication
 import org.mozilla.tv.firefox.telemetry.SentryIntegration
@@ -49,6 +50,17 @@ open class FirefoxApplication : LocaleAwareApplication() {
 
             // Enable crash reporting. Don't add anything above here because if it crashes, we won't know.
             SentryIntegration.init(this, serviceLocator.settingsRepo)
+
+            // Set up Glean by connecting it to the preference flag for data collection and calling
+            // the initialize() function.
+            with(serviceLocator.settingsRepo) {
+                dataCollectionEnabled.observeForever { collectionEnabled ->
+                    if (collectionEnabled != null) {
+                        Glean.setUploadEnabled(collectionEnabled)
+                    }
+                }
+            }
+            Glean.initialize(applicationContext)
 
             TelemetryIntegration.INSTANCE.init(this)
 
--- a/app/src/main/java/org/mozilla/tv/firefox/FirefoxApplication.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/FirefoxApplication.kt
@@ -51,15 +51,13 @@ open class FirefoxApplication : LocaleAwareApplication() {
             // Enable crash reporting. Don't add anything above here because if it crashes, we won't know.
             SentryIntegration.init(this, serviceLocator.settingsRepo)
 
-            with(serviceLocator.settingsRepo) {
-                dataCollectionEnabled.observeForever { collectionEnabled ->
-                    if (collectionEnabled != null) {
-                        // This needs to be called before Glean.initialize, or we risk 1) not
-                        // sending startup data, or 2) sending even when the user has toggled
-                        // off data collection
-                        Glean.setUploadEnabled(collectionEnabled)
-                    }
+            serviceLocator.settingsRepo.dataCollectionEnabled.observeForever { collectionEnabled ->
+                if (collectionEnabled != null) {
+                    // This needs to be called before Glean.initialize, or we risk 1) not
+                    // sending startup data, or 2) sending even when the user has toggled
+                    // off data collection
                 }
+                    Glean.setUploadEnabled(collectionEnabled)
             }
             Glean.initialize(applicationContext)
 
--- a/app/src/main/java/org/mozilla/tv/firefox/FirefoxApplication.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/FirefoxApplication.kt
@@ -51,11 +51,12 @@ open class FirefoxApplication : LocaleAwareApplication() {
             // Enable crash reporting. Don't add anything above here because if it crashes, we won't know.
             SentryIntegration.init(this, serviceLocator.settingsRepo)
 
-            // Set up Glean by connecting it to the preference flag for data collection and calling
-            // the initialize() function.
             with(serviceLocator.settingsRepo) {
                 dataCollectionEnabled.observeForever { collectionEnabled ->
                     if (collectionEnabled != null) {
+                        // This needs to be called before Glean.initialize, or we risk 1) not
+                        // sending startup data, or 2) sending even when the user has toggled
+                        // off data collection
                         Glean.setUploadEnabled(collectionEnabled)
                     }
                 }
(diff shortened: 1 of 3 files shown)


travis79:

* Verify and validate glean baseline ping side-by-side with existing telemetry

* Ensure the events ping is being sent on the correct expected schedule with the expected contents

N/A

N/A

This _only_ adds the Glean baseline ping ([link](https://github.com/mozilla-mobile/android-components/blob/master/components/service/glean/docs/pings/baseline.md))

TBD

* All Firefox for Fire TV users with who have opted-in to telemetry.

* Opt-out through settings.

* Glean

* Glean and potentially other mobile teams

11) Is there a third-party tool (i.e. not Telemetry) that you are proposing to use for this data collection? If so:

* No

app/proguard-rules.pro
```diff
@@ -84,3 +84,10 @@
 # A reference class used by androidx.concurrent.futures.AbstractFuture$UnsafeAtomicHelper
 # Currently consumed (in depenendecies) by mozilla.components.service.freboard.scheduler
 -dontwarn sun.misc.Unsafe
+
+####################################################################################################
+# Glean SDK
+####################################################################################################
+
+# Keep code generated from Glean Metrics
+-keep class org.mozilla.tv.firefox.GleanMetrics.** { *; }
```
mdboom: This may not be needed anymore since the introspection was removed from Glean.  *If* it's not a distraction, it might be worth experimenting with taking this out.

travis79: I was thinking that may have been the case but I was just being sure.  This _does_ still exist in Fenix currently.

Dexterp37: I agree with @mdboom , let's remove this and make sure to try a *final build* (the release one generated on CI for which proguard applies) on a device to make sure. I'm  confident it will work :-)

app/src/main/java/org/mozilla/tv/firefox/FirefoxApplication.kt
```diff
@@ -50,6 +51,16 @@ open class FirefoxApplication : LocaleAwareApplication() {
             // Enable crash reporting. Don't add anything above here because if it crashes, we won't know.
             SentryIntegration.init(this, serviceLocator.settingsRepo)
 
+            // Initialize Glean telemetry and hook it to the collection enabled flag
+            Glean.initialize(applicationContext)
+            with(serviceLocator.settingsRepo) {
+                dataCollectionEnabled.observeForever { collectionEnabled ->
+                    if (collectionEnabled != null) {
+                        Glean.setUploadEnabled(collectionEnabled)
```
Dexterp37: This should be called *before* `Glean.initialize`, otherwise we risk not sending pings at startup, if needed. Or, worse, sending pings if the flag is flipped to off :)

app/proguard-rules.pro
mdboom: Yeah.  We need to go back and remove it from Fenix, too.  I got a bit blocked by duplicating the problem locally, and I didn't want to push the change without testing locally first.

app/src/main/java/org/mozilla/tv/firefox/FirefoxApplication.kt
travis79: Yep, I forgot about that.

travis79:
Okay @Dexterp37 and @mdboom I've removed the ProGuard rule and sorted out the initialization order.

app/src/main/java/org/mozilla/tv/firefox/FirefoxApplication.kt
```diff
@@ -50,6 +51,17 @@ open class FirefoxApplication : LocaleAwareApplication() {
             // Enable crash reporting. Don't add anything above here because if it crashes, we won't know.
             SentryIntegration.init(this, serviceLocator.settingsRepo)
 
+            // Set up Glean by connecting it to the preference flag for data collection and calling
+            // the initialize() function.
+            with(serviceLocator.settingsRepo) {
+                dataCollectionEnabled.observeForever { collectionEnabled ->
```
Dexterp37: I don't know this API, so somebody from the firefox-tv team should tell us whether or not `setUploadEnabled` gets called at least once here, or only after the settings are modified. Let's get this sorted out before merging.

travis79: This is how they hook into it for the existing telemetry, I mimic'd that.  I can add an explicit set, I agree that we should explicitly set it here.

Dexterp37: I think it's fine to just double check with the reviewer that this works the way we expect :)

travis79: Actually, upon further investigation, this always gets hit on the first pass so the explicit getting of the value shouldn't be needed.

dnarcese:
@liuche Can you provide the data review?

app/src/main/java/org/mozilla/tv/firefox/FirefoxApplication.kt
Baron-Severin: Sorry guys, it's been a busy week.

This will be provided a value whenever `SettingsRepo` is instantiated, so it's safe to rely on here.

app/src/main/java/org/mozilla/tv/firefox/FirefoxApplication.kt
```diff
@@ -50,6 +51,17 @@ open class FirefoxApplication : LocaleAwareApplication() {
             // Enable crash reporting. Don't add anything above here because if it crashes, we won't know.
             SentryIntegration.init(this, serviceLocator.settingsRepo)
 
+            // Set up Glean by connecting it to the preference flag for data collection and calling
+            // the initialize() function.
```
Baron-Severin: Nit: It seems like comment mostly describes existing code, so I think we should be safe to remove it.

I do see a review comment above saying that `setUploadEnabled` needs to be set before `initialize`, and that seems like value context.  I think we should comment about that.

app/src/main/java/org/mozilla/tv/firefox/FirefoxApplication.kt
```diff
@@ -50,6 +51,17 @@ open class FirefoxApplication : LocaleAwareApplication() {
...
+            with(serviceLocator.settingsRepo) {
```
Baron-Severin: Nit: Since we only have one receiver here, I think it's a bit simpler to avoid using the `with` syntax here.

app/src/main/java/org/mozilla/tv/firefox/FirefoxApplication.kt
```diff
@@ -50,6 +51,17 @@ open class FirefoxApplication : LocaleAwareApplication() {
...
+                dataCollectionEnabled.observeForever { collectionEnabled ->
+                    if (collectionEnabled != null) {
+                        Glean.setUploadEnabled(collectionEnabled)
+                    }
+                }
+            }
+            Glean.initialize(applicationContext)
```
Baron-Severin: Nit: This function operates at a lower level of abstraction than the rest of `onCreate`, so I'd rather move it into a helper function.

liuche:
Since this will be sending the Baseline ping, can we include one change to the docs/Telemetry.md file, that links to the Glean Baseline ping?

Baron-Severin:
> @travis79 @Baron-Severin Since this will be sending the Baseline ping, can we include one change to the docs/Telemetry.md file, that links to the Glean Baseline ping?
> 
> Like what Fenix does here:
> https://github.com/mozilla-mobile/fenix/pull/979/files#diff-b6840e6fe12d2b4f3ce88beefb5c876fR3

Done

liuche:
This is mystifying, this PR used to pass CI, but now I'm seeing [this error](https://taskcluster-artifacts.net/PAgx66-nSnCPVy3lHV_93w/0/public/reports/tests/testGeckoDebugUnitTest/classes/org.mozilla.tv.firefox.components.locale.LocalesTest.html) with WorkManager not being initialized correctly.

```
java.lang.IllegalStateException: WorkManager is not initialized properly.  The most likely cause is that you disabled WorkManagerInitializer in your manifest but forgot to call WorkManager#initialize in your Application#onCreate or a ContentProvider.
	at androidx.work.WorkManager.getInstance(WorkManager.java:141)
	at mozilla.components.service.glean.scheduler.MetricsPingScheduler.schedulePingCollection$service_glean_release(MetricsPingScheduler.kt:76)
	at mozilla.components.service.glean.scheduler.MetricsPingScheduler.startupCheck(MetricsPingScheduler.kt:198)
	at mozilla.components.service.glean.GleanInternalAPI.initialize(Glean.kt:113)
	at mozilla.components.service.glean.GleanInternalAPI.initialize$default(Glean.kt:81)
	at org.mozilla.tv.firefox.FirefoxApplication.initGlean(FirefoxApplication.kt:80)
```

liuche:
Hi @Dexterp37 and @mdboom - could you take a quick look at why Glean might be failing on this WorkManager initialization? I don't think this test should be doing anything different from the other tests (and seems to be a RTL test in fact) so I'm wondering what could be happening. Any thoughts? Do we just need to initialize the WorkManager in our test setup?

@npark-mozilla do you know if something has changed in our test harness that some dependencies wouldn't be working anymore? This is especially curious because before the documentation change, this PR was green 🤔 
However, the workaround that @travis79 suggested is to add [this WorkManager initialization]([url](https://github.com/mozilla-mobile/android-components/blob/1c3a7a2e466c115d0daebe0c58ac2723d14b7baa/components/service/glean/src/test/java/mozilla/components/service/glean/GleanTest.kt#L63)) to that test.

npark-mozilla:
AFAIK, I do not think there was any harness change on bitbar or firebase, and we did not update the test dependencies recently.

Dexterp37:
> 
> 
> Hi @Dexterp37 and @mdboom - could you take a quick look at why Glean might be failing on this WorkManager initialization? I don't think this test should be doing anything different from the other tests (and seems to be a RTL test in fact) so I'm wondering what could be happening. Any thoughts? Do we just need to initialize the WorkManager in our test setup?

I'm on this. Let me take a look.

Dexterp37:
> Hi @Dexterp37 and @mdboom - could you take a quick look at why Glean might be failing on this WorkManager initialization? I don't think this test should be doing anything different from the other tests (and seems to be a RTL test in fact) so I'm wondering what could be happening. Any thoughts? Do we just need to initialize the WorkManager in our test setup?
> 
> However, the workaround that @travis79 suggested is to add this WorkManager initialization to that test.

So, I don't think we should be initializing the `WorkManager` in this case. Yes, it would solve the immediate problem, but this is only highlighting a different problem: tests are running with Telemetry turned on.

Since we don't have any test specifically testing Glean, we should probably turn it off in unit tests.

liuche:
Duplicated in and closed by #2267

PR #2174, opened 2019-05-03
No issue comment strings

Baron-Severin:
### Pull Request checklist

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -106,9 +106,16 @@
     <!-- Settings strings -->
     <string name="settings_cookies_dialog_title">Clear all cookies and site data</string>
     <string name="settings_cookies_dialog_content2">This will clear all cookies and site data. Clearing this information may sign you out of websites and delete web content that is stored offline.</string>
+    <!-- Text on a button that, when pressed, will clear all user data.
+    Mock: https://user-images.githubusercontent.com/1721875/47298531-59442f00-d618-11e8-98df-632f74e9ebc3.png -->
     <string name="settings_cookies_confirm">Clear now</string>
 
+    <!-- Title for a screen that explains the purpose of our telemetry, and contains a
+    switch that can toggle it on and off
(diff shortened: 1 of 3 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -106,9 +106,16 @@
     <!-- Settings strings -->
     <string name="settings_cookies_dialog_title">Clear all cookies and site data</string>
     <string name="settings_cookies_dialog_content2">This will clear all cookies and site data. Clearing this information may sign you out of websites and delete web content that is stored offline.</string>
+    <!-- Text on a button that, when pressed, will clear all user data.
```
liuche: These strings are actually already translated - I guess we could add these comments though!

app/src/main/res/values/strings.xml
```diff
@@ -115,8 +115,9 @@
      Mock: https://user-images.githubusercontent.com/1721875/47298694-b809a880-d618-11e8-9e18-1d6d47d2045d.png -->
     <string name="preference_mozilla_telemetry2">Send usage data</string>
     <!-- Description text that explains the purpose of our telemetry
-      Mock: https://user-images.githubusercontent.com/1721875/47298694-b809a880-d618-11e8-9e18-1d6d47d2045d.png -->
-    <string name="settings_telemetry_description">Mozilla strives to collect only what we need to provide and improve Firefox for Fire TV for everyone</string>
+      Mock: https://user-images.githubusercontent.com/1721875/47298694-b809a880-d618-11e8-9e18-1d6d47d2045d.png
+      %1$s will be replaced with the name of the app (e.g. Firefox for Fire TV) -->
+    <string name="settings_telemetry_description">Mozilla strives to collect only what we need to provide and improve %1$s for everyone</string>
```
liuche: One note here that you may not be aware of: if we need a string to be re-translated (which is happening in this case, because the string has changed), then you need to change the string name - usually we suffix it with a number (like 1). That way, it'll show up in the Pontoon dashboard.

liuche:
derp lol

PR #2176, opened 2019-05-06
Closes #2162: Fix channel telemetry

### LINKED ISSUE #2162: [2159] Fix channel telemetry
#2110 broke telemetry in two places
- `NavigationOverlayFragment#onContextItemSelected` needs to send a `homeTileRemovedEvent`
- `DefaultChannelAdapter#onBindViewHolder` needs to send a `homeTileClickEvent`

Both of these events expect a `PinnedTile` as input, and send its ID.  Make sure that the new implementation creates IDs that are the same as the previous one, or this will cause a discontinuity in our data.

### Acceptance Criteria
- Removing a pinned tile sends a telemetry event
- Clicking on a pinned tile sends a telemetry event
- Sent events are indistinguishable from events sent by our previous implementation (specifically, their IDs)

dnarcese:
### Pull Request checklist

### DIFF
--- a/app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/NavigationOverlayFragment.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/NavigationOverlayFragment.kt
@@ -51,6 +51,7 @@ import org.mozilla.tv.firefox.navigationoverlay.channels.DefaultChannelFactory
 import org.mozilla.tv.firefox.navigationoverlay.channels.SettingsChannelAdapter
 import org.mozilla.tv.firefox.navigationoverlay.channels.SettingsScreen
 import org.mozilla.tv.firefox.pocket.PocketViewModel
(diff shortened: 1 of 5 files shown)


dnarcese:
@Baron-Severin The only places we use the id's is for 

- bundled tile clicks - which the id's are strings so those still match
- collecting unique tile clicks per session - since this is just a count of unique tiles, the id's don't need to match

Is that the only verification needed or am I missing something else?

Baron-Severin:
> @Baron-Severin The only places we use the id's is for
> 
>     * bundled tile clicks - which the id's are strings so those still match
> 
>     * collecting unique tile clicks per session - since this is just a count of unique tiles, the id's don't need to match
> 
> 
> Is that the only verification needed or am I missing something else?

Yep, that's all I'm aware of.  I just want to make sure that this doesn't break the continuity of our data (e.g., a click event with field `id:youtube` instead has field `id youtube`, complicating dashboard queries).

Issue #2203, opened 2019-05-08
Overflowing carousel scroll to show up in the right side of the channel


psymoon:
### Why/User Benefit/User Problem
- We need an indicator to show scrollability for channels on the right side

### What / Requirements
- Implement an overflowing carousel scrolling on the right side of the channels similarly to what's available in the left

### Acceptance Criteria (how do I know when I’m done?)
- The above requirement is implemented and tested

PR #2221, opened 2019-05-09
#2203: Channel Scrolling updates.

psymoon:
### Pull Request checklist

### DIFF
--- a/app/src/gecko/java/org/mozilla/tv/firefox/ext/View.kt
+++ b/app/src/gecko/java/org/mozilla/tv/firefox/ext/View.kt
@@ -4,8 +4,12 @@
 
 package org.mozilla.tv.firefox.ext
 
+import android.content.Context
 import android.view.View
+import android.view.ViewGroup
+import androidx.core.view.updateLayoutParams
 import org.mozilla.geckoview.GeckoView
+import org.mozilla.tv.firefox.R
 
 private const val GECKO_VIEW_ID = 2147483645 // Int.MAX_VALUE - 2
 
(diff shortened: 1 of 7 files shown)


app/src/gecko/java/org/mozilla/tv/firefox/ext/View.kt
```diff
@@ -26,6 +30,27 @@ fun View.validateKnownViewById(): Int {
...
+/**
+ * Set the horizontal margins on the given view.
+ *
+ * We want to add padding to the beginning and end of the RecyclerView: ideally we'd just add
+ * paddingStart/End. Unfortunately, this causes a visual glitch as each card scrolls offscreen.
+ * Instead, we set the margins for the first and last card.
+ *
+ * See [DefaultChannelAdapter.onBindViewHolder] for an example
+ */
+fun View.setHorizontalMargins(context: Context, position: Int, itemCount: Int) {
```
Baron-Severin: Does this need to be declared in both System and Gecko?  The implementations seem to be the same, unless I'm missing something.  Would it be possible to merge these?

Baron-Severin: This functionality doesn't seem broadly applicable to the rest of the app.  I also feel that this is too narrow a use case for it to make sense polluting the View namespace.  Could we move this into a companion object for `ChannelTile`, or something similar?

Baron-Severin: At first glance, I would expect a fun called `setHorizontalMargins` to take in two ints and set them to the left and right margins for a view.  Could we rename this to better communicate that it's handling a much more specific job?  `setChannelMarginByPosition` or something?  That's super wordy, but you see the direction I'm trying to move.

psymoon: View ext already exists in both variants, moving to main would raise a compiler issue

mcomella: I don't understand: what compiler issue would you see?

mcomella: Nvm, looks like this was removed anyway.

Baron-Severin: @mcomella At any given time, there would have been two files with the same package statement & name (one from main, one from either system or GV).

Issue #2166, opened 2019-05-02
[2159] Support deleting tiles from channels


Baron-Severin:
### Why/User Benefit/User Problem
Currently, items may only be deleted from pinned tiles.  @brampitoyo, do we plan to support removing tiles from channels, now or in the near future?

### What / Requirements
If we want to support removing tiles from other channels:
- Adjust implementation to remove from other channels
- Allow configuration of channels to allow or deny this behavior

If we do not plan to support this
- Update `registerForContextMenu(channelsContainer)` found in `NavigationOverlayFragment#onViewCreated` to only show the "Remove" popup when a pinned tile is long pressed (currently shows up on any channel tile)

brampitoyo:
Yes. FFTV should have the ability to remove tiles from channels in the future.

(This is part of a larger work that will allow users to pin sites to specific channels and maybe even create their own channels).

Before this happens, it’s necessary for FFTV to first ship with enough tiles across enough categories. With enough tiles shipped, a need will arise to edit/remove/reorganise those tiles.

liuche:
Refactoring to support different channels is done, and this just needs UX support to style the dialog.

This is necessary for removing tiles from channels.

psymoon:
![screen](https://user-images.githubusercontent.com/3597594/57554035-556ec480-733e-11e9-9653-b7e8f077421f.png)

@brampitoyo This is what it currently looks like, red highlighted items are changeable (also the background). Alternatively, we can define a whole new layout.

brampitoyo:
@psymoon 

> […] red highlighted items are changeable

Let’s use this:

**Remove metacritic from Pinned sites?**

By having the category name, this dialogue will scale when the user tries to remove a tile in other categories.

Let’s change the button title and add a new one:

**[[Remove site]]** **[[Cancel]]**

> Alternatively, we can define a whole new layout.

I’m okay with using a modal dialogue, but want to be cognizant of other dialogues we’ve already used:

![homescreen](https://user-images.githubusercontent.com/51007/57599219-1644ab80-75aa-11e9-8f7d-a084f5f777f5.png)

![homescreen](https://user-images.githubusercontent.com/51007/57599247-2d839900-75aa-11e9-93bc-8f2ba9d579f7.png)

Instead of defining a new style, I wonder whether we should reuse the text and button style of the Pocket onboarding dialogue.

Our remove tile dialogue can look like this:

[...]

PR #2235, opened 2019-05-10
#2166: Use Dialog to abstract Channel removal [Do not merge until UX verified)

psymoon:
### Pull Request checklist

### DIFF
--- a/app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/NavigationOverlayFragment.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/NavigationOverlayFragment.kt
@@ -271,7 +271,7 @@ class NavigationOverlayFragment : Fragment() {
 
     private fun observePinnedTileRemoval(): Disposable {
         return pinnedTileChannel.removeTileEvents.subscribe { tileToRemove ->
-            pinnedTileChannel.removeContent(tileToRemove)
+            serviceLocator.channelRepo.removeChannelContent(tileToRemove)
         }
     }
 
--- a/app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/NavigationOverlayFragment.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/NavigationOverlayFragment.kt
@@ -215,6 +215,8 @@ class NavigationOverlayFragment : Fragment() {
             .addTo(compositeDisposable)
         observePinnedTiles()
             .addTo(compositeDisposable)
+        observePinnedTileRemoval()
+            .addTo(compositeDisposable)
         observeShouldDisplayPinnedTiles()
             .addTo(compositeDisposable)
         observePocketState()
@@ -269,6 +271,12 @@ class NavigationOverlayFragment : Fragment() {
         return navigationOverlayViewModel.pinnedTiles.subscribe {
             pinnedTileChannel.setTitle(it.title)
             pinnedTileChannel.setContents(it.tileList)
+        }
+    }
+
+    private fun observePinnedTileRemoval(): Disposable {
+        return pinnedTileChannel.removeTileEvents.subscribe { tileToRemove ->
+            pinnedTileChannel.removeContent(tileToRemove)
         }
     }
 
--- a/app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/NavigationOverlayFragment.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/NavigationOverlayFragment.kt
@@ -12,10 +12,8 @@ import android.os.Handler
 import android.os.Looper
 import android.preference.PreferenceManager
 import android.util.AttributeSet
-import android.view.ContextMenu
 import android.view.KeyEvent
 import android.view.LayoutInflater
-import android.view.MenuItem
 import android.view.View
 import android.view.ViewGroup
 import android.widget.ScrollView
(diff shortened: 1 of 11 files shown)


app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/channels/ChannelTile.kt
```diff
@@ -23,7 +23,8 @@ data class ChannelTile(
     val title: String,
     val setImage: (ImageView) -> Unit,
     val tileSource: TileSource,
-    val id: String
+    val id: String,
+    val type: TileType
```
Baron-Severin: This seems like it's the same concept as `TileSource`.  Could we merge these into one?

Baron-Severin: For the record, I am worried about these.  We've had a lot of unnecessary branching logic based on tile type in the past.  Merging them into `ChannelTile` wasn't easy, and I'm worried that we could be setting ourselves up to repeat that mistake.

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/channels/DefaultChannel.kt
```diff
@@ -21,5 +22,9 @@ class DefaultChannel(
         adapter.submitList(tileData)
     }
 
+    fun removeContent(tileData: ChannelTile) {
+        // TODO
+    }
```
Baron-Severin: This function shouldn't be needed, right?  It doesn't particularly mesh with our architecture either.  The display state after calling `removeContent` is based on previous state, whereas `setContent` handles both add and remove in a way that is stateless.

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/NavigationOverlayFragment.kt
```diff
@@ -258,6 +260,12 @@ class NavigationOverlayFragment : Fragment() {
         }
     }
 
+    private fun observePinnedTileRemoval(): Disposable {
+        return pinnedTileChannel.removeTileEvents.subscribe { tileToRemove ->
+            pinnedTileChannel.removeContent(tileToRemove)
+        }
```
Baron-Severin: Hm, so we listen to the adapter, react to that information in the fragment, then use it to call back into the adapter.  I think we can simplify this.  What if we remove `DefaultChannel#removeContent`, then call straight into the repo from here?

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/channels/DefaultChannelAdapter.kt
```diff
@@ -65,7 +66,18 @@ class DefaultChannelAdapter(
...
+
+                builder.setCancelable(true)
+                builder.setPositiveButton(
+                        context.resources.getString(R.string.homescreen_tile_remove)
+                ) { dialog, _ ->
+                    _removeEvents.onNext(tile)
+                    dialog.dismiss()
+                }
+
+                builder.create().show()
```
Baron-Severin: This is much, much nicer than our current solution.  I'll be happy to have this in the codebase.

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/channels/DefaultChannelAdapter.kt
```diff
@@ -65,13 +65,15 @@ class DefaultChannelAdapter(
             }
 
             itemView.setOnLongClickListener {
-                val builder = AlertDialog.Builder(context)
-                builder.setTitle(tile.url)
+                val builder = AlertDialog.Builder(context, R.style.DialogStyle)
+
+                val removeStr = context.resources.getString(R.string.homescreen_tile_remove)
+
+                val title = removeStr + " " + tile.title + "?"
```
Baron-Severin: We'll need to get this localized again (I believe this could change the grammar is some languages), but we should wait until we get our final string from UX.  Would you mind adding a comment to that effect?  (e.g., 'we need to localize this, but are waiting until we hear back from UX'

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/channels/DefaultChannel.kt
Baron-Severin: Later commits clarify what this is doing.  I think you can ignore this comment in favor of https://github.com/mozilla-mobile/firefox-tv/pull/2235#discussion_r283029432

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/NavigationOverlayFragment.kt
psymoon: Makes sense to me

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/channels/DefaultChannelAdapter.kt
```diff
@@ -69,6 +69,7 @@ class DefaultChannelAdapter(
 
                 val removeStr = context.resources.getString(R.string.homescreen_tile_remove)
 
+                // FIXME: We need to localize this, but are waiting until we hear back from UX
```
Baron-Severin: Nit: "waiting until we hear back from UX" -> "waiting until we have the final string from UX"

PR #2294, opened 2019-05-17
Closes #2274: Refactor bundle tile logic

### LINKED ISSUE #2274: Identify storage format for new category channels
Blocks implementation of new channels to homescreen like #2196.

The new homescreen channels need to bundle tiles so we'll need to figure out how we want to store them.

**Keep it simple** and consider how it interacts with our existing bundled tile format.

psymoon:
### Pull Request checklist

### DIFF
--- a/app/src/main/java/org/mozilla/tv/firefox/channels/BundleTilesStore.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/channels/BundleTilesStore.kt
@@ -7,6 +7,7 @@ package org.mozilla.tv.firefox.channels
 import android.content.Context
 import android.content.SharedPreferences
 import org.json.JSONArray
+import org.json.JSONObject
 import java.util.Collections
 
 private const val PREF_BUNDLE_TILES = "bundleTiles"
@@ -44,7 +45,7 @@ class BundleTilesStore(private val applicationContext: Context) {
         _sharedPreferences.edit().putStringSet(sharedPrefKey, blackList).apply()
     }
 
-    fun getBundledTiles(type: BundleType): JSONArray {
+    fun getBundledTiles(type: BundleType): List<JSONObject> {
         val jsonPath = when (type) {
             BundleType.PINNED_TILES -> PINNED_TILES_JSON_PATH
         }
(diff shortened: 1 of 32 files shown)


app/src/main/java/org/mozilla/tv/firefox/channels/BundleTilesStore.kt
```diff
@@ -0,0 +1,30 @@
...
+enum class BundleType {
+    PINNED_TILES
+}
+
+/**
+ * [BundleTilesStore] is responsible for fetching bundled tiles data from /assets/ with
+ * [BundleType]
+ */
+class BundleTilesStore(private val applicationContext: Application) {
+    fun getBundledTiles(type: BundleType): JSONArray {
```
Baron-Severin: While we're moving things around anyway, I feel like we should be doing type conversions inside this class.  We lose a lot of type safety and clarity by exposing a `JSONArray`, I'd rather do that conversion on the edge between the app and storage (that is, here).

TL;DR: I think this should be a private function, then we can expose a `getPinnedTiles` method that calls this, converts, and returns a `List<ChannelTile>` (or `List<PinnedTile>` if that's unworkable).

app/src/main/java/org/mozilla/tv/firefox/channels/BundleTilesStore.kt
```diff
@@ -19,12 +26,50 @@ enum class BundleType {
...
+            }
+        }
+
+        return tilesJSONArray
+    }
+
+    /**
+     * Used to handle removing bundle tiles by adding to its [BundleType] blacklist in SharedPreferences
+     */
+    fun addBundleTileToBlackList(type: BundleType, id: String) {
```
Baron-Severin: I like this solution.

app/src/main/java/org/mozilla/tv/firefox/channels/pinnedtile/PinnedTileRepo.kt
```diff
@@ -30,7 +29,7 @@ private const val PREF_HOME_TILES = "homeTiles"
  * @constructor loads the initial [_pinnedTiles] (a combination of custom and bundled tiles)
  */
 class PinnedTileRepo(
-    private val applicationContext: Application,
+    private val applicationContext: Context,
```
Baron-Severin: Nit: If we're moving this to `Context`, we should rename it to match.

app/src/main/java/org/mozilla/tv/firefox/channels/BundleTilesStore.kt
```diff
@@ -25,7 +24,7 @@ enum class BundleType {
  * [BundleTilesStore] is responsible for fetching bundled tiles data from /assets/ with
  * [BundleType]
  */
-class BundleTilesStore(private val applicationContext: Application) {
+class BundleTilesStore(private val applicationContext: Context) {
```
Baron-Severin: Same as above

> Nit: If we're moving this to Context, we should rename it to match.

app/src/main/java/org/mozilla/tv/firefox/channels/BundleTilesStore.kt
psymoon: Yes I agree with this, I'll give some think on it

Issue #2223, opened 2019-05-09
Fetch & Save Pocket Articles at Regular Intervals


athomasmoz:
### Why/User Benefit/User Problem

As a user, I want Pocket content to be updated regularly, so that I can find new videos to watch when I open Firefox

### Tasks

* Confirm increased cost with Pocket for fetching data for all users nightly
* Get default content from Pocket
* Either turn off foreground fetching of new articles and close #2187 , or ensure #2187 has been prioritized for a future sprint

### Acceptance Criteria (how do I know when I’m done?)

* A list of Pocket videos is available to the user every time they open the app
  * The first time a user opens Firefox, a default list of Pocket videos are displayed
  * Each day a user opens Firefox, new Pocket content is displayed
  * Fetching new data should happen at night, when user is least likely to be using the app (e.g. because they're sleeping; Bram guessed 3am-5am their time zone but we should double-check that)
* Pocket content is fetched & saved regularly, so that users do not need to wait for it when the app opens
  * Fetching content does not occur at the same time for all users, to avoid overloading (add randomness to scheduled times, back-off times; consider WorkManager)
  * Pinging Pocket's server nightly for each user does not drastically increase the cost

### Follow-ups
- [x] Close #2297 (videos update when user looks at them) if applicable
- [x] Close #2187 (foreground fetch interval) if applicable
- [x] Close #2261 (fetching for inactive users) if applicable
- [ ] See if there are any other issues to close

athomasmoz:
@athomasmoz to tackle conversations with Pocket on additional API calls

mcomella:
@brampitoyo It turns out the Android APIs don't make it easy to guarantee a task daily at specific time (e.g. 3am) or within specific hours (e.g. 3am - 5am). For the MVP, can we say we'll make a best effort attempt to update within those hours but only guarantee we'll update once per day?

fwiw, by "best effort", in our testing, the system defaults to updating within seconds of the time we specify but may not guarantee it will continue to do so forever.

Note to self: we may want to add telemetry to track whether this solution is good enough or not.

brampitoyo:
> …can we say we'll make a best effort attempt to update within those hours but only guarantee we'll update once per day?

That works for me. A once-a-day update is a good minimum, especially considering the fact that Pocket doesn’t update its video recommendations as often as they do with articles.

liuche:
Follow-ups filed, can close.

PR #2305, opened 2019-05-22
Issue #2223: end-to-end Pocket background fetch working 🔁

mcomella:
I've tested the following cases:
- The app on fresh install displays the bundled Pocket data
- The PocketVideoFetchScheduler says `userFetchTime` is some time tomorrow night (I logged the value using `SimpleDateFormat.getInstance().format(userFetchTime.time)`).

That means we should still manually test:
- [x] PocketVideoFetchScheduler schedules a job and the data actually gets updated
  - You should have to background and foreground the app for the data to get updated
  - I recommend overridng the time to the near future (e.g. in 2 min) to test whether this is working
- Failure cases
  - [x] No API key
  - [x] Non en-US locales
  - [ ] Failure to hit the server == retry in background job (retry delay should have some randomness)
  - [ ] Invalid json from server == just fail
  - [ ] Does Sentry catch crashes in PocketWorker?
  - [ ] What happens when exceptions are thrown inside the PocketWorker?

The tasks left open are:

- ^ The above testing
- Fix the automated tests
  - [x] Make them compile
  - [x] Make them pass on CI (if they don't already)
- [x] Get this reviewed
- Remove remaining old Pocket foreground fetch code
  - [ ] Rename PocketEndpoint?
  - [ ] Remove PocketRepoCache and comment why it's not necessary (b/c we only content in onStart)
  - [ ] Remove unused code
- [ ] Close the issues we already fixed (listed in comment 0 of the issue this closes)
- TODOs in code
  - [x] en-US only
  - [ ] Clean up comments in scheduler
- TODOs in pairing doc
  - [ ] Notify team that we're scheduling one-time fetches in onStart and trade-offs (see doc)
- [ ] (could be follow-up) Write automated tests
- [x] `WorkRequest.setBackoffCriteria`: we don't want to run into Pocket server problems again
- [x] Land final Pocket data
- [ ] ? Anything I've forgotten

---

@dnarcese The ways the code has changed since you last looked at it:
- I abstracted the sharedPreferences & validation logic into the PocketVideoStore so the VideoRepo & Worker don't need to know how data gets stored
- PocketVideoFetchWorker is much simpler because of ^: it just fetches and asks the store to save
- PocketVideoRepo is much simpler because of ^: it just gets data from the store and puts it into our architecture

### Pull Request checklist


app/src/androidTest/java/org/mozilla/tv/firefox/helpers/CustomPocketFeedStateProvider.kt
```diff
@@ -23,7 +23,9 @@ class CustomPocketFeedStateProvider {
 
     private val localeIsEnglish: () -> Boolean = { true }
 
-    private val pocketEndpoint = object : PocketEndpoint("VERSION", "www.mock.com".toUri(), localeIsEnglish) {
+    // Because of the endpoint overrides, the raw endpoint is unused so the arguments don't matter.
+    private val pocketEndpointRaw = PocketEndpointRaw("001", null)
```
Baron-Severin: I don't understand what this comment means.  If the arguments don't matter, can they be removed?  I'm not clear on what's going on here.

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketVideoStore.kt
```diff
@@ -0,0 +1,91 @@
...
+
+/**
+ * The minimum number of valid videos we must receive from the server if we want to display them.
+ * This number is set to the number of videos that appear on screen at the time of writing.
+ */
+private const val REQUIRED_POCKET_VIDEO_COUNT = 4
+
+/**
+ * Saves the Pocket video recommendations as a raw JSON String and loads them in data structures for the app.
+ * Bad data should saved so bad data should never be returned.
```
Baron-Severin: `Bad data should saved so bad data should never be returned` 

Should this say `Bad data should not be saved`?

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketVideoStore.kt
```diff
@@ -0,0 +1,91 @@
...
+     * @return the list of loaded videos. This should never happen but in case of error, the empty list is returned.
+     */
+    @AnyThread
+    fun load(): List<PocketViewModel.FeedItem> {
+        fun loadBundledTiles(): String = assets.open(BUNDLED_VIDEOS_PATH).bufferedReader().use { it.readText() }
+
+        val rawJSON =  sharedPrefs.getString(KEY_VIDEO_JSON, null) ?: loadBundledTiles()
+
+        val convertedVideos = convertJSONToPocketVideos(rawJSON)
+        if (convertedVideos == null) {
```
Baron-Severin: Our failure cases (null and size is too small) should probably be consistent between save and load.

app/src/main/java/org/mozilla/tv/firefox/utils/ServiceLocator.kt
```diff
@@ -84,7 +85,9 @@ open class ServiceLocator(val app: Application) {
     val screenshotStoreWrapper by lazy { PinnedTileImageUtilWrapper(app) }
     val formattedDomainWrapper by lazy { FormattedDomainWrapper(app) }
     val channelRepo by lazy { ChannelRepo(pinnedTileRepo) }
+    val pocketEndpointRaw by lazy { PocketEndpointRaw(appVersion, buildConfigDerivables.globalPocketVideoEndpoint) }
```
Baron-Severin: This doesn't seem like something that should be publicly exposed.  Looking at the finished code, `PocketEndpointRaw` is only used by `PocketEndpoint` and `PocketVideoFetchWorker`.  If we update `PocketVideoFetchScheduler` and `PocketVideoFetchWorker` to accept a `PocketEndpointRaw` as a param, `pocketEndpointRaw` could be a private param in `ServiceLocator`.

app/src/main/java/org/mozilla/tv/firefox/MainActivity.kt
```diff
@@ -62,7 +62,10 @@ class MainActivity : LocaleAwareAppCompatActivity(), OnUrlEnteredListener, Media
         val safeIntent = intent.toSafeIntent()
         webRenderComponents.notifyLaunchWithSafeIntent(safeIntent)
 
-        lifecycle.addObserver(serviceLocator.engineViewCache)
+        arrayOf(
+            serviceLocator.engineViewCache,
+            serviceLocator.pocketVideoFetchScheduler
+        ).forEach { lifecycle.addObserver(it) }
```
Baron-Severin: Nit: a little repetition here would actually be less code, and more immediately understandable.
```
lifecycle.addObserver(serviceLocator.engineViewCache)
lifecycle.addObserver(serviceLocator.pocketVideoFetchScheduler)
```

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketVideoFetchScheduler.kt
```diff
@@ -0,0 +1,85 @@
...
+        val constraints = Constraints.Builder()
+            .setRequiredNetworkType(NetworkType.CONNECTED)
+            .build()
+
+        // TODO: why do we want this from a UX perspective.
+        // todo: this explanation sucks.
+        // Our fetch scheduling scheme: when the user foregrounds the app, we schedule an update for the next calendar
+        // day (i.e. if today is the 2nd, we schedule for the 3rd) at a random time inside our fetch interval.
+        // it's still "tonight" but already after midnight, we
+        // attempt to schedule one for the following night.
```
Baron-Severin: Does this mean that a daily user would never receive Pocket updates?

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketRepoCache.kt
```diff
@@ -22,6 +22,7 @@ import io.reactivex.Observable
  * should. This lets us manually control when updates are allowed.
  */
 open class PocketRepoCache(repo: PocketVideoRepo) {
+    // TODO: no longer necessary. Be sure to explain why somewhere.
```
Baron-Severin: Since this isn't required anymore, can we just remove it?  It would make the code less confusing.

app/src/androidTest/java/org/mozilla/tv/firefox/helpers/CustomPocketFeedStateProvider.kt
mcomella: The test overrides any functionality using endpointRaw so the specifics of the arguments passed to the initializer of endpointRaw don't matter (because endpointRaw unused). The arguments can't be removed but maybe it'd be clearer to pass a mock in instead?

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketVideoStore.kt
mcomella: Yes

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketVideoStore.kt
mcomella: My thinking is that this should never happen but we don't want to crash (to prevent crash loops) so we should return a reasonable value that will put us in an undefined state.

Then again, that undefined state could be crashing so I guess we should explicitly handle this, e.g. by returning null.

app/src/main/java/org/mozilla/tv/firefox/utils/ServiceLocator.kt
mcomella: Agreed.

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketVideoFetchScheduler.kt
mcomella: What do you mean by a daily user? This comment is really bad and needs to be improved. :(

The only people who would not receive updates are people who leave the app open all the time without a screensaver (because onStart is never called again to schedule background fetching or push in new content).

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketRepoCache.kt
mcomella: It'd require some rejiggering of the tests so it didn't seem important, compared to the rest of the changes we needed to get done.

app/src/androidTest/java/org/mozilla/tv/firefox/helpers/CustomPocketFeedStateProvider.kt
Baron-Severin: Nevermind, didn't notice this was in the test directory.  No change necessary here.

app/src/main/java/org/mozilla/tv/firefox/utils/ServiceLocator.kt
dnarcese: Passing parameters to a `Worker` is more complicated because it needs to be passed as a `Data` object. Typically, only basic types like strings and ints are passed as data, not objects. Not sure if this is worth the complexity? Is there another way?

Baron-Severin: I don't understand, what do you mean about it needing to be passed as a data object?  

We could also talk about this on Slack or Zoom.

dnarcese:
Filed #2325 to finish the tasks.

Issue #2306, opened 2019-05-22
Use SVGs as tiles


liuche:
### Vision statement / What / Requirements
As part of #2280 we didn't have time to handle SVGs, so we used 193.5 x 108 pngs instead. Since our tiles are all svgs, we need to support them as SVGs

### Impact
Our resources for other Channel tiles are svgs.
Alternatively, we can get them exported as pngs, but that's a lot more work for UX.

### Acceptance criteria
Load SVGs for tiles
(Unfortunately, Picasso doesn't seem to do this right out of the box.)

Issue #2280, opened 2019-05-15
Update Pocket to Match Mocks


athomasmoz:
### Vision statement / What / Requirements

The current implementation doesn't match the mocks, which were created to provide the best experience for users and match the overall channels design. This ticket is to update the Pocket channel to match those [mocks](https://brampitoyo.github.io/people/bpitoyo/firefox-for-fire-tv/new-homescreen/i04/#artboard1)

### Impact

* Improved User Experience
* Consistency between channels

### Acceptance criteria

* The following updates have been made to the Pocket channel to match the mocks:
- [x] Full-tile image for each tile
- [ ] Update Pinned tiles to use [new full-tile svgs](https://drive.google.com/drive/folders/15IFUqAoukdbAC-UEIKxD-nvrR-JeoFaf)
- [x] Highlight focus ring
- [x] 3-line wrapped title (rather than a single title space)
- [x] Spacing between tile and titles
- [ ] [blocked] Video length (this isn't currently included in the Pocket API, but would be good to add! Filed #2299)

liuche:
Moving this into the sprint bc the mocks exist and we need it for shipping Pocket.

liuche:
Also, these mocks look different from the original ones that we had - there, we had a single title space that spanned a longer width to accommodate the longer titles of Pocket Stories. In this version of the mock each tile has its own title underneath.

We've already implemented the single long title, so I'll have to undo that here.

liuche:
@brampitoyo Do you think you could get us the SVG assets for the bundled Pinned Tiles? Otherwise the Pinned Tiles with the current icons would look very boxy (since the assets aren't full-tile size).
* Youtube
* Google Video Search
* IMDB
* Rotten Tomatoes
* Metacritic
* Fandango
* Hollywood Reporter
* Flickr
* Instagram
* Pinterest

liuche:
![Screenshot_1558490606](https://user-images.githubusercontent.com/1082253/58142234-22a8b400-7bfb-11e9-8e4d-23753d267a51.png)


Things that I didn't do: 
- [ ] [**MVP**] Need to add the new Pinned Tile images so the tiles are full-size - @bram is making these as pngs, so we should be able to just swap them in. I filed #2306 to handle this.
[...]

PR #2307, opened 2019-05-22
Pocket channel style

liuche:
This adds styling for the Pocket and Pinned Tile channels.

There is one outstanding item that needs to happen, which is add the new Pinned Tiles svgs. (@bram will add the link to the new png assets in the issue - we can just replace the whole list of tiles in `assets/bundled` and this should just work)

Things that I didn't do (that would be great to do): 
- [ ] [**MVP release blocker**] Need to add the new Pinned Tile images so the tiles are full-size - @bram is making these as pngs, so we should be able to just swap them in. I filed #2306 to handle this.
- [ ] Settings currently isn't a "default channel", so its scrolling behavior looks pretty different. HOWEVER all the rest of the tiles are now `ListRowView` (same as Settings) so it's probably not TOO much more work to unify the code now, so the scrolling and everything is consistent 🤔 

https://github.com/mozilla-mobile/firefox-tv/issues/2280#issuecomment-494624767

### Pull Request checklist

top-focus
instagram


app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/channels/DefaultChannelAdapter.kt
```diff
@@ -64,7 +66,18 @@ class DefaultChannelAdapter(
             ChannelTile.setChannelMarginByPosition(holder.itemView, context, position, itemCount)
             val tile = getItem(position)
             tile.setImage.invoke(imageView)
-            titleView.text = tile.title
+
+            // We handle Pocket tiles differently.
```
psymoon: It's reasonable with the way it is now and there's a time pressure for this, but we should consider polymorphism for `ChannelTile` instead of null checking:

```
when (tile) {
	is PocketTile -> // pocket stuff
	is PinnedTile -> // pinned_tile stuff
}
```

Baron-Severin: I would take a different approach on this one, personally.  I would rename these from `titleView`, `pocketTitle`, and `pocketAuthor` to something generic like `aboveChannelText`, `belowChannelHeader`, `belowChannelText` (or some better version of all of that).  Then we can give `ChannelTile` `String?` fields to match, and it simply sets if present, or hides if null.

I feel really strongly that the presentation layer should not be conditionally styling based on the type of tile.  This approach would also give us a lot more flexibility to mix and match these options for future content.

PR #2341, opened 2019-05-25
2195 killswitch layout

Baron-Severin:
### Pull Request checklist

Changelog entry is in the follow up story

### DIFF
--- a/app/src/androidTest/java/org/mozilla/tv/firefox/ui/screenshots/PinTileTests.java
+++ b/app/src/androidTest/java/org/mozilla/tv/firefox/ui/screenshots/PinTileTests.java
@@ -23,7 +23,10 @@ import org.junit.Rule;
 import org.junit.Test;
 import org.mozilla.tv.firefox.MainActivity;
 import org.mozilla.tv.firefox.R;
+import org.mozilla.tv.firefox.components.locale.LocaleManager;
 import org.mozilla.tv.firefox.helpers.MainActivityTestRule;
+
+import java.util.Locale;
 
 import tools.fastlane.screengrab.Screengrab;
 import tools.fastlane.screengrab.locale.LocaleTestRule;
@@ -61,8 +64,21 @@ public class PinTileTests extends ScreenshotTest {
     public void unpinTileFromContextMenu() {
         onView(allOf(withId(R.id.navUrlInput), isDisplayed(), hasFocus()));
 
-        mDevice.pressDPadDown();
-        mDevice.pressDPadDown();
+        Locale currentLocale = LocaleManager.getInstance()
+                .getCurrentLocale(mActivityTestRule.getActivity());
+
+        boolean pocketIsDisplayed = currentLocale.getLanguage().equals("en");
+        if (pocketIsDisplayed) {
+            // Navigate down to pinned tiles, to ensure that they are on screen. This
+            // is important because some of our tests are run on small devices, where
+            // not doing this can cause errors
+            mDevice.pressDPadDown();
+            mDevice.pressDPadDown();
+        } else {
+            // The pocket channel is only displayed in EN locales, so elsewhere we only
+            // need to nav down once.
+            mDevice.pressDPadDown();
+        }
 
         onView(withText(R.string.homescreen_unpin_tutorial_toast))
                 .inRoot(withDecorView(not(is(mActivityTestRule.getActivity().getWindow().getDecorView()))))
--- a/app/src/androidTest/java/org/mozilla/tv/firefox/ui/screenshots/PinTileTests.java
+++ b/app/src/androidTest/java/org/mozilla/tv/firefox/ui/screenshots/PinTileTests.java
@@ -64,10 +64,12 @@ public class PinTileTests extends ScreenshotTest {
     public void unpinTileFromContextMenu() {
         onView(allOf(withId(R.id.navUrlInput), isDisplayed(), hasFocus()));
 
-        Locale currentLocale = LocaleManager.getInstance()
+        final Locale currentLocale = LocaleManager.getInstance()
                 .getCurrentLocale(mActivityTestRule.getActivity());
 
-        boolean pocketIsDisplayed = currentLocale.getLanguage().equals("en");
+        final boolean pocketIsDisplayed = currentLocale.getLanguage().equals("en");
+        // TODO it would be better to condition this on view visibility, but I'm not sure
+        // how to do that here.
         if (pocketIsDisplayed) {
             // Navigate down to pinned tiles, to ensure that they are on screen. This
             // is important because some of our tests are run on small devices, where
--- a/app/src/main/java/org/mozilla/tv/firefox/architecture/KillswitchLayout.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/architecture/KillswitchLayout.kt
@@ -41,6 +41,14 @@ class KillswitchLayout : FrameLayout {
     @VisibleForTesting(otherwise = VisibleForTesting.NONE)
     var localeManager = LocaleManager.getInstance()
 
+    /**
+     * Allows visibility to be set if the language and country of the current locale
+     * matches any one of the allowed locales.
+     *
+     * An allowed locale with no country set (represented by an empty string, see
+     * javadoc on [Locale.getCountry]) will match with any country, so long as the
+     * language is correct.
+     */
     fun setRequirements(isAllowedByCurrentExperiment: Boolean, allowedInLocales: KillswitchLocales) {
         this.isAllowedByCurrentExperiment = isAllowedByCurrentExperiment
         this.allowedInLocales = allowedInLocales
@@ -86,14 +94,6 @@ class KillswitchLayout : FrameLayout {
     }
 }
 
-/**
- * Returns true if the language and country of the current locale matches any
- * one of the allowed locales.
- *
- * An allowed locale with no country set (represented by an empty string, see
- * javadoc on [Locale.getCountry]) will match with any country, so long as the
- * language is correct.
- */
 private fun Locale.languageAndMaybeCountryMatch(allowedLocales: Array<out Locale>?): Boolean {
     allowedLocales ?: return false
     return allowedLocales.any { allowed ->
--- a/app/src/main/java/org/mozilla/tv/firefox/architecture/KillswitchLayout.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/architecture/KillswitchLayout.kt
@@ -7,7 +7,7 @@ package org.mozilla.tv.firefox.architecture
 import android.content.Context
 import android.util.AttributeSet
 import android.view.View
-import android.widget.LinearLayout
+import android.widget.FrameLayout
 import androidx.annotation.VisibleForTesting
 import org.mozilla.tv.firefox.components.locale.LocaleManager
 import java.util.Locale
@@ -23,10 +23,10 @@ sealed class KillswitchLocales {
  *
  * This is done to prevent bugs similar to #2133. In that bug, a view was set to visibility
  * GONE when disabled, however its animation toggled that visibility back to VISIBLE.
- * [KillswitchLinearLayout] allows us to be sure that something will not be displayed if
+ * [KillswitchLayout] allows us to be sure that something will not be displayed if
  * requirements are not set by intercepting [setVisibility] calls.
  */
-class KillswitchLinearLayout : LinearLayout {
+class KillswitchLayout : FrameLayout {
     constructor(context: Context) : this(context, null)
     constructor(context: Context, attrs: AttributeSet?) : this(context, attrs, 0)
     constructor(context: Context, attrs: AttributeSet?, defStyleAttr: Int) : super(context, attrs, defStyleAttr)
@@ -57,7 +57,7 @@ class KillswitchLinearLayout : LinearLayout {
      */
     override fun setVisibility(visibility: Int) {
         this.desiredVisibility = visibility
-        val context = context?: return
+        val context = context ?: return
         val isAllowedByCurrentExperiment = this.isAllowedByCurrentExperiment
         val allowedInLocales = this.allowedInLocales
 
--- a/app/src/main/java/org/mozilla/tv/firefox/architecture/KillswitchLinearLayout.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/architecture/KillswitchLinearLayout.kt
@@ -8,6 +8,7 @@ import android.content.Context
 import android.util.AttributeSet
 import android.view.View
 import android.widget.LinearLayout
+import androidx.annotation.VisibleForTesting
 import org.mozilla.tv.firefox.components.locale.LocaleManager
 import java.util.Locale
 
@@ -30,10 +31,15 @@ class KillswitchLinearLayout : LinearLayout {
     constructor(context: Context, attrs: AttributeSet?) : this(context, attrs, 0)
     constructor(context: Context, attrs: AttributeSet?, defStyleAttr: Int) : super(context, attrs, defStyleAttr)
 
+    // This is the most recent visibility that client code attempted to set on
+    // this view, whether or not it was actually applied
     private var desiredVisibility = this.visibility
 
     private var isAllowedByCurrentExperiment: Boolean? = null
     private var allowedInLocales: KillswitchLocales? = null
+
+    @VisibleForTesting(otherwise = VisibleForTesting.NONE)
+    var localeManager = LocaleManager.getInstance()
 
     fun setRequirements(isAllowedByCurrentExperiment: Boolean, allowedInLocales: KillswitchLocales) {
         this.isAllowedByCurrentExperiment = isAllowedByCurrentExperiment
@@ -67,7 +73,7 @@ class KillswitchLinearLayout : LinearLayout {
 
         val allAllowed = allowedInLocales == KillswitchLocales.All
         val allowedLocales = (allowedInLocales as? KillswitchLocales.ActiveIn)?.locales
-        val currentLocale = LocaleManager.getInstance().getCurrentLocale(context)
+        val currentLocale = localeManager.getCurrentLocale(context)
 
         val currentLocaleIsAllowed = (allowedLocales != null &&
             allowedLocales.any { allowed ->
--- a/app/src/main/java/org/mozilla/tv/firefox/architecture/KillswitchLinearLayout.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/architecture/KillswitchLinearLayout.kt
@@ -1,0 +1,82 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.tv.firefox.architecture
+
+import android.content.Context
+import android.util.AttributeSet
+import android.view.View
+import android.widget.LinearLayout
+import org.mozilla.tv.firefox.components.locale.LocaleManager
+import java.util.Locale
+
+sealed class KillswitchLocales {
+    object All : KillswitchLocales()
+    class ActiveIn(vararg val locales: Locale) : KillswitchLocales()
+}
+
+/**
+ * Can be used to wrap some other [View] in order to prevent it from ever being shown if
+ * certain locale and experiment requirements are not met.
+ *
+ * This is done to prevent bugs similar to #2133. In that bug, a view was set to visibility
+ * GONE when disabled, however its animation toggled that visibility back to VISIBLE.
+ * [KillswitchLinearLayout] allows us to be sure that something will not be displayed if
+ * requirements are not set by intercepting [setVisibility] calls.
+ */
+class KillswitchLinearLayout : LinearLayout {
+    constructor(context: Context) : this(context, null)
+    constructor(context: Context, attrs: AttributeSet?) : this(context, attrs, 0)
+    constructor(context: Context, attrs: AttributeSet?, defStyleAttr: Int) : super(context, attrs, defStyleAttr)
+
+    private var desiredVisibility = this.visibility
+
+    private var isAllowedByCurrentExperiment: Boolean? = null
+    private var allowedInLocales: KillswitchLocales? = null
+
+    fun setRequirements(isAllowedByCurrentExperiment: Boolean, allowedInLocales: KillswitchLocales) {
+        this.isAllowedByCurrentExperiment = isAllowedByCurrentExperiment
+        this.allowedInLocales = allowedInLocales
+        visibility = desiredVisibility
+    }
+
+    /**
+     * Sets the visibility to gone if any of the following are true:
+     * 1) Locale/Experiment requirements have not been set
+     * 2) Current locale is not allowed
+     * 3) Client is not in a required experiment
+     *
+     * Otherwise, sets visibility as usual
+     */
+    override fun setVisibility(visibility: Int) {
+        this.desiredVisibility = visibility
+        val context = context
+        val isAllowedByCurrentExperiment = this.isAllowedByCurrentExperiment
+        val allowedInLocales = this.allowedInLocales
+
+        if (context == null) return
+
+        if (isAllowedByCurrentExperiment == null || allowedInLocales == null) {
+            return super.setVisibility(View.GONE)
+        }
+
+        if (!isAllowedByCurrentExperiment) {
+            return super.setVisibility(View.GONE)
+        }
+
+        val currentLocale = LocaleManager.getInstance().getCurrentLocale(context)
+        val allAllowed = allowedInLocales == KillswitchLocales.All
+        val allowedLocales = (allowedInLocales as? KillswitchLocales.ActiveIn)?.locales
+        val currentLocaleTag = currentLocale?.toLanguageTag()
+
+        val currentLocaleIsAllowed = (allowedLocales != null && currentLocaleTag != null &&
+            allowedLocales.contains(currentLocaleTag)) //TODO fix comparison
+
+        if (!allAllowed && !currentLocaleIsAllowed) {
+            return super.setVisibility(View.GONE)
+        }
+
+        super.setVisibility(visibility)
+    }
+}
--- a/app/src/main/java/org/mozilla/tv/firefox/architecture/KillswitchLinearLayout.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/architecture/KillswitchLinearLayout.kt
@@ -57,11 +57,9 @@ class KillswitchLinearLayout : LinearLayout {
      */
     override fun setVisibility(visibility: Int) {
         this.desiredVisibility = visibility
-        val context = context
+        val context = context?: return
         val isAllowedByCurrentExperiment = this.isAllowedByCurrentExperiment
         val allowedInLocales = this.allowedInLocales
-
-        if (context == null) return
 
         if (isAllowedByCurrentExperiment == null || allowedInLocales == null) {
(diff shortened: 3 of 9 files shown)


app/src/main/java/org/mozilla/tv/firefox/components/locale/LocaleManager.java
```diff
@@ -272,38 +271,38 @@ private void persistLocale(Context context, String localeCode) {
...
+    @NonNull
     public Locale getCurrentLocale(@NonNull Context context) {
         if (currentLocale != null) {
             return currentLocale;
         }
 
-        final String current = getPersistedLocale(context);
-        if (current == null) {
-            return null;
+        String current = getPersistedLocale(context);
```
mcomella: nit: final

app/src/main/java/org/mozilla/tv/firefox/components/locale/LocaleManager.java
```diff
@@ -272,38 +271,38 @@ private void persistLocale(Context context, String localeCode) {
...
+        if (current != null) {
+            currentLocale = Locales.parseLocaleCode(current);
```
mcomella: Thinking out loud: I'm assuming we're using Java `Locales` instead of Android resources directly so that we can support locales outside of what Android supports. However, in the short term, we don't intend to support these or an in-app locale switcher.

Should we just do away with the support for this legacy set-up? Notably, it doesn't look like Fenix uses these legacy classes either. Excepting minor cases like these, I guess it isn't bothering us so maybe it's not worth the time to remove?

app/src/main/java/org/mozilla/tv/firefox/components/locale/LocaleManager.java
```diff
@@ -272,38 +271,38 @@ private void persistLocale(Context context, String localeCode) {
...
         }
-        String language = Locales.getLanguage(current);
-        return language.toLowerCase(current).startsWith("en");
+
+        return currentLocale;
+    }
+
+    @NonNull
+    public Boolean currentLanguageIsEnglish(@NonNull Context context) {
+        return getCurrentLocale(context).getLanguage().equals("en");
```
mcomella: I think we want to be enabling for `en-US` only, not all english locales: I don't think this is right. I sent out an email and we can follow a follow-up if it's wrong.

app/src/main/java/org/mozilla/tv/firefox/architecture/KillswitchLinearLayout.kt
```diff
@@ -65,13 +65,16 @@ class KillswitchLinearLayout : LinearLayout {
...
         val allAllowed = allowedInLocales == KillswitchLocales.All
         val allowedLocales = (allowedInLocales as? KillswitchLocales.ActiveIn)?.locales
-        val currentLocaleTag = currentLocale?.toLanguageTag()
+        val currentLocale = LocaleManager.getInstance().getCurrentLocale(context)
 
-        val currentLocaleIsAllowed = (allowedLocales != null && currentLocaleTag != null &&
-            allowedLocales.contains(currentLocaleTag)) //TODO fix comparison
+        val currentLocaleIsAllowed = (allowedLocales != null &&
+            allowedLocales.any { allowed ->
+                allowed.language == currentLocale.language &&
```
mcomella: nit: this logic is hard to read. Can we add an extension function to equate them? e.g. `Locale.equalsLanguageAndCountry()`

app/src/main/java/org/mozilla/tv/firefox/architecture/KillswitchLinearLayout.kt
```diff
@@ -65,13 +65,16 @@ class KillswitchLinearLayout : LinearLayout {
...
-        val currentLocale = LocaleManager.getInstance().getCurrentLocale(context)
         val allAllowed = allowedInLocales == KillswitchLocales.All
         val allowedLocales = (allowedInLocales as? KillswitchLocales.ActiveIn)?.locales
-        val currentLocaleTag = currentLocale?.toLanguageTag()
+        val currentLocale = LocaleManager.getInstance().getCurrentLocale(context)
 
-        val currentLocaleIsAllowed = (allowedLocales != null && currentLocaleTag != null &&
-            allowedLocales.contains(currentLocaleTag)) //TODO fix comparison
+        val currentLocaleIsAllowed = (allowedLocales != null &&
+            allowedLocales.any { allowed ->
```
mcomella: nit: does `allowedLocales?.any { ... }` work? Then we can compare directly to the boolean and it'd make it all I little simpler, imo.

app/src/main/java/org/mozilla/tv/firefox/architecture/KillswitchLinearLayout.kt
```diff
@@ -0,0 +1,82 @@
...
+ * GONE when disabled, however its animation toggled that visibility back to VISIBLE.
+ * [KillswitchLinearLayout] allows us to be sure that something will not be displayed if
+ * requirements are not set by intercepting [setVisibility] calls.
+ */
+class KillswitchLinearLayout : LinearLayout {
+    constructor(context: Context) : this(context, null)
+    constructor(context: Context, attrs: AttributeSet?) : this(context, attrs, 0)
+    constructor(context: Context, attrs: AttributeSet?, defStyleAttr: Int) : super(context, attrs, defStyleAttr)
+
+    private var desiredVisibility = this.visibility
```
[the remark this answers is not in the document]
Baron-Severin: Without holding on to we would have an edge case:
-client code sets visibility to VISIBLE
-> requirements have not yet been set
-> visibility is set to GONE
-client code sets passing requirements

Expected: visibility is VISIBLE
Actual: visibility is GONE

Conceptually, `desiredVisibility` is whatever client code has set, and `visibility` is what has actually happened.  I think it's a fairly clear distinction, but maybe it could use a new name?

app/src/main/java/org/mozilla/tv/firefox/architecture/KillswitchLinearLayout.kt
```diff
@@ -0,0 +1,82 @@
...
+
+    fun setRequirements(isAllowedByCurrentExperiment: Boolean, allowedInLocales: KillswitchLocales) {
+        this.isAllowedByCurrentExperiment = isAllowedByCurrentExperiment
+        this.allowedInLocales = allowedInLocales
+        visibility = desiredVisibility
+    }
+
+    /**
+     * Sets the visibility to gone if any of the following are true:
+     * 1) Locale/Experiment requirements have not been set
```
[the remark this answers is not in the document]
Baron-Severin: Hm, I didn't want to enforce settings requirements and visibility in any particular order, so I didn't see that as an error.  It seems like this should get caught at dev time.  If requirements aren't set, the feature just won't ever show up.

app/src/main/java/org/mozilla/tv/firefox/components/locale/LocaleManager.java
Baron-Severin: Yeah, the `Locales` file is incredibly bloated and difficult to read.  I believe there were a number of edge cases found in Fennec, but it's entirely possible that at least some of them have been fixed by the framework by now.  I would love to get rid of this, but was trying to make narrower changes, rather than refactoring everything as I went.

But yes, I would very much like to get rid of that file.

app/src/main/java/org/mozilla/tv/firefox/architecture/KillswitchLinearLayout.kt
```diff
@@ -0,0 +1,82 @@
...
+            return super.setVisibility(View.GONE)
+        }
+
+        val currentLocale = LocaleManager.getInstance().getCurrentLocale(context)
+        val allAllowed = allowedInLocales == KillswitchLocales.All
+        val allowedLocales = (allowedInLocales as? KillswitchLocales.ActiveIn)?.locales
+        val currentLocaleTag = currentLocale?.toLanguageTag()
+
+        val currentLocaleIsAllowed = (allowedLocales != null && currentLocaleTag != null &&
+            allowedLocales.contains(currentLocaleTag)) //TODO fix comparison
```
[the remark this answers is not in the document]
Baron-Severin: Addressed in a later commit

app/src/main/java/org/mozilla/tv/firefox/channels/ChannelConfig.kt
```diff
@@ -16,19 +18,25 @@ data class ChannelConfig(
...
 ) {
     companion object {
         fun getPocketConfig(): ChannelConfig = ChannelConfig(
                 onClickTelemetry = { tile -> TELEMETRY.pocketVideoClickEvent(tile.id) },
                 // TODO focus telemetry should probably only be sent on focus gain, but this is
                 //  how our previous implementation worked. Keeping this to maintain data consistency
-                onFocusTelemetry = { tile, _ -> TELEMETRY.pocketVideoImpressionEvent(tile.id) }
+                onFocusTelemetry = { tile, _ -> TELEMETRY.pocketVideoImpressionEvent(tile.id) },
+                isEnabledInCurrentExperiment = true,
+                enabledInLocales = KillswitchLocales.ActiveIn(Locale.US)
```
mcomella: I think there's a redundant check here:
- `currentLanguageIsEnglish` is still used in the code as a dependency to Pocket code
- We are enabling a killswitch here.

As I note elsewhere, Pocket should be enabled for all english locales, not just Locale.US

app/src/main/java/org/mozilla/tv/firefox/components/locale/LocaleManager.java
mcomella: According to the email, we want:
- Pocket in all English locales
- New home channels in en-US only

We'll need to figure that out.

app/src/main/java/org/mozilla/tv/firefox/architecture/KillswitchLinearLayout.kt
```diff
@@ -0,0 +1,82 @@
...
+     */
+    override fun setVisibility(visibility: Int) {
+        this.desiredVisibility = visibility
+        val context = context
+        val isAllowedByCurrentExperiment = this.isAllowedByCurrentExperiment
+        val allowedInLocales = this.allowedInLocales
+
+        if (context == null) return
+
+        if (isAllowedByCurrentExperiment == null || allowedInLocales == null) {
```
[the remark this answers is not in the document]
Baron-Severin: Personal opinion, but I'm starting to dislike putting space-saving nullability stuff into multi-part boolean conditions.  They're valuable in super simple cases, but I feel like it's just hiding a logical branch here, and I think this function is already suffering from a death of a thousand simple cuts.  I'd rather it be a few lines longer, but be more explicit.

app/src/main/java/org/mozilla/tv/firefox/architecture/KillswitchLinearLayout.kt
```diff
@@ -0,0 +1,82 @@
...
+
+    private var desiredVisibility = this.visibility
+
+    private var isAllowedByCurrentExperiment: Boolean? = null
+    private var allowedInLocales: KillswitchLocales? = null
+
+    fun setRequirements(isAllowedByCurrentExperiment: Boolean, allowedInLocales: KillswitchLocales) {
+        this.isAllowedByCurrentExperiment = isAllowedByCurrentExperiment
+        this.allowedInLocales = allowedInLocales
+        visibility = desiredVisibility
```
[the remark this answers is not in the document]
Baron-Severin: Added a comment to `desiredVisibility` explaining its purpose.

app/src/main/java/org/mozilla/tv/firefox/architecture/KillswitchLinearLayout.kt
```diff
@@ -0,0 +1,82 @@
...
+
+    /**
+     * Sets the visibility to gone if any of the following are true:
+     * 1) Locale/Experiment requirements have not been set
+     * 2) Current locale is not allowed
+     * 3) Client is not in a required experiment
+     *
+     * Otherwise, sets visibility as usual
+     */
+    override fun setVisibility(visibility: Int) {
```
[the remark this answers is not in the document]
Baron-Severin: I wrote a number of tests for this fun, but I don't understand the second part of your comment.  Can we talk about it in person?

app/src/main/java/org/mozilla/tv/firefox/components/locale/LocaleManager.java
Baron-Severin: Thank you for clearing that up.  Should be an easy fix here.

app/src/main/java/org/mozilla/tv/firefox/channels/ChannelConfig.kt
Baron-Severin: Hm.  We will need to plan how to move `currentLanguageIsEnglish` from other code.  It was added to avoid sending Pocket requests that would never be shown to the user.  I think reworking that is out of scope for this story.

app/src/test/java/org/mozilla/tv/firefox/architecture/KillswitchLinearLayoutTest.kt
```diff
@@ -0,0 +1,124 @@
+package org.mozilla.tv.firefox.architecture
```
mcomella: nit: license

app/src/main/java/org/mozilla/tv/firefox/architecture/KillswitchLinearLayout.kt
```diff
@@ -30,11 +31,16 @@ class KillswitchLinearLayout : LinearLayout {
...
 
+    // This is the most recent visibility that client code attempted to set on
+    // this view, whether or not it was actually applied
     private var desiredVisibility = this.visibility
 
     private var isAllowedByCurrentExperiment: Boolean? = null
     private var allowedInLocales: KillswitchLocales? = null
 
+    @VisibleForTesting(otherwise = VisibleForTesting.NONE)
+    var localeManager = LocaleManager.getInstance()
```
mcomella: nit: I think it'd be more correct & less fragile if this came from the service locator but that might be too much boilerplate.

app/src/test/java/org/mozilla/tv/firefox/architecture/KillswitchLinearLayoutTest.kt
```diff
@@ -0,0 +1,124 @@
...
+    @Before
+    fun setup() {
+        val activityController = Robolectric.buildActivity(Activity::class.java)
+        killswitchLayout = KillswitchLinearLayout(activityController.get())
+    }
+
+    @Test
+    fun `GIVEN requirements are not set WHEN visibility is set THEN visibility should always be gone`() {
+        // Don't set requirements
+        visibilityIsAlwaysGone()
```
mcomella: nit: `assertVisibilityIsAlwaysGone`. Without following the naming convention, I have to double-check the definition to ensure it's doing what I expect.

app/src/test/java/org/mozilla/tv/firefox/architecture/KillswitchLinearLayoutTest.kt
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    @Test
+    fun `GIVEN all locales are allowed WHEN visibility is set THEN it should be applied`() {
+        killswitchLayout.setRequirements(true, KillswitchLocales.All)
+
+        visibilityIsApplied()
```
mcomella: nit: `assert...`

app/src/test/java/org/mozilla/tv/firefox/architecture/KillswitchLinearLayoutTest.kt
```diff
@@ -0,0 +1,124 @@
...
+
+    @Test
+    fun `GIVEN visibility is set WHEN requirements are set THEN visibility should match the earlier set values`() {
+        killswitchLayout.visibility = View.VISIBLE
+        assertEquals(View.GONE, killswitchLayout.visibility)
+
+        killswitchLayout.setRequirements(true, KillswitchLocales.All)
+        assertEquals(View.VISIBLE, killswitchLayout.visibility)
+
+        visibilityIsApplied()
```
mcomella: nit: I think it's good to test this last assertion too but it's conceptually a different test than the one that's described in the test name but with the same setup. However, I've never had a problem with this style of unit tests in practice so it seems fine.

app/src/test/java/org/mozilla/tv/firefox/architecture/KillswitchLinearLayoutTest.kt
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    @Test
+    fun `GIVEN locale and country are on the accepted list WHEN visibility is set THEN it should be applied`() {
+        val mockLocaleManager = mockk<LocaleManager>()
+        every { mockLocaleManager.getCurrentLocale(any()) } answers { Locale("es", "mx") }
```
mcomella: Locales also include scripts and variants: do we want to sanity check those states too?

app/src/test/java/org/mozilla/tv/firefox/architecture/KillswitchLinearLayoutTest.kt
```diff
@@ -0,0 +1,124 @@
...
+        killswitchLayout.setRequirements(true, KillswitchLocales.ActiveIn(
+            Locale("es"),
+            Locale("en")
+        ))
+
+        visibilityIsApplied()
+    }
+
+    private fun visibilityIsAlwaysGone() {
+        killswitchLayout.visibility = View.VISIBLE
```
mcomella: nit: I think you'll disagree :P but I find these blocks easier to read when written with a loop:
```kotlin
val visibilitiesToCheck = arrayOf(View.VISIBLE, View.INVISIBLE, View.GONE, View.VISIBLE)
visibilitiesToCheck.forEach { visibilityToCheck ->
    killswitchLayout.visibility = visibilityToCheck
    assertEquals(View.GONE, killswitchLayout.visibility)
}
```

To me, I can quickly skim to see it says, "It asserts View.GONE for all these visibilities" rather than having to read every line because the actual code may not follow my assumptions or may be written incorrectly.

app/src/main/java/org/mozilla/tv/firefox/architecture/KillswitchLinearLayout.kt
```diff
@@ -71,21 +69,38 @@ class KillswitchLinearLayout : LinearLayout {
             return super.setVisibility(View.GONE)
         }
 
-        val allAllowed = allowedInLocales == KillswitchLocales.All
+        if (allowedInLocales == KillswitchLocales.All) {
+            return super.setVisibility(visibility)
+        }
+
+
         val allowedLocales = (allowedInLocales as? KillswitchLocales.ActiveIn)?.locales
```
mcomella: We have no tests for this null code path: is there a way we can test this so we can guarantee this code doesn't unintentionally break in the future?

app/src/main/java/org/mozilla/tv/firefox/architecture/KillswitchLinearLayout.kt
```diff
@@ -71,21 +69,38 @@ class KillswitchLinearLayout : LinearLayout {
...
 
         super.setVisibility(visibility)
     }
 }
+
+/**
+ * Returns true if the language and country of the current locale matches any
+ * one of the allowed locales.
+ *
+ * An allowed locale with no country set (represented by an empty string, see
```
mcomella: Can we document this behavior on a non-private method for consumers of the layout? Maybe `setRequirements`?

app/src/main/java/org/mozilla/tv/firefox/channels/ChannelConfig.kt
```diff
@@ -29,7 +29,8 @@ data class ChannelConfig(
                 //  how our previous implementation worked. Keeping this to maintain data consistency
                 onFocusTelemetry = { tile, _ -> TELEMETRY.pocketVideoImpressionEvent(tile.id) },
                 isEnabledInCurrentExperiment = true,
-                enabledInLocales = KillswitchLocales.ActiveIn(Locale.US)
+                // Pocket is enabled in all countries, for any English locale
+                enabledInLocales = KillswitchLocales.ActiveIn(Locale("en", ""))
```
mcomella: nit: Use Locale.ENGLISH instead of hard-coding the string?

app/src/androidTest/java/org/mozilla/tv/firefox/ui/screenshots/PinTileTests.java
```diff
@@ -63,8 +66,21 @@ public void tearDown() {
     public void unpinTileFromContextMenu() {
         onView(allOf(withId(R.id.navUrlInput), isDisplayed(), hasFocus()));
 
-        mDevice.pressDPadDown();
-        mDevice.pressDPadDown();
+        Locale currentLocale = LocaleManager.getInstance()
```
mcomella: nit: final

app/src/androidTest/java/org/mozilla/tv/firefox/ui/screenshots/PinTileTests.java
```diff
@@ -63,8 +66,21 @@ public void tearDown() {
...
+                .getCurrentLocale(mActivityTestRule.getActivity());
+
+        boolean pocketIsDisplayed = currentLocale.getLanguage().equals("en");
```
mcomella: nit: instead of branching on locale (which duplicates logic), maybe we should branch based on whether or not the view is visible. Probably not worth your time: add a comment for when the test needs to change next, maybe?

app/src/main/res/layout/default_channel.xml
```diff
@@ -1,26 +1,32 @@
...
+            android:layout_marginEnd="@dimen/overlay_margin_end"
+            android:layout_marginTop="24dp"
+            android:layout_marginBottom="16dp"
+            android:fontFamily="@string/font_ember_regular"
+            android:textSize="24sp"
+            android:text="@string/menu_settings" />
 
     <!-- Note that, for any items in this RV, parent layouts should be FocusLossFrameLayout.
     This corrects a bug that is described in the kdoc of that class -->
     <androidx.leanback.widget.ListRowView
```
mcomella: nit: this needs to get indented.

app/src/main/java/org/mozilla/tv/firefox/architecture/KillswitchLayout.kt
```diff
@@ -23,10 +23,10 @@ sealed class KillswitchLocales {
  *
  * This is done to prevent bugs similar to #2133. In that bug, a view was set to visibility
  * GONE when disabled, however its animation toggled that visibility back to VISIBLE.
- * [KillswitchLinearLayout] allows us to be sure that something will not be displayed if
+ * [KillswitchLayout] allows us to be sure that something will not be displayed if
  * requirements are not set by intercepting [setVisibility] calls.
  */
-class KillswitchLinearLayout : LinearLayout {
+class KillswitchLayout : FrameLayout {
```
mcomella: nit: I wonder if making this a ConstraintLayout would be more useful, though it is more prescriptive so maybe not.

app/src/main/java/org/mozilla/tv/firefox/channels/ChannelConfig.kt
```diff
@@ -16,19 +18,26 @@ data class ChannelConfig(
...
     companion object {
         fun getPocketConfig(): ChannelConfig = ChannelConfig(
                 onClickTelemetry = { tile -> TELEMETRY.pocketVideoClickEvent(tile.id) },
                 // TODO focus telemetry should probably only be sent on focus gain, but this is
                 //  how our previous implementation worked. Keeping this to maintain data consistency
-                onFocusTelemetry = { tile, _ -> TELEMETRY.pocketVideoImpressionEvent(tile.id) }
+                onFocusTelemetry = { tile, _ -> TELEMETRY.pocketVideoImpressionEvent(tile.id) },
+                isEnabledInCurrentExperiment = true,
+                // Pocket is enabled in all countries, for any English locale
+                enabledInLocales = KillswitchLocales.ActiveIn(Locale("en", ""))
```
mcomella: I wonder if we should send telemetry describing which features are available to users rather than the data team having to piece it together (and thus duplicate logic) from our "which experiments is the user in" and "which locale is the user in" telemetry. Maybe file a bug to investigate?

app/src/main/java/org/mozilla/tv/firefox/architecture/KillswitchLinearLayout.kt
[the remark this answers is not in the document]
mcomella: The comment is helpful but I wish we could better encapsulate that in the name. I can't think of a better one but maybe something jargony for a pending operation like transactions? `pendingVisibility`?

app/src/main/java/org/mozilla/tv/firefox/architecture/KillswitchLinearLayout.kt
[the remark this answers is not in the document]
mcomella: fwiw, if we did enforce a contract on how this View is used, the View has fewer states, making the code simpler to understand, decreasing surface area for bugs. While it's not a lot of code right now so it's fairly negligible, it could be a bigger deal if the View adds more states later.

I don't have context on how this View is intended to be used to know what my preference is but I just wanted to share some more trade-offs I had thought of.

app/src/main/java/org/mozilla/tv/firefox/channels/ChannelConfig.kt
mcomella: Can you file a follow-up bug? If it needs to get done this sprint, please raise it.

app/src/main/java/org/mozilla/tv/firefox/architecture/KillswitchLinearLayout.kt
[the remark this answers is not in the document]
Baron-Severin: Coming back to this, I'm pretty sure it would crash whenever visibility was set in XML, since that would be applied right after inflation, and we don't have a chance to set requirements until later.

app/src/main/java/org/mozilla/tv/firefox/channels/ChannelConfig.kt
Baron-Severin: No, it works as is, but it's duplicated logic and a landmine for future changes.  We definitely don't need to block on it, which I think would make that issue an aspirational refactor.  I can add a TODO in code though.

app/src/test/java/org/mozilla/tv/firefox/architecture/KillswitchLinearLayoutTest.kt
Baron-Severin: That makes sense to me here, actually.  I'll make the change.

app/src/androidTest/java/org/mozilla/tv/firefox/ui/screenshots/PinTileTests.java
Baron-Severin: I'm sure it's possible, but I don't know how to do that in Espresso.  I'll add a comment.

app/src/main/java/org/mozilla/tv/firefox/architecture/KillswitchLayout.kt
Baron-Severin: I think `FrameLayout` is a good approach because it has minimal overhead, and consumers can use whatever layout they want inside it.

app/src/main/java/org/mozilla/tv/firefox/architecture/KillswitchLinearLayout.kt
Baron-Severin: Agreed, but it'd also be a fairly sizable change, which I'm trying to pull back from doing all the time.

PR #2368, opened 2019-06-03
Closes #2367: fix Robolectric.buildActivity

### LINKED ISSUE #2367: Fix initialization errors when calling `buildActivity(MainActivity` in robolectric
Currently, we have two platforms for testing: off-device tests and on-device tests. Unfortunately, we're forced to run all integration tests on-device – which take a long time to write, take a long time to run, and are hard to maintain. We can't write off-device tests for them because our MainActivity fails to initialize when using robolectric. The code that fails:
```kotlin
@RunWith(RobolectricTestRunner::class)
class MainActivityTest {

    private lateinit var controller: ActivityController<MainActivity>

    @Before
    fun setUp() {
        controller = Robolectric.buildActivity(MainActivity::class.java)
            .create()
            .start()
    }

    @Test
    fun `WHEN onStart is called THEN `() {
        controller.get()
    }
}
```

And the error it fails with:
```
java.lang.UnsupportedOperationException
	at android.webkit.WebViewFactory.getProvider(WebViewFactory.java:241)
	at android.webkit.WebSettings.getDefaultUserAgent(WebSettings.java:1266)
	at mozilla.components.browser.engine.system.SystemEngine$Companion.initDefaultUserAgent(SystemEngine.kt:111)
	at mozilla.components.browser.engine.system.SystemEngine$Companion.access$initDefaultUserAgent(SystemEngine.kt:103)
	at mozilla.components.browser.engine.system.SystemEngine.<init>(SystemEngine.kt:30)
	at org.mozilla.tv.firefox.webrender.WebRenderComponents$engine$2.invoke(WebRenderComponents.kt:35)
	at org.mozilla.tv.firefox.webrender.WebRenderComponents$engine$2.invoke(WebRenderComponents.kt:22)
	at kotlin.SynchronizedLazyImpl.getValue(LazyJVM.kt:74)
	at org.mozilla.tv.firefox.webrender.WebRenderComponents.getEngine(WebRenderComponents.kt)
```

The problem is that `WebSettings` is not mocked by robolectric so we can't call into it.

However, **I discovered an easy fix.** If we override `SystemEngine.defaultUserAgent` before creating the activity, the tests pass. This bug is to utilize this quick fix.

### Impact
Devs will no longer be blocked when trying to write tests `MainActivity` or other integration tests. This will allow us to cover more of our code with an easy and quick (to write and run) style of testing.

For example, had this been working in #2325, we could have replaced a manual TestRail test case we added with a few unit tests that would have taken only a few hours (or less) to write, saving much more time from QA in the long run.

### Acceptance criteria
- [x] `Robolectric.buildActivity` can be used to write tests

mcomella:
### Pull Request checklist

### DIFF
--- a/Screengrabfile
+++ b/Screengrabfile
@@ -26,4 +26,4 @@ exit_on_test_failure false
 skip_open_summary false
 reinstall_app true
 
-test_instrumentation_runner 'org.mozilla.tv.firefox.FirefoxTestRunner'
+test_instrumentation_runner 'org.mozilla.tv.firefox.FirefoxOnDeviceTestRunner'
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -26,7 +26,7 @@ android {
         versionCode 11 // This versionCode is "frozen" for local builds. For "release" builds we
         // override this with a generated versionCode at build time.
         versionName "4.2-LAT1"
-        testInstrumentationRunner "org.mozilla.tv.firefox.FirefoxTestRunner"
+        testInstrumentationRunner "org.mozilla.tv.firefox.FirefoxOnDeviceTestRunner"
         testInstrumentationRunnerArguments clearPackageData: 'true'
 
         multiDexEnabled true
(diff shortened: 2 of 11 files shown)


mcomella:
I tried rebasing but the debug overlay including the engine view version now causes the tests to fail: I'll have to figure out why that is.

mcomella:
@Baron-Severin FYI: I pushed a new commit on top to fix the rebase errors. I'm not too concerned about it because it follows an existing pattern.

Issue #2325, opened 2019-05-23
Follow-up to Pocket background fetch


dnarcese:
Follow-up tasks to #2223 
### Vision statement / What / Requirements
The tasks that were not finished in the original PR:

- Remove remaining old Pocket foreground fetch code
  - [ ] Rename PocketEndpoint?
  - [ ] Remove PocketRepoCache and comment why it's not necessary (b/c we only content in onStart)
  - [ ]  Remove unused code
-  [ ] Clean up comments in scheduler

- Automated tests
  -  [ ] No API key
  -  [ ]  Non en-US locales
  - [ ]  Failure to hit the server == retry in background job (retry delay should have some randomness)
  - [ ] Invalid json from server == just fail
  - [ ] Does Sentry catch crashes in PocketWorker?
  - [ ] What happens when exceptions are thrown inside the PocketWorker?

### Impact
Some unused code needs to be removed and this background fetch functionality needs to be tested.
### Acceptance criteria
All the above tasks are completed.

npark-mozilla:
Just my 2 cents:

So this is my understanding of the background fetch works (which could be wrong):

- Background fetch is scheduled typically between anytime during 3AM and 6AM

- When the app is started, it checks for the fetched video JSON, and schedules the next fetch time
- Initially there is a hardcoded video JSON (so there is always something to show on the overlay)
- Assuming it will always fetch the list of video from pocket server that’s different from what we currently have 
- We don’t have the record of the last successful fetch timestamp

If above is correct, it would be ideal if we have following grey-box tests that does the following, in addition to the usual unit tests. They cover pretty typical scenarios:

Checking whether update actually happens nominally:
- Opens app, checks the fetch is done, checks the next update schedule value is set.
- Change the next update schedule to currentTime + 1000ms
- wait to see whether pocketstore updates and refreshes overlay when reopened

Checking whether pocket updates past the specified time:
- Opens app, checks the fetch is done, checks the next update schedule value is set.
- Change the next update schedule to currentTime or past time (assuming this gets noticed by the app)
- Check whether the pocketstore updates (since the scheduled time is now in the past) and refreshes overlay when reopened

Checking whether app does not crash during network is down, and schedule reattempt 
- Opens app, checks the fetch is done, checks the next update schedule value is set.
- Disable wifi
- Change the next update schedule to currentTime + 1000ms
- wait too see the app does not crash after failed fetch attempt, and check backoff time is properly set

[...]

PR #2371, opened 2019-06-04
Issue #2325: Expose way for QA to change pocket fetch delay

dnarcese:
This allows QA to start the app through an intent with this command:
`adb shell am start -n org.mozilla.tv.firefox/org.mozilla.tv.firefox.MainActivity -a android.intent.action.MAIN --el qaFetchDelaySeconds 10`
where `qaFetchDelaySeconds` is in seconds.

The custom delay will only apply once. You need to run the command every time you want to use that custom delay.

### Pull Request checklist

### DIFF
--- a/app/src/main/java/org/mozilla/tv/firefox/IntentValidator.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/IntentValidator.kt
@@ -18,6 +18,7 @@ import org.mozilla.tv.firefox.telemetry.TelemetryIntegration
 import org.mozilla.tv.firefox.utils.UrlUtils
 
 private const val EXTRA_ACTIVE_EXPERIMENTS = "qaActiveExperiments"
+private const val EXTRA_FETCH_DELAY_KEY = "fetchDelay"
 
 data class ValidatedIntentData(val url: String, val source: Session.Source)
 
@@ -66,11 +67,11 @@ object IntentValidator {
                     TelemetryIntegration.INSTANCE.youtubeCastEvent()
                     return ValidatedIntentData(url = "https://www.youtube.com/tv?$dialParams", source = Session.Source.ACTION_VIEW)
                 }
-                // The fetch delay can be changed for testing purposes through an intent with custom flags
-                // isQA and fetchDelay. Where isQA is true/false and fetchDelay is the intended next fetch time
-                // in seconds.
-                else if (extras.keySet().contains(IS_QA_BUILD_KEY)) {
-                    PocketVideoFetchScheduler.setDelayForQA(extras.getLong(FETCH_DELAY_KEY))
+
+                // The fetch delay can be changed for testing purposes through an intent with custom flag
+                // fetchDelay. Where fetchDelay is the intended next fetch time in seconds.
+                else if (extras.containsKey(EXTRA_FETCH_DELAY_KEY)) {
+                    PocketVideoFetchScheduler.setDelayForQA(extras.getLong(EXTRA_FETCH_DELAY_KEY))
                 }
             }
             Intent.ACTION_VIEW -> {
--- a/app/src/main/java/org/mozilla/tv/firefox/IntentValidator.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/IntentValidator.kt
@@ -7,12 +7,13 @@ package org.mozilla.tv.firefox
 import android.content.Context
 import android.content.Intent
 import android.os.Bundle
-import androidx.annotation.VisibleForTesting
 import android.text.TextUtils
+import androidx.annotation.VisibleForTesting
 import mozilla.components.browser.session.Session
 import mozilla.components.service.fretboard.ExperimentDescriptor
 import mozilla.components.support.utils.SafeIntent
 import org.mozilla.tv.firefox.ext.serviceLocator
+import org.mozilla.tv.firefox.pocket.PocketVideoFetchScheduler
 import org.mozilla.tv.firefox.telemetry.TelemetryIntegration
 import org.mozilla.tv.firefox.utils.UrlUtils
 
@@ -58,10 +59,18 @@ object IntentValidator {
 
         when (intent.action) {
             Intent.ACTION_MAIN -> {
+                val extras = intent.extras ?: return null
+
                 val dialParams = intent.extras?.getString(DIAL_PARAMS_KEY) ?: return null
                 if (dialParams.isNotEmpty()) {
                     TelemetryIntegration.INSTANCE.youtubeCastEvent()
                     return ValidatedIntentData(url = "https://www.youtube.com/tv?$dialParams", source = Session.Source.ACTION_VIEW)
+                }
+                // The fetch delay can be changed for testing purposes through an intent with custom flags
+                // isQA and fetchDelay. Where isQA is true/false and fetchDelay is the intended next fetch time
+                // in seconds.
+                else if (extras.keySet().contains(IS_QA_BUILD_KEY)) {
+                    PocketVideoFetchScheduler.setDelayForQA(extras.getLong(FETCH_DELAY_KEY))
                 }
             }
             Intent.ACTION_VIEW -> {
--- a/app/src/main/java/org/mozilla/tv/firefox/IntentValidator.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/IntentValidator.kt
@@ -55,8 +55,8 @@ object IntentValidator {
     }
 
     fun validate(context: Context, intent: SafeIntent): ValidatedIntentData? {
-        setExperimentOverrides(intent, context)
-        changeFetchDelaySecondsForQA(intent, context)
+        setQAExperimentOverrides(intent, context)
+        setQAFetchDelayOverrides(intent, context)
 
         when (intent.action) {
             Intent.ACTION_MAIN -> {
@@ -88,7 +88,7 @@ object IntentValidator {
         return null
     }
 
-    private fun setExperimentOverrides(intent: SafeIntent, context: Context) {
+    private fun setQAExperimentOverrides(intent: SafeIntent, context: Context) {
         val experimentsArray = intent.extras?.getStringArray(EXTRA_ACTIVE_EXPERIMENTS) ?: return
         val fretboard = context.serviceLocator.fretboardProvider.fretboard
         fretboard.clearAllOverrides(context)
@@ -98,9 +98,9 @@ object IntentValidator {
         }
     }
 
-    private fun changeFetchDelaySecondsForQA(intent: SafeIntent, context: Context) {
+    private fun setQAFetchDelayOverrides(intent: SafeIntent, context: Context) {
         intent.extras?.getLong(EXTRA_FETCH_DELAY_KEY)?.let { delay ->
-            context.serviceLocator.pocketVideoFetchScheduler.setDelayForQA(delay)
+            context.serviceLocator.pocketVideoFetchScheduler.setQAFetchDelayOverride(delay)
         }
     }
 }
--- a/app/src/main/java/org/mozilla/tv/firefox/IntentValidator.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/IntentValidator.kt
@@ -13,6 +13,7 @@ import mozilla.components.browser.session.Session
 import mozilla.components.service.fretboard.ExperimentDescriptor
 import mozilla.components.support.utils.SafeIntent
 import org.mozilla.tv.firefox.ext.serviceLocator
+import org.mozilla.tv.firefox.pocket.PocketVideoFetchScheduler
 import org.mozilla.tv.firefox.telemetry.TelemetryIntegration
 import org.mozilla.tv.firefox.utils.UrlUtils
 
@@ -56,7 +57,7 @@ object IntentValidator {
 
     fun validate(context: Context, intent: SafeIntent): ValidatedIntentData? {
         setQAExperimentOverrides(intent, context)
-        setQAFetchDelayOverrides(intent, context)
+        setQAFetchDelayOverrides(intent, context.serviceLocator.pocketVideoFetchScheduler)
 
         when (intent.action) {
             Intent.ACTION_MAIN -> {
@@ -98,9 +99,9 @@ object IntentValidator {
         }
     }
 
-    private fun setQAFetchDelayOverrides(intent: SafeIntent, context: Context) {
+    private fun setQAFetchDelayOverrides(intent: SafeIntent, pocketVideoFetchScheduler: PocketVideoFetchScheduler) {
         intent.extras?.getLong(EXTRA_FETCH_DELAY_KEY)?.let { delay ->
-            context.serviceLocator.pocketVideoFetchScheduler.setQAFetchDelayOverride(delay)
+            pocketVideoFetchScheduler.setQAFetchDelayOverride(delay)
         }
     }
 }
--- a/app/src/main/java/org/mozilla/tv/firefox/IntentValidator.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/IntentValidator.kt
@@ -57,6 +57,7 @@ object IntentValidator {
 
     fun validate(context: Context, intent: SafeIntent): ValidatedIntentData? {
         setExperimentOverrides(intent, context)
+        changeFetchDelaySecondsForQA(intent)
 
         when (intent.action) {
             Intent.ACTION_MAIN -> {
@@ -66,12 +67,6 @@ object IntentValidator {
                 if (dialParams.isNotEmpty()) {
                     TelemetryIntegration.INSTANCE.youtubeCastEvent()
                     return ValidatedIntentData(url = "https://www.youtube.com/tv?$dialParams", source = Session.Source.ACTION_VIEW)
-                }
-
-                // The fetch delay can be changed for testing purposes through an intent with custom flag
-                // fetchDelay. Where fetchDelay is the intended next fetch time in seconds.
-                else if (extras.containsKey(EXTRA_FETCH_DELAY_KEY)) {
-                    PocketVideoFetchScheduler.setDelayForQA(extras.getLong(EXTRA_FETCH_DELAY_KEY))
                 }
             }
             Intent.ACTION_VIEW -> {
@@ -103,6 +98,12 @@ object IntentValidator {
 
         experimentsArray.forEach {
             fretboard.setOverride(context, ExperimentDescriptor(it), true)
+        }
+    }
+
+    private fun changeFetchDelaySecondsForQA(intent: SafeIntent) {
+        intent.extras?.getLong(EXTRA_FETCH_DELAY_KEY)?.let { delay ->
+            PocketVideoFetchScheduler.setDelayForQA(delay)
         }
     }
 }
--- a/app/src/main/java/org/mozilla/tv/firefox/IntentValidator.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/IntentValidator.kt
@@ -13,12 +13,11 @@ import mozilla.components.browser.session.Session
 import mozilla.components.service.fretboard.ExperimentDescriptor
 import mozilla.components.support.utils.SafeIntent
 import org.mozilla.tv.firefox.ext.serviceLocator
-import org.mozilla.tv.firefox.pocket.PocketVideoFetchScheduler
 import org.mozilla.tv.firefox.telemetry.TelemetryIntegration
 import org.mozilla.tv.firefox.utils.UrlUtils
 
 private const val EXTRA_ACTIVE_EXPERIMENTS = "qaActiveExperiments"
-private const val EXTRA_FETCH_DELAY_KEY = "fetchDelay"
+private const val EXTRA_FETCH_DELAY_KEY = "qaFetchDelaySeconds"
 
 data class ValidatedIntentData(val url: String, val source: Session.Source)
 
@@ -57,12 +56,10 @@ object IntentValidator {
 
     fun validate(context: Context, intent: SafeIntent): ValidatedIntentData? {
         setExperimentOverrides(intent, context)
-        changeFetchDelaySecondsForQA(intent)
+        changeFetchDelaySecondsForQA(intent, context)
 
         when (intent.action) {
             Intent.ACTION_MAIN -> {
-                val extras = intent.extras ?: return null
-
                 val dialParams = intent.extras?.getString(DIAL_PARAMS_KEY) ?: return null
                 if (dialParams.isNotEmpty()) {
                     TelemetryIntegration.INSTANCE.youtubeCastEvent()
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/mozilla/tv/firefox/IntentValidator.kt
```diff
@@ -30,6 +31,8 @@ data class ValidatedIntentData(val url: String, val source: Session.Source)
 object IntentValidator {
     @VisibleForTesting const val DIAL_PARAMS_KEY = "com.amazon.extra.DIAL_PARAM"
     private const val ACTIVE_EXPERIMENTS_KEY = "activeExperiments"
+    private const val IS_QA_BUILD_KEY = "isQA"
```
mcomella: nit: I prefer to declare private constants at the top level, which reduces the size of the object's block, making it clearer what is important in this class.

I think liuche was going to move the existing constant to the top level in a follow-up.

app/src/main/java/org/mozilla/tv/firefox/IntentValidator.kt
```diff
@@ -57,10 +60,18 @@ object IntentValidator {
 
         when (intent.action) {
             Intent.ACTION_MAIN -> {
-                val dialParams = intent.extras?.getString(DIAL_PARAMS_KEY) ?: return null
-                if (dialParams.isNotEmpty()) {
+                val extras = intent.extras ?: return null
+
+                if (extras.keySet().contains(DIAL_PARAMS_KEY)) {
```
mcomella: This is a change in behavior: `dialParams` is no longer checked for `isNotEmpty`. Given this is unrelated to your change, can we change it back to keep the functionality the same?

It seems like there should have been a test for this. :)

mcomella: nit: `extras.containsKey` is more efficient. However, since we sometimes need to fetch the parameter anyway, it's slightly more efficient and has less duplication to just fetch it and check for null. However, it's less explicit. I prefer to fetch directly but I could go either way.

app/src/main/java/org/mozilla/tv/firefox/IntentValidator.kt
```diff
@@ -57,10 +60,18 @@ object IntentValidator {
...
+                    val param = extras.getString(DIAL_PARAMS_KEY)
                     TelemetryIntegration.INSTANCE.youtubeCastEvent()
-                    return ValidatedIntentData(url = "https://www.youtube.com/tv?$dialParams", source = Session.Source.ACTION_VIEW)
+                    return ValidatedIntentData(url = "https://www.youtube.com/tv?$param", source = Session.Source.ACTION_VIEW)
+                }
+                // The fetch delay can be changed for testing purposes through an intent with custom flags
+                // isQA and fetchDelay. Where isQA is true/false and fetchDelay is the intended next fetch time
+                // in seconds.
+                else if (extras.keySet().contains(IS_QA_BUILD_KEY)) {
```
mcomella: This is not mutually exclusive with `DIAL_PARAMS_KEY`: does this have to be an `else if`? Since we may return from the block above, I think this should actually be before that block.

fwiw, there's a saying, "no else after return", that I never really understood. However, it's redundant to return in a previous block and then add, "else if": maybe that's why?

app/src/main/java/org/mozilla/tv/firefox/IntentValidator.kt
```diff
@@ -57,10 +60,18 @@ object IntentValidator {
...
-                if (dialParams.isNotEmpty()) {
+                val extras = intent.extras ?: return null
+
+                if (extras.keySet().contains(DIAL_PARAMS_KEY)) {
+                    val param = extras.getString(DIAL_PARAMS_KEY)
                     TelemetryIntegration.INSTANCE.youtubeCastEvent()
-                    return ValidatedIntentData(url = "https://www.youtube.com/tv?$dialParams", source = Session.Source.ACTION_VIEW)
+                    return ValidatedIntentData(url = "https://www.youtube.com/tv?$param", source = Session.Source.ACTION_VIEW)
+                }
+                // The fetch delay can be changed for testing purposes through an intent with custom flags
```
mcomella: nit: maybe we should pull this into a helper function. A general practice I follow is that if I'm writing a comment for code, I'll consider pulling it into a separate method. The reasons are:
- if the code is complicated enough to warrant a comment, it should probably be abstracted/isolated in a function
- the helper function name can often stand-in for a comment
- when comments are added, the code being commented is sometimes lower-level than the code around it and it's hard to read code that is at different levels of abstraction

app/src/main/java/org/mozilla/tv/firefox/IntentValidator.kt
```diff
@@ -57,10 +60,18 @@ object IntentValidator {
 
         when (intent.action) {
             Intent.ACTION_MAIN -> {
-                val dialParams = intent.extras?.getString(DIAL_PARAMS_KEY) ?: return null
-                if (dialParams.isNotEmpty()) {
+                val extras = intent.extras ?: return null
+
+                if (extras.keySet().contains(DIAL_PARAMS_KEY)) {
+                    val param = extras.getString(DIAL_PARAMS_KEY)
```
mcomella: nit: `dialParams` is clearer and it won't break the blame on the following line.

app/src/main/java/org/mozilla/tv/firefox/IntentValidator.kt
mcomella: nit: `containsKey(...)`

app/src/main/java/org/mozilla/tv/firefox/IntentValidator.kt
mcomella: Can you elaborate on the trade-offs of the isQA flag? It seems unnecessary to me: adding any single flag designed for QA is equivalent to adding the `isQA` flag, right?

app/src/main/java/org/mozilla/tv/firefox/IntentValidator.kt
```diff
@@ -30,6 +31,8 @@ data class ValidatedIntentData(val url: String, val source: Session.Source)
 object IntentValidator {
     @VisibleForTesting const val DIAL_PARAMS_KEY = "com.amazon.extra.DIAL_PARAM"
     private const val ACTIVE_EXPERIMENTS_KEY = "activeExperiments"
+    private const val IS_QA_BUILD_KEY = "isQA"
+    private const val FETCH_DELAY_KEY = "fetchDelay"
```
mcomella: nit: EXTRA_FETCH_DELAY or KEY_EXTRA_FETCH_DELAY -> it's helpful to prepend KEY or EXTRA so that it's really clear which constants are intended to be used where, if they're used in a consistent way.

mcomella: nit: I think the value, and the flag QA has to send, should be `qaFetchDelaySeconds`. I think prepending `qa` is important to tell users not to expect much from it and, if we keep `isQA`, to tell QA that that flag is needed to use this flag.

app/src/main/java/org/mozilla/tv/firefox/IntentValidator.kt
```diff
@@ -57,10 +60,18 @@ object IntentValidator {
...
+                    val param = extras.getString(DIAL_PARAMS_KEY)
                     TelemetryIntegration.INSTANCE.youtubeCastEvent()
-                    return ValidatedIntentData(url = "https://www.youtube.com/tv?$dialParams", source = Session.Source.ACTION_VIEW)
+                    return ValidatedIntentData(url = "https://www.youtube.com/tv?$param", source = Session.Source.ACTION_VIEW)
+                }
+                // The fetch delay can be changed for testing purposes through an intent with custom flags
+                // isQA and fetchDelay. Where isQA is true/false and fetchDelay is the intended next fetch time
+                // in seconds.
+                else if (extras.keySet().contains(IS_QA_BUILD_KEY)) {
+                    PocketVideoFetchScheduler.setDelayForQA(extras.getLong(FETCH_DELAY_KEY))
```
mcomella: Static state is really hard to reason about and should be avoided! For example, if our process does not die but the application is recreated (in practice, not sure if this ever happens), the static state will remain even though we have a new scheduler: ouch!

I think we can replace this with `context.serviceLocator.pocketFetchScheduler...`

mcomella: We should only do this if FETCH_DELAY_KEY exists, right?

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketVideoFetchScheduler.kt
```diff
@@ -53,6 +53,19 @@ class PocketVideoFetchScheduler(
...
+        // The custom delay will only run once.
+        if (IS_QA_BUILD) {
+            delay = DELAY_TIME_FOR_QA_MILLIS
+            workPolicy = ExistingWorkPolicy.REPLACE
+            IS_QA_BUILD = false
+        } else {
+            delay = getDelayUntilUpcomingNightFetchMillis(now, randLong)
+            workPolicy = ExistingWorkPolicy.KEEP
+        }
         // When the user foregrounds the app, we schedule a one time update for a random time inside our fetch interval
```
mcomella: nit: newline above. I generally always put a newline above a comment block, except if I'm starting an indent (e.g. for a function definition).

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketVideoFetchScheduler.kt
```diff
@@ -53,6 +53,19 @@ class PocketVideoFetchScheduler(
             .setRequiredNetworkType(NetworkType.CONNECTED)
             .build()
 
+        val delay: Long
+        val workPolicy: ExistingWorkPolicy
+
+        // This sets the custom fetch delay for testing purposes.
+        // The custom delay will only run once.
```
mcomella: nit: I'd put this comment above the values so we can better group the code together without misleading new lines.

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketVideoFetchScheduler.kt
```diff
@@ -105,6 +118,13 @@ class PocketVideoFetchScheduler(
         // Since this is a background job, we're in no rush and can wait a while.
         @VisibleForTesting(otherwise = PRIVATE) val BACKOFF_DELAY_MIN_MILLIS = TimeUnit.SECONDS.toMillis(30)
         @VisibleForTesting(otherwise = PRIVATE) val BACKOFF_DELAY_MAX_MILLIS = TimeUnit.SECONDS.toMillis(60)
+
+        private var DELAY_TIME_FOR_QA_MILLIS = TimeUnit.SECONDS.toMillis(5)
```
mcomella: Since this is a QA feature, maybe we're better off initializing this to null, or -1, and asserting this has been set before we these values. It'll help ensure that, while the code changes, our assumptions about how these values are being used won't change. This assumes we keep `isQA` or have another way of representing that state outside this value.

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketVideoFetchScheduler.kt
```diff
@@ -53,6 +53,19 @@ class PocketVideoFetchScheduler(
...
 
+        val delay: Long
+        val workPolicy: ExistingWorkPolicy
+
+        // This sets the custom fetch delay for testing purposes.
+        // The custom delay will only run once.
+        if (IS_QA_BUILD) {
+            delay = DELAY_TIME_FOR_QA_MILLIS
+            workPolicy = ExistingWorkPolicy.REPLACE
+            IS_QA_BUILD = false
```
mcomella: We might want to comment why we disabled this flag after it gets used once.

app/src/main/java/org/mozilla/tv/firefox/IntentValidator.kt
mcomella: I think we should do this for all intents, not just ACTION_MAIN.

mcomella:
As I mentioned on Zoom, tests would be nice but if we're pressed for time or it's too challenging, maybe not. I think I'm mostly concerned about the PocketFetchScheduler production code accidentally changing which I guess we test already, so again, maybe not.

app/src/main/java/org/mozilla/tv/firefox/IntentValidator.kt
```diff
@@ -57,16 +56,14 @@ object IntentValidator {
...
             Intent.ACTION_MAIN -> {
-                val extras = intent.extras ?: return null
-
-                if (extras.keySet().contains(DIAL_PARAMS_KEY)) {
-                    val param = extras.getString(DIAL_PARAMS_KEY)
+                val dialParams = intent.extras?.getString(DIAL_PARAMS_KEY) ?: return null
+                if (dialParams.isNotEmpty()) {
                     TelemetryIntegration.INSTANCE.youtubeCastEvent()
-                    return ValidatedIntentData(url = "https://www.youtube.com/tv?$param", source = Session.Source.ACTION_VIEW)
+                    return ValidatedIntentData(url = "https://www.youtube.com/tv?$dialParams", source = Session.Source.ACTION_VIEW)
```
mcomella: It's probably not worth your time to fix now but this is still going to break blame: you'd need to edit the existing commit that originally changed this.

app/src/main/java/org/mozilla/tv/firefox/IntentValidator.kt
```diff
@@ -57,6 +57,7 @@ object IntentValidator {
 
     fun validate(context: Context, intent: SafeIntent): ValidatedIntentData? {
         setExperimentOverrides(intent, context)
+        changeFetchDelaySecondsForQA(intent)
```
mcomella: nit: -> setFetchDelayOverrides ?

app/src/main/java/org/mozilla/tv/firefox/IntentValidator.kt
```diff
@@ -101,9 +98,9 @@ object IntentValidator {
         }
     }
 
-    private fun changeFetchDelaySecondsForQA(intent: SafeIntent) {
+    private fun changeFetchDelaySecondsForQA(intent: SafeIntent, context: Context) {
```
mcomella: super-nit: when `context` is a parameter, I like to put it first. Contexts are dangerous (memory leaks, access to everything, etc.) so it's important to highlight it when possible.

That being said, I think you should take the `PocketVideoFetchScheduler` into this method, rather than a Context: it's better to be explicit and scope access down as much as possible.

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketVideoFetchScheduler.kt
```diff
@@ -53,14 +53,15 @@ class PocketVideoFetchScheduler(
             .setRequiredNetworkType(NetworkType.CONNECTED)
             .build()
 
+        // This sets the custom fetch delay for testing purposes.
+        // The custom delay will only run once.
         val delay: Long
         val workPolicy: ExistingWorkPolicy
```
mcomella: nit: since you put the comment above, I'd remove this newline too: basically, the whole block, including the variable declarations, is performing that functionality so, to me, it makes sense to group it all together.

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketVideoFetchScheduler.kt
```diff
@@ -120,12 +126,8 @@ class PocketVideoFetchScheduler(
         @VisibleForTesting(otherwise = PRIVATE) val BACKOFF_DELAY_MIN_MILLIS = TimeUnit.SECONDS.toMillis(30)
         @VisibleForTesting(otherwise = PRIVATE) val BACKOFF_DELAY_MAX_MILLIS = TimeUnit.SECONDS.toMillis(60)
 
-        private var DELAY_TIME_FOR_QA_MILLIS = TimeUnit.SECONDS.toMillis(5)
+        private var DELAY_TIME_FOR_QA_MILLIS: Long? = null
```
mcomella: These variables should be moved into the class body, otherwise this is still static state.

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketVideoFetchScheduler.kt
```diff
@@ -110,6 +111,11 @@ class PocketVideoFetchScheduler(
         return userFetchTime.timeInMillis - now.timeInMillis
     }
 
+    fun setDelayForQA(seconds: Long) {
```
mcomella: nit: -> setQAFetchDelayOverride ?

I think "override" indicates that this will be influenced by an outside source while `setDelayForQA` could just be an internal method. I changed the word order to be consistent with the command line flag.

This function should be similarly named to the function in IntentValidator, imo.

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketVideoFetchScheduler.kt
```diff
@@ -53,14 +53,15 @@ class PocketVideoFetchScheduler(
...
         val delay: Long
         val workPolicy: ExistingWorkPolicy
 
-        // This sets the custom fetch delay for testing purposes.
-        // The custom delay will only run once.
         if (IS_QA_BUILD) {
-            delay = DELAY_TIME_FOR_QA_MILLIS
+            delay = checkNotNull(DELAY_TIME_FOR_QA_MILLIS) { "Fetch delay value must be set" }
             workPolicy = ExistingWorkPolicy.REPLACE
+            // Only allow this value to be used once so we do not end up spamming the Pocket servers
```
mcomella: nit: imo, I don't really like having a line of code, a line of comment, and a line of code so I'd put a newline above this or shorten the text and inline it. That being said, I don't have great reasons for that (I think it's because I like comments to "preface" the code) so I've been allowing it for other devs on the team.

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketVideoFetchScheduler.kt
mcomella: nit: maybe add a kdoc explaining this method? It's a little dangerous.

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketVideoFetchScheduler.kt
```diff
@@ -110,6 +111,11 @@ class PocketVideoFetchScheduler(
         return userFetchTime.timeInMillis - now.timeInMillis
     }
 
+    fun setDelayForQA(seconds: Long) {
+        IS_QA_BUILD = true
+        DELAY_TIME_FOR_QA_MILLIS = TimeUnit.SECONDS.toMillis(seconds)
```
mcomella: nit: -> qaFetchDelayOverrideMillis

I changed the word order to be consistent with the command line flag.

Related nit: ALL_CAPS is only used for constants, not for static state

app/src/main/java/org/mozilla/tv/firefox/IntentValidator.kt
```diff
@@ -55,6 +56,7 @@ object IntentValidator {
 
     fun validate(context: Context, intent: SafeIntent): ValidatedIntentData? {
         setExperimentOverrides(intent, context)
+        changeFetchDelaySecondsForQA(intent, context)
```
mcomella: nit: -> setQAFetchDelayOverrides for consistency ?

If you were you, I'd prepending "ExperimentOverrides" with QA too.

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketVideoFetchScheduler.kt
```diff
@@ -53,6 +53,21 @@ class PocketVideoFetchScheduler(
...
+        // This sets the custom fetch delay for testing purposes.
+        // The custom delay will only run once.
+        val delay: Long
+        val workPolicy: ExistingWorkPolicy
+
+        if (IS_QA_BUILD) {
+            delay = checkNotNull(DELAY_TIME_FOR_QA_MILLIS) { "Fetch delay value must be set" }
+            workPolicy = ExistingWorkPolicy.REPLACE
+            // Only allow this value to be used once so we do not end up spamming the Pocket servers
+            IS_QA_BUILD = false
```
mcomella: We should null out the delayMillis too, right? Otherwise we're being inconsistent how we handle the state.

It might be good to extract this into a "reset" helper function that is next to `setDelayForQA` so that it's visible what states are intended to be set and unset together.

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketVideoFetchScheduler.kt
```diff
@@ -27,6 +27,8 @@ private const val FETCH_UNIQUE_WORK_NAME = "PocketFetch"
 class PocketVideoFetchScheduler(
     private val isPocketEnabledByLocale: () -> Boolean
 ) : LifecycleObserver {
+    private var qaFetchDelayOverrideMillis: Long? = null
+    private var isQABuild = false
```
mcomella: nit: I don't love that these initial states are duplicated with the reset values in `resetQAFetchDelayOverrides`. However, most of the solutions I can think of have a lot of boilerplate or are misleading. Maybe:
```kotlin
    private var qaFetchDelayOverrideMillis: Long? = null
    private var isQABuild = false
    init { resetQAFetchDelayOverrides() } // replaces initial values above to avoid duplication.
```

What do you think?

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketVideoFetchScheduler.kt
```diff
@@ -111,9 +111,21 @@ class PocketVideoFetchScheduler(
...
+     * Used for QA to set the fetch delay to a specified number of seconds. This is set through
+     * a custom intent.
+     */
+    fun setQAFetchDelayOverride(seconds: Long) {
+        isQABuild = true
+        qaFetchDelayOverrideMillis = TimeUnit.SECONDS.toMillis(seconds)
+    }
+
+    /**
+     * Used to reset QA fetch delay values after each use so we do not end up spamming the Pocket servers.
```
mcomella: The first sentence of a function should always be what the function does, not how it's intended to be used. For example, "Resets fetch delay to production values. We do this so we do not spam the Pocket servers..." fwiw, sometimes it's hard to make this different from the method name but if you can, it's usually worth it.

The reasons for saying what the method does in the first sentence is:
- iirc, generated javadoc will only use the first sentence in the summary views (i.e. the list of all the methods; not sure what kdoc does)
- Devs won't necessarily read all of your kdoc so you should tell them ASAP what effects it will have on the program (does it have side effects? Does it just return values?)

And we generally want to avoid saying _how_ it's being used because the code can change and that comment can be set out of date. However, it's valuable to say _why_ it might be used (e.g. so we don't spam the Pocket servers) because that may affect the caller's decision to use it.

I hope this makes sense and is actually correct – I'm a little tired. :P

Here and above.

Issue #2320, opened 2019-05-23
Set Up Fretboard Experiment for TV Guide


athomasmoz:
### Why/User Benefit/User Problem

As a Product Manager, I want an A/B test set up for the TV Guide feature, so that I can measure whether it's providing value to users

### Acceptance Criteria (how do I know when I’m done?)
* [x] Set up an experiment placeholder where our changes can be added
* [x] Turn experiment to 0%
- [x] Come up with a plan with QA to test "channels on" and communicate it over email - Timebox this to max 1 day
- [x] **On June 3**: add channels to experiment #2320 (we want the categories to be visible for testing our release build before users see it so we can find regressions.) and pref to 0% for release builds

* Group A is the control (do not show new Channels or Channels onboarding)
* Group B will show:
  * Each of the 5 new categories
  * Onboarding screen
* ~~TV Guide experiment is set to X% (TBD from @athomasmoz) - Fretboard bug is #2320~~

### Not Included

* ~~This does not include turning on or running the experiment (that is #2321 )~~ That bug is closed as a dupe of this one.

liuche:
@athomasmoz has already started the conversation w/ Felix from DS to get percentages.

liuche:
Looking at whether we can pass in an experiment on the commandline (looking at [Fretboard docs](https://github.com/mozilla-mobile/android-components/blob/master/components/service/fretboard/README.md#setting-override-values) for “Setting override values”)

liuche:
Fennec uses `setOverride` for allowing live-toggling of experiments through about:experiments.

This uses [SharedPreferences](https://searchfox.org/mozilla-central/source/mobile/android/base/java/org/mozilla/gecko/switchboard/SwitchBoard.java#154) to store override values that get read by Fretboard/Switchboard [before loading experiment config](https://searchfox.org/mozilla-central/source/mobile/android/base/java/org/mozilla/gecko/switchboard/SwitchBoard.java#154).

We don't necessarily need the live overrides - and would in fact want to clear them each time (so that we don't mix up experiments).

Baron-Severin:
> Looking at whether we can pass in an experiment on the commandline (looking at [Fretboard docs](https://github.com/mozilla-mobile/android-components/blob/master/components/service/fretboard/README.md#setting-override-values) for “Setting override values”)

We will also have to look into mocking locale.

liuche:
With this patch, you can now specify a set of experiments to start with the following command:
[...]

PR #2375, opened 2019-06-05
Add commandline control over fretboard experiments

liuche:
I based this off of Daniela's PR, but that could be pulled out (and could optionally depend on the `isQA` flag).

This is meant to be QA only, and run from command-line with intent extras and isn't an expected user path.

See usage [here](https://github.com/mozilla-mobile/firefox-tv/issues/2320#issuecomment-498904888). I'd probably add this explanation to the README as well (or our Fretboard documentation).

A few things probably should be updated from previous TODOs in the code.

One thing I learned is that for experiments to run at all, they need to be on Kinto - otherwise, the experiment just shortcircuits before you can handle overrides, etc.

### Pull Request checklist

### DIFF
--- a/app/src/main/java/org/mozilla/tv/firefox/IntentValidator.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/IntentValidator.kt
@@ -10,7 +10,9 @@ import android.os.Bundle
 import androidx.annotation.VisibleForTesting
 import android.text.TextUtils
 import mozilla.components.browser.session.Session
+import mozilla.components.service.fretboard.ExperimentDescriptor
 import mozilla.components.support.utils.SafeIntent
+import org.mozilla.tv.firefox.ext.serviceLocator
 import org.mozilla.tv.firefox.telemetry.TelemetryIntegration
 import org.mozilla.tv.firefox.utils.UrlUtils
 
@@ -27,6 +29,7 @@ data class ValidatedIntentData(val url: String, val source: Session.Source)
  */
 object IntentValidator {
     @VisibleForTesting const val DIAL_PARAMS_KEY = "com.amazon.extra.DIAL_PARAM"
+    private const val ACTIVE_EXPERIMENTS_KEY = "activeExperiments"
 
     /**
      * Validate that [intent] contains all expected parameters.
@@ -50,9 +53,9 @@ object IntentValidator {
     }
 
     fun validate(context: Context, intent: SafeIntent): ValidatedIntentData? {
-        val action = intent.action
+        setExperimentOverrides(intent, context)
 
-        when (action) {
+        when (intent.action) {
             Intent.ACTION_MAIN -> {
                 val dialParams = intent.extras?.getString(DIAL_PARAMS_KEY) ?: return null
                 if (dialParams.isNotEmpty()) {
@@ -80,5 +83,16 @@ object IntentValidator {
             }
         }
         return null
+    }
+
+    private fun setExperimentOverrides(intent: SafeIntent, context: Context) {
+        intent.extras?.getStringArray(ACTIVE_EXPERIMENTS_KEY)?.let { strArray ->
+            val fretboard = context.serviceLocator.fretboardProvider.fretboard
+            fretboard.clearAllOverrides(context)
+
+            strArray.forEach {
+                fretboard.setOverride(context, ExperimentDescriptor(it), true)
+            }
+        }
     }
 }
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/mozilla/tv/firefox/IntentValidator.kt
```diff
@@ -81,4 +84,15 @@ object IntentValidator {
         }
         return null
     }
+
+    private fun setExperimentOverrides(intent: SafeIntent, context: Context) {
+        intent.extras?.getStringArray(ACTIVE_EXPERIMENTS_KEY)?.let { strArray ->
```
mcomella: nit: strArray -> experimentDescriptorOverrides, or similar

app/src/main/java/org/mozilla/tv/firefox/IntentValidator.kt
```diff
@@ -27,6 +29,7 @@ data class ValidatedIntentData(val url: String, val source: Session.Source)
  */
 object IntentValidator {
     @VisibleForTesting const val DIAL_PARAMS_KEY = "com.amazon.extra.DIAL_PARAM"
+    private const val ACTIVE_EXPERIMENTS_KEY = "activeExperiments"
```
mcomella: nit: I prefer to put private constants at the top level, to reduce the amount of code in the class body, which makes it clearer what is important in the class.

app/src/main/java/org/mozilla/tv/firefox/IntentValidator.kt
```diff
@@ -81,4 +84,15 @@ object IntentValidator {
         }
         return null
     }
+
+    private fun setExperimentOverrides(intent: SafeIntent, context: Context) {
+        intent.extras?.getStringArray(ACTIVE_EXPERIMENTS_KEY)?.let { strArray ->
+            val fretboard = context.serviceLocator.fretboardProvider.fretboard
+            fretboard.clearAllOverrides(context)
```
mcomella: If we receive an Intent while the app is foregrounded, we're going to clear all of our overrides: is that intended? I'd think maybe we want to clear our overrides in `onStop`, or `onCreate`, or something... Why is clearing necessary?

mcomella: I realized this only happens conditionally, when an intent is received and new extras are set, which makes sense: when overrides are received, clear all overrides. Should we document the behavior somewhere for QA?

liuche: I'll send an email out, and ask if there's another place QA would like that documentation (or put it in the README).

app/src/main/java/org/mozilla/tv/firefox/experiments/ExperimentsProvider.kt
```diff
@@ -61,22 +61,22 @@ class ExperimentsProvider(private val fretboard: Fretboard, private val context:
...
-//            expDescriptor.name.endsWith(ExperimentSuffix.A.value) -> false
-//            expDescriptor.name.endsWith(ExperimentSuffix.B.value) -> true
-//            else -> {
-//                Sentry.capture(ExperimentIllegalStateException("TV Guide Channels Illegal Branch Name"))
-//                false
-//            }
-//        }
+        val expDescriptor = checkBranchVariants(ExperimentConfig.TV_GUIDE_CHANNELS)
+        return when {
+            // The user is currently not part of the experiment
```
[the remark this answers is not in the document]
liuche: It looks like the `null` return value is a little overloaded in `ExperimentEvaluator` - this either means the experiment doesn't exist in the list of experiments, or that there is an override that sets the experiment branch to false.

The else statement handles the case when we receive a non-null experiment, but it's not one that we handle, e.g. we've updated the experiment on the Kinto server, but this client is outdated and doesn't recognize the new branch.

Added a clarifying comment.

app/src/main/java/org/mozilla/tv/firefox/experiments/ExperimentsProvider.kt
```diff
@@ -61,22 +61,22 @@ class ExperimentsProvider(private val fretboard: Fretboard, private val context:
...
-//            // The user is currently not part of the experiment
-//            expDescriptor == null -> false
-//            expDescriptor.name.endsWith(ExperimentSuffix.A.value) -> false
-//            expDescriptor.name.endsWith(ExperimentSuffix.B.value) -> true
-//            else -> {
-//                Sentry.capture(ExperimentIllegalStateException("TV Guide Channels Illegal Branch Name"))
-//                false
-//            }
-//        }
+        val expDescriptor = checkBranchVariants(ExperimentConfig.TV_GUIDE_CHANNELS)
```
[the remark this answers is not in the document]
liuche: I think the idea is that each experiment could have a different number of branches, etc, so it should be different per check.

PR #2380, opened 2019-06-06
Issue #2320 - Address experiment options nits.

liuche:
### Pull Request checklist

### DIFF
--- a/app/src/main/java/org/mozilla/tv/firefox/IntentValidator.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/IntentValidator.kt
@@ -16,6 +16,8 @@ import org.mozilla.tv.firefox.ext.serviceLocator
 import org.mozilla.tv.firefox.telemetry.TelemetryIntegration
 import org.mozilla.tv.firefox.utils.UrlUtils
 
+private const val EXTRA_ACTIVE_EXPERIMENTS = "qaActiveExperiments"
+
 data class ValidatedIntentData(val url: String, val source: Session.Source)
 
 /**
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/mozilla/tv/firefox/IntentValidator.kt
```diff
@@ -17,6 +17,7 @@ import org.mozilla.tv.firefox.telemetry.TelemetryIntegration
 import org.mozilla.tv.firefox.utils.UrlUtils
 
 data class ValidatedIntentData(val url: String, val source: Session.Source)
+private const val ACTIVE_EXPERIMENTS_KEY = "activeExperiments"
```
mcomella: nit: constants should be defined about classes. Also, imo classes & values should generally have newlines between them.

mcomella: nit: Maybe we should change the value to `qaActiveExperiments`? That will make it clear to QA (and pesky users!) that they should not have many expectations when using this functionality. I asked Daniela to make this change too.

mcomella: nit: The value should probably be `EXTRA_KEY_ACTIVE_EXPERIMENTS` or `KEY_ACTIVE_EXPERIMENTS`: it makes it more obvious how certain constants are to be used when it's prefixed. I asked Daniela to make this change too.

liuche: I'm reading this as "above classes" and "add a newline"

liuche: I'll update the wiki page too.

Issue #2338, opened 2019-05-24
Releng checklist - v3.10 (take 2)


dnarcese:
##### Sprint Release Owner: @mcomella 
 _Since last sprint we did not release v3.10, name these LAT's as 3.10-LAT3 and 3.10-LAT4_
#### Tuesday # 1

- [x] Upgrade a-c

#### Wednesday # 1

- [x] Export strings (*if applicable*) ([instructions](https://github.com/mozilla-mobile/firefox-tv/wiki/Localization#exporting-strings-for-translation))
  - Submitted https://github.com/mozilla-l10n/firefoxtv-l10n/pull/25
- [x] Create PR for bumping version (`X.X-LAT1`)
  - Merge to master
- [x] Cut LAT branch `releases/vX.X-LAT1`
- [x] Create GitHub pre-release build `vX.X-LAT1`([instructions](https://github.com/mozilla-mobile/firefox-tv/wiki/Releng-Checklist#final-builds))
- [x] Release LAT ([instructions](https://developer.amazon.com/docs/app-testing/live-app-testing-getting-started.html))

#### Wednesday # 2 (soft code freeze)

- [x] Create PR for bumping version (`X.X-LAT2`)
  - Merge to master
- [x] Cut LAT branch `releases/vX.X-LAT2`
- [x] Create GitHub pre-release build `vX.X-LAT2` ([instructions](https://github.com/mozilla-mobile/firefox-tv/wiki/Releng-Checklist#final-builds))
- [x] Release LAT ([instructions](https://developer.amazon.com/docs/app-testing/live-app-testing-getting-started.html))

#### Thursday # 2

- [x] [QA] run screenshot test and upload ([instructions](https://github.com/mozilla-mobile/firefox-tv/wiki/Localization#screenshots))
- [x] Upgrade a-c on master

#### Friday # 2 (hard code freeze)

- [x] [Product, QA, eng] sign-off for release

#### Monday # 3

- [x] Create PR for bumping version (`X.X`) and updating changelog
  - Update the CHANGELOG: glance over the changelog to see if any major
    features are missing and add the new release header (don't forget to add the diff links at the bottom)
  - Merge to master
- [x] Cut release branch `releases/vX.X`
- [x] Create GitHub release build `vX.X`([instructions](https://github.com/mozilla-mobile/firefox-tv/wiki/Releng-Checklist#final-builds))
[...]

PR #2387, opened 2019-06-06
Issue #2338: Update a-c to v0.55; suppress deprecation error.

mcomella:
### Pull Request checklist


mcomella:
> This is semi-related but maybe out of the scope for this PR, I think we should either update gecko version every a-c bump (since we rely our rendering implementation on a-c Engine) or remove gecko build unit testing in our CI (because I fear one day it would go "out-of-sync" and break).

I added a local note to follow-up on this: I'd also like for us to be upgrading our other non-a-c dependencies regularly.

Issue #2354, opened 2019-05-30
FF for fire tv crashes when trying to delete the last tile


abodea:
## Steps to reproduce
1. Launch FF for fire tv.
2. Select Pinned Tiles section.
3. Delete all the Pinned Tiles.
### Expected behavior
All the pinned tiles can be deleted from FF for fire tv.
### Actual behavior
When trying to delete the last tile, FF for fire tv crashes.
### Device information
* Fire TV device: **`Fire Stick 2nd gen`**
* Latest affected Firefox version: **`v3.10-LAT3`**
* Earliest affected Firefox version: **`v3.10-LAT3`**
### Notes
This issue is reproducible all the time. The screen remains frozen for like ~15 second and the crashes.
For more details please see the attachments: [logcat.txt](https://github.com/mozilla-mobile/firefox-tv/files/3237423/logcat.txt) and [video](https://drive.google.com/file/d/1kDKQJ7SnDOFvnP5sMITeB_jEGysisVyW/view?usp=sharing).
Note that I tried on the latest release v3.9 to delete all the tiles for 5 times and FF didn't crashed.

abodea:
Please note that this issue also occurs on the **`Fire TV Cube 4k.`**

Baron-Severin:
Looks like this is caused by an infinite loop of focus updates.  This can be shown by adding a log in a `doOnNext` on `NavigationOverlayViewModel#focusUpdate`.  These updates are blocking the main thread and causing an ANR.

Baron-Severin:
Focus is bouncing endlessly between the Pocket channel and one of the Pocket tiles.

Baron-Severin:
Comment in `FocusRepo#updateFocusStateIfNew`: `// If pinnedTiles is empty and current focus is on home_tile, we need to restore lost focus (this happens when you remove all tiles in the overlay)`.  With the channel update, all channel tiles have an ID of home_tile, which is causing this infinite loop.

abodea:
Verified as fixed in the latest Fire TV build `3.10-LAT4)`.

PR #2397, opened 2019-06-07
Closes #2379: fixes YouTube focus loss on menu press

### LINKED ISSUE #2379: YouTube videos/videos from Pocket are stuck when returning from browser overview menu
## Steps to reproduce
1. Launch FF for fire TV.
2. Select the pocket section.
3. Select any video from pocket and play it.
4. After the video starts tap on the menu button.
5. From menu tap back or menu button again.

### Expected behavior
The user is redirected to the video and the video keeps playing or it is able to be resumed.
### Actual behavior
The screen is stuck on the video screen.
### Device information
* Fire TV device: **`Fire Stick 2nd gen`**
* Latest affected Firefox version: **`v3.10-LAT4`**
* Earliest affected Firefox version: **`v3.10-LAT4`**
### 
Please note that this issue also occurs if the 
* `HINT_BAR_TEST("HintBar-2011"),`
* `TV_GUIDE_CHANNELS("TvGuideChannels-2195")` are active.
Attachments:
* [logcat.txt](https://github.com/mozilla-mobile/firefox-tv/files/3261942/logcat.txt)
* [video](https://drive.google.com/file/d/1ZoxCIHU_DoOiaaRzF2h1SGN1uALE1H8i/view?usp=sharing)
* Also, please note that this issue occurs even when entering a YouTube video from the Tile sections.

Baron-Severin:
RCA:
0b7e1b9 made a change where any time the active screen is the overlay and the EngineView receives focus, focus is moved to the NavUrlBar. However it looks like this is getting hit while switching between overlay and render fragments (because our internal representation of current screen and the actually active screen are not exactly in sync). This moves focus around while on YouTube, breaking its navigation.

The change made here will partially reintroduce #2354. Focus will be non-fatally lost (hitting buttons will bring it back), and no crash will occur. This is a less frequent case than #2379 and focus is extremely finicky, so it's an acceptable change.

### Pull Request checklist

Bug never reached users

### DIFF
--- a/app/src/main/java/org/mozilla/tv/firefox/focus/FocusRepo.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/focus/FocusRepo.kt
@@ -15,7 +15,6 @@ import org.mozilla.tv.firefox.ScreenController
 import org.mozilla.tv.firefox.ScreenControllerStateMachine.ActiveScreen
 import org.mozilla.tv.firefox.ext.validateKnownViewById
 import org.mozilla.tv.firefox.channels.pinnedtile.PinnedTileRepo
-import org.mozilla.tv.firefox.ext.ENGINE_VIEW_ID
 import org.mozilla.tv.firefox.pocket.PocketVideoRepo
 import org.mozilla.tv.firefox.session.SessionRepo
 import org.mozilla.tv.firefox.utils.URLs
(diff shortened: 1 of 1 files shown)


Baron-Severin:
@mcomella Focus changes are always risky, so I'm not entirely comfortable with this, but it seems safer than most.

Just verified the behavior on device, and it seems to work as expected.

Issue #2399, opened 2019-06-07
No longer collecting individual Pocket Clicks w/ Telemetry


liuche:
In our Pocket Tile click events, we're [only sending the Pocket video id impression/click](https://github.com/mozilla-mobile/firefox-tv/blob/master/app/src/main/java/org/mozilla/tv/firefox/channels/ChannelConfig.kt#L28) as part of the Pocket ping, which is separate from the rest of our telemetry (because we don't want to mix Pocket video ids with the rest of our user data).

We are still sending the number of unique pocket videos clicked. However we are not counting each click event.

This change would probably involve adding a tile click extra of type `pocket` in addition to already collecting bundled/custom/youtube.

## Steps to reproduce
Click on Pocket video tile.

### Expected behavior
Click collected.

### Actual behavior
Only unique Pocket clicks are collected in main telemetry. (And Pocket ping includes video clicks with video id.)

### Device information
* Fire TV device: all 
* Latest affected Firefox version: Not sure when we stopped collecting Pocket video clicks. At least since 3.8.
* Earliest affected Firefox version: ?
* Turbo Mode: NA

liuche:
From log statements, the ping looks like this:
`2019-06-10 11:29:42.907 3924-3924/org.mozilla.tv.firefox.debug E/LIUCHE: [2628,"action","click","home_tile","pocket"]`

Issue #2091, opened 2019-04-18
Remove YouTube tile telemetry


dnarcese:
### Vision statement / What / Requirements
This is a follow-up to #2014 and needs to be done in the next release (after 3.7.1).
- [ ] File a data bugzilla ticket to make sure the Youtube tile values are the same (and to report the discrepancy, if not)
- [ ] Remove the Youtube Tile Event

### Impact
In #2014, bundled tile clicks were collected as an extra, so the separate value for YouTube clicks is not needed anymore.
### Acceptance criteria
Bugzilla ticket is filed, and the telemetry event for YouTube clicks is removed.

liuche:
@SuYoungHong I'll cc you when we release our 3.8 release, and you can start looking for the new tiles telemetry from #2014 - if it looks good, we'll remove this probe.

mcomella:
Even if we're blocked to merge the code, we can still write the code: assigning to self.

mcomella:
> @SuYoungHong I'll cc you when we release our 3.8 release, and you can start looking for the new tiles telemetry from #2014 - if it looks good, we'll remove this probe.

I sent Su an email and will follow up.

mcomella:
I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1548420 for our data science request: we're now blocked on its completion.

mcomella:
Su followed up on Bugzilla: https://bugzilla.mozilla.org/show_bug.cgi?id=1548420#c1 I'll continue the conversation there.

mcomella:
Su and I met in person - https://bugzilla.mozilla.org/show_bug.cgi?id=1548420#c4 - I have an action item to briefly try to identify the bug that's causing us to undercount.

mcomella:
We deprioritized this issue in favor of urgent feature work: we'll come back and make a decision later.

mcomella:
I dug in for 30 minutes, searching for bugs in the source code on master and in v3.9 (where Su saw extra = null for a "bundled" tile): I couldn't find any bugs. I noticed the analysis may include debug builds so [I'm asking about that](https://bugzilla.mozilla.org/show_bug.cgi?id=1548420#c6).

PR #2400, opened 2019-06-07
No issue: Updating telemetry docs for data science.

liuche:
While working w/ the telemetry docs, Felix said that these are a few things to correct in our docs.

### Pull Request checklist

### DIFF
--- a/docs/telemetry.md
+++ b/docs/telemetry.md
@@ -55,7 +55,7 @@ The event ping contains a list of events ([see event format on readthedocs.io](h
 
 ### General
 
-| Event                                                | category   | method                  | object            | value      | extras.    |
+| Event                                                | category   | method                  | object            | value      | extra.    |
 |------------------------------------------------------|------------|-------------------------|-------------------|------------|------------|
 | Pocket channel: unique videos clicked per session    | aggregate  | click                   | pocket_video      | `<int>`    |            |
 | App opened from view intent                          | action     | view_intent             | app               |            |            |
@@ -64,7 +64,7 @@ The event ping contains a list of events ([see event format on readthedocs.io](h
 (*) This event is sent at the end of every session.
 
 ### Browser Overlay
-| Event                                  | category | method                | object       | value                    | extras.       |
+| Event                                  | category | method                | object       | value                    | extra.       |
 |----------------------------------------|----------|-----------------------|--------------|--------------------------|---------------|
 | Browser: back clicked                  | action   | click                 | menu         | back                     |               |
 | Browser: forward clicked               | action   | click                 | menu         | forward                  |               |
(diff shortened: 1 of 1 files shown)


liuche:
@mcomella maybe this is relevant to what Su could be seeing as a discrepancy in the youtube tile telemetry?

docs/telemetry.md
```diff
@@ -64,15 +64,15 @@ The event ping contains a list of events ([see event format on readthedocs.io](h
...
 | Browser: back clicked                  | action   | click                 | menu         | back                     |               |
 | Browser: forward clicked               | action   | click                 | menu         | forward                  |               |
 | Browser: refresh clicked               | action   | click                 | menu         | refresh                  |               |
 | Turbo mode switch clicked              | action   | change                | turbo_mode   | on/off (the new value)   |               |
 | Pin site switch clicked                | action   | change                | pin_page     | on/off (the new value)   | desktop_mode* |
 | Desktop mode switch clicked            | action   | change                | desktop_mode | on/off (the new value)   |               |
-| Tile clicked                           | action   | click                 | home_tile    | bundled/custom/youtube** | tile_id***    |
+| Tile clicked                           | action   | click                 | home_tile    | bundled/custom/youtube_tile** | tile_id***    |
```
mcomella: This can also be "pocket", right?

mcomella: In the note for `youtube_tile**`, it mentions this value is removed in v3.8 but that's not true. We're trying to remove it in #2091 and, as of v3.10, it's still in there and counting using the `youtube_tile` is preferred. That will probably change pending the resolution of #2091.

Do you think you could fix that up and request re-review? Alternatively, @ mention me in #2091 and I can do it as part of that bug because it's related to my work.

liuche: That's actually part of #2399 - we don't have pocket telemetry clicks in right now.

liuche: Sure, I can remove the 3.8 ending at least, but I'll leave it up to #2091 to update that value if we end up removing it.

liuche: Ah, this had some conflicts which is why this didn't have Pocket - rebased and fixed now.

docs/telemetry.md
```diff
@@ -83,7 +83,7 @@ The event ping contains a list of events ([see event format on readthedocs.io](h
 (*)When the pin site switch is clicked, the state (on/off) of the desktop mode switch is also sent.
 
 (**)YouTube tile clicks are a sub-category of bundled tile clicks- each click will count towards both values.
-This value existed from v3.2 to v3.8. For versions after v3.8, the tile_id is included as an extra, not as a value.
+This value existed from v3.2+, but is intended for removal. For versions after v3.8, the tile_id is included as an extra.
```
mcomella: nit: -> "For versions after v3.8, the tile_id is included as an extra but, as of v3.10, this new probe does not accurately match the original YouTube probe: see #2091 for a discussion"

It's not necessary to add but it could be helpful to make it clear that that may not be the final implementation. I wonder if that means counting clicks on our new channel tiles may be inaccurate... I'll follow up with Ashley.

mcomella: Actually, if I find a bug in the implementation, it's probably better for me to update that as part of #2091 so nevermind about updating it here.

PR #2417, opened 2019-06-11
2179 remove pocket onboarding

Baron-Severin:
### Pull Request checklist


app/src/main/java/org/mozilla/tv/firefox/channels/ChannelTile.kt
```diff
@@ -65,10 +68,40 @@ data class ChannelTile(
     }
 }
 
+sealed class ImageSetStrategy {
```
psymoon: nit: kdoc

app/src/main/java/org/mozilla/tv/firefox/channels/ChannelTile.kt
```diff
@@ -65,10 +68,40 @@ data class ChannelTile(
...
+            imageView.setImageResource(id)
+        }
+    }
+
+    // Note that ByPath can be used with either local paths or URLs
+    data class ByPath(val path: String) : ImageSetStrategy() {
+        override fun invoke(imageView: ImageView) {
+            PicassoWrapper.client
+                // TODO find a less brittle way to retrieve this path
+                .load(path)
```
psymoon: Does loading from path not require a placeholder?

app/src/main/java/org/mozilla/tv/firefox/channels/DefaultChannel.kt
```diff
@@ -26,9 +28,20 @@ class DefaultChannel(
...
     }
 
+    fun setSubtitle(subtitle: CharSequence?) {
+        subtitleView.visibility = if (subtitle.isNullOrEmpty()) {
+            View.GONE
+        } else {
+            View.VISIBLE
+        }
+
+        subtitleView.text = subtitle
```
psymoon: Do we still need to set the text if `subtitle` is empty or null?

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketViewModel.kt
```diff
@@ -23,11 +27,15 @@ const val POCKET_VIDEO_COUNT = 20
  * view should not have to perform any transformations on this data).
  */
 class PocketViewModel(
+    resources: Resources,
```
psymoon: Alternatively, we could extend `AndroidViewModel()`

app/src/main/java/org/mozilla/tv/firefox/channels/ChannelTile.kt
Baron-Severin: Seems like it should, doesn't it?  But it wasn't using one before.  I'll add a TODO.

app/src/main/java/org/mozilla/tv/firefox/channels/DefaultChannel.kt
Baron-Severin: It seems right to me to clear the text, in case we ever want to update these over time.  It doesn't really matter since the visibility will be GONE anyway, but this also avoids an extra conditional branch.

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketViewModel.kt
Baron-Severin: Can `AndroidViewModel` classes be tested without Robolectric?  I haven't checked.  Cause I do like that `Resources` can be easily mocked without using Robo, since it has such a dramatic impact on test start times.

PR #2422, opened 2019-06-11
2402 bump ac to 0.56.1

Baron-Severin:
### Pull Request checklist

### DIFF
--- a/app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/ToolbarUiController.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/ToolbarUiController.kt
@@ -4,7 +4,6 @@
 
 package org.mozilla.tv.firefox.navigationoverlay
 
-import android.content.Context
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/ToolbarUiController.kt
```diff
@@ -58,7 +57,7 @@ class ToolbarUiController(
             }
         }
 
-        val layoutInflater = layout.context.systemService<LayoutInflater>(Context.LAYOUT_INFLATER_SERVICE)
+        val layoutInflater = layout.context.getSystemService<LayoutInflater>()!!
```
mcomella: nit: we could get rid of the `!!` if we use `LayoutInflater.from(layout.context)`

mcomella:
I also confirmed that after this change that the memory leak appears to be fixed! Good job tracking that down!

PR #2439, opened 2019-06-14
2435 fix tiles

Baron-Severin:
### Pull Request checklist

The commit structure is not super clear, because this was a rush job.  The major fixes here are:
- Pinned tile deletions are now persisted between sessions
- Custom pinned tiles may now be deleted without crashing

### DIFF
--- a/app/src/main/java/org/mozilla/tv/firefox/channels/ChannelRepo.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/channels/ChannelRepo.kt
@@ -60,7 +60,11 @@ class ChannelRepo(
 
     fun removeChannelContent(tileData: ChannelTile) {
         when (tileData.tileSource) {
-            TileSource.BUNDLED, TileSource.CUSTOM -> {
+            TileSource.CUSTOM -> {
+                TelemetryIntegration.INSTANCE.homeTileRemovedEvent(tileData)
+                pinnedTileRepo.removePinnedTile(tileData.url)
+            }
+            TileSource.BUNDLED -> {
                 TelemetryIntegration.INSTANCE.homeTileRemovedEvent(tileData) // TODO: verify if we need news, sports and music tiles tracked
                 addBundleTileToBlackList(tileData.tileSource, tileData.id)
                 pinnedTileRepo.removePinnedTile(tileData.url)
--- a/app/src/main/java/org/mozilla/tv/firefox/channels/ChannelRepo.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/channels/ChannelRepo.kt
@@ -9,14 +9,18 @@ import android.content.Context
 import android.content.SharedPreferences
 import androidx.annotation.VisibleForTesting
 import io.reactivex.Observable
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/mozilla/tv/firefox/channels/ChannelRepo.kt
```diff
@@ -60,7 +60,11 @@ class ChannelRepo(
 
     fun removeChannelContent(tileData: ChannelTile) {
         when (tileData.tileSource) {
-            TileSource.BUNDLED, TileSource.CUSTOM -> {
+            TileSource.CUSTOM -> {
```
psymoon: This is totally fine, but maybe shorter and less duplicate to do:
```
TileSource.BUNDLED, TileSource.CUSTOM -> {
    TelemetryIntegration.INSTANCE.homeTileRemovedEvent(tileData)
    if (tileData.tileSource is TileSource.BUNDLED) addBundleTileToBlackList(tileData.tileSource, tileData.id)
    pinnedTileRepo.removePinnedTile(tileData.url)
}
```

app/src/main/java/org/mozilla/tv/firefox/channels/ChannelRepo.kt
```diff
@@ -89,10 +99,10 @@ class ChannelRepo(
             else -> throw NotImplementedError("other types shouldn't be able remove tiles")
         }
 
-        return _sharedPreferences.getStringSet(sharedPrefKey, Collections.emptySet())!!.toList()
+        return _sharedPreferences.getStringSet(sharedPrefKey, Collections.emptySet())!!
     }
 
-    private fun saveBlackList(source: TileSource, blackList: List<String>) {
+    private fun saveBlackList(source: TileSource, blackList: Set<String>) {
```
psymoon: This commit consists of mostly refactor work. So the root cause of the tile deletion not persisting issue is the usage of `List` instead of `Set`?

Baron-Severin: Most of the changes are actually functional, not code cleanliness, but I didn't have time to format the commits in a more understandable way.  Prior to this commit, bundled tiles were not hooked up to any blacklist, so pinned tile removals were not persisted.  This led to a lot of weird behavior.

The easiest way to hook them up was to use the existing `filterNotBlacklisted` method, but doing so meant moving some code around.  It's nice that some of it wound up in better places, but the goal was to make minimal changes to get the code to work.

The `List` -> `Set` change was my first theory on what was causing problems.  I still suspect there are some edge cases for which it's important, but I was asked to get this fix out by noon, so I didn't stop to verify which change fixed each problem.

app/src/main/java/org/mozilla/tv/firefox/channels/ChannelRepo.kt
Baron-Severin: Yeah, I was honestly unsure between the two choices.  Wound up going with my gut and making a quick decision, but I'm fine with us reevaluating and changing this one in the future.

PR #2453, opened 2019-06-24
2444 fix last tile deleted crash

Baron-Severin:
### Pull Request checklist

### DIFF
--- a/app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/NavigationOverlayFragment.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/NavigationOverlayFragment.kt
@@ -21,6 +21,7 @@ import android.widget.Toast
 import androidx.annotation.VisibleForTesting
 import androidx.annotation.VisibleForTesting.NONE
 import androidx.core.content.ContextCompat
+import androidx.core.view.isVisible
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/NavigationOverlayFragment.kt
```diff
@@ -286,22 +286,33 @@ class NavigationOverlayFragment : Fragment() {
         ).map { channel -> channel.forwardRemoveEventsToRepo() }
     }
 
-    private fun observeShouldDisplayPinnedTiles(): Disposable {
-        // We considered putting all common channel behavior into one reusable function (and may
-        // still in the future), but were concerned about a potential problem.
+    private fun observeChannelVisibility(): List<Disposable> {
+        // NOTE: for unknown reasons, removing the last tile in a channel crashes if it is visible.
```
mcomella: Ideally, we'd cover this with a test too but I know we're short on time and this test would be hard to write so I'm okay without it.

Baron-Severin: Yeah, and without knowing the root cause this would have to be a UI test, too.

PR #2472, opened 2019-07-01
WIP Add Glean Probes to Metrics (#2644)

kglazko:
### Pull Request checklist


mcomella:
@kglazko Do you intend to keep working on this? I was pruning our open PRs and I was wondering if leaving the PR open was the best course of action (if you're working on it, it is) or if we should close, post the WIP in the issue, and wait for the issue to be prioritized (if you're not working on it, this is probably the best course of action).

app/metrics.yaml
```diff
@@ -0,0 +1,75 @@
...
+    send_in_pings:
+      - baseline
+    lifetime: user
+    bugs:
+      - N/A
+    data_reviews:
+      - N/A
+    notification_emails:
+      - katglazko@gmail.com
+    expires: 2100-01-01
```
liuche: In general, these should be 6mo max out from the current date (and would also need to be listed in the [data review form](https://github.com/mozilla/data-review/blob/master/request.md)).

app/metrics.yaml
```diff
@@ -0,0 +1,75 @@
...
+      String of the item clicked in the menu which includes: Back, Forward, Reload, Exit Firefox
+    send_in_pings:
+      - baseline
+    lifetime: user
+    bugs:
+      - N/A
+    data_reviews:
+      - N/A
+    notification_emails:
+      - katglazko@gmail.com
```
liuche: Notification emails should be the team list (so everyone sees the expiring glean probes, rather than a single person being responsible).

app/metrics.yaml
```diff
@@ -0,0 +1,75 @@
...
+    expires: 2100-01-01
+
+  browser_back_controller:
+    type: event
+    description: >
+      String of the option clicked on the hardware controller which includes: Back
+    extra_keys:
+      source_of_click:
+        description: Explains where the action came from
```
liuche: For things like extras, exhaustive values (or ranges) is important - the goal is for this to be a non-technical output so someone can easily decide "yes this is okay to collect" or "no some of these values are too sensitive to collect".

app/build.gradle
```diff
@@ -447,6 +451,9 @@ if (project.hasProperty("coverage")) {
     }
 }
 
+
+apply from: 'https://github.com/mozilla-mobile/android-components/raw/master/components/service/glean/scripts/sdk_generator.gradle'
```
mcomella: I'm not sure what this does but I wonder if this is necessary: I'd think the glean team would come up with a better way to apply this.

That being said, if this is the recommended way, we should consider an alternative like downloading a copy of this and putting it in the repository. Applying scripts from the internet creates issues: it prevents reproducible builds and leaves developers/the app more open to vulnerabilities if this script changes and we're unaware of it.

kglazko: The optimal way according to Travis is like Fenix, see here: https://github.com/mozilla-mobile/fenix/blob/3253f01667b2b0757614880138c85fd94b8c357d/app/build.gradle#L464 I am currently using the same one as Reference Browser https://github.com/mozilla-mobile/reference-browser/blob/0a9d541b8f343914c3e10f13cc6b13369050dbc1/app/build.gradle#L297 also shown to me by Travis but he recommends changing it so it is using the version tag of AC that FFTV is using.

mcomella:
I'm not actively working on this so I'm going to unassign my reviewer status: let's leave the PR open so someone else can take a look later though. Sorry I never got around to this. :(

Issue #2445, opened 2019-06-14
Integrate FxA Accounts Component


athomasmoz:
### Why/User Benefit/User Problem

As an FFTV eng, I want to re-use the A-C component for accounts, so that I don't need to recreate & maintain a separate version for FFTV

### Acceptance Criteria (how do I know when I’m done?)
* FxA accounts component integrated into FFTV codebase
* Can load sign-in page

### Not included

* UI for signing in doesn't need to be polished #2441 

---

TODO:
- [x] Get credentials to allow sign in, as per https://github.com/mozilla-mobile/android-components/tree/master/components/service/firefox-accounts
- [x] Figure out which sign in API to use:
  - `feature-accounts` (seems to require multi tab so maybe not)
  - `FxaAccountManager` (seems right)
  - `FirefoxAccount`
- [x] Integrate component
- [ ] ? Improve documentation on how to integrate FxA

liuche:
Have never seen this component before, will look at Fenix codebase.

mcomella:
Status update: I've got an account logged in end-to-end with hacky code. Before we can merge, I should clean it up (into our architecture). Then we can parallelize:
- Fixing the sign in experience (bug is filed)
- Correctly updating the state as part of our architecture

mcomella:
Status update: open PR. We'll need to rebase on breaking changes in the next a-c release, I'd like to write tests, and I should help improve the docs though.

mcomella:
Closed in https://github.com/mozilla-mobile/firefox-tv/pull/2513; follow-ups in #2538.

mcomella:
qa-ready: the feature can be enabled with the `SendTab-2511:B` [flag](https://github.com/mozilla-mobile/firefox-tv/wiki/QA#fretboard-experiments). The implemented experience is defined in the acceptance criteria.

Limitations: 
- Sign in button uses placeholder icon in nav overlay. Currently, it's "CC" next to the "Exit Firefox" button
- After signing in, the behavior after clicking "Sign in" again is undefined.
- After signing in, the only artifact that indicates that the user is signed is the "Sign in success" status page. 
- Content is loaded into the tab the user is using to browse content. The sign in page acts like any other page the user loaded. This is not necessarily intended to be the final experience (e.g. we may clear those items from the history stack in the future and return the user to the site they last browsed to)

abodea:
Verified as fixed in the latest LAT 4.1-LAT2, I was able to activate the feature and I added bugs accordingly.

Issue #2462, opened 2019-06-25
Handle AC breaking changes to searchDefault field in search JSON


liuche:
## Steps to reproduce
FFTV is no longer overriding search codes with FFTV-specific ones, which were added in #1876
See dropoff: https://sql.telemetry.mozilla.org/queries/63329#162207

### Expected behavior
FFTV should override the search codes from AC in [SearchEngineManagerFactory](https://github.com/mozilla-mobile/firefox-tv/blob/master/app/src/main/java/org/mozilla/tv/firefox/search/SearchEngineManagerFactory.kt#L13), so that FFTV will use custom search codes rather than the defaults.

### Actual behavior
Replacement codes are not showing up, and from talking to AC this is likely due to AC changes to the search JSON list: https://github.com/mozilla-mobile/android-components/issues/2371

### Acceptance criteria
* Search codes can be overridden again (and the FFTV codes are sent, rather than the defaults)
* Regression tests for search code overrides. May need to ask AC if there aren't clear errors being thrown when overriding fails. To be explicit, the test should fail if upstream a-c changes break the search code overrides and the wrong search codes are used.

### Device information
* Fire TV device: all
* Latest affected Firefox version: master eef55eb 
* Earliest affected Firefox version: Apr 8, 2019
* Turbo Mode: N/A

liuche:
P1 because this affects our use of search codes.

liuche:
https://sql.telemetry.mozilla.org/queries/63329#162207 for changes in search code numbers

Baron-Severin:
It looks like we need a change to A-C in order to fix this.  I've opened https://github.com/mozilla-mobile/android-components/issues/3711.

In the meantime, I'm writing an integration/regression test to cover this case.

PR #2473, opened 2019-07-02
Issue #2445 - Upgrade AC to 1.0.0

liuche:
I also looked at upgrading the [GeckoView version](https://github.com/mozilla-mobile/android-components/blob/v1.0.0/buildSrc/src/main/java/Gecko.kt) but it seems like the current version doesn't build w/ GVDebug anyway so there didn't seem to be any point.
 
### Pull Request checklist


liuche:
> lgtm. Things we may want to look more into from the changelog:
> 
> > ```
> > Added Engine.version property (EngineVersion) for printing and comparing the version of the used engine.
> > ```
> 
> Should we use this instead of whatever we're using in our debug overlay?
We could use both - I'm running into some trouble accessing the Engine in the activity creation, and it's not working how I would expect.

> 
> > Added getProvidedDefaultSearchEngine to SearchEngineManager to return the provided default search engine or the first search engine if the default is not set. This allows use cases like #3344.
> 
> How do we get the default search engine? should we be using this?

We do a search engine substitution, so we don't want to use that at all. (Though currently it's broken #2462)

PR #2513, opened 2019-07-10
Issue #2445: Add FxA account sign in

mcomella:
Grisha is adding breaking changes to the API so we'll soon need to rebase on it (blocked on a-c 4.0.0 upgrade - might as well land this first): leaving this issue open. I'd also still like to:
- ~~Actually test the experiment config~~ Done.
- Add some tests (after the rebase)
- Improve the docs

---

Connected to #2445 ???

## Testing and Review Notes

- Enable with flag: `adb shell am start -n org.mozilla.tv.firefox.debug/org.mozilla.tv.firefox.MainActivity --esa qaActiveExperiments SendTab-2511`
- Click placeholder icon
- Sign in

## Checklist

- [x] Add **testing notes and/or screenshots** in PR description to help guide all potential reviewers
  - Follow-up PR after rebasing on breaking changes
  - Feature is not complete yet
  - Not ready for QA yet, I think

### DIFF
--- a/app/src/main/java/org/mozilla/tv/firefox/fxa/FxaLoginUseCase.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/fxa/FxaLoginUseCase.kt
@@ -1,0 +1,101 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.tv.firefox.fxa
+
+import android.net.Uri
+import androidx.fragment.app.FragmentManager
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.GlobalScope
+import kotlinx.coroutines.launch
+import mozilla.components.service.fxa.manager.FxaAccountManager
+import org.mozilla.tv.firefox.ScreenController
+import org.mozilla.tv.firefox.session.SessionRepo
+import org.mozilla.tv.firefox.telemetry.SentryIntegration
+
+private const val LOGTAG = "FxaLoginUseCase"
+
+/**
+ * Manages login with a Firefox Account (FxA).
+ *
+ * To begin the login flow, we call [beginLogin]: this loads the login page into the engine view. This login page
+ * does the heavy lifting: most error states (e.g. invalid credentials) are handled in the engine view and are not
+ * reported to us. On success, our observer in this class transmits the results from the engine view back to the FxA
+ * library. After that, control of FxA state is returned to the [fxaRepo].
+ *
+ * This class was created as a "super repo": it allows repos to interact with each other without having
+ * references to each other, eliminating possible circular dependencies.
+ */
+class FxaLoginUseCase(
+    private val fxaRepo: FxaRepo,
+    private val sessionRepo: SessionRepo,
+    private val screenController: ScreenController,
+    private val sentryIntegration: SentryIntegration = SentryIntegration
+) {
+
+    /**
+     * Opens the browser screen and loads the FxA login URL to begin the login flow.
+     */
+    fun beginLogin(fragmentManager: FragmentManager) {
+        // TODO: should we throw an error if we're already authenticated when this is called?
+        GlobalScope.launch(Dispatchers.Main) { // main thread: we modify UI state.
+            // a-c#3713: this await will never resume if the user is already logged in.
+            val loginUri = fxaRepo.beginLoginInternalAsync().await()
+
+            // This may be null when the FxA library fails to do things internally, mostly likely due to network issues.
+            if (loginUri == null) {
+                sentryIntegration.captureAndLogError(
+                    LOGTAG, IllegalStateException("beginAuthenticationAsync returned null loginUri"))
+                return@launch
+            }
+
+            // TODO: if user is already signing in, we should consider not reloading the page.
+            screenController.showBrowserScreenForUrl(fragmentManager, loginUri)
+        }
+    }
+
+    init {
+        attachFxaLoginSuccessObserver() // @Suppress cannot be used on an initializer so we call a helper method.
+    }
+
+    @Suppress("CheckResult") // no need to dispose: sessionRepo is active for the duration of the app.
+    private fun attachFxaLoginSuccessObserver() {
+        fun isLoginSuccessUri(uri: String): Boolean = uri.startsWith(FxaRepo.REDIRECT_URI)
+
+        fun extractLoginSuccessKeys(uriStr: String): LoginSuccessKeys? {
+            val uri = Uri.parse(uriStr)
+            val code = uri.getQueryParameter("code")
+            val state = uri.getQueryParameter("state")
+            return if (code != null && state != null) LoginSuccessKeys(code = code, state = state) else null
+        }
+
+        sessionRepo.state
+            .map { it.currentUrl }
+            .distinctUntilChanged()
+            .subscribe {
+                if (!isLoginSuccessUri(it)) {
+                    return@subscribe
+                }
+
+                val loginSuccessKeys = extractLoginSuccessKeys(it)
+                if (loginSuccessKeys == null) {
+                    // Since we received a login success URL, this is never expected. However, since this action is
+                    // controlled by a server, we don't want to crash the app so we log to Sentry instead.
+                    sentryIntegration.captureAndLogError(
+                        LOGTAG, IllegalStateException("Received success URI but success keys cannot be found"))
+                    return@subscribe
+                }
+
+                // TODO: do we want to pop these fxa pages from the user's browsing history?
+                fxaRepo.accountManager.finishAuthenticationAsync(loginSuccessKeys)
+            }
+    }
+
+    private data class LoginSuccessKeys(val code: String, val state: String)
+
+    private fun FxaAccountManager.finishAuthenticationAsync(keys: LoginSuccessKeys) {
+        @Suppress("DeferredResultUnused") // We don't care to wait until completion.
+        finishAuthenticationAsync(code = keys.code, state = keys.state)
+    }
+}
--- a/app/src/main/java/org/mozilla/tv/firefox/fxa/FxaLoginUseCase.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/fxa/FxaLoginUseCase.kt
@@ -10,11 +10,12 @@ import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.GlobalScope
 import kotlinx.coroutines.launch
 import mozilla.components.service.fxa.manager.FxaAccountManager
+import mozilla.components.support.base.log.logger.Logger
 import org.mozilla.tv.firefox.ScreenController
 import org.mozilla.tv.firefox.session.SessionRepo
 import org.mozilla.tv.firefox.telemetry.SentryIntegration
 
-private const val LOGTAG = "FxaLoginUseCase"
+private val logger = Logger("FxaLoginUseCase")
 
 /**
  * Manages login with a Firefox Account (FxA).
@@ -46,7 +47,7 @@ class FxaLoginUseCase(
             // This may be null when the FxA library fails to do things internally, mostly likely due to network issues.
             if (loginUri == null) {
                 sentryIntegration.captureAndLogError(
-                    LOGTAG, IllegalStateException("beginAuthenticationAsync returned null loginUri"))
+                    logger, IllegalStateException("beginAuthenticationAsync returned null loginUri"))
                 return@launch
             }
 
@@ -83,7 +84,7 @@ class FxaLoginUseCase(
                     // Since we received a login success URL, this is never expected. However, since this action is
                     // controlled by a server, we don't want to crash the app so we log to Sentry instead.
                     sentryIntegration.captureAndLogError(
-                        LOGTAG, IllegalStateException("Received success URI but success keys cannot be found"))
+                        logger, IllegalStateException("Received success URI but success keys cannot be found"))
                     return@subscribe
                 }
 
--- a/app/src/main/java/org/mozilla/tv/firefox/fxa/FxaLoginUseCase.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/fxa/FxaLoginUseCase.kt
@@ -6,6 +6,7 @@ package org.mozilla.tv.firefox.fxa
 
 import android.net.Uri
 import androidx.fragment.app.FragmentManager
+import io.reactivex.Observable
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.GlobalScope
 import kotlinx.coroutines.launch
@@ -78,23 +79,27 @@ class FxaLoginUseCase(
             return if (code != null && state != null) LoginSuccessKeys(code = code, state = state) else null
         }
 
-        sessionRepo.state
-            .map { it.currentUrl }
-            .distinctUntilChanged()
-            .subscribe {
-                if (!isLoginSuccessUri(it)) {
-                    return@subscribe
-                }
+        fun Observable<String>.filterMapLoginSuccessKeys(): Observable<LoginSuccessKeys> =
+            this.flatMap { url ->
+                val loginSuccessKeys = extractLoginSuccessKeys(url)
 
-                val loginSuccessKeys = extractLoginSuccessKeys(it)
-                if (loginSuccessKeys == null) {
+                if (loginSuccessKeys != null) {
+                    Observable.just(loginSuccessKeys)
+                } else {
                     // Since we received a login success URL, this is never expected. However, since this action is
                     // controlled by a server, we don't want to crash the app so we log to Sentry instead.
                     sentryIntegration.captureAndLogError(
                         logger, IllegalStateException("Received success URI but success keys cannot be found"))
-                    return@subscribe
+                    Observable.empty()
                 }
+            }
 
+        sessionRepo.state
+            .map { it.currentUrl }
+            .distinctUntilChanged()
+            .filter { isLoginSuccessUri(it) }
+            .filterMapLoginSuccessKeys()
+            .subscribe { loginSuccessKeys ->
                 // TODO: do we want to pop these fxa pages from the user's browsing history?
                 fxaRepo.accountManager.finishAuthenticationAsync(loginSuccessKeys)
             }
--- a/app/src/main/java/org/mozilla/tv/firefox/fxa/FxaLoginUseCase.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/fxa/FxaLoginUseCase.kt
@@ -25,8 +25,14 @@ private val logger = Logger("FxaLoginUseCase")
  * reported to us. On success, our observer in this class transmits the results from the engine view back to the FxA
  * library. After that, control of FxA state is returned to the [fxaRepo].
  *
- * This class was created as a "super repo": it allows repos to interact with each other without having
- * references to each other, eliminating possible circular dependencies.
+ * This class is designed to allows repos to interact with each other without references to each other,
+ * allowing each repo to operate independently of other repos, simplifying state.
+ *
+ * FxaRepo    SessionRepo
+ *    |            |
+ *     ------------
+ *           |
+ *    FxaLoginUseCase
  */
 class FxaLoginUseCase(
     private val fxaRepo: FxaRepo,
--- a/app/src/main/java/org/mozilla/tv/firefox/fxa/FxaLoginUseCase.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/fxa/FxaLoginUseCase.kt
@@ -63,7 +63,8 @@ class FxaLoginUseCase(
     }
 
     init {
-        attachFxaLoginSuccessObserver() // @Suppress cannot be used on an initializer so we call a helper method.
+        // @Suppress cannot be used on an initializer so we call a helper method.
+        attachFxaLoginSuccessObserver()
     }
 
     @Suppress("CheckResult") // no need to dispose: sessionRepo is active for the duration of the app.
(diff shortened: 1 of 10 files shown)


app/src/main/java/org/mozilla/tv/firefox/fxa/FxaRepo.kt
```diff
@@ -0,0 +1,123 @@
...
+import org.mozilla.tv.firefox.fxa.FxaRepo.AccountState.NEEDS_REAUTHENTICATION
+import org.mozilla.tv.firefox.fxa.FxaRepo.AccountState.NOT_AUTHENTICATED
+
+private const val LOGTAG = "FxaRepo"
+
+/**
+ * Manages Firefox Account (FxA) state. In particular, [accountStateDontUseMeYet] exposes the current sign in
+ * state of the user. If you want to initiate sign in, see [FxaLoginUseCase.beginLogin].
+ *
+ * Devs should use this class rather than interacting with the FxA library directly.
```
Baron-Severin: I know we haven't done this in the past, but it's less work than it sounds, and might be a good option here.

If we don't want to expose the FxA library to the rest of the app, maybe we should move this code into a separate Gradle module.  We can choose not to expose the dependency transitively, and instead only provide access to our repo.  

The setup/complexity for this is pretty minimal, it enforces some separation, and if we made it a habit it would even improve build times (not that those are a major concern for us yet).  I've been playing with this in a side app and I'm liking it quite a bit.

app/src/main/java/org/mozilla/tv/firefox/fxa/FxaRepo.kt
```diff
@@ -0,0 +1,123 @@
...
+     */
+    fun beginLoginInternalAsync(): Deferred<String?> {
+        return accountManager.beginAuthenticationAsync()
+    }
+
+    private inner class FirefoxAccountObserver : AccountObserver {
+        override fun onAuthenticated(account: OAuthAccount) {
+            Log.d(LOGTAG, "onAuthenticated")
+            // todo: is this correct?
+            val nextState = if (accountManager.accountProfile() != null) AUTHENTICATED else AUTHENTICATED_NO_PROFILE
```
Baron-Severin: Nit: this line's a bit long, it should probably be reformatted.

app/src/main/java/org/mozilla/tv/firefox/fxa/FxaRepo.kt
```diff
@@ -0,0 +1,123 @@
...
+     * instead of this method.
+     */
+    fun beginLoginInternalAsync(): Deferred<String?> {
+        return accountManager.beginAuthenticationAsync()
+    }
+
+    private inner class FirefoxAccountObserver : AccountObserver {
+        override fun onAuthenticated(account: OAuthAccount) {
+            Log.d(LOGTAG, "onAuthenticated")
+            // todo: is this correct?
```
Baron-Severin: It might be a good idea to reach out to someone on the FxA team to validate some of these suspicions (e.g., if `onAuthenticated` is called with a null profile, state is "successful authentication without a profile."  It's possible that it's some weird edge case in the service.

app/src/main/java/org/mozilla/tv/firefox/fxa/FxaRepo.kt
```diff
@@ -0,0 +1,123 @@
...
+    val accountManager: FxaAccountManager = newInstanceDefaultAccountManager(context)
+) {
+
+    enum class AccountState {
+        // TODO: Are these states accurate?
+        // TODO: Later, may need "failed to login": https://github.com/mozilla-mobile/android-components/issues/3712
+        AUTHENTICATED,
+        AUTHENTICATED_NO_PROFILE, // todo: necessary? profile seems to be fetched async so yes?
+        NEEDS_REAUTHENTICATION,
+        NOT_AUTHENTICATED
```
Baron-Severin: Is there any potential overlap between these states?  `NEEDS_REAUTHENTICATION` seems like it might be a substate of `NOT_AUTHENTICATED`, for example.

app/src/main/java/org/mozilla/tv/firefox/fxa/FxaRepo.kt
```diff
@@ -0,0 +1,123 @@
...
+        private const val CLIENT_ID = "a2270f727f45f648"
+        const val REDIRECT_URI = "https://accounts.firefox.com/oauth/success/$CLIENT_ID"
+
+        private fun newInstanceDefaultAccountManager(context: Context): FxaAccountManager {
+            return FxaAccountManager(
+                context,
+                Config.release(CLIENT_ID, REDIRECT_URI),
+                applicationScopes = arrayOf("profile", "https://identity.mozilla.com/apps/oldsync"), // todo: proper vals?
+                deviceTuple = DeviceTuple(
+                    name = "Firefox for Fire TV",
```
Baron-Severin: Nit: R.string.app_name_extended_fire

app/src/main/java/org/mozilla/tv/firefox/fxa/FxaLoginUseCase.kt
```diff
@@ -0,0 +1,101 @@
...
+        }
+    }
+
+    init {
+        attachFxaLoginSuccessObserver() // @Suppress cannot be used on an initializer so we call a helper method.
+    }
+
+    @Suppress("CheckResult") // no need to dispose: sessionRepo is active for the duration of the app.
+    private fun attachFxaLoginSuccessObserver() {
+        fun isLoginSuccessUri(uri: String): Boolean = uri.startsWith(FxaRepo.REDIRECT_URI)
```
Baron-Severin: Nit: this might be a tiny bit clearer as an extension function.

app/src/main/java/org/mozilla/tv/firefox/fxa/FxaLoginUseCase.kt
```diff
@@ -0,0 +1,101 @@
...
+                    // Since we received a login success URL, this is never expected. However, since this action is
+                    // controlled by a server, we don't want to crash the app so we log to Sentry instead.
+                    sentryIntegration.captureAndLogError(
+                        LOGTAG, IllegalStateException("Received success URI but success keys cannot be found"))
+                    return@subscribe
+                }
+
+                // TODO: do we want to pop these fxa pages from the user's browsing history?
+                fxaRepo.accountManager.finishAuthenticationAsync(loginSuccessKeys)
+            }
```
Baron-Severin: The conditional statements in this block amount to a `filter` and `filterMap`.  `filterMap` is an FP concept, but doesn't have its own operator, we accomplish it in Rx using a `flatMap`.  We can rewrite this as:

```
fun Observable<String>.filterMapSuccessKeys(): Observable<LoginSuccessKeys> =
    this.flatMap { url ->
        val loginSuccessKeys = extractLoginSuccessKeys(url)
        
        if (loginSuccessKeys != null) {
            Observable.just(loginSuccessKeys)
        } else {
            // Since we received a login success URL, this is never expected. However, since this action is
            // controlled by a server, we don't want to crash the app so we log to Sentry instead.
            sentryIntegration.captureAndLogError(
                LOGTAG, IllegalStateException("Received success URI but success keys cannot be found"))
            Observable.empty()
        }
    }

sessionRepo.state
    .map { it.currentUrl }
    .distinctUntilChanged()
    .filter { isLoginSuccessUri(it) }
    .filterMapSuccessKeys()
    .subscribe { loginSuccessKeys ->
        // TODO: do we want to pop these fxa pages from the user's browsing history?
        fxaRepo.accountManager.finishAuthenticationAsync(loginSuccessKeys)
    }
```

app/src/main/res/layout/fragment_navigation_overlay_top_nav.xml
```diff
@@ -48,6 +48,16 @@
...
 
+    <!-- This button is set visible if the user is in the experiment. -->
+    <!-- TODO: add content description dynamically -->
+    <!-- TODO: use production icon. -->
+    <ImageButton
+        android:id="@+id/fxaButton"
+        style="@style/NavigationButton"
+        android:src="@drawable/lb_ic_cc"
+        android:contentDescription="@null"
+        android:visibility="gone"/>
```
Baron-Severin: Up to you, but I would consider wrapping this in a `KillSwitchLayout` (for paranoia reasons).

app/src/main/java/org/mozilla/tv/firefox/fxa/FxaRepo.kt
```diff
@@ -26,6 +26,12 @@ import org.mozilla.tv.firefox.fxa.FxaRepo.AccountState.NOT_AUTHENTICATED
 
 private const val LOGTAG = "FxaRepo"
 
+private val APPLICATION_SCOPES = arrayOf(
+    // We don't use sync, however, if we later add the sync scope, already authenticated users
+    // will have to log in again. To avoid this in the future, we preemptively declare a sync scope.
+    "https://identity.mozilla.com/apps/oldsync"
```
Baron-Severin: Are there potentially any other scopes that we might want to use in the future?  Is there anyone we should ask about this?

app/src/main/java/org/mozilla/tv/firefox/fxa/FxaRepo.kt
```diff
@@ -0,0 +1,129 @@
...
+        // TODO: Later, may need "failed to login": https://github.com/mozilla-mobile/android-components/issues/3712
+        AUTHENTICATED,
+        AUTHENTICATED_NO_PROFILE, // todo: necessary? profile seems to be fetched async so yes?
+        NEEDS_REAUTHENTICATION,
+        NOT_AUTHENTICATED
+        // TODO: do we need initial state before state is fetched from disk?
+    }
+
+    // TODO: set state accurately, choose correct subject, set initial state, etc.
+    val accountStateDontUseMeYet = BehaviorSubject.create<AccountState>()
```
grigoryk: FWIW, in both Fenix and the reference-browser an initialized `accountManager` is exposed to the rest of the application, and UI bits that care about account state can query it and/or register to receive updates. This works pretty well.

It seems to me like you'll end-up with the same pattern here, but via your own intermediary. I'm not sure what value this adds (versus exposing `accountManager`), but perhaps it makes things easier for you.

My general recommendation would be to avoid modelling this after internals of the account manager, but rather build on top of data abstractions that it provides (e.g. we're logged in; we have some auth problems; we have a new version of the profile; etc). Generally speaking, any module you add on top of some other module should ideally introduce a higher level of abstraction. If it doesn't, there's a good chance that module isn't needed and just adds complexity.

app/src/main/java/org/mozilla/tv/firefox/fxa/FxaRepo.kt
mcomella: I think about this from time-to-time for our other integrations (Sentry, Picasso, Telemetry, etc.). However, I'm not sure the benefits:
- (potentially?) faster build times
- Clean separation of who should use what dependencies
- (I think there are other small benefits here too but I'm pressed for time)

Outweigh the downsides:
- Have to rewrite all data types from the library (e.g. can't easily expose just FxaAccountManager if we wanted to without exposing entire dependency)
- afaik, IDE does not make it clear why objects can't be imported (both app/ to thirdparty dependencies and our interface to app/) and newer devs may stumble on this
- Similar files (e.g. FxaRepo & SessionRepo) are in entirely separate file hierarchies and can be hard to browse in IDE or file system if you forget to take module boundaries into account
- More complex build code: it must be multi-module aware (and we struggle with build code already)
- Helper classes (such as utils) are not yet broken into a "common" gradle module so they're unavailable to our intermediate build modules/interfaces.

@Baron-Severin Briefly, what do you think? We should talk IRL for a longer conversation.

app/src/main/java/org/mozilla/tv/firefox/fxa/FxaRepo.kt
mcomella: This is intended for the follow-up bug: #2529. Sorry for not communicating this better.

app/src/main/java/org/mozilla/tv/firefox/fxa/FxaRepo.kt
mcomella: This will be resolved in #2516: we probably want to do more than use `R.string.app_name...`

app/src/main/java/org/mozilla/tv/firefox/telemetry/SentryIntegration.kt
```diff
@@ -60,7 +61,24 @@ object SentryIntegration {
...
+
+    /**
+     * Sends the given [exception] to the Sentry servers without crashing the app
+     * and logs its message at the error level.
+     *
+     * @see [Sentry.capture]
+     */
+    fun captureAndLogError(tag: String, exception: Exception) {
+        // Note: instead, by adding the Log4j dependency, we may be able to get Sentry to log captures to logcat automatically.
+        Log.e(tag, exception.message)
```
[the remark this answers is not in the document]
mcomella: Sure, seems reasonable: I think we should send an email out encouraging everyone to use the a-c logger instead of Android's with the reasons why and a suggested pattern. I did:
```kotlin
private val logger = Logger("tag")

public class Whatever...
```

@Baron-Severin Would you like to take that on?

app/src/main/java/org/mozilla/tv/firefox/fxa/FxaLoginUseCase.kt
```diff
@@ -0,0 +1,101 @@
...
+
+/**
+ * Manages login with a Firefox Account (FxA).
+ *
+ * To begin the login flow, we call [beginLogin]: this loads the login page into the engine view. This login page
+ * does the heavy lifting: most error states (e.g. invalid credentials) are handled in the engine view and are not
+ * reported to us. On success, our observer in this class transmits the results from the engine view back to the FxA
+ * library. After that, control of FxA state is returned to the [fxaRepo].
+ *
+ * This class was created as a "super repo": it allows repos to interact with each other without having
```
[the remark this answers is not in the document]
mcomella: That's what I meant by super repo but I can see that was confusing. :) I'll do the diagram instead.

app/src/main/java/org/mozilla/tv/firefox/fxa/FxaLoginUseCase.kt
```diff
@@ -0,0 +1,101 @@
...
+        GlobalScope.launch(Dispatchers.Main) { // main thread: we modify UI state.
+            // a-c#3713: this await will never resume if the user is already logged in.
+            val loginUri = fxaRepo.beginLoginInternalAsync().await()
+
+            // This may be null when the FxA library fails to do things internally, mostly likely due to network issues.
+            if (loginUri == null) {
+                sentryIntegration.captureAndLogError(
+                    LOGTAG, IllegalStateException("beginAuthenticationAsync returned null loginUri"))
+                return@launch
+            }
```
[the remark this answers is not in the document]
mcomella: No: @Baron-Severin how would you propose handling it? Shall we make it a follow-up? The options I can think of are:
- retry, but if this is network related, it's likely fail again
- Show a toast alerting the user to try again later

Implementing the second one sounds okay.

app/src/main/java/org/mozilla/tv/firefox/fxa/FxaLoginUseCase.kt
mcomella: I always feel uncomfortable creating String extension functions that only apply to uris encoded as Strings: these types of extension functions should be on the Uri data type and we should be passing around Uris. I don't think this greatly changes the readability, especially with your suggested Rx changes below, so I'll leave it as is for now.

app/src/main/java/org/mozilla/tv/firefox/fxa/FxaLoginUseCase.kt
mcomella: This is overall an improvement in readability: neat. :) I don't love how unintuitive flatMap as filterMap is, especially for new devs, but creating a new data type to handle the empty case sounds bad too.

I never would have guessed Rx was missing an operator. :P

app/src/main/java/org/mozilla/tv/firefox/fxa/FxaRepo.kt
mcomella: @grigoryk 

> It seems to me like you'll end-up with the same pattern here, but via your own intermediary.

If you're curious why we use this intermediary, it's a part of our MVVM architecture (essentially [the one Google recommends](https://developer.android.com/jetpack/docs/guide#recommended-app-arch)): we represent our model with Observable values that act as single-source-of-truth of declarative state, simplifying our UI updates (like React). Let me know if I should elaborate.

> and UI bits that care about account state can query it and/or register to receive updates. 

These patterns go against our architecture and we've decided against these patterns because:
- Querying introduces multiple sources of truth and update timing errors
- Registering for non-mutually exclusive state updates is 1) non-declarative so generally harder to reason about than a single state object and 2) if multiple components register for the same updates, there can be duplication in assumptions of logic (e.g. if we receive `onAuthenticated`, does each component remember to check if a profile is available?).

tbh, I'm surprised Fenix uses these patterns: iiuc, the MVI-inspired architecture is intended to represent all model state within the state tree, removing the classes of errors I mention above.

> My general recommendation would be to avoid modelling this after internals of the account manager, but rather build on top of data abstractions that it provides

That's my intention, yes – thank you for suggesting it. However, the internal FxA account state seems to be about as minimal as we can get: each of the states needs a separate representation in the UI.

> Generally speaking, any module you add on top of some other module should ideally introduce a higher level of abstraction. If it doesn't, there's a good chance that module isn't needed and just adds complexity.

Agreed. Our architecture does add complexity (through boilerplate and explicitly modeling dependency state) but it introduces simplicity by transforming it into declarative state and allowing us to look at model and UI independently.

btw, I appreciate your suggestions on best practices – keep them coming! 😄

app/src/main/java/org/mozilla/tv/firefox/fxa/FxaRepo.kt
```diff
@@ -0,0 +1,129 @@
...
+ * state of the user. If you want to initiate sign in, see [FxaLoginUseCase.beginLogin].
+ *
+ * Devs should use this class rather than interacting with the FxA library directly.
+ */
+class FxaRepo(
+    context: Context,
+    val accountManager: FxaAccountManager = newInstanceDefaultAccountManager(context)
+) {
+
+    enum class AccountState {
```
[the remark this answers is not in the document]
mcomella: These is addressed below.

app/src/main/java/org/mozilla/tv/firefox/fxa/FxaLoginUseCase.kt
```diff
@@ -0,0 +1,101 @@
...
+    private val sessionRepo: SessionRepo,
+    private val screenController: ScreenController,
+    private val sentryIntegration: SentryIntegration = SentryIntegration
+) {
+
+    /**
+     * Opens the browser screen and loads the FxA login URL to begin the login flow.
+     */
+    fun beginLogin(fragmentManager: FragmentManager) {
+        // TODO: should we throw an error if we're already authenticated when this is called?
```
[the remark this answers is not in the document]
mcomella: @Baron-Severin What do you mean by a success notification? We call this in response to a UI event and my reason for wanting to throw an error if we're already authenticated is that the UI should not be shown if we're already authenticated so we should never receive this event.

app/src/main/java/org/mozilla/tv/firefox/fxa/FxaRepo.kt
Baron-Severin: Sounds worth talking about in person.  I could go either way on it.

app/src/main/java/org/mozilla/tv/firefox/telemetry/SentryIntegration.kt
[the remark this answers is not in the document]
Baron-Severin: Sure, works for me.

app/src/main/java/org/mozilla/tv/firefox/fxa/FxaLoginUseCase.kt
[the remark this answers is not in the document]
Baron-Severin: Toast sounds fine to me for now.

app/src/main/java/org/mozilla/tv/firefox/fxa/FxaLoginUseCase.kt
Baron-Severin: To be clear, I meant a local extension function, not an app-wide one.  The problem here is that we're using Strings when we should probably use a more expressive type.  But I was already on the fence about this change anyway, and I'm happy to leave it as is.

app/src/main/java/org/mozilla/tv/firefox/fxa/FxaLoginUseCase.kt
Baron-Severin: Haha, yeah.  I consider sometimes writing a small `filterMap` operator that handles this part for us.  I've been waffling back and forth on that one for a while....

app/src/main/res/layout/fragment_navigation_overlay_top_nav.xml
mcomella: That makes sense to me and it's good to be consistent. Thanks for calling this out.

app/src/main/java/org/mozilla/tv/firefox/fxa/FxaLoginUseCase.kt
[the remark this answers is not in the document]
mcomella: I filed https://github.com/mozilla-mobile/firefox-tv/issues/2551

app/src/main/res/layout/fragment_navigation_overlay_top_nav.xml
mcomella: I tried wrapping this view and it doesn't just work so, instead of spending 30min debugging, I'm just going to leave it as is. I'm not concerned because this view's use case is trivial – it's on or it's off – and doesn't manage complex visibility state with animations like our previous experiment.

We can use this as a learning experiment, if we run into problems because of this later, we can be even more careful to use KillswitchLayout.

mcomella:
@Baron-Severin Did this need another review pass?

dnarcese:
@mcomella I am upgrading to a-c 4.0.0 today. Which one do you want to land first?

mcomella:
> @mcomella I am upgrading to a-c 4.0.0 today (#2554). Which one do you want to land first?

@dnarcese I landed your a-c upgrade already - thanks for checking in. :)

mcomella:
@Baron-Severin FYI: I added a small commit to rebase onto a-c 4.0.0, which changed the initial configuration slightly.

Issue #2503, opened 2019-07-09
Bitbar UI test fails with channels experiment enabled


npark-mozilla:
See
https://mozilla.testdroid.com/#testing/device-session/296208/758804/15814976

This still happens after the Amazon OS update screen is cleared. (Although now the error happens on different scripts)

### Device information
* Fire TV device: 4K stick / Gen 2 Stick
* Latest affected Firefox version: 3.11
* Earliest affected Firefox version: 3.11
* Turbo Mode: True

npark-mozilla:
Turns out, the new welcome screen for channels was not handled in the UI test, and the device in Bitbar was participating in the fretboard experiement.  In our UI test, this dialog should be handled during the first launch, along with the 'Enable Turbo' dialog as well.

npark-mozilla:
Above was taken care of with the `TV_ONBOARDING_SHOWN_PREF` to true, but another test fails while accessing settings, and it's because we hardcoded the number of keypresses to get to the settings in https://github.com/mozilla-mobile/firefox-tv/blob/master/app/src/androidTest/java/org/mozilla/tv/firefox/ui/robots/NavigationOverlayRobot.kt#L173.  When channels experiement is enabled, this would definitely fail to access the setting.

mcomella:
I believe all our UI tests are fixed now.

PR #2515, opened 2019-07-10
Fix Bitbar test failures in relation to Channels Experiment

npark-mozilla:
Connected to #2503

## Checklist

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -44,7 +44,7 @@ tasks:
           && ./tools/taskcluster/upload-coverage-report.sh
           && ./tools/taskcluster/download-firebase-sdk.sh
           && ./tools/taskcluster/google-firebase-testlab-login.sh
-          && ./tools/taskcluster/execute-firebase-test.sh system/debug app-system-debug model=Nexus9,version=25,orientation=landscape
+          && ./tools/taskcluster/execute-firebase-test.sh system/debug app-system-debug model=sailfish,version=25,orientation=landscape
       artifacts:
         'public/reports':
           type: 'directory'
--- a/app/src/androidTest/java/org/mozilla/tv/firefox/ui/robots/NavigationOverlayRobot.kt
+++ b/app/src/androidTest/java/org/mozilla/tv/firefox/ui/robots/NavigationOverlayRobot.kt
@@ -7,6 +7,7 @@ package org.mozilla.tv.firefox.ui.robots
 import android.net.Uri
 import android.os.Build
 import androidx.test.espresso.Espresso.onView
+import androidx.test.espresso.NoMatchingViewException
 import androidx.test.espresso.action.ViewActions.clearText
 import androidx.test.espresso.action.ViewActions.click
 import androidx.test.espresso.action.ViewActions.closeSoftKeyboard
@@ -29,6 +30,8 @@ import mozilla.components.support.android.test.espresso.assertIsDisplayed
 import mozilla.components.support.android.test.espresso.assertIsEnabled
 import mozilla.components.support.android.test.espresso.assertIsSelected
 import mozilla.components.support.android.test.espresso.click
+import mozilla.components.support.android.test.espresso.matcher.hasFocus
+import org.hamcrest.CoreMatchers.allOf
 import org.hamcrest.CoreMatchers.containsString
 import org.junit.Assert.assertTrue
 import org.mozilla.tv.firefox.R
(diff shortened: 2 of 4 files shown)


app/src/androidTest/java/org/mozilla/tv/firefox/ui/robots/NavigationOverlayRobot.kt
```diff
@@ -56,14 +59,18 @@ class NavigationOverlayRobot {
...
+        var settingsTileSelected = false
+
         device.apply {
-            // This will need to change if the button layout changes. However, such layout
-            // changes are infrequent, and updating this will be easy.
-            repeat(4) { pressDPadDown() }
+            while (!settingsTileSelected) {
+                try {
+                    onView(allOf(withId(R.id.settings_cardview), hasFocus(true))).assertIsDisplayed(true)
+                    settingsTileSelected = true
```
liuche: This looks good to me!

Though I'm a little confused why the number of clicks is different - doesn't our UI stay the same regardless of what device it's on?

app/src/androidTest/java/org/mozilla/tv/firefox/ui/robots/NavigationOverlayRobot.kt
```diff
@@ -171,11 +178,18 @@ class NavigationOverlayRobot {
         }
 
         fun linearNavigateToTelemtryTileAndOpen(interact: SettingsRobot.() -> Unit): SettingsRobot.Transition {
```
liuche: Hmm, I don't remember why I wrote it this way, but now that both of these are changing and actually look almost the same, maybe we should replace this with `linearNavigateToSettings` + `openTelemetryTile` - what do you think?

app/src/androidTest/java/org/mozilla/tv/firefox/ui/screenshots/SettingsTest.java
```diff
@@ -43,39 +45,48 @@ public void tearDown() {
...
-    public void showSettingsViews() throws InterruptedException {
-        onView(withId(R.id.navUrlInput)).check(matches(isDisplayed()));
-
-        // current settings list view
-        onView(withId(R.id.container_web_render)).check(matches(isDisplayed()));
-
+    public void showSettingsUsageData() throws InterruptedException {
         linearNavigateToSettingsChannel();
 
-        onView(withId(R.id.settings_tile_telemetry)).check(matches(isDisplayed()));
```
liuche: Should we move this `onView` check lower, to verify we're clicking into the Telemetry tile? (in case the tile order changes)

.taskcluster.yml
```diff
@@ -44,7 +44,7 @@ tasks:
           && ./tools/taskcluster/upload-coverage-report.sh
           && ./tools/taskcluster/download-firebase-sdk.sh
           && ./tools/taskcluster/google-firebase-testlab-login.sh
-          && ./tools/taskcluster/execute-firebase-test.sh system/debug app-system-debug model=Nexus9,version=25,orientation=landscape
+          && ./tools/taskcluster/execute-firebase-test.sh system/debug app-system-debug model=sailfish,version=25,orientation=landscape
```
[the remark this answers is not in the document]
npark-mozilla: this is Pixel 1, not a virtual device, but an actual hardware. still using API 25.  I looked at the video for Nexus 9, and it was lagging badly. I remember it had that problem before with virtual devices, so I switched to an hardware that supports API 25

app/src/androidTest/java/org/mozilla/tv/firefox/ui/robots/NavigationOverlayRobot.kt
npark-mozilla: No, when the channel experiment is enabled, there are more rows between Pocket tile row and Settings menu row. since it gets enabled randomly, we have to dynamically adjust the number of DPADDOWN clicks.

app/src/androidTest/java/org/mozilla/tv/firefox/ui/robots/NavigationOverlayRobot.kt
npark-mozilla: Yes I was thinking that, but they belong to a different robot, so you can't reuse `linearNavigateToSettings` there.  We should refactor somehow, but that's outside the scope of this PR I think.

app/src/androidTest/java/org/mozilla/tv/firefox/ui/screenshots/SettingsTest.java
npark-mozilla: Since there is another test that checks the layout of the UI test, and this test only collects screenshots, I figured I could remove all the element checks to simplify it. (we manually check screenshots to make sure they match the title) We could click the DOM with the right tile ID instead of using `pressDPadCenter()` if we want to be sure.  Should I change it to work that way?

PR #2545, opened 2019-07-13
2543 Only show TV Guide in EN-US

wmichelle024:
## Checklist

- [x] Add **QA labels** on the associated issue (not this PR; `qa-ready` or `qa-denied`)

### DIFF
--- a/app/src/main/java/org/mozilla/tv/firefox/ext/Locale.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/ext/Locale.kt
@@ -3,7 +3,6 @@ package org.mozilla.tv.firefox.ext
 import org.mozilla.tv.firefox.architecture.KillswitchLocales
 import java.util.Locale
 
-
 /**
  * Returns true if the following is true for _any_ of the elements in [allowedLocales]:
  * 1) Language == [this] language
--- a/app/src/main/java/org/mozilla/tv/firefox/ext/Locale.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/ext/Locale.kt
@@ -1,6 +1,7 @@
 package org.mozilla.tv.firefox.ext
 
-import java.util.*
+import java.util.Locale
+
 
 /**
  * Returns true if the following is true for _any_ of the elements in [allowedLocales]:
@@ -13,6 +14,6 @@ fun Locale.languageAndMaybeCountryMatch(allowedLocales: Array<out Locale>?): Boo
         val languageMatches = allowed.language == this.language
         val countryMatches = allowed.country.isEmpty() ||
                 allowed.country == this.country
-        return languageMatches && countryMatches
+        return@any languageMatches && countryMatches
     }
 }
--- a/app/src/main/java/org/mozilla/tv/firefox/ext/Locale.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/ext/Locale.kt
@@ -1,0 +1,14 @@
+package org.mozilla.tv.firefox.ext
+
+import java.util.*
+
+
+fun Locale.languageAndMaybeCountryMatch(allowedLocales: Array<out Locale>?): Boolean {
+    allowedLocales ?: return false
+    return allowedLocales.any { allowed ->
+        val languageMatches = allowed.language == this.language
+        val countryMatches = allowed.country.isEmpty() ||
+                allowed.country == this.country
+        return languageMatches && countryMatches
+    }
+}
--- a/app/src/main/java/org/mozilla/tv/firefox/ext/Locale.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/ext/Locale.kt
@@ -2,7 +2,11 @@ package org.mozilla.tv.firefox.ext
 
 import java.util.*
 
-
+/**
+ * Returns true if the following is true for _any_ of the elements in [allowedLocales]:
+ * 1) Language == [this] language
+ * 2) No country selected || [this] country is selected country
+ */
 fun Locale.languageAndMaybeCountryMatch(allowedLocales: Array<out Locale>?): Boolean {
     allowedLocales ?: return false
     return allowedLocales.any { allowed ->
--- a/app/src/main/java/org/mozilla/tv/firefox/ext/Locale.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/ext/Locale.kt
@@ -1,5 +1,6 @@
 package org.mozilla.tv.firefox.ext
 
+import org.mozilla.tv.firefox.architecture.KillswitchLocales
 import java.util.Locale
 
 
@@ -15,5 +16,12 @@ fun Locale.languageAndMaybeCountryMatch(allowedLocales: Array<out Locale>?): Boo
         val countryMatches = allowed.country.isEmpty() ||
                 allowed.country == this.country
         return@any languageMatches && countryMatches
+    }
+}
+
+fun Locale.languageAndMaybeCountryMatch(allowedLocales: KillswitchLocales): Boolean {
+    return when (allowedLocales) {
+        is KillswitchLocales.All -> true
+        is KillswitchLocales.ActiveIn -> this.languageAndMaybeCountryMatch(allowedLocales.locales)
     }
 }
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/mozilla/tv/firefox/ext/Locale.kt
```diff
@@ -0,0 +1,14 @@
+package org.mozilla.tv.firefox.ext
```
mcomella: nit: needs license

app/src/main/java/org/mozilla/tv/firefox/ext/Locale.kt
```diff
@@ -2,7 +2,11 @@ package org.mozilla.tv.firefox.ext
 
 import java.util.*
 
-
+/**
+ * Returns true if the following is true for _any_ of the elements in [allowedLocales]:
```
mcomella: optional nit: this comment might be harder to read because of how closely it mirrors the code. Also, mirroring the code means the comment is liable to get outdated. However, I understand why communicating this behavior to the caller is important and I'm not sure there's a better solution. I wonder if this comment would be simpler (e.g. more declarative, don't have to wrap my head around a loop) in prose-like english.

app/src/test/java/org/mozilla/tv/firefox/ext/LocaleKtTest.kt
```diff
@@ -0,0 +1,65 @@
+package org.mozilla.tv.firefox.ext
```
mcomella: nit: needs license

app/src/main/java/org/mozilla/tv/firefox/ext/Locale.kt
```diff
@@ -13,6 +14,6 @@ fun Locale.languageAndMaybeCountryMatch(allowedLocales: Array<out Locale>?): Boo
         val languageMatches = allowed.language == this.language
         val countryMatches = allowed.country.isEmpty() ||
                 allowed.country == this.country
-        return languageMatches && countryMatches
+        return@any languageMatches && countryMatches
```
mcomella: Omg. 🤣

app/src/main/java/org/mozilla/tv/firefox/ext/Locale.kt
```diff
@@ -17,3 +18,10 @@ fun Locale.languageAndMaybeCountryMatch(allowedLocales: Array<out Locale>?): Boo
         return@any languageMatches && countryMatches
     }
 }
+
+fun Locale.languageAndMaybeCountryMatch(allowedLocales: KillswitchLocales): Boolean {
```
mcomella: nit: I wonder if we should point the kdoc for this function to the other function.

app/src/test/java/org/mozilla/tv/firefox/ext/LocaleKtTest.kt
```diff
@@ -62,4 +67,45 @@ class LocaleKtTest {
...
+        assertTrue(Locale.ENGLISH.languageAndMaybeCountryMatch(allowed))
+        assertTrue(Locale.GERMAN.languageAndMaybeCountryMatch(allowed))
+        assertTrue(Locale.CHINESE.languageAndMaybeCountryMatch(allowed))
+    }
+
+    @Test
+    fun `WHEN KillswitchLocale ActiveIn is passed THEN overload should be called`() {
+        // Mock extension functions in this class
+        // If this test fails, check if this class has been moved, causing this string to be incorrect
+        mockkStatic("org.mozilla.tv.firefox.ext.LocaleKt")
```
mcomella: nit: We can put the string in a constant at the top of the file.

Thanks for the handy comment but unfortunately if we ever call this function again, we'll be duplicating it a lot. :\

app/src/test/java/org/mozilla/tv/firefox/ext/LocaleKtTest.kt
```diff
@@ -62,4 +67,45 @@ class LocaleKtTest {
...
+                Locale.CHINA,
+                Locale.ENGLISH,
+                Locale.GERMAN,
+                Locale.CHINESE
+        )
+
+        locales.forEach { it.languageAndMaybeCountryMatch(allowed) }
+
+        // Verify that languageAndMaybeCountryMatch is called with an array of locales (i.e., the overload)
+        locales.forEach { verify(exactly = 1) { it.languageAndMaybeCountryMatch(any<Array<Locale>>()) } }
```
mcomella: nit: can we match `allowed` instead of `any<Array<Locale>>())`?

app/src/main/java/org/mozilla/tv/firefox/utils/Settings.kt
```diff
@@ -51,10 +52,11 @@ class Settings private constructor(context: Context) {
 
     fun shouldShowTVOnboarding(localeManager: LocaleManager, context: Context): Boolean {
```
mcomella: Should we have a test for this? Perhaps as a follow-up.

app/src/main/java/org/mozilla/tv/firefox/utils/Settings.kt
```diff
@@ -51,10 +52,11 @@ class Settings private constructor(context: Context) {
...
         val channelConfig = ChannelConfig.getTvGuideConfig(context)
+        val currentLocale = localeManager.getCurrentLocale(context)
 
         return !preferences.getBoolean(ChannelOnboardingActivity.TV_ONBOARDING_SHOWN_PREF, false) &&
                 channelConfig.isEnabledInCurrentExperiment &&
-                localeManager.currentLanguageIsEnglish(context)
+                currentLocale.languageAndMaybeCountryMatch(channelConfig.enabledInLocales)
```
mcomella: The logic in these two lines (isEnabled && languageAndMaybeCountryMatch) is duplicated in KillswitchLayout.setVisibility. The latter implementation is more complex but I wonder if there's a way we can consolidate them. Can you investigate in the follow-up for the nits?

app/src/main/java/org/mozilla/tv/firefox/ext/Locale.kt
Baron-Severin: There's got to be a better way to describe this one, but we spent a considerable amount of time trying to find it without any luck.  This was the clearest we could manage.

app/src/main/java/org/mozilla/tv/firefox/ext/Locale.kt
Baron-Severin: Yep....

This didn't cause any production bugs btw, but would have if we hadn't caught it here.

app/src/test/java/org/mozilla/tv/firefox/ext/LocaleKtTest.kt
Baron-Severin: Yeah, unfortunately true.  I'm not 100% happy with this one, but didn't see a better way to do it.

app/src/test/java/org/mozilla/tv/firefox/ext/LocaleKtTest.kt
Baron-Severin: We could match `allowed.locales`, but I think it would be a slightly less useful test.  I think we want to verify that the other method gets called exactly once, regardless of any other implementation changes.

app/src/main/java/org/mozilla/tv/firefox/utils/Settings.kt
Baron-Severin: Agreed.  The first fix I wanted to make for this was creating a new class that was experiment and locale aware, that would return a one line `shouldDisplayX` for each feature.  But there were are a few organizational complications there, as well as some refactor work to clarify the difference between that class and `ExperimentProvider`.  We decided to scale back the work, since it felt a little bit aspirational.

app/src/main/java/org/mozilla/tv/firefox/utils/Settings.kt
Baron-Severin: I started writing a test for this, but it involves mocking context several layers deep.  The function is mostly declarative, so I don't think it's worth the time investment as is.  I think it would be better to use the time to separate out the logic entirely, as discussed in https://github.com/mozilla-mobile/firefox-tv/pull/2545#discussion_r303185118.

app/src/main/java/org/mozilla/tv/firefox/utils/Settings.kt
mcomella: In that case, it'd be good to add comments to both sections of code mentioning the duplication and where to find it. That way if one changes, there's a larger chance the other will also change.

Issue #2386, opened 2019-06-06
Investigation: Video Player Issues


athomasmoz:
### Why/User Benefit/User Problem

As a user, I can't play videos on the any of the sports sites from Alexa's top sites

### Acceptance Criteria (how do I know when I’m done?)

* We have identified causes for the video playback not working on sport sites

### Output

* A list of probable causes for the issues we see with video playback in the sports section. Each cause includes
  * Which sites are affected
  * What's the next action to take for this (logging a bug (Amazon, site compat), doing a workaround or bug fix)
* Verify this is Chromium 70

### Not included

* Implementing workarounds is not included

### Timebox

4 days

devinreams:
Additional details provided on one specific issue related to handling of certificates (private/confidential): 

https://docs.google.com/document/d/1esS8L4UuXgxwlEkjeMDGZjch7YMIAhAjT_vCE3qeST0/edit?usp=sharing

liuche:
Note to self: also investigate whether mimicking Silk's User Agent would fix any of these?

kglazko:
STR
Compare video playback quality of properly formatted site: https://kglazko.github.io/testsite/clean.html
and improperly formatted site: https://kglazko.github.io/testsite/

Expected:
Video on improperly formatted site plays normally in fullscreen mode like on Silk. 
* The page is locked and doesn't allow side-to-side scrolling. 
* The video plays correctly and is not cut off minimized
* The video expands to fill the screen when fullscreen and is not cut off

Actual:
* Video on improperly formatted site isn't locked and allows side-to-side scrolling
* Video will often be cut off or not centered when going into full screen with large areas of black showing similar to the affected disapproved sites.

Notes:
On clean version, this doesn't happen. Need to investigate why Silk "locks" the improperly formatted page and doesn't allow side-to-side scrolling and why we don't do that. It is possible that formatting of the page somehow affects the rendering of the fullscreen video.

mcomella:
To follow up on Kate's research, maybe we can use `WebSettings.setLoadWithOverviewMode` or `WebSettings.getUseWideViewPort` to fix this issue. Perhaps there are other relevant methods in `WebView`, `WebSettings`, `WebChromeClient`, or `WebViewClient` too.

dnarcese:
Things I investigated:
[...]

Issue #2526, opened 2019-07-11
[2386] Mitigate shifted videos


dnarcese:
Related to #2386

### Acceptance Criteria

* Shift in videos on GolfDigest is fixed
* Golf Digest and other sites in the MVP categories list are retested
* Any sites where video playback is fixed are added to the homescreen

### Steps to reproduce
Visit https://video.golfdigest.com/watch/golf-digest-stephen-curry

### Expected behavior
Video is centred in container while minimized and fullscreen.
### Actual behavior
Video is only on right half minimized and in top right corner in fullscreen.

Investigating the css for video player on https://video.golfdigest.com/watch/golf-digest-stephen-curry:
<img width="209" alt="image-20190710173213342" src="https://user-images.githubusercontent.com/11432165/61006971-ae0c0f00-a339-11e9-95da-0fd7d6e78c98.png">

It seems that the `transform` property is not being respected because checking and unchecking that value does not change anything. But changing the `left` value moves the video. 
Unchecking both the `transform` and `left` properties puts the video in the correct place. Severin mentioned `transform/left` is a common tactic to center content in CSS.

This was discovered because the video was offset in the WebView by an amount similar to the `<video>` CSS property `offsetLeft`: maybe we can use these properties to determine when we should or should not be trying to fix this issue on a given web page.

### Potential options for fixes
* Inject JS to handle CSS and remove both `transform` and `left`
* Other options?

This should be timeboxed to _____.

### Device information
* Fire TV device: ?
* Latest affected Firefox version: ?
* Earliest affected Firefox version: ?
* Turbo Mode: ?

Baron-Severin:
As a hacky testing solution, I injected a script that would find all `<video>` elements and clear their `transform` and `left` properties once per second.  I picked out five sites that previously showed videos out of position (Sbnation, CNBC, NY Post, BBC, and TIME), and two that previously worked (CNN, Fox News).

Video problems on CNBC were fixed.

Video positions on all other sites (both those that worked and failed) seemed to be unaffected.

Baron-Severin:
I tested all sports sites that either 1) previously worked, or 2) failed due to weird video placement.  These were my results

Previously broken (offset videos)
CBS sports - fixed
NHL - not fixed
Deadspin - couldn't even get any videos to start
Ncaa - full screened, still cut off. Spent time trying to fix, no luck
Sbnation - fixed
WWE - not fixed. Full screened video is still offcenter
[...]

PR #2557, opened 2019-07-17
Issue #2386: partial fix for shifted videos

Baron-Severin:
See commit for more details regarding what was attempted and how effective the changes were.

Connected to #???

## Testing and Review Notes

- Using a release build (for Chromium 70)
- Test sites from [here](https://docs.google.com/spreadsheets/d/1uuYQzG5RZKeB1UIdVXiPIacumJb3NN_YA_plKLF0nBA/edit#gid=0) for compatibility
- We should expect no regressions, and a few improvements (earlier testing showed improvements on 2/13 tested sites)

## Checklist

Only partially satisfied, but this is the best we've managed
- [X] Add **testing notes and/or screenshots** in PR description to help guide all potential reviewers
- [X] Add a **CHANGELOG entry** if applicable
- [X] Add **QA labels** on the associated issue (not this PR; `qa-ready` or `qa-denied`)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ It diverges in the following ways:
 ### Changed
 
 ### Fixed
-- Videos on many sites are displayed offset, with a black strip pushing content off of the page.  Solved one of the issues that led to that, fixing some broken videos (#2386)
+- Prevent some videos from being offset to the right, with a black strip on the left. (#2526)
 
 ## 4.0 - 2019-07-15
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ It diverges in the following ways:
 ### Changed
 
 ### Fixed
+- Partial fix for poorly formatted videos (#2386)
 
 ## 4.0 - 2019-07-15
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ It diverges in the following ways:
 ### Changed
 
 ### Fixed
-- Partial fix for poorly formatted videos (#2386)
+- Videos on many sites are displayed offset, with a black strip pushing content off of the page.  Solved one of the issues that led to that, fixing some broken videos (#2386)
 
 ## 4.0 - 2019-07-15
 
--- a/app/src/main/java/org/mozilla/tv/firefox/ext/Js.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/ext/Js.kt
@@ -53,7 +53,7 @@ var _firefoxTV_isPlaybackStateObserverLoaded;
 
     function onDOMChangedForVideos() {
         addPlaybackStateListeners();
-        adjustVideoCSS();
+        removeNegativeTranslationCSSCentering();
         syncPlaybackState();
     }
 
@@ -91,15 +91,31 @@ var _firefoxTV_isPlaybackStateObserverLoaded;
         javaInterface.syncPlaybackState(isVideoPresent, isPlaying, positionSeconds, playbackRate);
     }
 
-    function adjustVideoCSS() {
+    /*
+     * When a video has left=50% & transform=translateX(-50%) properties (a common CSS centering
+     * trick), videos are pushed outside their containers: if we see similar properties, we remove
+     * them as a workaround. #2526
+     */
+    function removeNegativeTranslationCSSCentering() {
         const rawVideoElements = document.getElementsByTagName('video');
         // Extract elements from HTMLCollection and collect them into an Array
         const videoElements = new Array(...rawVideoElements);
 
         videoElements.forEach((vid) => {
-            vid.style.left = '0px';
-            vid.style.transform = 'translate(0px, 0px)';
-            console.log('FFTV - adjusted video tag CSS')
+            const style = window.getComputedStyle(vid);
+            const left = parseFloat(style.left); /* style.left is e.g. "320px" */
+
+            /* The matrix's 4th column, 1st row (m41) represents the X translation.
+             * See: https://stackoverflow.com/a/42267468
+             */
+            const translateX = new DOMMatrix(style.transform).m41;
+
+            const translateNegatesLeft = left + translateX === 0;
+            if (translateNegatesLeft) {
+                vid.style.left = '0px';
+                vid.style.transform = 'translate(0px, 0px)';
+                console.log('FFTV - adjusted video tag CSS')
+            }
         });
     }
 
--- a/app/src/main/java/org/mozilla/tv/firefox/ext/Js.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/ext/Js.kt
@@ -97,24 +97,21 @@ var _firefoxTV_isPlaybackStateObserverLoaded;
      * them as a workaround. #2526
      */
     function removeNegativeTranslationCSSCentering() {
-        const rawVideoElements = document.getElementsByTagName('video');
-        // Extract elements from HTMLCollection and collect them into an Array
-        const videoElements = new Array(...rawVideoElements);
+        const videoElements = document.querySelectorAll('video');
 
         videoElements.forEach((vid) => {
             const style = window.getComputedStyle(vid);
             const left = parseFloat(style.left); /* style.left is e.g. "320px" */
 
             /* The matrix's 4th column, 1st row (m41) represents the X translation.
-             * See: https://stackoverflow.com/a/42267468
-             */
+             * See: https://stackoverflow.com/a/42267468 */
             const translateX = new DOMMatrix(style.transform).m41;
 
             const translateNegatesLeft = left + translateX === 0;
             if (translateNegatesLeft) {
                 vid.style.left = '0px';
                 vid.style.transform = 'translate(0px, 0px)';
-                console.log('FFTV - adjusted video tag CSS')
+                console.log('FFTV - adjusted video tag CSS');
             }
         });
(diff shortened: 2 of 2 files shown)


app/src/main/java/org/mozilla/tv/firefox/ext/Js.kt
```diff
@@ -90,6 +91,18 @@ var _firefoxTV_isPlaybackStateObserverLoaded;
         javaInterface.syncPlaybackState(isVideoPresent, isPlaying, positionSeconds, playbackRate);
     }
 
+    function adjustVideoCSS() {
+        const rawVideoElements = document.getElementsByTagName('video');
+        // Extract elements from HTMLCollection and collect them into an Array
+        const videoElements = new Array(...rawVideoElements);
```
mcomella: nit: `querySelectorAll('video')` will return a `NodeList` that has a `forEach` method on it, invalidating the need to convert the data type. See elsewhere in this file.

If you don't use that, `Array.from(rawVideoElements)`, as a function call, is probably more intuitive for us than its `...` syntactic equivalent here.

app/src/main/java/org/mozilla/tv/firefox/ext/Js.kt
```diff
@@ -54,6 +54,7 @@ var _firefoxTV_isPlaybackStateObserverLoaded;
     function onDOMChangedForVideos() {
```
mcomella: THe issue number on this commit is incorrect: it should be #2526.

app/src/main/java/org/mozilla/tv/firefox/ext/Js.kt
```diff
@@ -90,6 +91,18 @@ var _firefoxTV_isPlaybackStateObserverLoaded;
...
     }
 
+    function adjustVideoCSS() {
+        const rawVideoElements = document.getElementsByTagName('video');
+        // Extract elements from HTMLCollection and collect them into an Array
+        const videoElements = new Array(...rawVideoElements);
+
+        videoElements.forEach((vid) => {
+            vid.style.left = '0px';
+            vid.style.transform = 'translate(0px, 0px)';
```
mcomella: It seems risky to modify the CSS unconditionally in the event that web devs are doing something less innocuous using these CSS parameters. My thinking was that we'd only do this if we saw the conditions that seem to cause problems: that is `left 50% + translateX -50%` (or maybe something slightly broader).

What do you think? Does that seem reasonable?

app/src/main/java/org/mozilla/tv/firefox/ext/Js.kt
```diff
@@ -54,6 +54,7 @@ var _firefoxTV_isPlaybackStateObserverLoaded;
     function onDOMChangedForVideos() {
         addPlaybackStateListeners();
         syncPlaybackState();
+        adjustVideoCSS();
```
mcomella: It's _possible_ adjusting the video CSS will change the playback state: perhaps we should do this first, before the other functions.

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ It diverges in the following ways:
 ### Changed
 
 ### Fixed
+- Partial fix for poorly formatted videos (#2386)
```
mcomella: Can you be more specific? Why are they poorly formatted and what did we fix?

app/src/main/java/org/mozilla/tv/firefox/ext/Js.kt
Baron-Severin: Seems reasonable to me.

CHANGELOG.md
Baron-Severin: I've normally kept these short, assuming that if someone wanted more information they would just look at the commit message.  But I can expand it.

mcomella: I think they should be short but informative: ideally, a person who is experiencing this bug can read this and know if it addressed their specific fullscreen problem or another one.

Baron-Severin: If I recall correctly, this only resolves the issue for 2 out of 13 sites where it's experienced, so that's unfortunately going to be next to impossible in this case.  I did add some more detail though.

app/src/main/java/org/mozilla/tv/firefox/ext/Js.kt
Baron-Severin: This turned out to be a lot more frustrating than expected.  Finding `left` involves dividing `computedLeft` by the `width` of the parent element, which is annoying but fairly trivial.  Checking transform is substantially more complex (querying `transform` on one of these broken sites returns the string `"matrix(1, 0, 0, 1, -362, 0)"`, and I'm not even yet sure if that pattern holds true for all examples).

I spent a full day QAing this change, and it didn't seem to negatively impact any sites.  We need a rendering engine fix for the long term, but for the short term I don't think it's a good use of time to keep building up CSS injection logic.  This code is necessary for us right now, but it's just inherently really, really bad.  I think we should ship this code, or cut the feature and wait on the rendering engine fix.

mcomella: via https://stackoverflow.com/a/42267468, it looks like we can get the translateX pixel value (through the transform matrix) by:
```js
var style = window.getComputedStyle(myElement);
var matrix = new DOMMatrix(style.transform);
console.log('translateX: ', matrix.m41);
```

afaict (disclaimer: I'm slightly distracted), we can identify if it's "translateX(-50%)" by comparing this translateX value against `style.transformOrigin` (where we're translating from) which looks like, "362px 203.633px" (this value might be equivalent to left, I'm not sure). If transformOrigin X is equivalent to that previous translate X value, we're likely in a `translateX(-50%)` situation. I think that's everything we need.

> I spent a full day QAing this change, and it didn't seem to negatively impact any sites.

Without conditionally making this change, we have a high probability of breaking page layout on some site. I'm concerned that the sites that break will affect our non-YT users and they'll leave. Given that our users have a low frequency of reporting bugs and our limited ability to test a wide range of sites, I'm also concerned that we won't find these bugs for a while, encouraging more users to leave.

> but for the short term I don't think it's a good use of time to keep building up CSS injection logic.

I don't think it's a much larger investment in time to add the conditional statement. I agree with your sentiment – that CSS injection is fragile, risky, and generally not a good idea – but I think the conditional change actually makes it less risky and a better change because it'll affect fewer pages and we can still confirm it fixes the specific sites we expect it to.

Let's follow-up IRL. I think we should pair through building the conditional statement to increase our confidence in the solution.

Baron-Severin: You're right, and that's a better approach.  Thank you for suggesting it, we can work through it later today.

app/src/main/java/org/mozilla/tv/firefox/ext/Js.kt
```diff
@@ -91,15 +91,31 @@ var _firefoxTV_isPlaybackStateObserverLoaded;
...
-    function adjustVideoCSS() {
+    /*
+     * When a video has left=50% & transform=translateX(-50%) properties (a common CSS centering
+     * trick), videos are pushed outside their containers: if we see similar properties, we remove
+     * them as a workaround. #2526
+     */
+    function removeNegativeTranslationCSSCentering() {
         const rawVideoElements = document.getElementsByTagName('video');
         // Extract elements from HTMLCollection and collect them into an Array
         const videoElements = new Array(...rawVideoElements);
```
mcomella: nit: see previous comment: looks like this wasn't changed.

I realized this syntax is also impossible to google search so it's actually really bad to keep. :P Otherwise I might say it wasn't worth it.

app/src/main/java/org/mozilla/tv/firefox/ext/Js.kt
```diff
@@ -91,15 +91,31 @@ var _firefoxTV_isPlaybackStateObserverLoaded;
...
         videoElements.forEach((vid) => {
-            vid.style.left = '0px';
-            vid.style.transform = 'translate(0px, 0px)';
-            console.log('FFTV - adjusted video tag CSS')
+            const style = window.getComputedStyle(vid);
+            const left = parseFloat(style.left); /* style.left is e.g. "320px" */
+
+            /* The matrix's 4th column, 1st row (m41) represents the X translation.
+             * See: https://stackoverflow.com/a/42267468
+             */
```
mcomella: nit: should be moved to previous line.

app/src/main/java/org/mozilla/tv/firefox/ext/Js.kt
```diff
@@ -91,15 +91,31 @@ var _firefoxTV_isPlaybackStateObserverLoaded;
...
+            const translateX = new DOMMatrix(style.transform).m41;
+
+            const translateNegatesLeft = left + translateX === 0;
+            if (translateNegatesLeft) {
+                vid.style.left = '0px';
+                vid.style.transform = 'translate(0px, 0px)';
+                console.log('FFTV - adjusted video tag CSS')
```
mcomella: nit: semicolon

app/src/main/java/org/mozilla/tv/firefox/ext/Js.kt
Baron-Severin: Oh geez, sorry.  I swear I fixed that....

Baron-Severin:
> lgtm: paired through the final fix and tested on device together.
> 
> Do you mind verifying the 2nd of the 13 sites to make sure that works too?

Verified, looks good.

Issue #2441, opened 2019-06-14
Account Sign-in Page


athomasmoz:
### Why/User Benefit/User Problem

As an FFTV user, I want to log in to my Firefox account so that I can send a tab to my TV from another device once implemented

### Acceptance Criteria (how do I know when I’m done?)

* If user goes to log in page on FFTV,
  * They can dpad navigate to the input fields
  * Input fields will pop up keyboard

### Tasks

* Enable dpad navigation to the input fields
* Verify keyboard pop ups for input fields, placeholder matches hint on input
* Verify clicking next from email address will automatically advance to password field. If not, log a ticket in FxA Github (Shane Tomlinson)
* Determine if there's additional work needed to sign in (comment or log bugs)
* Determine if there's additional work needed to handle errors (comment or log bugs)
* Determine if there's additional work needed to persist sign-in across app restarts (comment or log bugs)

### Not included

* Entry point for sign in from home screen
* Sign in errors should already be implemented in the webview, so should not require additional implementation from our team

liuche:
If we need any changes to the FxA signin flow (like auto-advance of input fields), we should ping A-S team (Shane Tomlinson).

Baron-Severin:
* ~Enable dpad navigation to the input fields~
Done.  I'll push the PR up shortly
* ~Verify keyboard pop ups for input fields, placeholder matches hint on input~
Keyboard opens, placeholders match
* ~Verify clicking next from email address will automatically advance to password field. If not, log a ticket in FxA Github (Shane Tomlinson)~
Opened https://github.com/mozilla/fxa/issues/1735
* ~Determine if there's additional work needed to sign in (comment or log bugs)~
Login seems to work as expected
* ~Determine if there's additional work needed to handle errors (comment or log bugs)~
Opened https://github.com/mozilla/fxa/issues/1737
* ~Determine if there's additional work needed to persist sign-in across app restarts (comment or log bugs)~
Seems to work out of the box (after stopping and restarting, hitting the CC button calls onAuthenticated and onProfileUpdated)

Also opened https://github.com/mozilla/fxa/issues/1734 as a meta bug

Baron-Severin:
@athomasmoz I think we should probably block on https://github.com/mozilla/fxa/issues/1736, probably not the other two.  There are videos and descriptions in the issues so you can see for yourself, they're all linked in [the meta bug](https://github.com/mozilla/fxa/issues/1734).

athomasmoz:
[...]

PR #2564, opened 2019-07-19
Issue #2441: enable linear navigation on FxA login page

Baron-Severin:
Connected to #???

## Checklist

### DIFF
--- a/app/src/main/java/org/mozilla/tv/firefox/ext/String.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/ext/String.kt
@@ -85,3 +85,6 @@ val String.isUriYouTubeTV: Boolean
 val String.isUriYouTubeTvVideo: Boolean
     get() = this.isUriYouTubeTV &&
             this.contains("/watch/")
+
+val String.isUriMozSignIn: Boolean
+    get() = this.startsWith("https://accounts.firefox.com/authorization")
--- a/app/src/main/java/org/mozilla/tv/firefox/ext/String.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/ext/String.kt
@@ -86,5 +86,5 @@ val String.isUriYouTubeTvVideo: Boolean
     get() = this.isUriYouTubeTV &&
             this.contains("/watch/")
 
-val String.isUriMozSignIn: Boolean
+val String.isUriFxaSignIn: Boolean
     get() = this.startsWith("https://accounts.firefox.com/authorization")
--- a/app/src/main/java/org/mozilla/tv/firefox/webrender/cursor/CursorModel.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/webrender/cursor/CursorModel.kt
@@ -16,6 +16,7 @@ import io.reactivex.subjects.PublishSubject
 import org.mozilla.tv.firefox.ScreenControllerStateMachine
 import org.mozilla.tv.firefox.ScreenControllerStateMachine.ActiveScreen.WEB_RENDER
 import org.mozilla.tv.firefox.ext.isKeyCodeSelect
+import org.mozilla.tv.firefox.ext.isUriMozSignIn
 import org.mozilla.tv.firefox.ext.isUriYouTubeTV
 import org.mozilla.tv.firefox.ext.toDirection
 import org.mozilla.tv.firefox.framework.FrameworkRepo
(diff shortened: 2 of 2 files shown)


app/src/main/java/org/mozilla/tv/firefox/ext/String.kt
```diff
@@ -85,3 +85,6 @@ val String.isUriYouTubeTV: Boolean
 val String.isUriYouTubeTvVideo: Boolean
     get() = this.isUriYouTubeTV &&
             this.contains("/watch/")
+
+val String.isUriMozSignIn: Boolean
+    get() = this.startsWith("https://accounts.firefox.com/authorization")
```
mcomella: Should this also include the success screen? `accounts.firefox.com/oauth/success`?

app/src/main/java/org/mozilla/tv/firefox/webrender/cursor/CursorModel.kt
```diff
@@ -134,7 +135,9 @@ class CursorModel(
     ) { activeScreen, isVoiceViewEnabled, sessionState ->
         // We only display the cursor when the web content is active.
         val isWebRenderActive = activeScreen == WEB_RENDER
-        val doesWebpageHaveOwnNavControls = sessionState.currentUrl.isUriYouTubeTV || isVoiceViewEnabled
+        val doesWebpageHaveOwnNavControls = sessionState.currentUrl.isUriYouTubeTV ||
+            sessionState.currentUrl.isUriMozSignIn ||
```
mcomella: nit: I can see this list of sites growing larger: should we try to group them into a collection? e.g.:
```kotlin
val sitesWithOwnNavControls = setOf(
    YOUTUBE_URI,
    FXA_SYNC_IN_URI
)

...

val doesWebpageHaveOwnNavControls = sessionState.currentUrl in sitesWithOwnNavControls ||
        isVoiceViewEnabled
```

app/src/main/java/org/mozilla/tv/firefox/ext/String.kt
```diff
@@ -85,3 +85,6 @@ val String.isUriYouTubeTV: Boolean
 val String.isUriYouTubeTvVideo: Boolean
     get() = this.isUriYouTubeTV &&
             this.contains("/watch/")
+
+val String.isUriMozSignIn: Boolean
```
mcomella: nit: `isUriFxaSignIn` - both to be consistent with terminology in other files & just in case Mozilla introduces another layer of sign in (uh oh!).

mcomella: I feel weird exposing this on all strings but I don't have a better solution right now. It was a little better for YouTube because we do have many branches for YouTube but I doubt we'll have a similar case here.

app/src/main/java/org/mozilla/tv/firefox/ext/String.kt
Baron-Severin: I don't think they need to really interact with that screen, right?  Might be confusing if there isn't anything to focus.

app/src/main/java/org/mozilla/tv/firefox/ext/String.kt
Baron-Severin: Good thinking.

Baron-Severin: Agreed, but I think consistency is the more important element here.  We can keep it in mind and try to work on a better solution.

Baron-Severin: Agreed, but I think consistency is the more important element here.  We can keep it in mind and try to work on a better solution.

app/src/main/java/org/mozilla/tv/firefox/webrender/cursor/CursorModel.kt
Baron-Severin: Good point, but `isVoiceViewActive` doesn't fit that pattern, so it'll split this check up a bit.  I think this is a problem we can fix when it becomes an issue; it shouldn't be a hard refactor.

app/src/main/java/org/mozilla/tv/firefox/ext/String.kt
mcomella: Have you tried it? If there's nothing to focus, maybe nothing happens. I think the cursor can be confusing too: maybe they'll try to click on the "Success" icon.

Then again, once users hit this page, we should redirect them off of it so maybe this is a non-issue.

mcomella: Actually, yeah - we should just pick something and leave it for now because users probably won't see this page at all.

Baron-Severin: I ran through the whole process and didn't notice anything strange about this page, but we can take a closer look before we deploy it to users.  It's a quick and simple tweak.

PR #2571, opened 2019-07-22
Fix hanging tests

Baron-Severin:
Connected to #???

## Checklist

### DIFF
--- a/app/src/test/java/org/mozilla/tv/firefox/pocket/PocketVideoRepoTest.kt
+++ b/app/src/test/java/org/mozilla/tv/firefox/pocket/PocketVideoRepoTest.kt
@@ -7,10 +7,13 @@ package org.mozilla.tv.firefox.pocket
 import io.mockk.MockKAnnotations
 import io.mockk.every
 import io.mockk.impl.annotations.MockK
+import io.reactivex.android.plugins.RxAndroidPlugins
 import io.reactivex.observers.TestObserver
+import io.reactivex.schedulers.Schedulers
 import io.reactivex.subjects.BehaviorSubject
 import org.junit.Assert.assertEquals
 import org.junit.Before
+import org.junit.BeforeClass
 import org.junit.Ignore
 import org.junit.Test
 import org.junit.runner.RunWith
@@ -20,6 +23,15 @@ import org.robolectric.RobolectricTestRunner
 
 @RunWith(RobolectricTestRunner::class)
 class PocketVideoRepoTest {
+
+    companion object {
+        @JvmStatic
+        @BeforeClass
+        fun beforeClass() {
+            RxAndroidPlugins.setInitMainThreadSchedulerHandler { Schedulers.trampoline() }
+            RxAndroidPlugins.setMainThreadSchedulerHandler { Schedulers.trampoline() }
+        }
+    }
 
     private lateinit var repo: PocketVideoRepo
     @MockK private lateinit var videoStore: PocketVideoStore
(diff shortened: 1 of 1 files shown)


app/src/test/java/org/mozilla/tv/firefox/pocket/PocketVideoRepoTest.kt
```diff
@@ -7,10 +7,13 @@ package org.mozilla.tv.firefox.pocket
 import io.mockk.MockKAnnotations
```
mcomella: nit: I'd like to put your RCA & lessons learned in version control: GH may fail to exist while the repository remains.

There's a double-standard here – why don't we put everything in VC then? As a compromise for my time and readability, I think I try to put any results & conclusions into VC (even if it's duplicated) while the investigative details remain in the issues.

app/src/test/java/org/mozilla/tv/firefox/pocket/PocketVideoRepoTest.kt
```diff
@@ -21,6 +24,15 @@ import org.robolectric.RobolectricTestRunner
 @RunWith(RobolectricTestRunner::class)
 class PocketVideoRepoTest {
 
+    companion object {
+        @JvmStatic
+        @BeforeClass
+        fun beforeClass() {
+            RxAndroidPlugins.setInitMainThreadSchedulerHandler { Schedulers.trampoline() }
```
mcomella: nit: comment why this is necessary, and if it's not possible to extract this into a Rule, why not

app/src/test/java/org/mozilla/tv/firefox/pocket/PocketVideoRepoTest.kt
```diff
@@ -105,12 +103,12 @@ class PocketVideoRepoTest {
     @Test
     fun `GIVEN pocket key is valid WHEN getting a new instance THEN feed state starts inactive`() {
         val repo = PocketVideoRepo.newInstance(videoStore, { true }, isPocketKeyValid = true)
-        assertEquals(FeedState.Inactive, repo.feedState.blockingFirst())
+        repo.feedState.test().assertValue(FeedState.Inactive)
```
mcomella: nit: would it be more idiomatic to have a `feedStateTestObs` member var?

mcomella:
As discussed IRL?, it'd be ideal to extract this into an `@Rule`, however.

Baron-Severin:
> As discussed IRL?, it'd be ideal to extract this into an `@Rule`, however.

Pulled this out into a function that can be called from `@BeforeClass` in https://github.com/mozilla-mobile/firefox-tv/pull/2627.  I also pinged a few friends and asked them to look at it, in case I'm missing some landmine here.

app/src/test/java/org/mozilla/tv/firefox/pocket/PocketVideoRepoTest.kt
Baron-Severin: It looks like these two tests are creating a different repo and subject than the ones used in the `@Before` block, so that wouldn't work here.  That is a little bit confusing though, I'll add a comment to that effect.

Baron-Severin: It looks like these two tests are creating a different repo and subject than the ones used in the `@Before` block, so that wouldn't work here.  That is a little bit confusing though, I'll add a comment to that effect.

PR #2572, opened 2019-07-23
Closes #2529: Expose fxa state in architecture

### LINKED ISSUE #2529: Expose Firefox Account signed in state in architecture

As an FFTV user, I want FFTV to hold my logged in state, so that I can receive tabs from other devices I'm logged in to

* FxA library Firefox Account state (e.g. signed in, signed out, re-auth needed, etc.) is represented in application architecture
* There are unit tests to ensure we don't break the functionality in future

### Tasks

* Make sure signed in state is represented in our architecture
  * mcomella wrote some skeleton code to do this but he's not certain it's correct.
  * Specifically, the state transmitted from the AccountObserver (in FxaRepo) callbacks should be represented in the Observable `FxaRepo.accountStateDontUseMeYet`.
    * That way, NavigationOverlayFragment can observe the account state and change the behavior of the sign in/out button anyway
* Write unit tests to verify correctness of sign in state against FxA library

dnarcese:
Connected to #2529

## Checklist

### DIFF
--- a/app/src/main/java/org/mozilla/tv/firefox/fxa/FxaRepo.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/fxa/FxaRepo.kt
@@ -24,7 +24,6 @@ import org.mozilla.tv.firefox.fxa.FxaRepo.AccountState.AUTHENTICATED_WITH_PROFIL
 import org.mozilla.tv.firefox.fxa.FxaRepo.AccountState.AUTHENTICATED_NO_PROFILE
 import org.mozilla.tv.firefox.fxa.FxaRepo.AccountState.NEEDS_REAUTHENTICATION
 import org.mozilla.tv.firefox.fxa.FxaRepo.AccountState.NOT_AUTHENTICATED
-import org.mozilla.tv.firefox.fxa.FxaRepo.AccountState.INITIAL
 
 private val logger = Logger("FxaRepo")
 
@@ -49,16 +48,15 @@ class FxaRepo(
         // TODO: Later, may need "failed to login": https://github.com/mozilla-mobile/android-components/issues/3712
         AUTHENTICATED_WITH_PROFILE, // After the profile is fetched async
         AUTHENTICATED_NO_PROFILE, // Before the profile is fetched async
-        NEEDS_REAUTHENTICATION, // When an auth error occurs
-        NOT_AUTHENTICATED, // When logged out
-        INITIAL // Behaves the same as NOT_AUTHENTICATED
+        NEEDS_REAUTHENTICATION,
+        NOT_AUTHENTICATED // Initial state
     }
 
     @VisibleForTesting(otherwise = VisibleForTesting.NONE)
     val accountObserver = FirefoxAccountObserver()
 
     // TODO: set state accurately, choose correct subject, set initial state, etc.
-    val accountState = BehaviorSubject.createDefault(INITIAL)
+    val accountState = BehaviorSubject.createDefault(NOT_AUTHENTICATED)
 
     init {
         accountManager.register(accountObserver)
--- a/app/src/main/java/org/mozilla/tv/firefox/fxa/FxaRepo.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/fxa/FxaRepo.kt
@@ -77,6 +77,7 @@ class FxaRepo(
         return accountManager.beginAuthenticationAsync()
     }
 
+    @VisibleForTesting(otherwise = VisibleForTesting.NONE)
     inner class FirefoxAccountObserver : AccountObserver {
         /**
          * The account profile is fetched asynchronously.
--- a/app/src/main/java/org/mozilla/tv/firefox/fxa/FxaRepo.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/fxa/FxaRepo.kt
@@ -19,10 +19,11 @@ import mozilla.components.concept.sync.Profile
 import mozilla.components.service.fxa.DeviceConfig
 import mozilla.components.service.fxa.manager.FxaAccountManager
 import mozilla.components.support.base.log.logger.Logger
-import org.mozilla.tv.firefox.fxa.FxaRepo.AccountState.AUTHENTICATED
+import org.mozilla.tv.firefox.fxa.FxaRepo.AccountState.AUTHENTICATED_WITH_PROFILE
 import org.mozilla.tv.firefox.fxa.FxaRepo.AccountState.AUTHENTICATED_NO_PROFILE
 import org.mozilla.tv.firefox.fxa.FxaRepo.AccountState.NEEDS_REAUTHENTICATION
 import org.mozilla.tv.firefox.fxa.FxaRepo.AccountState.NOT_AUTHENTICATED
+import org.mozilla.tv.firefox.fxa.FxaRepo.AccountState.INITIAL
 
 private val logger = Logger("FxaRepo")
 
@@ -33,7 +34,7 @@ private val APPLICATION_SCOPES = setOf(
 )
 
 /**
- * Manages Firefox Account (FxA) state. In particular, [accountStateDontUseMeYet] exposes the current sign in
+ * Manages Firefox Account (FxA) state. In particular, [accountState] exposes the current sign in
  * state of the user. If you want to initiate sign in, see [FxaLoginUseCase.beginLogin].
  *
  * Devs should use this class rather than interacting with the FxA library directly.
@@ -44,17 +45,16 @@ class FxaRepo(
 ) {
 
     enum class AccountState {
-        // TODO: Are these states accurate?
         // TODO: Later, may need "failed to login": https://github.com/mozilla-mobile/android-components/issues/3712
-        AUTHENTICATED,
-        AUTHENTICATED_NO_PROFILE, // todo: necessary? profile seems to be fetched async so yes?
-        NEEDS_REAUTHENTICATION,
-        NOT_AUTHENTICATED
-        // TODO: do we need initial state before state is fetched from disk?
+        AUTHENTICATED_WITH_PROFILE, // After the profile is fetched async
+        AUTHENTICATED_NO_PROFILE, // Before the profile is fetched async
+        NEEDS_REAUTHENTICATION, // When an auth error occurs
+        NOT_AUTHENTICATED, // When logged out
+        INITIAL // Behaves the same as NOT_AUTHENTICATED
     }
 
     // TODO: set state accurately, choose correct subject, set initial state, etc.
-    val accountStateDontUseMeYet = BehaviorSubject.create<AccountState>()
+    val accountState = BehaviorSubject.createDefault(INITIAL)
 
     init {
         accountManager.register(FirefoxAccountObserver())
@@ -74,30 +74,32 @@ class FxaRepo(
     }
 
     private inner class FirefoxAccountObserver : AccountObserver {
+        /**
+         * The account profile is fetched asynchronously.
+         * There is no way to transition from [NOT_AUTHENTICATED] directly to [AUTHENTICATED_WITH_PROFILE];
+         * [AUTHENTICATED_NO_PROFILE] is always a state in between.
+         */
         override fun onAuthenticated(account: OAuthAccount) {
             logger.debug("onAuthenticated")
-            // todo: is this correct?
-            val nextState = if (accountManager.accountProfile() != null) {
-                AUTHENTICATED
-            } else {
-                AUTHENTICATED_NO_PROFILE
-            }
-            accountStateDontUseMeYet.onNext(nextState)
+            accountState.onNext(AUTHENTICATED_NO_PROFILE)
         }
 
         override fun onAuthenticationProblems() {
             logger.debug("onAuthenticationProblems")
-            accountStateDontUseMeYet.onNext(NEEDS_REAUTHENTICATION)
+            accountState.onNext(NEEDS_REAUTHENTICATION)
         }
 
         override fun onLoggedOut() {
             logger.debug("onLoggedOut")
-            accountStateDontUseMeYet.onNext(NOT_AUTHENTICATED)
+            accountState.onNext(NOT_AUTHENTICATED)
         }
 
+        /**
+         * This is called when the profile is first fetched after sign-in.
+         */
         override fun onProfileUpdated(profile: Profile) {
             logger.debug("onProfileUpdated")
-            // todo: update state.
+            accountState.onNext(AUTHENTICATED_WITH_PROFILE)
         }
     }
 
--- a/app/src/main/java/org/mozilla/tv/firefox/fxa/FxaRepo.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/fxa/FxaRepo.kt
@@ -5,6 +5,7 @@
 package org.mozilla.tv.firefox.fxa
 
 import android.content.Context
+import androidx.annotation.VisibleForTesting
 import androidx.lifecycle.ProcessLifecycleOwner
 import io.reactivex.subjects.BehaviorSubject
 import kotlinx.coroutines.Deferred
@@ -53,11 +54,14 @@ class FxaRepo(
         INITIAL // Behaves the same as NOT_AUTHENTICATED
     }
 
+    @VisibleForTesting(otherwise = VisibleForTesting.NONE)
+    val accountObserver = FirefoxAccountObserver()
+
     // TODO: set state accurately, choose correct subject, set initial state, etc.
     val accountState = BehaviorSubject.createDefault(INITIAL)
 
     init {
-        accountManager.register(FirefoxAccountObserver())
+        accountManager.register(accountObserver)
         accountManager.registerForDeviceEvents(FirefoxDeviceEventsObserver(), ProcessLifecycleOwner.get(),
             autoPause = false /* Avoid pausing even when the app is backgrounded. */)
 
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/mozilla/tv/firefox/fxa/FxaRepo.kt
```diff
@@ -44,17 +45,16 @@ class FxaRepo(
...
-        AUTHENTICATED,
-        AUTHENTICATED_NO_PROFILE, // todo: necessary? profile seems to be fetched async so yes?
-        NEEDS_REAUTHENTICATION,
-        NOT_AUTHENTICATED
-        // TODO: do we need initial state before state is fetched from disk?
+        AUTHENTICATED_WITH_PROFILE, // After the profile is fetched async
+        AUTHENTICATED_NO_PROFILE, // Before the profile is fetched async
+        NEEDS_REAUTHENTICATION, // When an auth error occurs
+        NOT_AUTHENTICATED, // When logged out
+        INITIAL // Behaves the same as NOT_AUTHENTICATED
```
Baron-Severin: If this behaves the same as `NOT_AUTHENTICATED`, can the two states be merged?  It would be good to have fewer.

Even if the class itself needs to know whether it's in one or the other, I feel like we shouldn't expose that complexity to the rest of the app.

app/src/main/java/org/mozilla/tv/firefox/fxa/FxaRepo.kt
```diff
@@ -44,17 +45,16 @@ class FxaRepo(
...
     enum class AccountState {
-        // TODO: Are these states accurate?
         // TODO: Later, may need "failed to login": https://github.com/mozilla-mobile/android-components/issues/3712
-        AUTHENTICATED,
-        AUTHENTICATED_NO_PROFILE, // todo: necessary? profile seems to be fetched async so yes?
-        NEEDS_REAUTHENTICATION,
-        NOT_AUTHENTICATED
-        // TODO: do we need initial state before state is fetched from disk?
+        AUTHENTICATED_WITH_PROFILE, // After the profile is fetched async
+        AUTHENTICATED_NO_PROFILE, // Before the profile is fetched async
```
Baron-Severin: I'm a little bit confused by this comment.  Is this state active while a fetch is in progress?  That seems counterintuitive, given the name.

app/src/main/java/org/mozilla/tv/firefox/fxa/FxaRepo.kt
```diff
@@ -44,17 +45,16 @@ class FxaRepo(
...
+        NEEDS_REAUTHENTICATION, // When an auth error occurs
+        NOT_AUTHENTICATED, // When logged out
+        INITIAL // Behaves the same as NOT_AUTHENTICATED
     }
 
     // TODO: set state accurately, choose correct subject, set initial state, etc.
-    val accountStateDontUseMeYet = BehaviorSubject.create<AccountState>()
+    val accountState = BehaviorSubject.createDefault(INITIAL)
```
Baron-Severin: Nit: we don't want to publicly expose this `Subject`, that would let the rest of the app push state to it.  Could you update this to follow the pattern of exposing the `.hide` val?  See `SessionRepo.state` for an example

app/src/main/java/org/mozilla/tv/firefox/fxa/FxaRepo.kt
```diff
@@ -44,17 +45,16 @@ class FxaRepo(
...
         // TODO: Later, may need "failed to login": https://github.com/mozilla-mobile/android-components/issues/3712
-        AUTHENTICATED,
-        AUTHENTICATED_NO_PROFILE, // todo: necessary? profile seems to be fetched async so yes?
-        NEEDS_REAUTHENTICATION,
-        NOT_AUTHENTICATED
-        // TODO: do we need initial state before state is fetched from disk?
+        AUTHENTICATED_WITH_PROFILE, // After the profile is fetched async
+        AUTHENTICATED_NO_PROFILE, // Before the profile is fetched async
+        NEEDS_REAUTHENTICATION, // When an auth error occurs
+        NOT_AUTHENTICATED, // When logged out
```
Baron-Severin: Nit: the comments on `NOT_AUTHENTICATED` and `NEEDS_REAUTHENTICATION` seem redundant, I think we could get rid of them.

app/src/main/java/org/mozilla/tv/firefox/fxa/FxaRepo.kt
```diff
@@ -73,7 +77,7 @@ class FxaRepo(
         return accountManager.beginAuthenticationAsync()
     }
 
-    private inner class FirefoxAccountObserver : AccountObserver {
+    inner class FirefoxAccountObserver : AccountObserver {
```
Baron-Severin: Nit: we could use a `VisibleForTesting` annotation on this as well.

app/src/test/java/org/mozilla/tv/firefox/fxa/FxaRepoTest.kt
```diff
@@ -0,0 +1,74 @@
...
+    @MockK private lateinit var profile: Profile
+
+    private lateinit var fxaRepo: FxaRepo
+    private lateinit var accountState: BehaviorSubject<FxaRepo.AccountState>
+    private lateinit var accountStateTestObs: TestObserver<FxaRepo.AccountState>
+
+    @Before
+    fun setUp() {
+        MockKAnnotations.init(this)
+        accountState = BehaviorSubject.create()
```
Baron-Severin: It looks like we're missing a commit here, because `accountState` isn't hooked up to the class.  This will be what's causing CI to fail.

app/src/test/java/org/mozilla/tv/firefox/fxa/FxaRepoTest.kt
```diff
@@ -0,0 +1,74 @@
...
+import io.reactivex.subjects.BehaviorSubject
+import mozilla.components.concept.sync.OAuthAccount
+import mozilla.components.concept.sync.Profile
+import mozilla.components.service.fxa.manager.FxaAccountManager
+import org.junit.Before
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+
+@RunWith(RobolectricTestRunner::class)
```
Baron-Severin: Since we aren't using real `Context`, we don't need to use Robolectric here.  I tried removing it, and it dropped down the execution time for the class from ~13.2 seconds to ~7.5s.

app/src/test/java/org/mozilla/tv/firefox/fxa/FxaRepoTest.kt
```diff
@@ -0,0 +1,74 @@
...
+class FxaRepoTest {
+
+    @MockK private lateinit var context: Context
+    @MockK(relaxed = true) private lateinit var accountManager: FxaAccountManager
+    @MockK private lateinit var account: OAuthAccount
+    @MockK private lateinit var profile: Profile
```
Baron-Severin: Nit: `profile`, `context`, and `account` are only used in one function each, could we move their setup to those funs?  The syntax for that would look like `val context = mockk<Context>()`.

It makes the test a little easier to read when there's less setup up top.

app/src/main/java/org/mozilla/tv/firefox/fxa/FxaRepo.kt
dnarcese: Ya this is the first state after logging in. It is still in authenticated state but the profile is being fetched. Then once the profile is fetched, it moves to authenticated_with_profile. Is there a different name that you think would make it more intuitive?

mcomella: I can see what Severin means: it sounds like AUTHENTICATED_NO_PROFILE and AUTHENTICATED_WITH_PROFILE are two different terminal states. The comments help but I can't think of a better name quickly. Maybe AUTHENTICATED_BEFORE_PROFILE? AUTHENTICATED_FETCHING_PROFILE?

Can you make the inline comments kdoc comments above each enum state?

app/src/main/java/org/mozilla/tv/firefox/fxa/FxaRepo.kt
```diff
@@ -44,20 +46,21 @@ class FxaRepo(
 ) {
 
     enum class AccountState {
-        // TODO: Are these states accurate?
         // TODO: Later, may need "failed to login": https://github.com/mozilla-mobile/android-components/issues/3712
-        AUTHENTICATED,
-        AUTHENTICATED_NO_PROFILE, // todo: necessary? profile seems to be fetched async so yes?
+        AUTHENTICATED_WITH_PROFILE, // After the profile is fetched async
+        AUTHENTICATED_NO_PROFILE, // Before the profile is fetched async
```
mcomella: Is it possible for an account to never have a profile? Both in terms of 1) profile failed to fetch and 2) account doesn't have a profile (e.g. maybe legacy accounts?). How would we represent this state? Would the account just stay in the AUTHENTICATED_NO_PROFILE state?

Can we document this is some way? e.g. kdoc on each enum to explain the states further.

app/src/main/java/org/mozilla/tv/firefox/fxa/FxaRepo.kt
```diff
@@ -44,20 +46,21 @@ class FxaRepo(
...
         NEEDS_REAUTHENTICATION,
-        NOT_AUTHENTICATED
-        // TODO: do we need initial state before state is fetched from disk?
+        NOT_AUTHENTICATED // Initial state
     }
 
-    // TODO: set state accurately, choose correct subject, set initial state, etc.
-    val accountStateDontUseMeYet = BehaviorSubject.create<AccountState>()
+    @VisibleForTesting(otherwise = VisibleForTesting.NONE)
```
mcomella: nit: typically I important each VisibleForTesting value so this can be: `@VisibleForTesting(otherwise = NONE)`

app/src/main/java/org/mozilla/tv/firefox/fxa/FxaRepo.kt
```diff
@@ -44,20 +46,21 @@ class FxaRepo(
...
+    val accountObserver = FirefoxAccountObserver()
+
+    private val _accountState: BehaviorSubject<AccountState> = BehaviorSubject.createDefault(NOT_AUTHENTICATED)
```
mcomella: No action now but if we start with a NOT_AUTHENTICATED state and the user is already logged in, the initial UI might flash from the NOT_AUTHENTICATED to the AUTHENTICATED state (for an example, you can see similar flashing states in Fenix a lot). This is why it can be useful to have a pre-init state (either an explicit value or no value).

However, we'll solve this problem when we come to it.

app/src/main/java/org/mozilla/tv/firefox/fxa/FxaRepo.kt
```diff
@@ -73,31 +76,34 @@ class FxaRepo(
         return accountManager.beginAuthenticationAsync()
     }
 
-    private inner class FirefoxAccountObserver : AccountObserver {
+    @VisibleForTesting(otherwise = VisibleForTesting.NONE)
+    inner class FirefoxAccountObserver : AccountObserver {
+        /**
+         * The account profile is fetched asynchronously.
```
mcomella: Perhaps this comment should be on the enum kdoc instead and we should reference it here. It can probably also we an inline comment rather than function kdoc.

app/src/main/java/org/mozilla/tv/firefox/fxa/FxaRepo.kt
```diff
@@ -73,31 +76,34 @@ class FxaRepo(
...
+         * There is no way to transition from [NOT_AUTHENTICATED] directly to [AUTHENTICATED_WITH_PROFILE];
+         * [AUTHENTICATED_NO_PROFILE] is always a state in between.
+         */
         override fun onAuthenticated(account: OAuthAccount) {
             logger.debug("onAuthenticated")
```
mcomella: nit: I'd like to remove these eventually so if they're not useful now, feel free to remove them.

app/src/test/java/org/mozilla/tv/firefox/fxa/FxaRepoTest.kt
```diff
@@ -0,0 +1,72 @@
...
+        MockKAnnotations.init(this)
+        val context = mockk<Context>()
+        fxaRepo = FxaRepo(context, accountManager)
+        accountState = fxaRepo.accountState
+        accountStateTestObs = accountState.test()
+    }
+
+    @Test
+    fun `WHEN fxa repo is initialized THEN the account observer is registered once`() {
+        verify(exactly = 1) { accountManager.register(fxaRepo.accountObserver) }
```
mcomella: Cool - thanks for writing these tests for me. :P

app/src/test/java/org/mozilla/tv/firefox/fxa/FxaRepoTest.kt
```diff
@@ -0,0 +1,72 @@
...
+        accountStateTestObs.assertValueAt(1, FxaRepo.AccountState.NOT_AUTHENTICATED)
+    }
+
+    @Test
+    fun `WHEN on profile update callback is called THEN account state is authenticated with profile`() {
+        val profile = mockk<Profile>()
+        fxaRepo.accountObserver.onProfileUpdated(profile)
+        accountStateTestObs.assertValueAt(1, FxaRepo.AccountState.AUTHENTICATED_WITH_PROFILE)
+    }
+}
```
mcomella: nit: need newline at end of file

app/src/test/java/org/mozilla/tv/firefox/fxa/FxaRepoTest.kt
```diff
@@ -0,0 +1,72 @@
...
+    fun `WHEN on logout callback is called THEN account state is not authenticated`() {
+        fxaRepo.accountObserver.onLoggedOut()
+        accountStateTestObs.assertValueAt(1, FxaRepo.AccountState.NOT_AUTHENTICATED)
+    }
+
+    @Test
+    fun `WHEN on profile update callback is called THEN account state is authenticated with profile`() {
+        val profile = mockk<Profile>()
+        fxaRepo.accountObserver.onProfileUpdated(profile)
+        accountStateTestObs.assertValueAt(1, FxaRepo.AccountState.AUTHENTICATED_WITH_PROFILE)
```
mcomella: No action needed but FYI. We're not defensively asserting that the FxA library will move through the states in an expected way (e.g. not_authenticated -> authenticated_no_profile -> authenticated_with_profile). For example, we may wish to record an error if the library ever pushes us through a state transition we didn't expect (e.g. not_authenticated -> authenticated_with_profile) so that way we can uncover bugs in our implementation.

The tests pointed this out for me: we never set an previous state and assert the transition: we just call the related callback and assert the output value.

app/src/main/java/org/mozilla/tv/firefox/fxa/FxaRepo.kt
```diff
@@ -73,31 +76,34 @@ class FxaRepo(
         return accountManager.beginAuthenticationAsync()
     }
 
-    private inner class FirefoxAccountObserver : AccountObserver {
+    @VisibleForTesting(otherwise = VisibleForTesting.NONE)
+    inner class FirefoxAccountObserver : AccountObserver {
+        /**
+         * The account profile is fetched asynchronously.
+         * There is no way to transition from [NOT_AUTHENTICATED] directly to [AUTHENTICATED_WITH_PROFILE];
```
mcomella: Is this true? What happens if the user is already authenticated? Is the profile saved to disk?

app/src/main/java/org/mozilla/tv/firefox/fxa/FxaRepo.kt
```diff
@@ -73,31 +76,34 @@ class FxaRepo(
...
         override fun onLoggedOut() {
             logger.debug("onLoggedOut")
-            accountStateDontUseMeYet.onNext(NOT_AUTHENTICATED)
+            _accountState.onNext(NOT_AUTHENTICATED)
         }
 
+        /**
+         * This is called when the profile is first fetched after sign-in.
+         */
         override fun onProfileUpdated(profile: Profile) {
```
mcomella: I don't think it matters in the current implementation but have we taken into account that this may be called multiple times? e.g. if the user changes their avatar, I'd expect this to be called.

dnarcese: From looking at the FxA code, this only gets called once.

app/src/main/java/org/mozilla/tv/firefox/fxa/FxaRepo.kt
dnarcese: Looks like account is saved but profile is not. So even if account is saved, the profile needs to be fetched.

app/src/main/java/org/mozilla/tv/firefox/fxa/FxaRepo.kt
dnarcese: To me, fetching the profile seems to be visually the same state as having no profile. It is possible that the profile is null, and in that case this would be the state. Unless we want to do something visually different while it is fetching, makes sense for me to keep them in the same state.

app/src/main/java/org/mozilla/tv/firefox/fxa/FxaRepo.kt
mcomella: It only gets called in one location but it's not guaranteed to only be called once. For example, the public method `FxaAccountManager.updateProfileAsync` queues a `FetchProfile` event, which will fetch the profile and queue a `FetchedProfile` event which will call `onProfileUpdated`.

@dnarcese Is the current code resilient to this being called multiple times?

app/src/main/java/org/mozilla/tv/firefox/fxa/FxaRepo.kt
mcomella: Seems reasonable. :)

Issue #2395, opened 2019-06-07
Remove FocusRepo


mcomella:
### Vision statement / What / Requirements
`FocusRepo` is complex and hard to change: is there a simpler solution? In theory, the Android framework should be doing focus handling for us but the `FocusRepo` seems to reimplement a lot of this framework functionality (disclaimer: I'm not that familiar with the code). I'd like to explore removing it and replacing it with something simpler.

### Impact
Without making this change:
- we may continue to suffer from obscure focus bugs
- changing the focus handling code will be slow, challenging, and may result in regressions (disclaimer: I'm not that familiar with the code)

By making this change:
- we'll undo the issues above
- we may introduce regressions
- for all we know, the replacement could be harder to maintain (but I think not)

### Acceptance criteria
- [ ] FocusRepo is removed
- [ ] Focus behavior is equivalent to, or better than, existing behavior
  - [ ] This solution is simpler than FocusRepo
  - [ ] Go through open focus bugs: are they fixed with these changes? Are they more complicated?
- [ ] If applicable, any telemetry built around focus is working correctly

---

I wrote a proof-of-concept ([here is the branch](https://github.com/mcomella/firefox-tv/tree/host-poc-rm-focusrepo)) that removes FocusRepo without adding additional code. Testing on an emulator, there are two discrepancies from the existing codebase:
- `NavigationOverlay` is not focused when opened
- Moving up from nav url bar to page actions incorrectly focuses the "Exit Firefox" button

These seem mostly straightforward to fix to me so, assuming there are not many more significant focus issues, I don't think the complexity of the 350+ lines I removed is necessary.

---

I noticed that we have a few `<requestFocus` tags in XML that steal focus in some unexpected situations: removing them may simplify our focus handling code. See my explanation here for more: https://github.com/mozilla-mobile/firefox-tv/pull/2455#pullrequestreview-262617788

---

I do not consider this an aspirational refactor (which we don't file) because the proof-of-concept immediately shows the benefits of these changes and I think it's more than aspirational: I think we should do it next time we need to change focus code.

mcomella:
@Baron-Severin @psymoon Do you have any thoughts on this, given my findings from the proof-of-concept?

psymoon:
Hey @mcomella, I support this 💯 . Here is the list of major problems `FocusRepo` was addressing:
1. Pocket error state focus handling (error button vs. recyclerView) - https://github.com/mozilla-mobile/firefox-tv/issues/1130
[...]

PR #2609, opened 2019-07-26
Issue #2395 - FocusRepo Removal 

wmichelle024:
Connected to #???

## Checklist


app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/NavigationOverlayFragment.kt
```diff
@@ -271,6 +276,18 @@ class NavigationOverlayFragment : Fragment() {
...
 
+    private fun observeToolbarState(): Disposable {
+        return navigationOverlayViewModel.sessionState
+                .subscribe { state ->
+                    navUrlInput.nextFocusUpId = when {
+                        state.backEnabled -> R.id.navButtonBack
+                        state.forwardEnabled -> R.id.navButtonForward
+                        state.currentUrl != URLs.APP_URL_HOME -> R.id.navButtonReload
+                        else -> R.id.turboButton
+                    }
```
Baron-Severin: There are a few issues here.  The biggest one is that we're duplicating logic.  If the code in `ToolbarVM` changed, this would fall out of sync and cause weird focus bugs.  Another is that this logic shouldn't live in the View at all, because it isn't its responsibility.  Let's work together on Monday to find a better place for this to live.

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/NavigationOverlayViewModel.kt
```diff
@@ -47,4 +46,15 @@ class NavigationOverlayViewModel(
...
             .distinctUntilChanged()
+
+    val focusView : Observable<Int> = screenController.currentActiveScreen
+            .buffer(2,1)
+            .filter{ (_, currentScreen) -> currentScreen == ActiveScreen.NAVIGATION_OVERLAY }
+            .map { (prevScreen, _) ->
+                when(prevScreen) {
+                    ActiveScreen.WEB_RENDER -> R.id.navUrlInput
+                    ActiveScreen.SETTINGS -> R.id.settings_tile_telemetry
+                    else -> View.NO_ID
```
Baron-Severin: Could you replace the `else` with `ActiveScreen.NAVIGATION_OVERLAY`?  That way if we add any screens in the future, this will be a compilation error and we'll know to update it.  With the `else` branch it would just work, which could lead to bad behavior when `View.NO_ID` was emitted.

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/NavigationOverlayViewModel.kt
```diff
@@ -57,4 +55,14 @@ class NavigationOverlayViewModel(
...
                 }
             }
+
+    val toolBarFocus : Observable<Int> = toolbarViewModel.state
+            .map{state ->
+                when {
+                    state.backEnabled -> R.id.navButtonBack
+                    state.forwardEnabled -> R.id.navButtonForward
+                    state.refreshEnabled -> R.id.navButtonReload
+                    else -> R.id.turboButton
```
Baron-Severin: Hm, this is not entirely clear to me.  This emits a new ID any time toolbar state changes.  I see that it's used to set `navUrlInput.nextFocusUpId`, but that isn't really made clear to me by the name `toolBarFocus`.  Could you rename this to better communicate what it's doing (which is emitting the ID of the currently active toolbar button that is furthest to the left)?

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/NavigationOverlayViewModel.kt
```diff
@@ -65,4 +66,13 @@ class NavigationOverlayViewModel(
                     else -> R.id.turboButton
                 }
             }
+
+    val toolBarFocusListener = View.OnKeyListener { v: View?, keyCode: Int, _: KeyEvent? ->
+        if (keyCode == KeyEvent.KEYCODE_DPAD_LEFT) {
+            v?.requestFocus()
+            true
+        } else {
+            false
```
Baron-Severin: I think we should probably take a different approach to this, for a few reasons.  There's a rule of thumb in Android dev that `requestFocus` should only be used after everything else has failed, for a number of reasons (it's uninutitive to change user focus, depending on the exact timing it might show a flash of focus elsewhere before moving, etc).  I also don't love that we're handling the same kind of problem differently for different views (some are using `nextFocus*Id`, some are intercepting dpad events).  I think we need to find a different solution to this.

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/NavigationOverlayViewModel.kt
wmichelle024: I'm getting a Kotlin compiler warning (`Enum argument can be null in Java, but exhaustive when contains no null branch`) when I remove the else case. Any way I can get around this?

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/NavigationOverlayFragment.kt
```diff
@@ -294,6 +271,17 @@ class NavigationOverlayFragment : Fragment() {
         ).map { channel -> channel.forwardRemoveEventsToRepo() }
     }
 
+    private fun observeToolbarState(): Disposable {
+        return navigationOverlayViewModel.leftmostActiveToolBarId
+                .subscribe { viewToFocus ->
```
Baron-Severin: Nit: `viewToFocus` doesn't seem to be accurate here.  Could you rename?

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/NavigationOverlayFragment.kt
```diff
@@ -294,6 +271,17 @@ class NavigationOverlayFragment : Fragment() {
         ).map { channel -> channel.forwardRemoveEventsToRepo() }
     }
 
+    private fun observeToolbarState(): Disposable {
```
Baron-Severin: Nit: could we rename this one to `observeToolbarFocusability`, or something similar?  To distinguish this from code that sets when buttons are enabled and such.

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/NavigationOverlayFragment.kt
```diff
@@ -294,6 +271,17 @@ class NavigationOverlayFragment : Fragment() {
...
+        return navigationOverlayViewModel.leftmostActiveToolBarId
+                .subscribe { viewToFocus ->
+                    // Reset previous left most active toolbar button's nextFocusLeftID
+                    rootView?.findViewById<View>(navUrlInput.nextFocusUpId)?.nextFocusLeftId = -1
+                    rootView?.findViewById<View>(viewToFocus)?.nextFocusLeftId = viewToFocus
+
+                    navUrlInput.nextFocusUpId = viewToFocus
```
Baron-Severin: Nit: the logic here is pretty simple, but there's a lot of noise that makes it a little hard to read.  I think a comment between 278 & 279 would help for clarity, something like `// Disable left clicks on viewToFocus `

Baron-Severin:
I did some testing on this, and it looks good.  Once these last few nits have been cleaned up I think we'll be good to merge.  Let's make sure to ping QA about this one though, they should definitely take a pass at it.

Issue #2643, opened 2019-08-02
Land strings for Accounts & Receive Tab


athomasmoz:
### Why/User Benefit/User Problem

This story is to land our strings for Accounts & Receive tab so that localization can start. Strings will not be visible to users until we launch the new features

### Acceptance Criteria (how do I know when I’m done?)

- [x] Final strings have been landed
- [x] All strings have been sent for localization ([l10n pr](https://github.com/mozilla-l10n/android-l10n/pull/63))

### Finalized Strings

https://docs.google.com/spreadsheets/d/1z852T9AKhFpQFq8JcX633-zsC81vxAkKYKyilwdistU/edit?ts=5d423a76#gid=0

liuche:
I think we should pre-land these strings and export them for L10N, and leave off the one string that we're not sure about. I've marked the one w/ Firefox Preview in the spreadsheet.

athomasmoz:
Thanks, and when we launch it will still be Preview, so we're safe to use that.

liuche:
Let's try to land these by 8/9 and exported so L10N has enough time to localized for us.

mcomella:
@dnarcese Do you know if we need to run a manual export (using the old system) on our strings or if l10n will automate the export (with the new system)? Can you summarize the current state of the world? As per liuche's comment, we'd like to export these strings ASAP so I'm hoping something is in place. :)

CC @flodolo

flodolo:
Redirecting to @pike who's covering for @Delphine this week

Pike:
We started with `android-l10n` in https://github.com/mozilla-l10n/android-l10n/pull/59.

I'll create the first import back from the strings in a moment.

Pike:
PR #2673 is up :-)

dnarcese:
@mcomella We just need to land strings on master, and l10n team will do regular exports to android-l10n repo.

dnarcese:
l10n does not want to rush merging these in on a Friday (it is late night their time) - it will get merged on Monday (cc @liuche)

dnarcese:
All strings have now been exported.

PR #2649, opened 2019-08-06
Issue #2643: Add FxA strings

dnarcese:
## Checklist

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -215,8 +215,9 @@
 
     <!-- Firefox Accounts Strings-->
     <!-- Description for Sent tab buttons (in navigation toolbar & settings, both on the homescreen)
-     that will launch the Firefox Account sign-in experience so users can receive Firefox tabs on their Fire TV. -->
-    <string name="fxa_navigation_item_new">New! Send Firefox tabs to Fire TV</string>
+     that will launch the Firefox Account sign-in experience so users can receive Firefox tabs on their Fire TV.
+     %1$s will be replaced with Firefox. -->
+    <string name="fxa_navigation_item_new">New! Send %1$s tabs to Fire TV</string>
     <!-- Description for Sent tab button in navigation toolbar
      that will launch the Firefox Account settings after the user is signed in. -->
     <string name="fxa_navigation_item_signed_in">Sent tab settings</string>
@@ -226,16 +227,18 @@
     <!-- Header for the Firefox Account sign-in pre-boarding, where we introduce the send tab feature. -->
     <string name="fxa_preboarding_header">See your tabs on the big screen</string>
     <!-- Instruction text for the Firefox Account sign-in pre-boarding, where we introduce the send tab feature.
-    %1$s will be replaced with the name of our app (i.e. Firefox for Fire TV) -->
-    <string name="fxa_preboarding_instruction">Send a tab from anywhere you use Firefox and see it here. Sign in to %1$s to get started.</string>
+    %1$s will be replaced with Firefox.
+    %2$s will be replaced with the name of our app (i.e. Firefox for Fire TV) -->
+    <string name="fxa_preboarding_instruction">Send a tab from anywhere you use %1$s and see it here. Sign in to %2$s to get started.</string>
     <!-- Sign-in button text for the Firefox Account sign-in pre-boarding, where we introduce the send tab feature. -->
     <string name="fxa_preboarding_primary_button">Sign in</string>
     <!-- Text on the cancel button for the Firefox Account sign-in pre-boarding, where we introduce the send tab feature. -->
     <string name="fxa_preboarding_secondary_button">Not now</string>
     <!-- Header for Firefox Accounts onboarding screen, which is shown after signing in. -->
     <string name="fxa_onboarding_header">Your tabs. Now ready for prime time.</string>
-    <!-- Instruction text for Firefox Accounts onboarding screen, which is shown after signing in. -->
-    <string name="fxa_onboarding_instruction">You’re ready to receive tabs. To get started, select Send Tab to Device on Firefox for desktop and iOS or Share on Firefox Preview. Then choose Fire TV and your tab will appear in just a moment.</string>
+    <!-- Instruction text for Firefox Accounts onboarding screen, which is shown after signing in.
+     %1$s will be replaced with Firefox. -->
+    <string name="fxa_onboarding_instruction">You’re ready to receive tabs. To get started, select Send Tab to Device on %1$s for desktop and iOS or Share on Firefox Preview. Then choose Fire TV and your tab will appear in just a moment.</string>
     <!-- Text on the dismiss button for Firefox Accounts onboarding screen, which is shown after signing in. -->
     <string name="fxa_onboarding_primary_button">Ok, got it</string>
     <!-- Toast shown on webpage when tab is sent from another device
@@ -246,8 +249,9 @@
     <!-- Body text for Firefox Accounts settings page
     %1$s will be replaced with the signed-in user account email. -->
     <string name="fxa_settings_body">Signed in as %1$s</string>
-    <!-- Button text for seeing the onboarding screen on the Firefox Accounts settings page -->
-    <string name="fxa_settings_primary_button">Firefox tabs</string>
+    <!-- Button text for seeing the onboarding screen on the Firefox Accounts settings page.
+     %1$s will be replaced with Firefox. -->
+    <string name="fxa_settings_primary_button">%1$s tabs</string>
     <!-- Button text to sign out from the Firefox Accounts settings page -->
     <string name="fxa_settings_secondary_button">Sign out</string>
 </resources>
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -184,10 +184,6 @@
     to notify user that Firefox has requested the default "non-desktop version" from the site -->
     <string name="notification_request_non_desktop_site">Returned to non-desktop version</string>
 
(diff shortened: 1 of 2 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -217,4 +213,40 @@
      When the described button is pressed, the overlay will open and the browser will be hidden underneath it -->
     <string name="hint_press_menu_to_open_overlay_a11y">Press the menu button for %1$s Home, search, settings, and more</string>
 
+    <!-- FxA Strings-->
+    <!-- Content description for a button that will prompt a user to sign in, in order to receive tabs pushed from FF
+    on another device -->
+    <string name="fxa_login_a11y">Receive tabs</string>
+    <!-- Heading for settings menu item button and CTA link on new update -->
```
mcomella: I'm not sure this description is super helpful to localizers: what's a CTA link? What's a new update? What happens when the button is pressed? Also, "Heading" isn't quite accurate: it's more of a button description. I'm thinking something like:

"Description for Receive Tab buttons (in navigation toolbar & settings, both on the homescreen) that will launch a Firefox Account sign-in experience so users can receive desktop Firefox tabs on their Fire TV".

app/src/main/res/values/strings.xml
```diff
@@ -217,4 +213,40 @@
...
+    <string name="fxa_settings_item_new">New! Firefox tabs on Fire TV</string>
+    <!-- Heading for settings menu item button and CTA link when signed in -->
+    <string name="fxa_settings_item_signed_in">Sent tab settings</string>
+    <!-- Heading for settings menu item button and CTA link when need to sign in again -->
+    <string name="fxa_settings_item_sign_in_again">Sent tabs - sign in again</string>
+    <!-- FxA pre-boarding header -->
```
mcomella: This localization note doesn't add anything over the string name. Perhaps we should give a brief description of what pre-boarding is? fwiw, I think this needs to be duplicated in all of the pre-boarding strings. For example,

"Heading for the Firefox Account sign-in pre-boarding, where we introduce the send tab feature."

app/src/main/res/values/strings.xml
```diff
@@ -217,4 +213,40 @@
...
+    <string name="fxa_preboarding_header">See your tabs on the big screen</string>
+    <!-- FxA pre-boarding instruction text -->
+    <string name="fxa_preboarding_instruction">Send a tab from anywhere you use Firefox and see it here. Sign in to Firefox for Fire TV to get started.</string>
```
mcomella: nit: I wonder if we should be subbing the `@string/app_name_extended_fire` instead of hardcoding, "FFTV". Can you check with Delphine what the best practice is?

app/src/main/res/values/strings.xml
```diff
@@ -217,4 +213,40 @@
...
+    <!-- FxA pre-boarding sign in button -->
+    <string name="fxa_preboarding_primary_button">Sign in</string>
+    <!-- FxA pre-boarding not now button -->
+    <string name="fxa_preboarding_secondary_button">Not now</string>
+    <!-- FxA onboarding header -->
```
mcomella: Should we tell them onboarding is shown after sign in?

app/src/main/res/values/strings.xml
mcomella: I think we should type out Firefox Account: I don't know that localizers know what FxA is.

app/src/main/res/values/strings.xml
```diff
@@ -217,4 +213,40 @@
...
+    <string name="fxa_onboarding_header">Your tabs. Now ready for prime time.</string>
+    <!-- FxA onboarding instruction text -->
+    <string name="fxa_onboarding_instruction">You’re ready to receive tabs. To get started, select Send Tab to Device on Firefox for desktop and iOS or Share on Firefox Preview. Then choose Fire TV and your tab will appear in just a moment.</string>
+    <!-- FxA onboarding primary button -->
+    <string name="fxa_onboarding_primary_button">Ok, got it</string>
+    <!-- Toast shown on webpage when tab is sent from another device
+     %1$s will be replaced with the device name where the tab was sent from. -->
+    <string name="fxa_tab_sent_toast">Tab sent from %1$s</string>
+    <!-- FxA settings page header -->
+    <string name="fxa_settings_header">Sent tabs</string>
```
mcomella: Maybe we should call this, "sent_tabs_settings_header" because that's all that's in here right now. We can change it to, `fxa...` when it actually hosts more fxa data.

app/src/main/res/values/strings.xml
```diff
@@ -217,4 +213,40 @@
...
+    <!-- FxA settings page body text
+    %1$s will be replaced with the signed-in user account email. -->
+    <string name="fxa_settings_body">Signed in as %1$s</string>
+    <!-- FxA settings page primary button -->
```
mcomella: I'm on the fence with this one but I'm not sure localizers are familiar with, "Primary button" and "secondary button" concepts.

mcomella:
> However, I think we can do better about providing useful names & localization notes.

To be explicit, can you make sure all of the strings have good names & localization notes (following the suggestions I provide in my comments), even the ones I haven't commented on?

app/src/main/res/values/strings.xml
dnarcese: Wouldn't it be easier to just leave the string names with `fxa`? Because I think if we change the name don't we have to get them re-localized?

mcomella: After your comment, I notice now that there are some strings that will likely need to be relocalized (e.g. `fxa_settings_header` = "Sent tabs") and some that will likely not be (e.g. `fxa_settings_body` = "Signed in as ...").

I say we stick with `fxa_*` so that, if this does because an FxA settings screen and we don't need to relocalize some strings, we don't have a mix of string identifiers with different prefix conventions (`sent_tab_*` vs `fxa_*`).

Thanks for raising this.

app/src/main/res/values/strings.xml
```diff
@@ -217,4 +213,44 @@
      When the described button is pressed, the overlay will open and the browser will be hidden underneath it -->
     <string name="hint_press_menu_to_open_overlay_a11y">Press the menu button for %1$s Home, search, settings, and more</string>
 
+    <!-- Firefox Accounts Strings-->
+    <!-- Description for Sent tab buttons (in navigation toolbar & settings, both on the homescreen)
+     that will launch the Firefox Account sign-in experience so users can receive Firefox tabs on their Fire TV.
+     %1$s will be replaced with Firefox. -->
+    <string name="fxa_navigation_item_new">New! Send %1$s tabs to Fire TV</string>
+    <!-- Description for Sent tab button in navigation toolbar
+     that will launch the Firefox Account settings after signed in -->
```
mcomella: nit: ...after the user is signed in

app/src/main/res/values/strings.xml
```diff
@@ -217,4 +213,44 @@
      When the described button is pressed, the overlay will open and the browser will be hidden underneath it -->
     <string name="hint_press_menu_to_open_overlay_a11y">Press the menu button for %1$s Home, search, settings, and more</string>
 
+    <!-- Firefox Accounts Strings-->
+    <!-- Description for Sent tab buttons (in navigation toolbar & settings, both on the homescreen)
+     that will launch the Firefox Account sign-in experience so users can receive Firefox tabs on their Fire TV.
+     %1$s will be replaced with Firefox. -->
+    <string name="fxa_navigation_item_new">New! Send %1$s tabs to Fire TV</string>
+    <!-- Description for Sent tab button in navigation toolbar
```
mcomella: nit: should this be consistent with the other button strings?

app/src/main/res/values/strings.xml
```diff
@@ -217,4 +213,44 @@
...
+    %2$s will be replaced with the name of our app (i.e. Firefox for Fire TV) -->
+    <string name="fxa_preboarding_instruction">Send a tab from anywhere you use %1$s and see it here. Sign in to %2$s to get started.</string>
+    <!-- Sign-in button text for the Firefox Account sign-in pre-boarding, where we introduce the send tab feature. -->
+    <string name="fxa_preboarding_primary_button">Sign in</string>
+    <!-- Text on the cancel button for the Firefox Account sign-in pre-boarding, where we introduce the send tab feature. -->
+    <string name="fxa_preboarding_secondary_button">Not now</string>
+    <!-- Header for Firefox Accounts onboarding screen, which is shown after signing in. -->
+    <string name="fxa_onboarding_header">Your tabs. Now ready for prime time.</string>
+    <!-- Instruction text for Firefox Accounts onboarding screen, which is shown after signing in. -->
+    <string name="fxa_onboarding_instruction">You’re ready to receive tabs. To get started, select Send Tab to Device on %1$s for desktop and iOS or Share on Firefox Preview. Then choose Fire TV and your tab will appear in just a moment.</string>
```
mcomella: nit: in localization note, need to declare which string we're substituting

app/src/main/res/values/strings.xml
```diff
@@ -217,4 +213,44 @@
...
+    <!-- Text on the dismiss button for Firefox Accounts onboarding screen, which is shown after signing in. -->
+    <string name="fxa_onboarding_primary_button">Ok, got it</string>
+    <!-- Toast shown on webpage when tab is sent from another device
+     %1$s will be replaced with the device name where the tab was sent from. -->
+    <string name="fxa_tab_sent_toast">Tab sent from %1$s</string>
+    <!-- Description for Sent tabs button on the homescreen and header text for Firefox Accounts settings page. -->
```
mcomella: -> "Description for header text for Firefox Accounts Ssettings page"

This isn't for the "Sent tabs button on the homescreen" (though the string is consistent with it) so we should remove the reference entirely or optionally mention it's consistent with other strings.

Issue #2541, opened 2019-07-12
[2386] mp4 videos are squashed/distorted when scrolling & in fullscreen (4K Pendant)


dnarcese:
Meta: #2386.

_Similar issue on 4K Stick & Stick gen 2: #2540_

## Steps to reproduce
1.  Visit https://www.ncaa.com/video  (or YouTube.com #2552, or http://mcomella.xyz/test/video.html for minimal CSS)
2. Fullscreen
### Expected behavior
Video fills whole screen.
### Actual behavior
Video is squished into the top 1/5 of screen, and rest is black.
### Device information
* Fire TV device: 4k (pendant)
* Latest affected Firefox version: ?
* Earliest affected Firefox version: ?
* Turbo Mode: ?

### Notes

* Minimal test case in raw WebView: https://docs.google.com/document/d/1WdioHdXZi9V9BbGcDKfuKYpJEX1JYzz54JRDkAjImTQ/edit#
* Here's [a video](https://drive.google.com/file/d/1_kvYvYTp2tnbSKKCaM3vYRGhJcLnWxEA/view?usp=sharing) of the issue on YouTube.com
* More detail available in #2552 
* #2540 is a very similar but different problem that I can only reproduce on Stick gen 2 & 4K stick

oathzed:
I have a good representation of this with pictures: https://github.com/mozilla-mobile/firefox-tv/issues/2552

mcomella:
I was able to reproduce intermittently on ncaa without turbo mode; I was unable to play video content with turbo mode enabled and filed https://github.com/mozilla-mobile/firefox-tv/issues/2555. I was also able to reproduce on desktop YouTube (thanks for the like oathzed!).

Like #2552, I can confirm the video unexpectedly "smooshes" (distorts) when scrolling down the page and that fullscreening will sometimes cut off content.

mcomella:
I was also able to reproduce the fullscreen issue on the Stick gen 2 though the smooshing is not present. Also, smooshing on ncaa appears to be intermittent on the 4K pendant too.

mcomella:
@dnarcese How frequently are you able to reproduce this on ncaa?

@athomasmoz I'm having trouble reproducing this consistently on ncaa which, pending Daniela's response, makes me think we may wish to deprioritize this particular issue, unless it affects another site we care about.

I've found I'm able to reproduce this consistently on YouTube desktop when scrolling the screen down slightly, as per #2552.

---

[...]

Issue #2540, opened 2019-07-12
[2386] mp4 videos are translated offscreen in fullscreen after scrolling (4K Stick/Stick gen 2)


dnarcese:
Meta: #2386.

_Similar issue on 4K pendant: #2541_

## Steps to reproduce

1. Visit  https://www.travelandleisure.com/video (or https://youtube.com/?app=desktop&persist=1 or http://mcomella.xyz/test/video.html for minimal CSS)
2. Scroll the video offscreen **vertically**
3. Fullscreen 

Second example:  https://www.sbnation.com/videos the video is in a popup and when you scroll the webpage behind it the fullscreen video position changes. Also cc.com

### Expected behavior
Fullscreen video format fills the whole screen.
### Actual behavior
The fullscreen video is cropped to where it was scrolled

### Device information
* Fire TV device: 4k stick, stick gen 2
* Latest affected Firefox version: ?
* Earliest affected Firefox version: ?
* Turbo Mode: ?

### Notes
* Here's [a video of the STR on youtube.com](https://drive.google.com/file/d/1dKD9z6MI8oXgYSO-okXIz-VnBf2NRXlR/view?usp=sharing).
* Given the similarity to the issue affecting the 4K pendant, I assume this would also work in a raw WebView minimal test case. However, I have not tested it.
* [Minimal test case document](https://docs.google.com/document/d/1jh_dSSxXVhg6NIxe-wN9SnVn2F1MNCRk4FFTIrjAgKE/edit#)

mcomella:
I tried reproducing this issue on the 4K pendant and 2nd gen stick but I was unable to. However, I was also unable to reproduce https://github.com/mozilla-mobile/firefox-tv/issues/2541#issuecomment-515251226 today, which I could before, so I'll try again tomorrow.

mcomella:
I was able to reproduce on my video test page with minimal CSS – http://mcomella.xyz/test/video.html – which makes me think that our options are similar to the similar 4K pendant issue https://github.com/mozilla-mobile/firefox-tv/issues/2541#issuecomment-513985586

mcomella:
We can also reproduce this for videos on cc.com.

mcomella:
Calling `window.scrollTo(0, nonFullscreenScrollPosition);` while the video is fullscreen corrects the video position in fullscreen. I've been able to create a hack fix to correct the issue. It is:
- Cache the scroll position before fullscreening
[...]

PR #2661, opened 2019-08-09
Issue #2540, #2541: add mp4 video translation workaround

mcomella:
@dnarcese Can you additionally review my Remote Settings changes? It's necessary to enable this feature. Note that I haven't had the ability to test the Remote Settings code thoroughly because the experiment needs to exist before you can use command line overrides.

This code is confusing because it uses multiple fairly-arbitrary delays to make it work: let me know if you run into trouble.

---

This is a partial workaround that fixes:
- My test page: http://mcomella.xyz/test/video.html
- https://www.travelandleisure.com/video
- sbnation.com/videos

On the 4K Stick, 2nd gen Stick, and 4K pendant. However, the following sites are not fixed:
- youtube.com desktop (I assume there's either a CSS property that prevents this or specific code that runs for YouTube on the Fire OS WebView)

Caveats to the workaround:
- If the site is not finished loading (as considered by internal state), the hack won't work

## Checklist

  - This issue is not being closed
  - Hard to test: this is a UI issue (hard to test) in WebView (harder to test) that works around Fire OS-specific behavior (harder to test).

### DIFF
--- a/app/src/main/java/org/mozilla/tv/firefox/experiments/ExperimentConfig.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/experiments/ExperimentConfig.kt
@@ -16,7 +16,13 @@ enum class ExperimentConfig(val value: String) {
     AA_TEST("AAtest-1675"),
     HINT_BAR_TEST("HintBar-2011"),
     TV_GUIDE_CHANNELS("TvGuideChannels-2195"),
-    SEND_TAB("SendTab-2511")
+    SEND_TAB("SendTab-2511"),
+
+    /**
+     * This is not an experiment. If Amazon deploys a fix for this bug, our workaround
+     * may break it: we use this flag as an option to disable this workaround remotely.
+     */
+    MP4_VIDEO_WORKAROUND("Mp4VideoWorkaround-2540")
 }
 
 /**
--- a/app/src/main/java/org/mozilla/tv/firefox/ext/Js.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/ext/Js.kt
@@ -4,6 +4,8 @@
 
 package org.mozilla.tv.firefox.ext
 
+import mozilla.components.browser.session.Session
+
 object Js {
     const val CACHE_VAR = "_firefoxForFireTvPreviouslyFocusedElement"
     const val CACHE_JS = "var $CACHE_VAR = document.activeElement;"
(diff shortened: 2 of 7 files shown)


app/src/main/java/org/mozilla/tv/firefox/experiments/ExperimentConfig.kt
```diff
@@ -16,7 +16,13 @@ enum class ExperimentConfig(val value: String) {
...
     HINT_BAR_TEST("HintBar-2011"),
     TV_GUIDE_CHANNELS("TvGuideChannels-2195"),
-    SEND_TAB("SendTab-2511")
+    SEND_TAB("SendTab-2511"),
+
+    /**
+     * This is not an experiment. If Amazon deploys a fix for this bug, our workaround
+     * may break it: we use this flag as an option to disable this workaround remotely.
+     */
+    MP4_VIDEO_WORKAROUND("Mp4VideoWorkaround-2540")
```
dnarcese: Good idea putting it behind a flag.

app/src/main/java/org/mozilla/tv/firefox/ext/Js.kt
```diff
@@ -17,12 +19,79 @@ object Js {
...
+var $CACHED_SCROLL_POSITION;
+(function () {
+    /* We delay restoring the cached value because WebView may override the value we restore. This creates a race
+     * condition & is imperfect: on a slow page, sometimes WebView will override us anyway. This value was set through
+     * manual testing. I did not find better solutions.
+     *
+     * We use the cached scroll position from when this method is initially called so that we're less likely to use one
+     * of WebView's incorrect positions (see OBSERVE_SCROLL_POSITION). */
+    var nonFullscreenPageScrollPosition = $CACHED_SCROLL_POSITION;
+    window.setTimeout(() => {
```
dnarcese: Why do we need a timeout here and why did you choose a value longer than in `OBSERVE_SCROLL_POSITION`?

app/src/main/java/org/mozilla/tv/firefox/ext/Js.kt
```diff
@@ -17,12 +19,79 @@ object Js {
...
+    _firefoxTV_isScrollPositionObserverLoaded = true;
+
+    console.log('adding scroll event listener');
+    window.addEventListener('scroll', (e) => {
+        /* During the transition to fullscreen, the WebView may or may not change the scroll position to an incorrect
+         * value so we add a short delay before caching. This creates a race condition & is imperfect: on a slow page,
+         * it's possible the incorrect value will cache before we restore the fullscreen value so we'll restore the
+         * incorrect position. Also, if the user is fast enough, they can open fullscreen before we cache the value.
+         * This value was set through manual testing. I did not find better solutions. */
+        if (typeof _firefoxTV_lastTimeout !== 'undefined') {
```
dnarcese: Why do we need to clear the timeout here but not in `UPDATE_FULLSCREEN_SCROLL_POSITION`?

mcomella: It's an optimization: we could allow the timeout to fire for every scroll event and the fix would still work correctly but given how fast scroll events fire, there could be a non-trivial impact on the JS thread event queue.

I'll add a comment.

app/src/main/java/org/mozilla/tv/firefox/ext/Js.kt
mcomella: I describe the need for a timeout in my comment:

> We delay restoring the cached value because WebView may override the value we restore.

In other words, when fullscreen is pressed, the WebView sometimes updates the scroll position of the page. We delay a short duration so that we can ensure the WebView scroll position update occurs before we write our final scroll position. The duration we need was identified through manual testing.

We choose a different value from `OBSERVE_SCROLL_POSITION` because each timeout is solving a different race condition with a different necessary delay time.

I can see now my comment was unclear: I'll clarify it in a way similar to what I wrote here. Let me know if that's not good enough.

PR #2673, opened 2019-08-09
Import strings from android-l10n

Pike:
State: mozilla-l10n/android-l10n@ab580f757c76ed323a5c81ef92e62c9253efe673

This is a large diff, but most of it is comments and whitespace.

@mcomella, can you review this and merge?

Also, can you give me permission to tag PRs? We're using the `import-l10n` tag across repositories to have some visibility into what's going on across the repos.

### DIFF
--- a/app/src/main/res/values-de/strings.xml
+++ b/app/src/main/res/values-de/strings.xml
@@ -1,78 +1,217 @@
-<?xml version='1.0' encoding='utf-8'?>
-<resources>
-  <string name="firefox_tv_brand_name">Firefox TV</string>
-  <string name="firefox_tv_brand_name_short">Firefox</string>
-  <string name="pocket_brand_name">Pocket</string>
-  <string name="action_cancel">Abbrechen</string>
-  <string name="action_ok">OK</string>
-  <string name="urlbar_hint">Suchen oder Adresse eingeben</string>
-  <string name="menu_settings">Einstellungen</string>
-  <string name="menu_about">Über</string>
-  <string name="preference_language">Sprache</string>
-  <string name="preference_language_systemdefault">Systemstandard</string>
-  <string name="preference_category_privacy">Datenschutz</string>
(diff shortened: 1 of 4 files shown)


mcomella:
> Also, can you give me permission to tag PRs?

I added the l10n team within our org with "Triage" permissions: let me know if that didn't work for you.

app/src/main/res/values-de/strings.xml
```diff
@@ -1,78 +1,217 @@
-<?xml version='1.0' encoding='utf-8'?>
```
mcomella: I notice only 4 files have changed: why aren't the remaining locales affected? Specifically:
- ja
- es-rES
- it

mcomella:
@Pike Please merge once you've reviewed & responded to the comments, assuming there is no further action necessary. Thanks for setting this up!

Pike:
Triage would probably be good enough, but that doesn't allow us to merge or create labels. If you'd create the `import-l10n` label and merged the PRs on your side, then the Triage group is good enough. Also, we'll want `blocks-quarantine` on issues, to highlight issues that block us from merging the quarantine on our side. FWIW, on the other android projects, the reviewer also merges for us. That's easier for us as we're not deeply involved in other engineering tasks on these projects.

As for obsolete and untouched files, we're now localizing the strings.xml files directly, instead of doing a conversion step. That means, if there's no change by a localizer, the file won't change at all.

I created a tag for fire tv at https://pontoon.mozilla.org/projects/android-l10n/tags/fire-tv/ (CC @Delphine so that she knows) where you can see latest activity. `it` and `ja` have a suggestion, which is waiting for review.

mcomella:
@Pike As requested, I created the `import-l10n` and `blocks-quarantine` labels. For now, we'll keep your permissions as "Triage" and the reviewer should merge all the pull requests.

Issue #2528, opened 2019-07-11
FFTV Support for ADM


athomasmoz:
### Why/User Benefit/User Problem

As an Amazon device user, I want to see tabs sent from my other devices on my TV right away, without having to ask for them on my TV

### Acceptance Criteria (how do I know when I’m done?)

* ADM is supported in FFTV
* [Graceful degradation](https://developer.amazon.com/docs/adm/integrate-your-app.html#gracefully-degrade-if-adm-is-unavailable) is supported in FFTV

### Not included

* FFTV support, which is covered in separate ticket

dnarcese:
Followups:
- #2712 
- #2713

Issue #2653, opened 2019-08-07
Use only one http client


dnarcese:
### Vision statement / What / Requirements
Currently we are using both `HttpUrlConnectionClient` and `OkHttpClient` in different places in the code. 
### Impact
More consistency if we only use one. Also, there may be certain benefits of one over the other.
### Acceptance criteria
- One library is decided on
- All usages of the other library are replaced

mcomella:
I have a WIP in https://github.com/mcomella/firefox-tv/tree/wip-2653-one-http-client

PR #2694, opened 2019-08-15
Issue #2528: Add ADM

dnarcese:
## Checklist


app/src/main/java/org/mozilla/tv/firefox/FirefoxApplication.kt
```diff
@@ -70,6 +74,24 @@ open class FirefoxApplication : LocaleAwareApplication() {
                 updateExperiments()
             }
 
+            // This is needed for Rust components to make HTTP network requests
+            Megazord.init()
+            RustHttpConfig.setClient(lazy { HttpUrlConnectionWrapper.client })
+
+            var admAvailable = false
+            try {
+                Class.forName("com.amazon.device.messaging.ADM")
```
jonalmeida: I'm not sure how much I trust the Amazon [guide re graceful degradation](https://developer.amazon.com/docs/adm/integrate-your-app.html#gracefully-degrade-if-adm-is-unavailable). This check should technically pass since we are bundling the ADM jar that contains this class even if the device is not supported. 

It seems like there is also a [`ADM.isSupported`](https://s3-us-west-1.amazonaws.com/devportal-reference-docs/adm/API-Reference/reference/com/amazon/device/messaging/ADM.html#isSupported()) as well that might be an alternative or can be a secondary check in this block as well:

```kotlin
admAvailable = ADM.isSupported()
```

app/src/main/java/org/mozilla/tv/firefox/FirefoxApplication.kt
```diff
@@ -70,6 +74,24 @@ open class FirefoxApplication : LocaleAwareApplication() {
                 updateExperiments()
             }
 
+            // This is needed for Rust components to make HTTP network requests
+            Megazord.init()
+            RustHttpConfig.setClient(lazy { HttpUrlConnectionWrapper.client })
+
+            var admAvailable = false
```
jonalmeida: nit: We can probably re-introduce that setup method for push now that there are more checks that need to be done. 😅

app/build.gradle
```diff
@@ -216,6 +216,7 @@ dependencies {
     androidTestImplementation "org.jetbrains.kotlin:kotlin-reflect:${kotlin_version}"
     compileOnly files('libs/amazon-device-messaging-1.0.1.jar')
     testImplementation files('libs/amazon-device-messaging-1.0.1.jar')
+    testImplementation "org.mozilla.appservices:full-megazord-forUnitTests:0.37.1"
```
mcomella: Where did this version come from? Can you add a link so folks know where to find the latest version?

Also, what happens if this dependency gets out of sync with the dependency automatically provided by a-c? Can you follow-up with the app-services team?

Also also, this seems unrelated to ADM so probably shouldn't be grouped with it.

app/src/test/java/org/mozilla/tv/firefox/helpers/FirefoxRobolectricTestRunner.kt
```diff
@@ -0,0 +1,29 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.tv.firefox.helpers
+
+import org.mozilla.tv.firefox.helpers.shadows.ShadowADMMessageHandlerBase
+import org.mozilla.tv.firefox.helpers.shadows.ShadowADMMessageReceiver
+import org.robolectric.RobolectricTestRunner
+import org.robolectric.annotation.Config
+
+class FirefoxRobolectricTestRunner(testClass: Class<*>) : RobolectricTestRunner(testClass) {
```
mcomella: nit: add class kdoc

app/src/test/java/org/mozilla/tv/firefox/helpers/FirefoxRobolectricTestRunner.kt
```diff
@@ -0,0 +1,29 @@
...
+
+    /**
+     * Add ADM class shadows to global robolectric test runner config.
```
mcomella: nit: The first line of this kdoc is redundant to the code. The second line is far from the code where it happens and is likely to not get updated if the code gets updated.

Can we remove the first line and put the second line above the related code?

app/src/main/AndroidManifest.xml
```diff
@@ -41,7 +41,7 @@
         <!-- For ADM -->
         <amazon:enable-feature
             android:name="com.amazon.device.messaging"
-            android:required="true"/>
+            android:required="false"/>
```
mcomella: Can you add a comment to explain what this does? e.g. "if false, graceful degradation requried. if true, (and I'm not sure if this is true), APK will not be available to devices without ADM."

app/src/main/java/org/mozilla/tv/firefox/FirefoxApplication.kt
mcomella: > This check should technically pass since we are bundling the ADM jar that contains this class even if the device is not supported.

The ADM jar is compileOnly and the classes are provided on device so I assume their check would work.

That being said, `ADM.isSupported()` sounds much better if it's available on all devices.

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ It diverges in the following ways:
 [Semantic Versioning](http://semver.org/spec/v2.0.0.html) for bug fix releases.
 
 ## Unreleased
+### Added
+- Add support for Amazon Device Messaging (#2528)
```
mcomella: nit: remove "Add". The CHANGELOG is supposed to read like, "Added ... Support for Amazon Device Messaging".

Also, this isn't user facing yet so I'd preface with, "(disabled by default) Support for Amazon..."

app/src/main/java/org/mozilla/tv/firefox/FirefoxApplication.kt
```diff
@@ -78,19 +78,7 @@ open class FirefoxApplication : LocaleAwareApplication() {
...
-            } catch (e: ClassNotFoundException) {
-                android.util.Log.i(DEFAULT_LOGTAG, "ADM is not available on this device.")
-            }
-
-            // Only install push feature if ADM is available on this device
-            if (admAvailable) {
-                // This installs the [pushFeature] as a singleton
-                PushProcessor.install(serviceLocator.pushRepo.pushFeature)
-            }
+            initPush()
```
mcomella: Can this happen after `enableStrictMode`?

app/src/main/java/org/mozilla/tv/firefox/FirefoxApplication.kt
```diff
@@ -89,18 +90,14 @@ open class FirefoxApplication : LocaleAwareApplication() {
...
     private fun initPush() {
-        var admAvailable = false
-        try {
-            Class.forName("com.amazon.device.messaging.ADM")
-            admAvailable = true
-        } catch (e: ClassNotFoundException) {
-            android.util.Log.i(DEFAULT_LOGTAG, "ADM is not available on this device.")
-        }
-
+        var admAvailable = ADM(applicationContext).isSupported
```
mcomella: nit: you can inline this

app/src/main/java/org/mozilla/tv/firefox/FirefoxApplication.kt
```diff
@@ -89,18 +90,14 @@ open class FirefoxApplication : LocaleAwareApplication() {
...
         // Only install push feature if ADM is available on this device
```
mcomella: nit: -> "Only use push feature..."

I'm not sure what, "install push feature" means.

app/src/main/java/org/mozilla/tv/firefox/FirefoxApplication.kt
```diff
@@ -89,18 +90,14 @@ open class FirefoxApplication : LocaleAwareApplication() {
...
         if (admAvailable) {
+            Log.log(tag = DEFAULT_LOGTAG, message = "ADM not available")
             // This installs the [pushFeature] as a singleton
```
mcomella: nit: this comment is redundant to the implementation, which this calling code shouldn't be aware of anyway.

app/src/main/java/org/mozilla/tv/firefox/FirefoxApplication.kt
```diff
@@ -89,18 +90,14 @@ open class FirefoxApplication : LocaleAwareApplication() {
...
-            Class.forName("com.amazon.device.messaging.ADM")
-            admAvailable = true
-        } catch (e: ClassNotFoundException) {
-            android.util.Log.i(DEFAULT_LOGTAG, "ADM is not available on this device.")
-        }
-
+        var admAvailable = ADM(applicationContext).isSupported
         // Only install push feature if ADM is available on this device
         if (admAvailable) {
+            Log.log(tag = DEFAULT_LOGTAG, message = "ADM not available")
```
mcomella: The logger is intended to be used like https://groups.google.com/a/mozilla.com/forum/#!search/android$20logger/amazon-dev/6BIH4epQGU8/6AoGkzt5BAAJ

Also, the android Log statement below should be replaced with the a-c logger too.

app/src/test/java/org/mozilla/tv/firefox/helpers/shadows/ShadowADM.kt
```diff
@@ -0,0 +1,28 @@
...
+    /**
+     * [ADM] has RuntimeException stub in the compileOnly constructor
+     * which hinders unit testing. This overrides to exclude super() and the exception stub.
+     */
+    @Implementation
+    @Suppress("UNUSED_PARAMETER")
+    fun __constructor__(var1: Context) {}
+
+    @Implementation
+    fun isSupported() {}
```
mcomella: Why is this necessary? Add a comment. Should there be a return value here?

app/src/main/java/org/mozilla/tv/firefox/FirefoxApplication.kt
```diff
@@ -78,6 +89,18 @@ open class FirefoxApplication : LocaleAwareApplication() {
...
 
+    private fun initPush() {
+        var admAvailable = ADM(applicationContext).isSupported
+        // Only install push feature if ADM is available on this device
+        if (admAvailable) {
+            Log.log(tag = DEFAULT_LOGTAG, message = "ADM not available")
+            // This installs the [pushFeature] as a singleton
+            PushProcessor.install(serviceLocator.pushRepo.pushFeature)
+        } else {
+            android.util.Log.i(DEFAULT_LOGTAG, "ADM is not available on this device.")
```
mcomella: If ADM is not available, that means send tab is unavailable on this device. Can you file a follow-up to determine:
- Are there any devices we ship to without ADM?
- How do we want to handle devices without ADM? This includes people who sideload our builds to non-Amazon platforms and whatever we find out above
- If we don't want to handle it, add a Sentry warning so we know how many users are hitting this

mcomella: nit: this should probably be a warning, not an info.

app/build.gradle
```diff
@@ -209,6 +214,9 @@ dependencies {
     }
 
     androidTestImplementation "org.jetbrains.kotlin:kotlin-reflect:${kotlin_version}"
+    compileOnly files('libs/amazon-device-messaging-1.0.1.jar')
```
mcomella: a-c already includes this dependency which means there's a big chance this could go out of sync. Can you file a follow-up bug?

app/src/main/java/org/mozilla/tv/firefox/fxa/ADMService.kt
```diff
@@ -0,0 +1,12 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.tv.firefox.fxa
+
+import mozilla.components.lib.push.amazon.AbstractAmazonPushService
+import com.amazon.device.messaging.ADMMessageReceiver
+
+class ADMService : AbstractAmazonPushService() {
```
mcomella: We need code to integrate the API key, right? File a follow-up.

app/src/main/java/org/mozilla/tv/firefox/fxa/FxaRepo.kt
```diff
@@ -80,6 +82,10 @@ class FxaRepo(
 
         @Suppress("DeferredResultUnused") // No value is returned & we don't need to wait for this to complete.
         accountManager.initAsync() // If user is already logged in, the appropriate observers will be triggered.
+
+        SendTabFeature(accountManager, pushRepo.pushFeature) { _, _ ->
```
mcomella: Can this get GC'd? Instead of doing the research, the quickest solution is to keep it in a local reference just to be safe (and comment why we do that).

mcomella:
wrt the start-up crash, I had some ideas on Slack:

> wrt my review, I had an idea. If the branches you take never reference the Amazon APIs on non-Amazon API devices, it might not crash. I think this is why they use reflection to check if ADM is available. For example:
```kotlin
if (isADMAvailable) {
    // use Amazon APIs here
} else {
    // Do not use Amazon APIs here
}
```
> We'll have to be extra careful not to call any of the APIs. If it's possible, I think the classloader solution I mentioned (injecting stubs if it's not available, otherwise using implementations on device) is more fool-proof but it's more complicated and there are a lot of unknowns so I think we should go for this quicker solution, assuming it works.
> If you put all calls to the API behind a single object, e.g. ADMIntegration, and only reference that object inside isADMAvailable blocks (using reflection), I think that'd be the best structure. Ideally, ADMIntegration would be behind a build module so no one accidentally uses the APIs outside of that but that can be a later change
> That ADMIntegration could live in the ServiceLoader
> If you disagree with the approach or structure, let me know

app/build.gradle
dnarcese: I just used the latest version suggested from IDE. Should I add the link to where it is in maven?

dnarcese: Or to the github file line where it is declared?

app/src/main/java/org/mozilla/tv/firefox/fxa/ADMService.kt
dnarcese: The api key is only needed for debug builds. Amazon provides an api key for the release builds. Where would be the best place to document for devs that you need to include the debug key? Is a comment on this class an appropriate place?

app/src/main/java/org/mozilla/tv/firefox/fxa/FxaRepo.kt
dnarcese: The default lifecycle for `SendTabFeature` is the whole application

app/build.gradle
mcomella: Preferably, if app-services has documentation with latest version, point to that. Otherwise, maybe point to the github file (I'd say maven first but it should be hosted on maven.mozilla.org which isn't really a full web server).

app/src/main/java/org/mozilla/tv/firefox/fxa/ADMService.kt
mcomella: Probably:
- Above the "SendTabFeature" declaration (i.e. where this key is required for the feature)
- In the readme
- Send an email to the team notifying them of this behavior

app/src/main/java/org/mozilla/tv/firefox/fxa/FxaRepo.kt
mcomella: What do you mean by the default lifecycle?

I'm concerned that there may be no strong references to SendTabFeature and it may get garbage collected. a-c implements some of their listeners/observers with weak references so if those are the only references, SendTabFeature & it's observers may get GC'd.

app/src/main/java/org/mozilla/tv/firefox/FirefoxApplication.kt
```diff
@@ -70,6 +73,12 @@ open class FirefoxApplication : LocaleAwareApplication() {
                 updateExperiments()
             }
 
+            // This is needed for Rust components to make HTTP network requests
+            RustHttpConfig.setClient(lazy { HttpUrlConnectionWrapper.client })
```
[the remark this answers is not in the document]
dnarcese: This can be taken care of in #2653. It's not a simple swap right now.

app/src/main/java/org/mozilla/tv/firefox/fxa/FxaRepo.kt
jonalmeida: @mcomella The SendTabFeature accepts [a lifecycle owner](https://github.com/mozilla-mobile/android-components/blob/master/components/feature/sendtab/src/main/java/mozilla/components/feature/sendtab/SendTabFeature.kt#L46) that defaults to [ProcessLifecycleOwner](https://developer.android.com/reference/android/arch/lifecycle/ProcessLifecycleOwner) which should result in the feature never getting GC'd unless the lifecycle owner is destroyed.

app/src/main/java/org/mozilla/tv/firefox/FirefoxApplication.kt
jonalmeida: > The ADM jar is compileOnly and the classes are provided on device so I assume their check would work.

Nice, I didn't notice that fire TV needed to do that as well.

app/build.gradle
jonalmeida: > a-c already includes this dependency which means there's a big chance this could go out of sync.

We use `compileOnly` as well so we build against the same JAR, but don't include it either. I've filed https://github.com/mozilla-mobile/android-components/issues/4157 so that we expose the SDK we compile against to make it easier for consuming apps to do the same.

app/src/main/java/org/mozilla/tv/firefox/fxa/FxaRepo.kt
jonalmeida: Granted, it is a strange API we were trying to figure out how to reference during implementation since it doesn't need any references and only observes other features/managers.

app/src/main/java/org/mozilla/tv/firefox/fxa/PushRepo.kt
```diff
@@ -0,0 +1,26 @@
...
+ */
+class PushRepo(
+    context: Context
+) {
+    companion object {
+        private const val senderId = "fftv"
+    }
+
+    val admService by lazy { ADMService() }
+    private val pushConfig = PushConfig(senderId = senderId, serviceType = ServiceType.ADM)
```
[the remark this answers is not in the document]
dnarcese: The server has already setup that value. Issue for improving docs -> https://github.com/mozilla-mobile/android-components/issues/4159

app/src/main/java/org/mozilla/tv/firefox/FirefoxApplication.kt
jonalmeida: > If ADM is not available, that means send tab is unavailable on this device

@mcomella @dnarcese FYI, this is not true. When you 'sync' in FxA you can get new tabs from SendTabFeature.

What did before push support was added, was to do a poll `onStart` to check for new tabs, and it would follow the same flow as push if the service is unavailable.

app/src/main/java/org/mozilla/tv/firefox/fxa/FxaRepo.kt
mcomella: > The SendTabFeature accepts a lifecycle owner that defaults to ProcessLifecycleOwner which should result in the feature never getting GC'd unless the lifecycle owner is destroyed.

I wasn't sure that `ProcessLifecycleOwner` observers are guaranteed to not be GC'd but thinking back to how I used them in the past (as lambdas) and that the `Lifecycle.addObserver` method does not mention weak references, I now see that that should be fine. Additionally, our `ObserverRegistry.observers` keeps a strong reference to our observer.

app/build.gradle
```diff
@@ -218,10 +218,10 @@ dependencies {
...
     testImplementation "org.mozilla.appservices:full-megazord-forUnitTests:0.37.1"
 
-    compileOnly files('libs/amazon-device-messaging-1.0.1.jar')
-    testImplementation files('libs/amazon-device-messaging-1.0.1.jar')
-    // See https://github.com/mozilla/application-services/blob/master/.buildconfig-android.yml#L1 for latest version
-    testImplementation "org.mozilla.appservices:full-megazord-forUnitTests:0.37.1"
+    // Download from https://developer.amazon.com/sdk-download
+    // This library has not been updated since 2013, so it might be unlikely that the version will change
+    compileOnly files("libs/amazon-device-messaging-${adm_version}.jar")
+    testImplementation files("libs/amazon-device-messaging-${adm_version}.jar")
```
mcomella: I'm not sure what's best here. Now that we fixed the startup crash, we might be able to change this to `testCompileOnly` and remove the shadows. However, if we ever try to run tests that use these classes, the tests might crash without the shadows and the runtime dependencies.

Any thoughts? Maybe it's better just to add a comment and leave it how it is because it works now and we can go add more user value. :)

If you think it's appropriate to change this, can you file a bug for someone to change this into a-c, which also uses `testImplementation`?

mcomella: I also wonder if we should put all the Robolectric shadows in an a-c test dependency... I'm not sure how to import/use those though. Maybe file an a-c bug and don't bother doing it now?

app/src/main/java/org/mozilla/tv/firefox/FirefoxApplication.kt
mcomella: Thanks for clarifying, yes I mispoke. That being said, we have historically assumed all devices have push and weren't intending to support polling so we'll need to follow up on what to do for these users. It sounds like adding support is minimal effort so maybe that's what we should do.

Now that I think about it, adding polling support enables us to test on emulators too, which could be valuable.

This can be a follow-up.

app/src/main/java/org/mozilla/tv/firefox/FirefoxApplication.kt
```diff
@@ -29,6 +30,7 @@ import org.mozilla.tv.firefox.utils.ServiceLocator
 import org.mozilla.tv.firefox.webrender.WebRenderComponents
 
 private const val DEFAULT_LOGTAG = "FFTV"
+private val logger = Logger(DEFAULT_LOGTAG)
```
mcomella: nit: you can just inline the log tag

app/src/main/java/org/mozilla/tv/firefox/FirefoxApplication.kt
```diff
@@ -68,10 +69,16 @@ open class FirefoxApplication : LocaleAwareApplication() {
             initGlean()
             TelemetryIntegration.INSTANCE.init(this)
             initFretboard()
-            serviceLocator.admIntegration.initPush()
 
             enableStrictMode()
 
+            // For now, ignore the violations.
+            // android-components issue #4166:
+            // https://github.com/mozilla-mobile/android-components/issues/4166
```
mcomella: nit: I don't think you need the link & the bug number. I prefer the bug numbers: `a-c#4166`, which is similar to GitHub's cross-repository link syntax. For example: https://github.com/mozilla-mobile/android-components/issues/4166

app/build.gradle
```diff
@@ -215,7 +215,8 @@ dependencies {
 
     androidTestImplementation "org.jetbrains.kotlin:kotlin-reflect:${kotlin_version}"
 
-    // See https://github.com/mozilla/application-services/blob/master/.buildconfig-android.yml#L1 for latest version
+    // See https://github.com/mozilla-mobile/android-components/blob/master/buildSrc/src/main/java/Dependencies.kt#L29
```
mcomella: This line is going to change so can we provide a link to the file and a constant name instead? e.g.:
```kotlin
// For the latest version, see `Versions.mozilla_appservices` in
// <link>

File: app/src/main/java/org/mozilla/tv/firefox/fxa/ADMIntegration.kt [hunk 24]
```diff
@@ -0,0 +1,69 @@
...
+            Class.forName("com.amazon.device.messaging.ADM")
+            ADM(app).isSupported
+        } catch (e: ClassNotFoundException) {
+            logger.warn("ADM is not available on this device.")
+            false
+        }
+    }
+
+    private val pushFeature by lazy {
+        AutoPushFeature(app, ADMService(), PushConfig(senderId = "fftv", serviceType = ServiceType.ADM))
```
mcomella: nit: can we put `fftv` into a constant so it's less likely to get accidentally changed?

app/build.gradle
dnarcese: I think I'm confused how this relates to the startup crash. If the shadows are removed, aren't the tests going to fail on every PR?

mcomella: > If the shadows are removed, aren't the tests going to fail on every PR?

I'm not sure – we can try it! :)

My thinking was that Robolectric crashes for the same reason the app crashes on startup (it tries to load classes whose implementations are not available) and so by fixing the startup crash, we fixed the issue, provided those APIs are not accessed. However, it could be that's not the case (e.g. Robolectric has additional startup logic around class loading that will crash even though the app doesn't) and we can't remove the shadows.

Then again, we may need to load those Amazon classes in testing, even if it's just to mock them, so maybe it doesn't make sense to remove the classes at all. I'm not sure. Sorry if this is unclear but we can talk on Zoom to work through it if you'd like.

app/src/debug/AndroidManifest.xml
```diff
@@ -26,4 +28,67 @@
     <uses-permission android:name="android.permission.CHANGE_CONFIGURATION" tools:ignore="ProtectedPermissions" />
 
     <!-- *** Fastlane permissions end *** -->
+
+
```
mcomella: nit: excess newline

app/src/debug/AndroidManifest.xml
```diff
@@ -26,4 +28,67 @@
...
+    <!-- *** ADM permissions (temporarily in debug only) *** -->
+    <!-- This permission ensures that no other application can intercept your ADM messages. -->
+    <permission
+        android:name="${applicationId}.permission.RECEIVE_ADM_MESSAGE"
+        android:protectionLevel="signature"/>
+    <uses-permission android:name="${applicationId}.permission.RECEIVE_ADM_MESSAGE" />
+    <uses-permission android:name="com.amazon.device.messaging.permission.RECEIVE" />
+    <!-- *** ADM permissions end *** -->
+
+    <application
```
mcomella: I don't want to duplicate the application declaration: it's possible someone will update the application in one file but not the other and we'll:
- (if they change prod) Spend a lot of time debugging because their change doesn't show up in debug
- (if they change debug) Not ship something we intend to that shows up on debug builds

I think our options are:
- The manifests are supposed to be "merged", right? I think that means if we create an application with no attributes and no activities, it'll merge the attributes from the higher-priority manifest into this one, right?
- Do not include the application at all in debug, leaving the new amazon services and features in the production manifest

I'd say go with 2 but I'm concerned about the `android:permission` attribute in the `<receiver>` and the `<amazon:enable-feature` tag and if these might force us to accept a permissions bump. Do you think you could try to merge the manifests with less duplication (option 1)? If you spend more than 30min on this, maybe we should pair or think whether it's worth it.

dnarcese: I can do option 1 but I get a warning on the <application> tag that `App is not indexable by Google Search. Consider adding one activity...`. But it doesn't actually break anything and is only in the debug build so this should be okay.

dnarcese: Ok nevermind that fails TC on lint check :(

app/lint.xml
```diff
@@ -34,4 +34,7 @@
     <!-- We don't want to fail the build, nor do we want to to pollute our informational output. -->
     <issue id="Typos" severity="informational"/>
 
+    <!--  For debug manifest. This is only a temporary change.  -->
```
mcomella: nit: we may replace this with `tools:ignore`: see my comment below.

If we don't, can you elaborate why this occurs with the debug manifest? If it's a temporary change, we should probably add a bug # to explain why we can remove it

app/src/debug/AndroidManifest.xml
```diff
@@ -4,7 +4,9 @@
    - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:tools="http://schemas.android.com/tools"
-    tools:ignore="ImpliedTouchscreenHardware,MissingLeanbackLauncher,MissingLeanbackSupport">
+    tools:ignore="ImpliedTouchscreenHardware,MissingLeanbackLauncher,MissingLeanbackSupport"
+    xmlns:amazon="http://schemas.amazon.com/apk/res/android"
+    package="org.mozilla.tv.firefox">
     <!-- ImpliedTouchscreenHardware, MissingLeanbackLauncher, and MissingLeanbackSupport
```
mcomella: I didn't notice this comment before and I think it applies here too: if it works, I think we should only ignore `GoogleAppIndexing` in the debug manifest. I think this is worth it because that means our change doesn't need to be temporary.

Sorry if my IRL comments mislead you :\

app/src/main/AndroidManifest.xml
```diff
@@ -4,6 +4,7 @@
    - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
           xmlns:tools="http://schemas.android.com/tools"
+          xmlns:amazon="http://schemas.amazon.com/apk/res/android"
```
mcomella: nit: unnecessary now.

PR #2696, opened 2019-08-15
Issue #2643: FxA preboarding string change

dnarcese:
Another string change was made :) 
## Checklist


app/src/main/res/values/strings.xml
```diff
@@ -230,6 +230,10 @@
     %1$s will be replaced with Firefox.
     %2$s will be replaced with the name of our app (i.e. Firefox for Fire TV) -->
     <string name="fxa_preboarding_instruction">Send a tab from anywhere you use %1$s and see it here. Sign in to %2$s to get started.</string>
```
mcomella: Should we remove this string then?

dnarcese: I was going to leave it for now just in case the new string doesn't get localized in time. Do you think that's a good idea?

mcomella: Seems reasonable. File an issue to remove it though!

dnarcese: Done. #2705

Issue #2490, opened 2019-07-05
Accounts Button on Toolbar


athomasmoz:
### Why/User Benefit/User Problem

As a FFTV user, I want an account button on the toolbar, so I can easily find my account details and sign out

### Acceptance Criteria (how do I know when I’m done?)
#2529 must be done first because it exposes sign in state to the UI.

* There is an account button in the toolbar for users who are logged in
  * Button matches design
  * Icon description is "sent tab settings"
  * Clicking on the button does nothing until accounts page is ready
* Event is tracked for account button clicks

### Design

Text: "Sent tab settings"
https://brampitoyo.github.io/people/bpitoyo/firefox-for-fire-tv/receive-tabs/i03/#artboard10
<img width="960" alt="image" src="https://user-images.githubusercontent.com/43346208/61807572-a7c46b00-adee-11e9-8e9a-1580b4ff17db.png">

### Not included

* Building out the accounts page
* Localization, which will be done once strings are finalized

athomasmoz:
@brampitoyo can you add a link to the default asset (if user doesn't have a profile picture)

brampitoyo:
Icon assets for default profile:

[SVG](https://cl.ly/54b5aaf4a4b2/avatar-authenticated-no-picture.svg)

PNG: ![avatar-authenticated-no-picture](https://user-images.githubusercontent.com/51007/61846732-5e624180-aefc-11e9-9836-a31eb7539c6e.png)

PNG@2x: <img width="24" src="https://user-images.githubusercontent.com/51007/61846735-61f5c880-aefc-11e9-825f-bcdca93eed35.png">

mcomella:
@brampitoyo One mild concern I have with the current design is that users may feel uncomfortable seeing their profile photo, especially if it's their face, on the homescreen. Do you have any thoughts? I'm less concerned about this than I was on desktop – it's smaller and isn't around _all_ the time – so if you're fine, I'm fine. :)

mcomella:
I noticed there are issues when using custom FxA avatars:
- They are not round like the mocks
- There is no selected state when they are highlighted

See the screenshot for more:
![image](https://user-images.githubusercontent.com/759372/63900216-b89b6a00-c9b4-11e9-80f4-0fe0c372319d.png)

@psymoon Can you address these remaining issues in this bug?

mcomella:
[...]

PR #2702, opened 2019-08-19
#2490: Account button

psymoon:
## Checklist


app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/NavigationOverlayFragment.kt
```diff
@@ -257,6 +260,20 @@ class NavigationOverlayFragment : Fragment() {
         activity!!.moveTaskToBack(true)
     }
 
+    private fun observeAccountState(): Disposable {
+        return serviceLocator.fxaRepo.accountState.subscribe {
+            when (it) {
+                FxaRepo.AccountState.AUTHENTICATED_WITH_PROFILE -> Unit //TODO
+                FxaRepo.AccountState.AUTHENTICATED_NO_PROFILE ->
+                    fxaButton.setImageResource(R.drawable.ic_avatar_authenticated_no_picture)
```
mcomella: The state flow always transitions from one to the next:
- NOT_AUTHENTICATED
- AUTHENTICATED_NO_PROFILE (since the profile is fetched async after authentication so we don't have it at first)
- AUTHENTICATED_WITH_PROFILE

I'm concerned that we'll see a flicker when moving from AUTHENTICATED_NO_PROFILE to AUTHENTICATED_WITH_PROFILE. We should address this.

Maybe the simplest way would be to move the logic into the model: we could add a state for "fetching profile", which delays for 5 seconds or until the profile is fetched – whichever comes first. If it times out, it can move to a "failed to fetch profile" state (though we'd need to check when the profile fetch might be retryed by sync).

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/NavigationOverlayFragment.kt
```diff
@@ -264,11 +264,15 @@ class NavigationOverlayFragment : Fragment() {
...
-                FxaRepo.AccountState.AUTHENTICATED_NO_PROFILE ->
+                FxaRepo.AccountState.AUTHENTICATED_NO_PROFILE -> {
                     fxaButton.setImageResource(R.drawable.ic_avatar_authenticated_no_picture)
+                    fxaButton.contentDescription = resources.getString(R.string.fxa_navigation_item_signed_in2)
+                }
                 FxaRepo.AccountState.NEEDS_REAUTHENTICATION,
-                FxaRepo.AccountState.NOT_AUTHENTICATED ->
+                FxaRepo.AccountState.NOT_AUTHENTICATED -> {
                     fxaButton.setImageResource(R.drawable.ic_fxa_login)
+                    fxaButton.contentDescription = resources.getString(R.string.fxa_navigation_item_new)
```
mcomella: This String needs a substitution

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/NavigationOverlayFragment.kt
```diff
@@ -261,9 +261,20 @@ class NavigationOverlayFragment : Fragment() {
...
 
     private fun observeAccountState(): Disposable {
-        return serviceLocator.fxaRepo.accountState.subscribe {
+        val fxaRepo = serviceLocator.fxaRepo
+
+        return fxaRepo.accountState.subscribe {
             when (it) {
-                FxaRepo.AccountState.AUTHENTICATED_WITH_PROFILE -> Unit //TODO
+                FxaRepo.AccountState.AUTHENTICATED_WITH_PROFILE -> {
+                    fxaRepo.accountManager.accountProfile()?.let { profile ->
```
mcomella: If the accountProfile is null, we should log a Sentry error.

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/NavigationOverlayFragment.kt
```diff
@@ -268,7 +269,11 @@ class NavigationOverlayFragment : Fragment() {
                 FxaRepo.AccountState.AUTHENTICATED_WITH_PROFILE -> {
                     fxaRepo.accountManager.accountProfile()?.let { profile ->
                         if (profile.avatar != null) {
```
mcomella: You should check `url` is non-null here instead of below. In the current code, if url is null but avatar is not, we're not going to update the icon properly.

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/NavigationOverlayFragment.kt
```diff
@@ -268,7 +269,11 @@ class NavigationOverlayFragment : Fragment() {
...
-                            // TODO: need to fetch avatar
+                            profile.avatar?.url?.let { url ->
+                                PicassoWrapper.client
+                                    .load(url)
```
mcomella: We need to define an error case: what if we're unable to fetch the icon from the server?

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/NavigationOverlayFragment.kt
```diff
@@ -268,7 +269,11 @@ class NavigationOverlayFragment : Fragment() {
                 FxaRepo.AccountState.AUTHENTICATED_WITH_PROFILE -> {
```
mcomella: nit: you could import `AccountState.*` to improve readability:
```kotlin
AUTHENTICATED_WITH_PROFILE -> {
    // Do a thing
}

File: app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/NavigationOverlayFragment.kt [hunk 7]
```diff
@@ -257,6 +261,39 @@ class NavigationOverlayFragment : Fragment() {
...
+                FxaRepo.AccountState.AUTHENTICATED_NO_PROFILE -> {
+                    fxaButton.setImageResource(R.drawable.ic_avatar_authenticated_no_picture)
+                    fxaButton.contentDescription = resources.getString(R.string.fxa_navigation_item_signed_in2)
+                }
+                FxaRepo.AccountState.NEEDS_REAUTHENTICATION,
+                FxaRepo.AccountState.NOT_AUTHENTICATED -> {
+                    fxaButton.setImageResource(R.drawable.ic_fxa_login)
+                    fxaButton.contentDescription = resources.getString(R.string.fxa_navigation_item_new)
+                }
+                null -> Unit
```
mcomella: The account state should never be null, right? Rx doesn't allow it and we wouldn't ever want to get into this state anyway. Is it necessary to check null (because Java interop)? If so, why don't we throw an IllegalStateException here instead?

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/NavigationOverlayFragment.kt
```diff
@@ -257,6 +261,39 @@ class NavigationOverlayFragment : Fragment() {
...
+                    }
+                }
+                FxaRepo.AccountState.AUTHENTICATED_NO_PROFILE -> {
+                    fxaButton.setImageResource(R.drawable.ic_avatar_authenticated_no_picture)
+                    fxaButton.contentDescription = resources.getString(R.string.fxa_navigation_item_signed_in2)
+                }
+                FxaRepo.AccountState.NEEDS_REAUTHENTICATION,
+                FxaRepo.AccountState.NOT_AUTHENTICATED -> {
+                    fxaButton.setImageResource(R.drawable.ic_fxa_login)
+                    fxaButton.contentDescription = resources.getString(R.string.fxa_navigation_item_new)
```
mcomella: We need to update the hover-over hint too, right?

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/NavigationOverlayFragment.kt
```diff
@@ -257,6 +261,39 @@ class NavigationOverlayFragment : Fragment() {
...
+                        }
+                    }
+                }
+                FxaRepo.AccountState.AUTHENTICATED_NO_PROFILE -> {
+                    fxaButton.setImageResource(R.drawable.ic_avatar_authenticated_no_picture)
+                    fxaButton.contentDescription = resources.getString(R.string.fxa_navigation_item_signed_in2)
+                }
+                FxaRepo.AccountState.NEEDS_REAUTHENTICATION,
+                FxaRepo.AccountState.NOT_AUTHENTICATED -> {
+                    fxaButton.setImageResource(R.drawable.ic_fxa_login)
```
mcomella: nit: we can de-duplicate the assignment logic by putting the "all resources are local" case into a helper function:
```kotlin
function setIconAndContentDescription() {
    fxaButton.setImageResource(...)
    fxaButton.contentDescription = ...
}

NOT_AUTHENTICATED -> {
    setIconAndContentDescription(...)
}
```

This may seem pedantic but I think it helps prevent one branch from assigning one asset but not the other.

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/NavigationOverlayFragment.kt
```diff
@@ -51,10 +51,12 @@ import org.mozilla.tv.firefox.channels.DefaultChannel
 import org.mozilla.tv.firefox.channels.DefaultChannelFactory
```
mcomella: Can you address the PR checklist items?

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/NavigationOverlayFragment.kt
psymoon: Good idea, I'll ask Grisha about some of the edge cases with nullability

mcomella: @psymoon On second thought, you can add the non-null Profile into the Account State Observable and never have to deal with the null state.

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/NavigationOverlayFragment.kt
psymoon: @mcomella to what? I don't see it in the story

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/NavigationOverlayFragment.kt
psymoon: I like it

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/NavigationOverlayFragment.kt
mcomella: If you look at the String definition, it has a placeholder for "Firefox" (which is mentioned in the localization note):
https://searchfox.org/mozilla-mobile/rev/1bc47e1452899910a17191a3c9cd5084dbad0401/firefox-tv/app/src/main/res/values/strings.xml#220

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/NavigationOverlayFragment.kt
psymoon: @mcomella Did you mean the tooltip? They appear to be set with `contentDescription` - https://github.com/mozilla-mobile/firefox-tv/blob/master/app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/ToolbarUiController.kt#L68

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/NavigationOverlayFragment.kt
psymoon: Filed https://github.com/mozilla-mobile/firefox-tv/issues/2735

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/NavigationOverlayFragment.kt
mcomella: I was not aware, sgtm.

mcomella: Not necessary to change but I wonder if we should make that more literal, e.g. a function like, `setTooltipAndContentDescription`, instead of just setting the content description.

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/NavigationOverlayFragment.kt
mcomella: Good idea to file a follow-up to keep the scope down here.

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/NavigationOverlayFragment.kt
```diff
@@ -257,6 +264,37 @@ class NavigationOverlayFragment : Fragment() {
...
+                                override fun onSuccess() {}
+                            })
+                    }
+                }
+                AccountState.AuthenticatedNoProfile -> {
+                    fxaButton.setImageResource(R.drawable.ic_avatar_authenticated_no_picture)
+                    fxaButton.contentDescription = resources.getString(R.string.fxa_navigation_item_signed_in2)
+                }
+                AccountState.NeedsReauthentication,
+                AccountState.NotAuthenticated -> {
```
mcomella: nit: put these on the same line or put a new line above these two states. Otherwise, it's easy to not notice that both of these states apply to the lambda below.

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/NavigationOverlayFragment.kt
```diff
@@ -257,6 +264,37 @@ class NavigationOverlayFragment : Fragment() {
         activity!!.moveTaskToBack(true)
     }
 
+    private fun observeAccountState(): Disposable {
+        val fxaRepo = serviceLocator.fxaRepo
+
+        return fxaRepo.accountState.subscribe { accountState ->
+            when (accountState) {
+                is AccountState.AuthenticatedWithProfile -> {
+                    val fxaProfile = accountState.profile
```
mcomella: nit: there's a lot of nesting here: maybe we can move the code inside this lambda into a helper function?

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/NavigationOverlayFragment.kt
```diff
@@ -257,6 +264,37 @@ class NavigationOverlayFragment : Fragment() {
...
+                    if (fxaProfile.avatar != null) {
```
mcomella: What happens when the profile avatar is null? We need to set an image, right?

My assumption is that, if the states transition how we expect them to, we'll still be using the image from the last state (authenticated_no_profile) so, in practice, it'll look fine to users in the current implementation. However, this logic couples the previous state to the current state, making you consider all states the code can be in instead of just this one state when making changes to this state.

Instead, we should act like this state exists in isolation and handle all of its cases: i.e. we should set an image if `fxaProfile.avatar` is not null or if it's null, even if the value we set is redundant to the previous state. That way, we don't have to consider the previous state at all: the amount of code we need to consider when making changes to this code is reduced.

In general, Rx (and functional programming) want you to consider functions/states in isolation (including this one!) to reduce the amount of code/state you need to consider at one time.

---

That being said, it looks like Severin is updating the `FxaProfile` to return an `ImageSetStrategy` instead of an avatar URL https://github.com/mozilla-mobile/firefox-tv/pull/2723/files#diff-6b880929405068b06513952ce832061eR169 that already encapsulates the logic to convert null URLs to default avatars and such. **Once you rebase this code, I think this explicitly setting a value for avatar == null will be taken care of.** I left the explanation above in to hopefully provide more context on some Rx best practices!

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/NavigationOverlayFragment.kt
```diff
@@ -257,6 +264,37 @@ class NavigationOverlayFragment : Fragment() {
...
+                        PicassoWrapper.client
+                            .load(fxaProfile.avatar.url)
+                            .into(fxaButton, object : Callback {
+                                override fun onError(e: Exception?) {
+                                    Sentry.capture(e)
```
mcomella: I don't think we need to capture these errors: presumably, some amount of these requests will fail due to network or server errors so the number we receive will be kind of arbitrary.

It could give us value if we notice the number goes from a small amount to a large amount – i.e. everyone is starting to fail requests – but we're not currently equipped to do regression analysis like this. Furthermore, I think we'll be able to catch these types of failures through manual testing with TestRail.

mcomella: As per my comment above about explicitly handling all states, if we fail to set the image, we should set an image here too, right?

I notice the `ImageSetStrategy` doesn't provide an option for an image to insert on error: maybe you could expand it to do so? If not, please file a follow-up bug for that!

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/NavigationOverlayFragment.kt
mcomella: As per my comment above, if we're explicitly handling all states to decouple them from previous states, we should set the contentDescription here too, right?

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/NavigationOverlayFragment.kt
```diff
@@ -257,6 +264,37 @@ class NavigationOverlayFragment : Fragment() {
...
 
+    private fun observeAccountState(): Disposable {
+        val fxaRepo = serviceLocator.fxaRepo
+
+        return fxaRepo.accountState.subscribe { accountState ->
+            when (accountState) {
+                is AccountState.AuthenticatedWithProfile -> {
+                    val fxaProfile = accountState.profile
+                    if (fxaProfile.avatar != null) {
+                        PicassoWrapper.client
```
mcomella: wrt my "explicitly handling all states" comments, one way to make sure we're performing all assignments in all states is to avoid duplicating the final UI update code:
```kotlin
val contentDescription: String
val avatarPath: String
when (accountState) {
    is AccountState.AuthenticatedWithProfile -> {
        contentDescription = ...
        avatarPath = ...
    }
    ...
}

PicassoWrapper.client.load(avatarPath).into(fxaButton)
fxaButton.contentDescription = contenDescription
```

Then, when a new state is added, the compiler checks that they made the assignment in their new state and the UI update code will happen automatically.

In practice, it doesn't always work out easily because the behavior may be different (e.g. setting image from resource vs. from URL) but you can often use lambdas and other types, like our `ImageSetStrategy`, to make it work.

I suggest you use this pattern here – it's ideal – but you don't have to.

Issue #2491, opened 2019-07-05
Receive Tabs with Firefox in the foreground


athomasmoz:
### Why/User Benefit/User Problem

As an FFTV users, I want to receive a tab from Firefox on another device, so that I can watch/listen to content on my TV without having to navigate to the same page on my TV that I already have open on my other device

### Acceptance Criteria (how do I know when I’m done?)

BLOCK TICKET ON PUSH #2528 IF POLLING DOESN'T AUTOMATICALLY WORK
(This should work, and I think backgrounding then foregrounding the app should trigger a poll, but please verify)

* User can receive a tab on their TV from Fenix, iOS, or Desktop Firefox
* Tab will overtake current content
* A toast is displayed to the user to indicate what's happened
* User can go back & forward through their history stack after sending a tab, and that tab will be included
* A receive tab count is tracked, so that we can determine usage of the feature
  * Event does not include details on what the tab is
  * Can we track type of device? ie. desktop / android / iOS
  * After speaking with Ashley, we want to record when a tab is received and visible to the user. If a push is invalid (e.g. no `entries`), we can report those errors to Sentry. If there are many of those, we can investigate and adjust
* If a tab is sent when Firefox is not open, nothing happens

### Design

Finalized string: "Tab sent from [device name]"
[Mock](https://brampitoyo.github.io/people/bpitoyo/firefox-for-fire-tv/receive-tabs/i03/#artboard8)

### Not included

* Queuing tab#2654 tab will overtake the currently displayed page
* Receiving tabs in the background: #2654

liuche:
Until Push is working, the app can be backgrounded and resumed to trigger a fetch (rather than waiting for the 15 min polling).

mcomella:
I learned automatic a-c polling is going to be removed soon. However, you can trigger it manually:

> so, for now just call accountManager.authenticatedAccount()?.deviceConstellation()?.refreshDeviceStateAsync() when you want to check for new tabs (e.g. in onResume)

mcomella:
I have a WIP that receives tabs end-to-end. I need to clean up the code and address the other acceptance criteria.

mcomella:
@brampitoyo If we receive multiple tabs to open, what should we do? Options:
- Pick one of the tabs to open. Note: tabs aren't guaranteed to be received in order so we can't say, "Open the last one", but I think we can make a best effort
[...]

PR #2708, opened 2019-08-20
Issue #2491: receive tabs

mcomella:
N.B.: as of the latest a-c upgrade, the sign-in button does nothing.
In order to test this patch, you must manually add the `RustHttpConfig`.

## Checklist

  - Still need to add telemetry, handle multiple received tabs, test when the app is backgrounded (requires push), and test the back/forward stack (hard to do without push)
  - Didn't land yet
  - This PR doesn't close the issue


mcomella:
> I think we want to use the `SendTabFeature` which gives us back the device info and tab data, instead of implementing our own `DeviceEventsObserver`. The setup isn't merged yet but its here: https://github.com/mozilla-mobile/firefox-tv/pull/2694/files#diff-44a6506f2008849f2160267d8ac8cb5aR46

Yep – from my perspective, the code is pretty much the same (receive a tab and call a callback) so I figured I'd land this and we could fix it when your code landed or vice versa.

PR #2710, opened 2019-08-20
No issue: Init rustHttp client

dnarcese:
## Checklist

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -150,6 +150,7 @@ dependencies {
     implementation "org.mozilla.components:service-firefox-accounts:$moz_components_version"
     implementation "org.mozilla.components:support-ktx:$moz_components_version"
     implementation "org.mozilla.components:support-utils:$moz_components_version"
+    implementation "org.mozilla.components:support-rusthttp:$moz_components_version"
     implementation "org.mozilla.components:service-glean:$moz_components_version"
 
     systemImplementation "org.mozilla.components:browser-engine-system:$moz_components_version"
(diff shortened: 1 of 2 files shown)


app/build.gradle
```diff
@@ -209,6 +210,9 @@ dependencies {
     }
 
     androidTestImplementation "org.jetbrains.kotlin:kotlin-reflect:${kotlin_version}"
+
+    // See https://github.com/mozilla/application-services/blob/master/.buildconfig-android.yml#L1 for latest version
+    testImplementation "org.mozilla.appservices:full-megazord-forUnitTests:0.37.1"
```
mcomella: I used the link and determined there's a newer version as of 23 hours ago. That being said, I don't think it'll break our code to be a little out-of-date so in the name of forward progress, I'm just going to land this.

jonalmeida: This was my mistake when I spoke with @dnarcese offline to document where the version number should come from.

It should actually follow the version that A-C is using: https://github.com/mozilla-mobile/android-components/blob/master/buildSrc/src/main/java/Dependencies.kt#L29

mcomella: @dnarcese Can you fix the link?

dnarcese: yup done [here](https://github.com/mozilla-mobile/firefox-tv/pull/2694/commits/b9aad86592ec754ba68e1e604c2a74d3644a8c6b)

pocmo: Also heads up: With the next release we ship a `support-test-appservices` component that will include the right `forUnitTests` version as a transitive dependency so that you do not have to sync them manually anymore.

mcomella: I filed https://github.com/mozilla-mobile/firefox-tv/issues/2738 to do the replacement.

PR #2715, opened 2019-08-20
Pass profile data along with `AuthenticatedWithProfile` state

Baron-Severin:
## Checklist

### DIFF
--- a/app/src/test/java/org/mozilla/tv/firefox/fxa/FxaRepoTest.kt
+++ b/app/src/test/java/org/mozilla/tv/firefox/fxa/FxaRepoTest.kt
@@ -74,6 +74,6 @@ class FxaRepoTest {
     fun `WHEN on profile update callback is called THEN account state is authenticated with profile`() {
         val profile = Profile("uid", "email", null, "displayName")
         fxaRepo.accountObserver.onProfileUpdated(profile)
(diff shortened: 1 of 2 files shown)


app/src/test/java/org/mozilla/tv/firefox/fxa/FxaRepoTest.kt
```diff
@@ -72,8 +72,8 @@ class FxaRepoTest {
 
     @Test
     fun `WHEN on profile update callback is called THEN account state is authenticated with profile`() {
-        val profile = mockk<Profile>()
+        val profile = Profile("uid", "email", null, "displayName")
         fxaRepo.accountObserver.onProfileUpdated(profile)
-        accountStateTestObs.assertValueAt(1, FxaRepo.AccountState.AuthenticatedWithProfile(profile))
+        accountStateTestObs.assertValueAt(1, FxaRepo.AccountState.AuthenticatedWithProfile(profile.toDomainObject()))
```
mcomella: nit: Should we just pass in `FxaProfile`? Otherwise we're technically testing whatever we're using this function to test and the `toDomainObject` function

Baron-Severin: Ideally I think `onProfileUpdated` will be doing some validation soon (which will need its own tests), so I don't think we can pass an `FxaProfile`.  I can update this to just check that the type is correct though, then we'll test the actual validation later.

PR #2747, opened 2019-08-26
2495 receive tab pre-boarding

Baron-Severin:
@mcomella 
### Reviewer notes

Known issues with this code
- There is a short flash of home screen between the 2nd and 3rd onboarding screens (if this approach is accepted, I will open a follow up)
- I tried to take the simplest solutions and avoid overengineering, largely because a lot of this code probably __shouldn't__ be easy to write (e.g., I didn't want an easy way to communicate between activities, like a bus).  However I may have gone too far the other way.  __I am extremely open to suggestions for better approaches here__.

## Checklist

- [X] Add thorough **tests** or an explanation of why it does not
UI changes only
- [X] Add a **CHANGELOG entry** if applicable


app/src/main/res/layout/receive_tab_preboarding.xml
```diff
@@ -0,0 +1,93 @@
+<?xml version="1.0" encoding="utf-8"?>
```
mcomella: I don't think `receive_tabs_preboarding` should be in the nodpi folder (though I don't entirely understand it). From [SO](https://stackoverflow.com/questions/34156957/what-is-the-difference-between-anydpi-and-nodpi), nodpi is:

> These are density-independent resources. The system does not scale resources tagged with this qualifier, regardless of the current screen's density.

Which means that the asset will appear have a different density-independent size at different density levels thus appearing smaller or larger on screen at different density levels. What we want to happen is that they fill the same amount of space of on the screen regardless of the density level.

Our other assets are in xhdpi and it seems to be working fine: maybe we should just put this in there.

app/src/main/java/org/mozilla/tv/firefox/MainActivity.kt
```diff
@@ -177,6 +152,32 @@ class MainActivity : LocaleAwareAppCompatActivity(), OnUrlEnteredListener, Media
     override fun onResume() {
         super.onResume()
         TelemetryIntegration.INSTANCE.startSession(this)
+
+        val safeIntent = intent.toSafeIntent()
+        // Skip onboarding if turbo mode is set via intent. This is used in automated perf testing.
```
mcomella: super-nit: I prefer when the comments are on top of a code block (though this may be many years of convention) and I haven't seen many code bases where comments in between code are a valid convention. This isn't well thought through but I think:
- the comments are more noticeable at the top
- If a block of code is so confusing that it needs a comment on its own line, that block of code should be grouped separately from nearby code. However, I often see:
```kotlin
unrelatedCode.doesSomethingUnrelated()
// Comment for a confusing block of code
confusingCode.doesSomethingConfusing()
moreLinesOfConfusingCode.goHere()
```

In our particular case, the comment is applying to the entire block of code, rather than to the if statement it's situated above so to me it makes more sense to put it at the top of the block.

app/src/main/java/org/mozilla/tv/firefox/MainActivity.kt
```diff
@@ -158,25 +157,23 @@ class MainActivity : LocaleAwareAppCompatActivity(), OnUrlEnteredListener, Media
         // See #1881 for details
```
mcomella: nit: can we add a comment to explain how this onboarding experience is intended to work? i.e. we open onboarding, then return to MainActivity so a new onboarding is shown, then return to MainActivity, etc. so we see all of them because this is in onResume?

fwiw, I do very much prefer this approach to launching onboarding to the previous one.

app/src/main/java/org/mozilla/tv/firefox/MainActivity.kt
```diff
@@ -158,25 +157,23 @@ class MainActivity : LocaleAwareAppCompatActivity(), OnUrlEnteredListener, Media
...
-                val onboardingIntent = Intent(this@MainActivity, OnboardingActivity::class.java)
-                startActivity(onboardingIntent)
+            val onboardingIntent = when {
+                settings.shouldShowTurboModeOnboarding() -> {
+                    Intent(this@MainActivity, OnboardingActivity::class.java)
+                }
+                settings.shouldShowTVOnboarding(localeManager, this) -> {
+                    Intent(this@MainActivity, ChannelOnboardingActivity::class.java)
+                }
+                // Receive tab must _always_ be the last onboarding shown. This is because it
```
mcomella: nit: as per my earlier comment, I'd put a newline above this

app/src/main/java/org/mozilla/tv/firefox/MainActivity.kt
```diff
@@ -158,25 +157,23 @@ class MainActivity : LocaleAwareAppCompatActivity(), OnUrlEnteredListener, Media
...
+                // can take the user to sign in
+                settings.shouldShowReceiveTabsOnboarding() -> {
+                    Intent(this@MainActivity, ReceiveTabOnboardingActivity::class.java)
+                }
+                else -> null
             }
+            if (onboardingIntent != null) startActivity(onboardingIntent)
```
mcomella: nit: we generally add a newline after a `{ }` because it makes it harder to overlook the code afterwards. Also, it generally makes sense that a large if statement should be grouped together and the following code is less related to the if statement so should start it's own group.

app/src/main/java/org/mozilla/tv/firefox/MainActivity.kt
```diff
@@ -158,25 +157,23 @@ class MainActivity : LocaleAwareAppCompatActivity(), OnUrlEnteredListener, Media
         // See #1881 for details
         if (!safeIntent.hasExtra("TURBO_MODE")) {
```
mcomella: nit: can we put the onboarding code in a helper method? It's starting to get long and complicated. fwiw, I generally try to put non-trivial calls in `onCReate`/`onResume`/etc. into helper functions to make them read like high level methods:
```kotlin
initCrashReporting()
initTelemetry()
// etc.
```

app/src/main/java/org/mozilla/tv/firefox/onboarding/ReceiveTabPreboardingActivity.kt
```diff
@@ -0,0 +1,66 @@
...
+            R.string.receive_tab_onboarding_description,
+            resources.getString(R.string.firefox_tv_brand_name_short),
+            resources.getString(R.string.firefox_tv_brand_name)
+        )
+
+        buttonSignIn.setOnClickListener {
+            application.registerActivityLifecycleCallbacks(object : Application.ActivityLifecycleCallbacks {
+                override fun onActivityResumed(activity: Activity?) {
+                    if (activity is MainActivity) {
+                        serviceLocator.fxaLoginUseCase.beginLogin(activity.supportFragmentManager)
```
mcomella: Why is it necessary to put beginLogin in `ActivityLifecycleCallbacks`? What happens if you just call the method?

app/src/main/java/org/mozilla/tv/firefox/onboarding/ReceiveTabPreboardingActivity.kt
```diff
@@ -0,0 +1,66 @@
...
+        setContentView(R.layout.receive_tab_preboarding)
+
+        descriptionText.text = resources.getString(
+            R.string.receive_tab_onboarding_description,
+            resources.getString(R.string.firefox_tv_brand_name_short),
+            resources.getString(R.string.firefox_tv_brand_name)
+        )
+
+        buttonSignIn.setOnClickListener {
+            application.registerActivityLifecycleCallbacks(object : Application.ActivityLifecycleCallbacks {
```
mcomella: nit: can we make an `ActivityLifecycleCallbacksAdapter` with default implementations to simplify this call site? I wonder if anything already exists, e.g. in ktx, for this.

app/src/main/java/org/mozilla/tv/firefox/onboarding/ReceiveTabPreboardingActivity.kt
Baron-Severin: I couldn't find any better way to get a reference to the `MainActivity`, at least not any that aren't super abusable.  I'm open to better options though.

mcomella: Discussed IRL: we'll add an observable instead. It's unlikely to get abused because most of our app is in one Activity and it maybe doesn't make sense to overly obscure this code to prevent abuse if it results in a lot of added complexity and unconventional code.

app/src/main/res/layout/receive_tab_preboarding.xml
```diff
@@ -0,0 +1,93 @@
...
+        android:layout_height="wrap_content"
+        android:layout_marginStart="76dp"
+        android:layout_marginTop="24dp"
+        android:layout_marginEnd="76dp"
+        android:adjustViewBounds="true"
+        android:contentDescription="@string/content_description_receive_tab_onboarding_graphic"
+        app:layout_constraintEnd_toEndOf="parent"
+        app:layout_constraintStart_toStartOf="parent"
+        app:layout_constraintTop_toBottomOf="@+id/titleText"
+        app:srcCompat="@drawable/receive_tabs_preboarding" />
```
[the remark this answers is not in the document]
Baron-Severin: I honestly know nothing about it, but I saw it somewhere and thought it sounded like a better plan.

app/src/main/res/layout/receive_tab_preboarding.xml
```diff
@@ -0,0 +1,93 @@
...
+
+    <ImageView
+        android:id="@+id/receiveTabGraphic"
+        android:layout_width="0dp"
+        android:layout_height="wrap_content"
+        android:layout_marginStart="76dp"
+        android:layout_marginTop="24dp"
+        android:layout_marginEnd="76dp"
+        android:adjustViewBounds="true"
+        android:contentDescription="@string/content_description_receive_tab_onboarding_graphic"
```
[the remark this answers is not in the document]
Baron-Severin: We don't.  But it's easy to give one to every image, and I think following that policy makes it less likely that we will make a mistake somewhere and leave an image that should have a CD without one.

app/src/main/res/layout/receive_tab_preboarding.xml
```diff
@@ -0,0 +1,93 @@
...
+        android:fontFamily="@string/font_ember_regular"
+        android:textColor="@color/photonGrey10_a80p"
+        android:textSize="18sp"
+        app:layout_constraintEnd_toEndOf="parent"
+        app:layout_constraintStart_toStartOf="parent"
+        app:layout_constraintTop_toBottomOf="@+id/receiveTabGraphic" />
+
+    <Button
+        android:id="@+id/buttonSignIn"
+        style="@style/Widget.AppCompat.Button.Borderless"
```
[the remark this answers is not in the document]
Baron-Severin: I used the same style as our onboarding activity.

app/src/main/res/layout/receive_tab_preboarding.xml
```diff
@@ -0,0 +1,93 @@
...
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:layout_marginTop="32dp"
+        android:layout_marginEnd="56dp"
+        android:background="@drawable/turbo_button_enable"
+        android:fontFamily="@string/font_ember_regular"
+        android:minWidth="@dimen/onboarding_button_width"
+        android:minHeight="@dimen/onboarding_button_height"
+        android:nextFocusUp="@+id/buttonNotNow"
```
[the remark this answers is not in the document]
Baron-Severin: Good point on VV.  I kept this working the same way it does in our turbo onboarding activity, so if it turns out that we need to change this we should remove it there as well.

app/src/main/java/org/mozilla/tv/firefox/MainActivity.kt
Baron-Severin: Spoke about this in person.  I find the comment to apply to the following line, but I can see how it does also somewhat apply to the entire block.  I'm happy to move it.

app/src/main/java/org/mozilla/tv/firefox/onboarding/ReceiveTabPreboardingActivity.kt
Baron-Severin: Looks like we're changing this approach, so resolving this comment.

app/src/main/java/org/mozilla/tv/firefox/onboarding/ReceiveTabPreboardingActivity.kt
Baron-Severin: Per our conversation, replacing this with a bus.

app/src/main/res/layout/receive_tab_preboarding.xml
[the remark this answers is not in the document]
Baron-Severin: It looks like it works with everything that `src` handles, but also accepts vectors.

app/src/main/res/layout/receive_tab_preboarding.xml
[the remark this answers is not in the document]
mcomella: But it's not easy for our volunteer localizers who need to translate a bunch of unused strings. The linter will catch us if we don't explicitly provide a contentDescription or explicitly ignore the warning so now I'm even more convinced we shouldn't be adding a contentDescription everywhere.

We could ask liuche if we need a second opinion.

PR #2772, opened 2019-08-29
Fixes #2494: fixes avatar images not being rounded

### LINKED ISSUE #2494: Account Page & Signout

As an FFTV user, I want to see who's logged into FFTV and log out of my account when I'm not using it, so that others in my family can log in to send tabs

* Account page showing username & profile image
* Account page is shown when user is logged in and clicks on the toolbar icon
* Option to see send tab tutorial (links to nothing for now, since we don't have that screen yet)
* User is logged out when they click on the disconnect/logout button
* Clicks on each button on page is tracked (~~back~~, disconnect, receive tab tutorial)
* Initially focused icon should be "Receive Tabs" button

### Design

* Strings should match what's already in the app:
  * Title: "Sent tabs"
  * User image: "Signed in as <line break>[username@email.com]"
  * Tabs button: "Firefox tabs"
  * Sign out button: "Sign out "
* [Mock](https://brampitoyo.github.io/people/bpitoyo/firefox-for-fire-tv/receive-tabs/i03/#artboard12)

### Not included

* Localization. This will come later

Baron-Severin:
## Checklist

- [] Confirm the **acceptance criteria** is fully satisfied in the issue(s) this PR will close

### DIFF
--- a/app/src/main/java/org/mozilla/tv/firefox/channels/ChannelTile.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/channels/ChannelTile.kt
@@ -11,6 +11,8 @@ import android.view.ViewGroup
 import android.widget.ImageView
 import androidx.core.view.updateLayoutParams
 import androidx.recyclerview.widget.RecyclerView
+import com.squareup.picasso.RequestCreator
+import com.squareup.picasso.Transformation
 import org.mozilla.tv.firefox.R
 import org.mozilla.tv.firefox.ext.getDimenPixelSize
 import org.mozilla.tv.firefox.utils.PicassoWrapper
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/mozilla/tv/firefox/channels/ChannelTile.kt
```diff
@@ -85,9 +87,27 @@ data class ChannelTile(
...
+    protected var transformation: Transformation? = null
+
+    // This cannot be done during strategy instantiation because it will often require
+    // information about the target ImageView (particularly its size)
+    fun setTransformation(transformation: Transformation): ImageSetStrategy {
+        this.transformation = transformation
+        return this
+    }
+
+    protected fun RequestCreator.applyTransformationIfNotNull(transformation: Transformation?): RequestCreator = this
```
dnarcese: nit: Can this function be simplified for readability to 
```
protected fun RequestCreator.applyTransformationIfNotNull(transformation: Transformation?): RequestCreator {
    transformation?.let { transform(it) }
    return this
}
```

Baron-Severin: There is one small difference between those two code blocks. In the suggested version, we don't return the result of transform. Usually the Java builder pattern uses a mutable builder, so it would do the same thing, but it isn't necessarily guaranteed to.  We can change it to this, though, which I think preserves the intent of your request:

```kotlin
protected fun RequestCreator.applyTransformationIfNotNull(transformation: Transformation?): RequestCreator {
    transformation?.let { return transform(it) }
    return this
}

```

PR #2785, opened 2019-09-04
Issue #2490: Fix account image in toolbar

dnarcese:
I also noticed that the icons in the FxA button were the wrong colour because of an alpha applied - so I removed the alpha.
## Checklist

### DIFF
--- a/app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/NavigationOverlayFragment.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/NavigationOverlayFragment.kt
@@ -55,6 +55,7 @@ import org.mozilla.tv.firefox.fxa.FxaRepo.AccountState
 import org.mozilla.tv.firefox.pocket.PocketViewModel
 import org.mozilla.tv.firefox.telemetry.MenuInteractionMonitor
 import org.mozilla.tv.firefox.telemetry.UrlTextInputLocation
+import org.mozilla.tv.firefox.utils.RoundCornerTransformation
 import org.mozilla.tv.firefox.utils.ServiceLocator
 import org.mozilla.tv.firefox.utils.ViewUtils
 import org.mozilla.tv.firefox.widget.InlineAutocompleteEditText
@@ -266,7 +267,9 @@ class NavigationOverlayFragment : Fragment() {
             .subscribe { accountState ->
                 when (accountState) {
                     is AccountState.AuthenticatedWithProfile -> {
-                        accountState.profile.avatarSetStrategy.invoke(fxaButton)
+                        accountState.profile.avatarSetStrategy
+                            .setTransformation(RoundCornerTransformation(fxaButton.width.toFloat()))
+                            .invoke(fxaButton)
                         fxaButton.contentDescription = resources.getString(R.string.fxa_navigation_item_signed_in2)
                     }
                     AccountState.AuthenticatedNoProfile -> {
--- a/app/src/main/res/layout/fragment_navigation_overlay_top_nav.xml
+++ b/app/src/main/res/layout/fragment_navigation_overlay_top_nav.xml
@@ -53,8 +53,10 @@
         tools:ignore="ContentDescription"
         android:id="@+id/fxaButton"
         style="@style/NavigationButton"
+        android:scaleType="fitXY"
         android:src="@drawable/ic_fxa_login"
-        android:visibility="gone" />
+        android:visibility="gone"
+        android:tint="@null"/>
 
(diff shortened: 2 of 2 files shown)


app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/NavigationOverlayFragment.kt
```diff
@@ -266,8 +267,11 @@ class NavigationOverlayFragment : Fragment() {
             .subscribe { accountState ->
                 when (accountState) {
                     is AccountState.AuthenticatedWithProfile -> {
-                        accountState.profile.avatarSetStrategy.invoke(fxaButton)
+                        accountState.profile.avatarSetStrategy
+                            .setTransformation(RoundCornerTransformation(fxaButton.width.toFloat()))
+                            .invoke(fxaButton)
                         fxaButton.contentDescription = resources.getString(R.string.fxa_navigation_item_signed_in2)
+                        fxaButton.imageTintList = null
```
Baron-Severin: Do we want the image tint list in other blocks? If not, can we remove this in the XML instead of here?

app/src/main/res/drawable/ic_avatar_authenticated_no_picture.xml
```diff
@@ -6,6 +6,5 @@
   <path
       android:pathData="m12,22c-5.5228,0 -10,-4.4772 -10,-10s4.4772,-10 10,-10 10,4.4772 10,10c-0.0061,5.5203 -4.4797,9.9939 -10,10zM17.6655,14.2092c-1.4812,-1.7094 -3.5322,0.5908 -5.6658,0.5908 -2.1322,0 -4.1846,-2.3002 -5.6658,-0.5908 -0.42,0.4844 -0.4536,1.19 -0.0938,1.7192 1.2558,1.848 3.3558,3.0716 5.7596,3.0716 2.4038,0 4.5038,-1.2236 5.7596,-3.0716 0.3612,-0.5292 0.3262,-1.2348 -0.0938,-1.7192zM12,13.5c2.3474,0 4.25,-1.9026 4.25,-4.25 0,-2.3474 -1.9026,-4.25 -4.25,-4.25 -2.3474,0 -4.25,1.9026 -4.25,4.25 0,2.3474 1.9026,4.25 4.25,4.25z"
       android:fillColor="#f9f9fa"
-      android:fillAlpha=".8"
```
Baron-Severin: Could you run the new colors by @brampitoyo to make sure they're what he intended?  Also, are there any other SVGs we should be updating this way?

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/NavigationOverlayFragment.kt
dnarcese: Yes, we need it on the other buttons.

app/src/main/res/drawable/ic_avatar_authenticated_no_picture.xml
dnarcese: I'm going to remove this change and file a follow-up.

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/NavigationOverlayFragment.kt
Baron-Severin: I was unclear, by other blocks I meant the other conditions of the when expression.  Right now it's going to keep its tint list until eventually going through `AccountState.AuthenticatedWithProfile`, but then lose it and never regain it.  So any of the other states could look two different ways depending on the flow that the user goes through.

Instead of nulling this out here, can we null it in the XML so that it's always the same for this button?

dnarcese: I think that we have to do it in code, because the style on the button applies its own tint. I tested signing in and signing out and the icon colour doesn't change so I think this line is okay? Adding back any tints can be taken care of in the follow-up since it's more related to the icons rather than the profile image.

Baron-Severin: I believe all we need to do is add `android:tint="@null"` to the button.  iirc, the more specific rule (on the view) will override the more general one (the style).

More generally, this is safe _right now_, since the style isn't changing anything we particularly care about, but that could change if the style were ever updated.  We've taken great pains to make sure that the data and the UI map 1:1, so it's worth putting in extra effort to avoid creating a situation where stale view state winds up changing our UI.

PR #2792, opened 2019-09-04
No issue: update README to indicate we share api_key.txt.

mcomella:
We don't want to link directly to the keys because, had we done so, external
folks might request access and that could lead to someone accidentally
providing these folks access.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,7 +135,11 @@ API services currently supported are:
 * pocket_key
 
 ##### ADM API key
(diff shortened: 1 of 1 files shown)


dnarcese:
I just thought of the fact that any outside of our team won't be able to add their own debug key, because they don't have access to our app on the dev portal. So, should the instructions be more tailored to the core team?

mcomella:
@dnarcese Good catch: I updated my PR with an update that mentions ADM testing is for core team members only.

PR #2796, opened 2019-09-05
 Upgrades .taskcluster.yml to V1, adds decision task

mitchhentges:
The point of this PR is to modernize your `.taskcluster.yml` and to create a "decision task" that's written in Python. This decision task gives us flexibility to have more advanced logic, such emailing upon a build finishing or changing behaviour based on release tag

## Checklist

- [ ] Update docker image to include `taskcluster` python module (installation delay can be > 10 minutes, I don't want to slow down your PRs)

@mcomella and @JohanLorenzo for review

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -1,160 +1,124 @@
-# This Source Code Form is subject to the terms of the Mozilla Public
-# License, v. 2.0. If a copy of the MPL was not distributed with this
-# file, You can obtain one at http://mozilla.org/MPL/2.0/.
-
-version: 0
-allowPullRequests: collaborators
+version: 1
+policy:
+  pullRequests: collaborators
 tasks:
-###############################################################################
-# Task: Pull requests
-#
-# Triggered whenever a pull request is opened or updated.
-#
-# - Build the app (all flavors)
-# - Run unit tests
-# - Run code quality tools (findbugs, lint, checkstyle etc.)
-###############################################################################
-  - provisionerId: '{{ taskcluster.docker.provisionerId }}'
-    workerType: '{{ taskcluster.docker.workerType }}'
-    extra:
-      github:
-        events:
-          - pull_request.opened
-          - pull_request.edited
-          - pull_request.synchronize
-          - pull_request.reopened
-    scopes:
-    - "secrets:get:project/mobile/firefox-tv/tokens"
-    payload:
-      maxRunTime: 3600
-      deadline: "{{ '2 hours' | $fromNow }}"
-      image: 'mozillamobile/firefox-tv:2.3'
-      command:
-        - /bin/bash
-        - '--login'
-        - '-cx'
-        - >-
-          export TERM=dumb
-          && git fetch {{ event.head.repo.url }} {{ event.head.repo.branch }}
-          && git config advice.detachedHead false
-          && git checkout {{event.head.sha}}
-          && yes | sdkmanager --licenses
-          && ./gradlew -PisPullRequest clean assembleSystem assembleAndroidTest lint checkstyle ktlint pmd detekt test
-          && ./gradlew -Pcoverage jacocoSystemDebugTestReport
-          && ./tools/taskcluster/upload-coverage-report.sh
-          && ./tools/taskcluster/download-firebase-sdk.sh
-          && ./tools/taskcluster/google-firebase-testlab-login.sh
-          && ./tools/taskcluster/execute-firebase-test.sh system/debug app-system-debug model=sailfish,version=25,orientation=landscape
-      artifacts:
-        'public/reports':
-          type: 'directory'
-          path: '/opt/firefox-tv/app/build/reports'
-          expires: "{{ '1 week' | $fromNow }}"
-
-      features:
-          taskclusterProxy: true
-
-    metadata:
-      name: Firefox for Amazon's Fire TV - Build - Pull Request
-      description: Building Firefox for Amazon's Fire TV (via Gradle) - triggered by a pull request.
-      owner: '{{ event.head.user.email }}'
-      source: '{{ event.head.repo.url }}'
-###############################################################################
-# Task: Master builds
-#
-# Triggered whenever something is pushed/merged to the master branch.
-#
-# - Build the app (all flavors)
-# - Run unit tests
-# - Run code quality tools (findbugs, lint, checkstyle etc.)
-# - Upload app and test build APKs to Bitbar Cloud and run UI tests
-#
-# - HACK: Used {{ event.head.repo.url }} instead of origin because origin routes to
-#   the Focus Android repo
-###############################################################################
-  - provisionerId: '{{ taskcluster.docker.provisionerId }}'
-    workerType: '{{ taskcluster.docker.workerType }}'
-    extra:
-      github:
-        events:
-          - push
-        branches:
-          - master
-    scopes:
-    - "secrets:get:project/mobile/firefox-tv/tokens"
-    payload:
-      maxRunTime: 3600
-      deadline: "{{ '2 hours' | $fromNow }}"
-      image: 'mozillamobile/firefox-tv:2.3'
-      command:
-        - /bin/bash
-        - '--login'
-        - '-cx'
-        - >-
-          export TERM=dumb
-          && git fetch {{ event.head.repo.url }}
-          && git config advice.detachedHead false
-          && git checkout {{event.head.sha}}
-          && yes | sdkmanager --licenses
-          && ./gradlew -PisPullRequest clean assembleSystem assembleAndroidTest lint checkstyle ktlint pmd detekt test
-          && python ./tools/taskcluster/get-bitbar-token.py
-          && python ./tools/taskcluster/execute-bitbar-test.py system/debug app-system-debug
-      artifacts:
-        'public':
-          type: 'directory'
-          path: '/opt/firefox-tv/app/build/reports'
-          expires: "{{ '1 week' | $fromNow }}"
-
-      features:
-          taskclusterProxy: true
-
-    metadata:
-      name: Firefox for Amazon's Fire TV - Build - Master
-      description: Building Firefox for Amazon's Fire TV (via Gradle) - Master
-      owner: '{{ event.head.user.email }}'
-      source: '{{ event.head.repo.url }}'
-
-###############################################################################
-# Task: Release builds
-#
-# Triggered when a new tag or release is published (in any branch)
-#
-# - Build (unsigned) release versions of the app with release translations and
-#   secrets.
-###############################################################################
-  - provisionerId: '{{ taskcluster.docker.provisionerId }}'
-    workerType: '{{ taskcluster.docker.workerType }}'
-    extra:
-      github:
-        events:
-          - release
-    scopes:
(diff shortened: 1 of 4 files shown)


mcomella:
FYI: we have a release on Monday so I don't want to merge this before then but I'll try to get a review done sooner.

tools/taskcluster/decision_task.py
```diff
@@ -0,0 +1,66 @@
+import argparse
```
mcomella: Are these files similar to the Fenix ones? If so, I wonder if we should be sharing the files with some kind of dependency manager instead of duplicating them across repos (though it does add an additional point of failure so maybe it's not worth it).

.taskcluster.yml
```diff
@@ -1,160 +1,124 @@
...
+      else: ${event.repository.html_url}
+    head_rev:
+      $if: 'tasks_for == "github-pull-request"'
+      then: ${event.pull_request.head.sha}
+      else:
+        $if: 'tasks_for == "github-push"'
+        then: ${event.after}
+        else: ${event.release.tag_name}
+    trust_level:
+      $if: 'tasks_for in ["github-push", "github-release"] && event.repository.html_url == "https://github.com/mozilla-mobile/fenix"'
```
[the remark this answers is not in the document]
mitchhentges: Oof, good eye

tools/taskcluster/decision_task.py
```diff
@@ -0,0 +1,70 @@
...
+    return (taskcluster.slugId(), builder.craft_release_task(tag)),
+
+
+def main():
+    parser = argparse.ArgumentParser()
+    subparsers = parser.add_subparsers(dest='command')
+
+    pr_parser = subparsers.add_parser('pull-request')
+    pr_parser.add_argument('author')
+    pr_parser.add_argument('branch')
```
JohanLorenzo: What's the driver for changing the CLI API? As far I understand, Chain of Trust expects the env var `MOBILE_HEAD_REF` to rebuild the context. This will be an issue whenever we turn push-apk on.

Not a blocker.

tools/taskcluster/decision_task.py
```diff
@@ -0,0 +1,70 @@
...
+        ordered_tasks = master(builder)
+    elif command == 'release':
+        builder = TaskBuilder('firefox-tv@mozilla.com', repo_url, commit, task_group_id)
+        ordered_tasks = release(builder, result.tag)
+    else:
+        raise ValueError('A command ("pull-request", "master" or "release") must be provided to '
+                         'the decision task')
+
+    full_task_graph = schedule_task_graph(ordered_tasks)
+    with open('task-graph.json', 'w') as f:
```
JohanLorenzo: Nit: Why don't we use 
```py
    populate_chain_of_trust_task_graph(full_task_graph)
    populate_chain_of_trust_required_but_unused_files()
```

anymore? Their names provide more context about what's going on. Not a blocker

tools/taskcluster/tasks.py
```diff
@@ -0,0 +1,143 @@
...
+                'name': name,
+                'description': '',
+                'owner': self.owner,
+                'source': f'{self.repo_url}/raw/{self.commit}/.taskcluster.yml',
+            }
+        }
+
+
+def schedule_task_graph(ordered_tasks):
+    queue = taskcluster.Queue({'rootUrl': os.environ['TASKCLUSTER_PROXY_URL']})
```
JohanLorenzo: Nit: default to `https://taskcluster.net` for smoother local run.

tools/taskcluster/tasks.py
```diff
@@ -0,0 +1,143 @@
...
+import os
+
+import taskcluster
+
+
+def artifact(artifact_type, absolute_path):
+    return {
+        'type': artifact_type,
+        'path': absolute_path,
+        'expires': taskcluster.stringDate(taskcluster.fromNow('1 week'))
```
JohanLorenzo: Nit: I guess we want to extend the duration to 1 year once the code reaches master. 1 week seems to short.

tools/taskcluster/tasks.py
```diff
@@ -0,0 +1,143 @@
...
+            f'Firefox for Fire TV - Release build {tag}',
+            script,
+            ['secrets:get:project/mobile/firefox-tv/tokens'],
+            {
+                'public': artifact('directory', '/opt/firefox-tv/app/build/outputs/apk')
+            }
+        )
+
+    def _craft_shell_task(self, name, script, scopes, artifacts):
+        single_command = ' && '.join([line.strip() for line in script.split('\n') if line.strip()])
```
JohanLorenzo: 👍

tools/taskcluster/tasks.py
```diff
@@ -0,0 +1,143 @@
...
+            '--login',
+            '-cx',
+            f'export TERM=dumb && {single_command}'
+        ]
+
+        return {
+            'taskGroupId': self.task_group_id,
+            'provisionerId': 'aws-provisioner-v1',
+            'schedulerId': 'taskcluster-github',
+            'workerType': 'github-worker',
```
JohanLorenzo: Okay to use github-worker for the sake of this PR. Let's move it to `mobile-X-b-fftv` as soon as possible, since we're shipping product off these workers.

tools/taskcluster/decision_task.py
```diff
@@ -0,0 +1,70 @@
...
+import argparse
+import json
+import os
+
+import taskcluster
+
+from tasks import schedule_task_graph, TaskBuilder
+
+
+def pr(builder, branch):
```
JohanLorenzo: Fun fact, I'm moving `builder` to a parameter, for the taskgraph conversion too!

tools/taskcluster/decision_task.py
mitchhentges: Ooh, good point about Chain of Trust. My rationale for this change is that it becomes more obvious which values are required, and in which situations - for example, I don't think the `MOBILE_HEAD_REF` is required (by the decision task) except for PRs.

I'm using dep signing and push workers right now without Chain of Trust issues, but that's probably because they're dep. If I run into issues when I use production scriptworkers, I'll add the environment variables accordingly.

tools/taskcluster/decision_task.py
mitchhentges: Right! I kept a comment for the "unused files", but I didn't leave a comment for `task-graph.json`, good eye.
I didn't write the functions because I wasn't sure that they added enough non-documentation-y value :smile:

tools/taskcluster/tasks.py
mitchhentges: :open_mouth: then it'll ... work locally? Does it just use the authentication environment variables, then it's happy? Nice!

tools/taskcluster/tasks.py
mitchhentges: Can do! I just copied this from the existing `.taskcluster.yml`, but it makes sense to extend it

tools/taskcluster/decision_task.py
mitchhentges: Yeah, it can arguably be global, but that's only because `TASK_ID`, `MOBILE_HEAD_REPOSITORY` and `MOBILE_HEAD_REV` are global environment variables, which is :unamused:. It's interesting that `taskgraph` requires it to be a parameter, though :thinking:

tools/taskcluster/tasks.py
mitchhentges: Hmm, it's still not happy with credentials, but now it's more obvious why it's failing. :+1: to this default

.taskcluster.yml
```diff
@@ -1,160 +1,127 @@
...
+      $if: 'tasks_for == "github-pull-request"'
+      then: ${event.pull_request.head.sha}
+      else:
+        $if: 'tasks_for == "github-push"'
+        then: ${event.after}
+        else: ${event.release.tag_name}
+    trust_level:
+      $if: 'tasks_for in ["github-push", "github-release"] && event.repository.html_url == "https://github.com/mozilla-mobile/firefox-tv"'
+      # Trust Level number corresponds to the "trust level" constants from the main Firefox repository.
+      # 3 represents trusted work (it's been reviewed and landed, or it's work by a trusted dev in-repo) while
```
mcomella: Do you recommend core devs work in repo then? What does this trust level affect?

.taskcluster.yml
```diff
@@ -1,160 +1,124 @@
...
+    repository:
+      $if: 'tasks_for == "github-pull-request"'
+      then: ${event.pull_request.head.repo.html_url}
+      else: ${event.repository.html_url}
+    head_rev:
+      $if: 'tasks_for == "github-pull-request"'
+      then: ${event.pull_request.head.sha}
+      else:
+        $if: 'tasks_for == "github-push"'
+        then: ${event.after}
```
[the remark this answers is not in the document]
mcomella: wfm

.taskcluster.yml
```diff
@@ -1,160 +1,127 @@
...
+                    git fetch ${repository} ${event.pull_request.head.ref}
+                    && git config advice.detachedHead false
+                    && git checkout FETCH_HEAD
+                    && apt update
+                    && apt install -y python3-pip
+                    && pip3 install taskcluster
+                    && python3 tools/taskcluster/decision_task.py pull-request ${event.sender.login}@users.noreply.github.com ${event.pull_request.head.ref}
+              metadata:
+                owner: ${event.sender.login}@users.noreply.github.com
+      - $if: 'tasks_for == "github-push" && event["ref"] == "refs/heads/master"'
```
mcomella: So we don't do CI for our pushes to our release branches then? `releases/*`? That might be fine – we run CI on all PRs we uplift, run CI when building the release build, and generally don't check merges to `releases/*` anyway – but I just wanted to be explicit.

tools/taskcluster/tasks.py
```diff
@@ -0,0 +1,143 @@
...
+            f'Firefox for Fire TV - Release build {tag}',
+            script,
+            ['secrets:get:project/mobile/firefox-tv/tokens'],
+            {
+                'public': artifact('directory', '/opt/firefox-tv/app/build/outputs/apk')
+            }
+        )
+
+    def _craft_shell_task(self, name, script, scopes, artifacts):
+        single_command = ' && '.join([line.strip() for line in script.split('\n') if line.strip()])
```
mcomella: As I mentioned before, this isn't the same as reading a script: if that's intended, maybe we should add a comment?

mcomella: Going to merge and if you want to change this, please do so in a follow-up.

.taskcluster.yml
mitchhentges: Ah, good question! Core devs should still follow the same style (do work in separate branch, PR it in).

This trust level affects signing and pushing apks - we shouldn't use our signing key or amazon credentials unless we can "trust" the commit (and we know it can be trusted if it's part of this repo).

.taskcluster.yml
mitchhentges: That's correct :100:

tools/taskcluster/tasks.py
mitchhentges: Oh, I misunderstood. Fortunately, I'm adjusting this in follow-up work to actually interpret `script` as a real `bash` script

PR #2802, opened 2019-09-05
Closes #2776: Add wake_lock permission to manifest

### LINKED ISSUE #2776: Add WAKE_LOCK permission
ADM requires the `WAKE_LOCK` permission `<uses-permission android:name="android.permission.WAKE_LOCK"/>`, but this permission is already brought in from a dependency so it wasn't added explicitly the first time. It would be more clear and safe (in case it gets removed from dependency) if we add it ourselves.
### Impact
If the dependency removes the permission, ADM will not work.
### Acceptance criteria
Add `<uses-permission android:name="android.permission.WAKE_LOCK"/>` in the manifest.

dnarcese:
This permission is needed for ADM. It is currently brought in by a dependency, but it is safer if we also add it ourselves.

## Checklist

### DIFF
--- a/app/src/debug/AndroidManifest.xml
+++ b/app/src/debug/AndroidManifest.xml
@@ -36,6 +36,8 @@
         android:protectionLevel="signature"/>
     <uses-permission android:name="${applicationId}.permission.RECEIVE_ADM_MESSAGE" />
     <uses-permission android:name="com.amazon.device.messaging.permission.RECEIVE" />
+    <!-- This permission is also brought in by androidx.work:work-runtime dependency. -->
+    <uses-permission android:name="android.permission.WAKE_LOCK"/>
     <!-- *** ADM permissions end *** -->
 
     <application>


app/src/debug/AndroidManifest.xml
```diff
@@ -36,6 +36,7 @@
         android:protectionLevel="signature"/>
     <uses-permission android:name="${applicationId}.permission.RECEIVE_ADM_MESSAGE" />
     <uses-permission android:name="com.amazon.device.messaging.permission.RECEIVE" />
+    <uses-permission android:name="android.permission.WAKE_LOCK"/>
```
mcomella: nit: can we add a comment mentioning this is included by the dependency? You can run a release build and check `app/build/outputs/logs/manifest-merger-system-release-report.txt` to see which dependency added it and include the name of it in the comment.

I think the comment will be helpful, just in case anyone tries to remove the permission and expects for it to go away for app users.

PR #2822, opened 2019-09-10
Update a-c to 12.0.0

dnarcese:
Changelog: https://mozac.org/changelog/#1200
## Checklist


app/src/main/java/org/mozilla/tv/firefox/fxa/FxaLoginUseCase.kt
```diff
@@ -109,6 +111,6 @@ class FxaLoginUseCase(
 
     private fun FxaAccountManager.finishAuthenticationAsync(keys: LoginSuccessKeys) {
         @Suppress("DeferredResultUnused") // We don't care to wait until completion.
-        finishAuthenticationAsync(code = keys.code, state = keys.state)
+        finishAuthenticationAsync(FxaAuthData(AuthType.Signin, keys.code, keys.state))
```
mcomella: This could conceivably also be, `AuthType.Recovered` (if there was an auth problem): how did we select `SignIn`?

Also, why do we need to pass this in? Shouldn't the a-c library know the current state and which one we're transitioning to? This API doesn't seem right to me...

PR #2840, opened 2019-09-16
Automation: enhanced release task group

mitchhentges:
:rotating_light: I can't test the signing and pushing-to-amazon without landing the code here first since it uses security-sensitive release tokens. So, there _probably will be breakage_ when this lands, which will require follow-up PRs to be landed :rotating_light: 

-----

This is the "cool" PR :smile::
* APKs to attach to Github Releases are now automatically signed
* New releases are automatically created in Amazon as an "upcoming release"
* Once automation is done, an email goes to `firefox-tv` with links to:
    * The github release
    * The signed APK for the github release
    * The Amazon Developer Console page for FFTV

## Checklist

### DIFF
--- a/tools/taskcluster/decision_task.py
+++ b/tools/taskcluster/decision_task.py
@@ -21,9 +21,11 @@ def master(builder):
 
 def release(builder, tag, is_staging):
     build_task_id = taskcluster.slugId()
+    sign_task_id = taskcluster.slugId()
     return (
         (build_task_id, builder.craft_release_build_task(tag)),
-        (taskcluster.slugId(), builder.craft_sign_for_github_task(build_task_id, is_staging)),
+        (sign_task_id, builder.craft_sign_for_github_task(build_task_id, is_staging)),
+        (taskcluster.slugId(), builder.craft_email_task(sign_task_id, tag)),
         (taskcluster.slugId(), builder.craft_amazon_task(build_task_id, is_staging)),
     )
 
--- a/tools/taskcluster/decision_task.py
+++ b/tools/taskcluster/decision_task.py
@@ -19,8 +19,13 @@ def master(builder):
     return (taskcluster.slugId(), builder.craft_master_task()),
 
 
-def release(builder, tag):
-    return (taskcluster.slugId(), builder.craft_release_build_task(tag)),
+def release(builder, tag, is_staging):
+    build_task_id = taskcluster.slugId()
+    return (
+        (build_task_id, builder.craft_release_build_task(tag)),
+        (taskcluster.slugId(), builder.craft_sign_for_github_task(build_task_id, is_staging)),
+        (taskcluster.slugId(), builder.craft_amazon_task(build_task_id, is_staging)),
+    )
 
 
 def main():
@@ -49,8 +54,9 @@ def main():
         builder = TaskBuilder(result.author, repo_url, commit, task_group_id)
         ordered_tasks = master(builder)
     elif command == 'release':
+        is_staging = repo_url != 'https://github.com/mozilla-mobile/firefox-tv'
         builder = TaskBuilder('firefox-tv@mozilla.com', repo_url, commit, task_group_id)
-        ordered_tasks = release(builder, result.tag)
+        ordered_tasks = release(builder, result.tag, is_staging)
     else:
         raise ValueError('A command ("pull-request", "master" or "release") must be provided to '
                          'the decision task')
--- a/tools/taskcluster/decision_task.py
+++ b/tools/taskcluster/decision_task.py
@@ -20,7 +20,7 @@ def master(builder):
 
 
 def release(builder, tag):
-    return (taskcluster.slugId(), builder.craft_release_task(tag)),
+    return (taskcluster.slugId(), builder.craft_release_build_task(tag)),
 
 
 def main():
--- a/tools/taskcluster/decision_task.py
+++ b/tools/taskcluster/decision_task.py
@@ -22,11 +22,12 @@ def master(builder):
 def release(builder, tag, is_staging):
     build_task_id = taskcluster.slugId()
     sign_task_id = taskcluster.slugId()
+    push_task_id = taskcluster.slugId()
     return (
         (build_task_id, builder.craft_release_build_task(tag)),
         (sign_task_id, builder.craft_sign_for_github_task(build_task_id, is_staging)),
-        (taskcluster.slugId(), builder.craft_email_task(sign_task_id, tag)),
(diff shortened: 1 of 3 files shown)


mitchhentges:
@JohanLorenzo would you be able to review this?

tools/taskcluster/tasks.py
```diff
@@ -84,41 +87,123 @@ def craft_release_task(self, tag):
...
-        single_command = ' && '.join([line.strip() for line in script.split('\n') if line.strip()])
+    def _craft_shell_task(self, name, script, scopes, artifacts, *, dependencies=(), chain_of_trust=False):
+        trimmed_script = '\n'.join([line.strip() for line in script.split('\n') if line.strip()])
         bash_command = [
             '/bin/bash',
             '--login',
-            '-cx',
-            f'export TERM=dumb && {single_command}'
+            '-c',
+            'cat <<"SCRIPT" script.sh && bash -e script.sh\n'
```
JohanLorenzo: Bug: `cat: script.sh: No such file or directory` => `cat` tries to concatenate both the here-document and `script.sh`. I don't believe this is the right approach, for now.

mitchhentges: This should be `cat <<"SCRIPT > script.sh && bash -e script.sh`

tools/taskcluster/decision_task.py
```diff
@@ -19,8 +19,13 @@ def master(builder):
     return (taskcluster.slugId(), builder.craft_master_task()),
 
 
-def release(builder, tag):
-    return (taskcluster.slugId(), builder.craft_release_build_task(tag)),
+def release(builder, tag, is_staging):
+    build_task_id = taskcluster.slugId()
+    return (
+        (build_task_id, builder.craft_release_build_task(tag)),
+        (taskcluster.slugId(), builder.craft_sign_for_github_task(build_task_id, is_staging)),
```
mcomella: Does `taskcluster.slugId()` generate new ids? Or should be you using `build_task_id` here and below?

tools/taskcluster/tasks.py
```diff
@@ -84,10 +84,11 @@ def craft_release_build_task(self, tag):
             ['secrets:get:project/mobile/firefox-tv/tokens'],
             {
                 'public': artifact('directory', '/opt/firefox-tv/app/build/outputs/apk')
-            }
+            },
+            chain_of_trust=True,  # Needed for sign and push task verification
         )
 
-    def _craft_shell_task(self, name, script, scopes, artifacts):
+    def _craft_shell_task(self, name, script, scopes, artifacts, *, chain_of_trust=False):
```
mcomella: What does a `*` parameter do? Is that a typo? I've only seen it used to unpack containers/keyword args like (`def function(*container)` (or whatever it's called).

tools/taskcluster/tasks.py
```diff
@@ -111,14 +108,65 @@ def _craft_shell_task(self, name, script, scopes, artifacts):
...
                     'taskclusterProxy': True,
+                    'chainOfTrust': chain_of_trust,
                 }
+            }
+        })
+
+    def craft_sign_for_github_task(self, build_task_id, is_staging):
+        return self._craft_base_task('Sign for Github', {
+            'provisionerId': 'scriptworker-prov-v1',
+            'workerType': 'mobile-signing-dep-v1' if is_staging else 'mobile-signing-v1',
```
mcomella: nit: `mobile-signing-dep-v1` - did you mean `mobile-signing-dev-v1`? Or maybe `mobile-signing-debug-v1`? I'm not sure what dep is.

Also, since both of these strings need to get updated when the version changes, I wonder if string format args or storing the version in a variable would be more robust than an if expression (though I guess they're maybe less readable in python)

tools/taskcluster/tasks.py
```diff
@@ -91,39 +91,45 @@ def craft_release_build_task(self, tag):
...
             script,
             [f'notify:email:{NOTIFY_EMAIL_ADDRESS}'],
             {},
-            dependencies=[sign_task_id],
+            dependencies=[sign_task_id, push_task_id],
         )
 
     def _craft_shell_task(self, name, script, scopes, artifacts, *, dependencies=(), chain_of_trust=False):
-        trimmed_script = ' '.join([line.strip() for line in script.split('\n') if line.strip()])
+        trimmed_script = '\n'.join([line.strip() for line in script.split('\n') if line.strip()])
```
mcomella: It looks like we remove all the \n from the script string and then rejoin it with \n: should we just use the script string directly?

tools/taskcluster/decision_task.py
mitchhentges: `taskcluster.slugId()` generates new ids: the first item of each of these tuples is the task id associated with each task.

You can imagine it like:
```
return (
    ('A', build_task()),
    ('B', sign_task('A')),
    ('C', push_task('A')),
    ('D', email_task('B', 'C')),
)
```

tools/taskcluster/tasks.py
mitchhentges: It's a neat python feature that forces parameters after the `*` to be named.
So, I can't do: `_craft_shell_task('name', script, [], artifacts, True)`
I have to do: `_craft_shell_task('name', script, [], artifacts, chain_of_trust=True)`

tools/taskcluster/tasks.py
mitchhentges: Heh, I've been stockholm syndrome'd into forgetting how opaque some of these terms and values are.

"dep" is an old releng term referring to a kind of staging/testing work. It has a relation to the old "depend vs clobber" days, which I'm unfamiliar with.
That worker (`mobile-signing-v1` or `mobile-signing-dep-v1`) is very long lived. When updates happen to the signing software, we update the software on that worker (the "version" in the name does not change). The `v1` in the name probably exists in case we make a radical new signing worker that is completely incompatible with the old worker, and we need a way to differentiate.

Since this happens infrequently, hard-coding the `v1` is effective.

tools/taskcluster/tasks.py
mitchhentges: We could use the script string directly, but by `strip()`-ing the extra spacing from the python heredoc values, we make the build tasks simpler (the "command" payload will have less irrelevant whitespace)

mcomella: Maybe comment to clarify that removing whitespace is the intent? I'm surprised there isn't a built-in python function to do this already. :P

mitchhentges: Good call!

PR #2850, opened 2019-09-17
[merge after release] Closes #2844: Use recommended tracking protection policy

### LINKED ISSUE #2844: Use TrackingProtectionPolicy.recommended()
Previously, the `recommended` tracking protection policy was not blocking ads correctly. It has since been fixed, so we should use it. https://github.com/mozilla-mobile/android-components/issues/4191
### Impact
Our tracking lists may get outdated if we are hardcoding the tracking categories instead of using a recommended list.
### Acceptance criteria
[The policy ](https://github.com/mozilla-mobile/firefox-tv/blob/71e4cfcdd5b40d4b3c433a7d48a0cabc457f4713/app/src/main/java/org/mozilla/tv/firefox/utils/Settings.kt#L89) should be set to `recommended`

dnarcese:
The new TP lists add some new categories, so using the recommended policy will make sure we aren't missing categories.
## Checklist

### DIFF
--- a/app/src/main/java/org/mozilla/tv/firefox/utils/Settings.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/utils/Settings.kt
@@ -91,18 +91,7 @@ class Settings private constructor(context: Context) {
     val trackingProtectionPolicy: TrackingProtectionPolicy
         get() {
             return if (isBlockingEnabled) {
-                TrackingProtectionPolicy.select(
-                    // We want to use TrackingCategory.RECOMMENDED but it doesn't block ads properly:
-                    // a-c#4191. We write out the values in RECOMMENDED manually below and it works
-                    // properly.
-                    trackingCategories = arrayOf(
-                        TrackingCategory.AD,
-                        TrackingCategory.ANALYTICS,
-                        TrackingCategory.SOCIAL,
-                        TrackingCategory.TEST
-                    )
-                    // We use the default cookiePolicy.
-                )
+                TrackingProtectionPolicy.recommended()
             } else {
                 TrackingProtectionPolicy.select(
                     // These defaults are from TrackingProtectionPolicy.none().


app/src/main/java/org/mozilla/tv/firefox/utils/Settings.kt
```diff
@@ -91,18 +91,7 @@ class Settings private constructor(context: Context) {
...
-                        TrackingCategory.AD,
-                        TrackingCategory.ANALYTICS,
-                        TrackingCategory.SOCIAL,
-                        TrackingCategory.TEST
-                    )
-                    // We use the default cookiePolicy.
-                )
+                TrackingProtectionPolicy.recommended()
             } else {
                 TrackingProtectionPolicy.select(
```
mcomella: This can also be switched to `TrackingProtectionPolicy.none()` now that SafeBrowsing is not a part of this.

app/src/main/java/org/mozilla/tv/firefox/utils/Settings.kt
```diff
@@ -91,18 +91,7 @@ class Settings private constructor(context: Context) {
...
-                    // properly.
-                    trackingCategories = arrayOf(
-                        TrackingCategory.AD,
-                        TrackingCategory.ANALYTICS,
-                        TrackingCategory.SOCIAL,
-                        TrackingCategory.TEST
-                    )
-                    // We use the default cookiePolicy.
-                )
+                TrackingProtectionPolicy.recommended()
```
mcomella: This adds the following tracking protection categories:
- Mozilla_social
- Cryptomining
- Fingerprinting

Adding tracking protection categories adds some risk that sites users are trying to use will break: can you follow up with Ashley that this is an acceptable risk for the upcoming release? We could postpone it to the following release though we might want to wait even another release if we're intending to take on the updated Disconnect lists from a-c next release, to make it clearer what might have caused the errors.

dnarcese: @athomasmoz

dnarcese: Let's wait until next release!

PR #2869, opened 2019-09-18
Closes #2861: Show FxA onboarding on every login

### LINKED ISSUE #2861: Show receive tab onboarding screen every time a user logs in
## Steps to reproduce

1. Open FFTV
2. Sign in to your account

### Expected behavior

Onboarding screen is shown

### Actual behavior

Onboarding screen is only shown if it's the first time you have signed in on that device

### UX note

It’s important to show the onboarding every time, because there are only a few scenarios where users need to sign out and back in:

1. To fix problem. Something doesn’t work, so better try again.
2. Because another person needs to send tabs.

We hope not to have too many problems in scenario 1. This leaves us with scenario 2, where a new user may not know how to send tabs and need to see the onboarding.

### Device information
* Fire TV device: ?
* Latest affected Firefox version: 4.5-LAT
* Turbo Mode: ?

dnarcese:
The main problem after making the onboarding screen show every time was that it also starting showing up on startup when already logged in. This happened because we set the default account state to `NotAuthenticated`, so on startup, it would emit `NotAuthenticated` followed by `AuthenticatedNoProfile`. To fix this, I added a default account state.
## Checklist

### DIFF
--- a/app/src/main/java/org/mozilla/tv/firefox/fxa/FxaRepo.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/fxa/FxaRepo.kt
@@ -76,13 +76,14 @@ class FxaRepo(
          */
         object AuthenticatedNoProfile : AccountState() // Before the profile is fetched async
         object NeedsReauthentication : AccountState()
-        object NotAuthenticated : AccountState() // Initial state
+        object NotAuthenticated : AccountState()
+        object Initial : AccountState()
     }
 
     @VisibleForTesting(otherwise = NONE)
     val accountObserver = FirefoxAccountObserver()
 
-    private val _accountState: BehaviorSubject<AccountState> = BehaviorSubject.createDefault(NotAuthenticated)
+    private val _accountState: BehaviorSubject<AccountState> = BehaviorSubject.createDefault(AccountState.Initial)
     val accountState: Observable<AccountState> = _accountState.hide()
 
     val receivedTabs: Observable<Consumable<FxaReceivedTab>> = admIntegration.receivedTabsRaw
--- a/app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/NavigationOverlayFragment.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/NavigationOverlayFragment.kt
@@ -288,14 +288,27 @@ class NavigationOverlayFragment : Fragment() {
                         fxaButton.contentDescription =
                             resources.getString(R.string.fxa_navigation_item_new,
                                 resources.getString(R.string.app_name))
+                        resetFxaOnboardingShown()
                     }
                     AccountState.NeedsReauthentication -> {
                         fxaButton.setImageResource(R.drawable.ic_fxa_needs_reauthentication)
                         fxaButton.contentDescription =
                             resources.getString(R.string.fxa_navigation_item_sign_in_again)
+                        resetFxaOnboardingShown()
                     }
                 }
             }
+    }
+
+    /**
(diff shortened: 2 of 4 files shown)


app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/NavigationOverlayFragment.kt
```diff
@@ -288,16 +288,29 @@ class NavigationOverlayFragment : Fragment() {
...
                         fxaButton.contentDescription =
                             resources.getString(R.string.fxa_navigation_item_sign_in_again)
+                        resetFxaOnboardingShown()
                     }
                 }
             }
     }
 
+    /**
+     * We need to reset the state of FxA onboarding screen to ensure that we do not show onboarding on every startup (#2861).
```
mcomella: nit: the first sentence of a kdoc should be a brief summary of what the thing does. The reason for this is that this one-line summary is used by the kdoc generation tools (or at least it was for javadoc) to provide a summary and clicking on it will provide the extended text.

Not that we generate kdoc but it's good to follow convention and, if we did ever generate kdoc, it could be useful.

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/NavigationOverlayFragment.kt
```diff
@@ -288,16 +288,29 @@ class NavigationOverlayFragment : Fragment() {
                         fxaButton.contentDescription =
                             resources.getString(R.string.fxa_navigation_item_new,
                                 resources.getString(R.string.app_name))
+                        resetFxaOnboardingShown()
```
mcomella: fwiw, I think a slightly simpler approach would be to launch the onboarding from [the login success callback](https://searchfox.org/mozilla-mobile/rev/84667ea7969a4c48afba2f9a17e0b2940a33db01/firefox-tv/app/src/main/java/org/mozilla/tv/firefox/fxa/FxaLoginUseCase.kt#116) – so we don't need to keep track of a boolean – but perhaps it's not worth changing now.

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/NavigationOverlayFragment.kt
```diff
@@ -288,16 +288,29 @@ class NavigationOverlayFragment : Fragment() {
...
                             resources.getString(R.string.fxa_navigation_item_sign_in_again)
+                        resetFxaOnboardingShown()
                     }
                 }
             }
     }
 
+    /**
+     * We need to reset the state of FxA onboarding screen to ensure that we do not show onboarding on every startup (#2861).
+     * It will get reset whenever there are FxA authentication problems or the user logs out.
```
mcomella: This line is redundant to the code (and separated from where it actually happens so this comment is unlkely to stay up to date) so I'd remove it.

app/src/main/java/org/mozilla/tv/firefox/fxa/FxaRepo.kt
```diff
@@ -76,13 +76,14 @@ class FxaRepo(
          */
         object AuthenticatedNoProfile : AccountState() // Before the profile is fetched async
         object NeedsReauthentication : AccountState()
-        object NotAuthenticated : AccountState() // Initial state
+        object NotAuthenticated : AccountState()
+        object Default : AccountState() // Initial state
```
mcomella: nit: this comment is redundant to the code so I'd remove it.

Also, this state could be named, `InitialState` or `Initial`.

PR #2892, opened 2019-09-23
Update automation to use taskgraph

mitchhentges:
## Questions for Johan:
* How should I handle getting the `GIT_TAG` and `HEAD_SHA`? I bet grabbing it from the environment isn't the right move
* Should I set up docker in-tree rather that continuing to use `mozillamobile/firefox-tv`?

## Checklist

- [x] [`build-puppet` PR is landed](https://github.com/mozilla-releng/build-puppet/pull/609)


.taskcluster.yml
```diff
@@ -1,127 +1,264 @@
...
+                  then: {$eval: as_slugid("decision_task")}
+                  else:
+                      $if: 'tasks_for == "cron"'
+                      then: '${ownTaskId}'
+          in:
+              $if: 'tasks_for in ["github-pull-request", "github-release", "action", "cron"] || tasks_for == "github-push" && head_branch[:10] != "refs/tags/"'
+              then:
+                  $let:
+                      level:
+                          $if: 'tasks_for in ["github-push", "github-release", "action", "cron"] && repoUrl == "https://github.com/mozilla-mobile/firefox-tv"'
```
JohanLorenzo: 👍

taskcluster/ci/config.yml
```diff
@@ -0,0 +1,47 @@
+---
```
JohanLorenzo: Note for future readers: I see no treeherder config has ever been done on this project. Since it's going into maintenance mode, I don't think it's worth reporting there.

taskcluster/ci/build/kind.yml
```diff
@@ -0,0 +1,64 @@
+---
+loader: taskgraph.loader.transform:loader
+
+transforms:
+  - firefox_tv_taskgraph.build:transforms
+  - taskgraph.transforms.task:transforms
+
+jobs:
+  pr:
+    label: Firefox for Amazon's Fire TV - Build - Pull Request
```
JohanLorenzo: Nit: we can move this sentence as the description and let taskgraph generate the label for us. Here it would call it: `build-pr`. Same thing for the other tasks.

Can be done in a follow up.

taskcluster/firefox_tv_taskgraph/build.py
```diff
@@ -0,0 +1,38 @@
...
+
+NOTIFY_EMAIL_ADDRESS = 'firefox-tv@mozilla.com'
+
+
+@transforms.add
+def build_task(config, tasks):
+    for task in tasks:
+        script = task["worker"]["script"].format(
+            repo_url=config.params["head_repository"],
+            commit=os.environ.get("MOBILE_HEAD_REV"),  # TODO use params?
```
JohanLorenzo: Yup! 

```suggestion
            commit=config.params["head_rev"]
```

taskcluster/firefox_tv_taskgraph/build.py
```diff
@@ -0,0 +1,38 @@
...
+        bash_command = [
+            "/bin/bash",
+            "--login",
+            "-c",
+            "cat <<'SCRIPT' > ../script.sh && bash -e ../script.sh\n"
+            "export TERM=dumb\n{}\nSCRIPT".format(script)
+        ]
+
+        task["worker"] = {
+            "max-run-time": 3600,
```
JohanLorenzo: Nit: `max-run-time` and `docker-image` can be moved to `build/kind.yml` since they're just plain data.

taskcluster/firefox_tv_taskgraph/build.py
```diff
@@ -0,0 +1,38 @@
...
+            "docker-image": "mozillamobile/firefox-tv:2.3",
```
JohanLorenzo: Followup: Build the docker image in-tree.

taskcluster/firefox_tv_taskgraph/build.py
```diff
@@ -0,0 +1,38 @@
...
+            "command": bash_command,
+            "artifacts": task["worker"].get("artifacts", []),
```
JohanLorenzo: Nit: this can be replace by: `task["worker"].setdefault("artifacts", [])`, anywhere in the transform.

taskcluster/firefox_tv_taskgraph/email.py
```diff
@@ -0,0 +1,39 @@
...
+
+@transforms.add
+def email_task(config, tasks):
+    tag = os.environ['GIT_TAG']
+    content = 'Automation for this release is ready. Please: \n' \
+              '* Download the APK and attach it to the [Github release](https://github.com/mozilla-mobile/firefox-tv/releases/tag/{})\n'.format(tag) + \
+              '* [Deploy the new release on Amazon](https://developer.amazon.com/apps-and-games/console/app/amzn1.devportal.mobileapp.7f334089688646ef8953d041021029c9/release/amzn1.devportal.apprelease.4ca3990c43f34101bf5729543343747a/general/detail)'
+
+    for task in tasks:
+        to_address = task["worker"]["to-address"]
```
JohanLorenzo: Nit: Can be replaced by: 

```suggestion
        to_address = task["worker"].pop("to-address")
```

taskcluster/firefox_tv_taskgraph/email.py
```diff
@@ -0,0 +1,39 @@
...
+
+transforms = TransformSequence()
+
+
+NOTIFY_EMAIL_ADDRESS = 'firefox-tv@mozilla.com'
+
+
+@transforms.add
+def email_task(config, tasks):
+    tag = os.environ['GIT_TAG']
```
JohanLorenzo: Bug: This line will break on any decision task that doesn't define GIT_TAG. I'd put:

```suggestion
    tag = os.environ.get('GIT_TAG', "UNDEFINED_TAG")
```

This way, if `target_tasks.py` is bugged and outputs this task in a PR, let's say, the email would be clear something is odd.

taskcluster/firefox_tv_taskgraph/email.py
```diff
@@ -0,0 +1,39 @@
...
+        del task["worker"]["to-address"]
+        task["scopes"] = ["queue:route:notify.email.{}.on-completed".format(to_address)]
+        task["routes"] = ["notify.email.{}.on-completed".format(to_address)]
+        task["extra"] = {
+            "notify": {
+                "email": {
+                    "content": content,
+                    "subject": "Release {} is ready for deployment".format(tag),
+                    "link": {
+                        "href": {"artifact-reference": "<sign/public/build/target.apk>"},
```
JohanLorenzo: 👍

taskcluster/firefox_tv_taskgraph/job.py
```diff
@@ -0,0 +1,69 @@
...
+    scopes = taskdesc.setdefault("scopes", [])
+    scopes.extend(["secrets:get:{}".format(secret(["name"]) for secret in secrets)])
+
+    run["cwd"] = "{checkout}"
+    run["using"] = "run-task"
+    configure_taskdesc_for_run(config, job, taskdesc, job["worker"]["implementation"])
+
+
+def _extract_command(run):
+    pre_gradle_commands = [["taskcluster/scripts/install-sdk.sh"]]
```
JohanLorenzo: Nit: I don't think we need to install the SDK if we're using the pre-built docker images.

taskcluster/ci/signing/kind.yml
```diff
@@ -0,0 +1,24 @@
...
+  - taskgraph.transforms.task:transforms
+
+jobs:
+  push:
+    label: Sign for Github
+    description: ''
+    worker-type: signing
+    run-on-tasks-for: [github-release]
+    dependencies:
+      build: 'Firefox for Fire TV - Release build'
```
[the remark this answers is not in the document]
mitchhentges: How should I handle this for the `email` case, where it has multiple dependencies?
I'm worried that if I use the `single_dep` loader for this kind, it will make dependency-configuration too inconsistent between kinds.

I'm indifferent though. If you've considered ^ already and you'd prefer using `single_dep`, that's :+1: and I'll implement it in this PR

taskcluster/firefox_tv_taskgraph/build.py
mitchhentges: I'll do this as part of this PR

taskcluster/ci/signing/kind.yml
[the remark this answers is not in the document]
tomprince: There are a variety of options for specifying depdencies.

Probably the two most relevant here are:
- `single_dep`: This is used when you want a downstream task for each task in the upstream kind, and wnat to pull data for the downstream task from the upstream task. Stuff downstream of builds is a common example of this; we'll have a variety of build types, and then for each build type, we'll have a series of downstream tasks for each, that depend on the build and have the same platform and stuff.
- [`release-deps`](https://searchfox.org/mozilla-central/source/taskcluster/taskgraph/transforms/release_deps.py): This adds all upstream tasks matching some attributes as dependencies of downstream tasks. This is used for stuff that needs to run after all the platform specific stuff, including things like notifications. (Note: this probably doesn't have the best name, and the m-c one has much more logic than we'd want for our current github projects).

.taskcluster.yml
```diff
@@ -1,125 +1,264 @@
...
+              branch: taskgraph
+              revision: 0c5a68749f9a7672a7e56604b69a7bd41b036614
+          trustDomain: mobile
+      in:
+          $let:
+              # Github events have this stuff in different places...
+              ownerEmail:
+                  $if: 'tasks_for == "github-push"'
+                  then: '${event.pusher.email}'
+                  # Assume Pull Request
```
mcomella: nit: this seems inaccurate because there are both PRs and releases below.

.taskcluster.yml
```diff
@@ -1,125 +1,264 @@
...
+                          then: '${repository.project}'
+              head_branch:
+                  $if: 'tasks_for == "github-pull-request"'
+                  then: ${event.pull_request.head.ref}
+                  else:
+                      $if: 'tasks_for == "github-push"'
+                      then: ${event.ref}
+                      else:
+                          $if: 'tasks_for == "github-release"'
+                          then: '${event.release.target_commitish}'
```
mcomella: `target_commitish`: is that really a part of the API?

.taskcluster.yml
```diff
@@ -1,125 +1,264 @@
...
+                              $if: 'tasks_for in ["cron", "action"]'
+                              then: '${push.revision}'
+              ownTaskId:
+                  $if: '"github" in tasks_for'
+                  then: {$eval: as_slugid("decision_task")}
+                  else:
+                      $if: 'tasks_for == "cron"'
+                      then: '${ownTaskId}'
+          in:
+              $if: 'tasks_for in ["github-pull-request", "github-release", "action", "cron"] || tasks_for == "github-push" && head_branch[:10] != "refs/tags/"'
```
mcomella: There's no else branch: what happens if the `if` check fails? Maybe we should comment about the intent here.

.taskcluster.yml
```diff
@@ -1,125 +1,264 @@
...
+              then:
+                  $let:
+                      level:
+                          $if: 'tasks_for in ["github-push", "github-release", "action", "cron"] && repoUrl == "https://github.com/mozilla-mobile/firefox-tv"'
```
mcomella: Perhaps nothing to change here but it kind of sucks how tied to GitHub all these configuration files are. :\ I wonder if we'd want to move our projects back to m-c at some point.

taskcluster/firefox_tv_taskgraph/__init__.py
```diff
@@ -0,0 +1,16 @@
+from __future__ import absolute_import, print_function, unicode_literals
```
mcomella: nit: not sure what was decided by licenses for the files in this commit

taskcluster/firefox_tv_taskgraph/email.py
```diff
@@ -0,0 +1,39 @@
...
```
mcomella: Are these brackets the correct syntax?

taskcluster/firefox_tv_taskgraph/job.py
```diff
@@ -0,0 +1,68 @@
...
+    scopes = taskdesc.setdefault("scopes", [])
+    scopes.extend(["secrets:get:{}".format(secret(["name"]) for secret in secrets)])
+
+    run["cwd"] = "{checkout}"
+    run["using"] = "run-task"
+    configure_taskdesc_for_run(config, job, taskdesc, job["worker"]["implementation"])
+
+
+def _extract_command(run):
+    pre_gradle_commands = [[
```
mcomella: nit: have you considered the trade-offs of integrating whatever commands you want to run "pre/post-gradle" as a Gradle task? tbh, I'm not sure if it's actually worth it but here's what I came up with:
- Pro: It consolidates our build logic in a single build system/set of scripts
- Pro: It (maybe) makes it easier for folks to run all the build commands
- Pro: if other releng tasks are created in Gradle, you can create task graphs (lol) and dependencies between tasks
- Con: releng might be less familiar with it
- Con: doesn't support other non-Android platforms as well

.taskcluster.yml
```diff
@@ -1,127 +1,264 @@
...
+                              $merge:
+                                  - MOBILE_BASE_REPOSITORY: '${baseRepoUrl}'
+                                    MOBILE_HEAD_REPOSITORY: '${repoUrl}'
+                                    MOBILE_HEAD_REF: '${head_branch}'
+                                    MOBILE_HEAD_REV: '${head_sha}'
+                                    MOBILE_REPOSITORY_TYPE: git
+                                    TASKGRAPH_BASE_REPOSITORY: https://hg.mozilla.org/ci/taskgraph
+                                    TASKGRAPH_HEAD_REPOSITORY: https://hg.mozilla.org/ci/${taskgraph.branch}
+                                    TASKGRAPH_HEAD_REV: ${taskgraph.revision}
+                                    TASKGRAPH_REPOSITORY_TYPE: hg
```
mcomella: We're not hg: is this correct?

tomprince: The taskgraph repository (which this is configuration for cloning) is still in hg. This also clones firefox-tv above (`MOBILE_*` prefixes), which has `git` as the repository type.

.taskcluster.yml
tomprince: This file is largely identical between all the taskgraph using github project, and the differences are mostly (only mostly, unfortunately) confined to the stanza at the beginning of this file. We have a similar somewhat standardized config for hg.mozilla.org products, that more or less shares the same initial stanza. So, migrate to hg.mozilla.org shouldn't be too painful in that regard.

---
A couple of additional details:
- We are just starting to use taskgraph outside of mozilla-central, so are still figuring out what abstraction we need for specifying `.taskcluster.yml` files, the variation beyond the initial stanza should go down over time.
- We've talked about various ways to avoid needing a mostly duplicated `.taskcluster.yml`, and being able to just have a simpler format for specifying things. We have some ideas on how to specify that, but require support from the taskcluster platform. Given that that team is dealing with several migration at the moment, it will be awhile before something better is implemented.

.taskcluster.yml
mitchhentges: Good eye

.taskcluster.yml
mitchhentges: [Affirmative](https://developer.github.com/v3/activity/events/types/#webhook-payload-example-33) :)

.taskcluster.yml
mitchhentges: Yeah, it sounds like the previous intent of TaskCluster was to act as the CI system (that just happened to talk to Github as a backend), but teams were asking for access to Github properties/features faster than the TC team planned, so this iteration of TaskCluster has more of a "pass-through" role

taskcluster/firefox_tv_taskgraph/__init__.py
mitchhentges: :/ right, I'll paste those in

taskcluster/firefox_tv_taskgraph/email.py
mitchhentges: This is a `taskgraph`-specific construct. If it sees an object with the key `"artifact-reference"`, it'll parse the value and link tasks accordingly:
```
<sign/public/build/target.apk>
 ^^^^ ^^^^^^^^^^^^^^^^^^^^^^^
 task |
      |
      |
      path to artifact on the task
```

taskcluster/firefox_tv_taskgraph/job.py
mitchhentges: That's a good question! I think that groovy/kotlin code is more maintainable and testable than a shell command. However, this taskgraph-ification is a move to consolidate the automation in our mobile projects, and we used the pre/post-gradle-command mechanism in the Fenix project.
I'm interested in improving this further as we evolve automation here

.taskcluster.yml
mitchhentges: :eyes:

mitchhentges:
This is ready to merge once [this `build-puppet` job lands](https://github.com/mozilla-releng/build-puppet/pull/609)

Issue #2805, opened 2019-09-05
[FxA] Return user to page they started on after sign in


mcomella:
### Why/User Benefit/User Problem
We have one WebView and we use the WebView to sign into FxA so whatever page the user was looking at gets replaced by the sign in process, even after successful sign in. It's in their back stack so in theory they can still press back to see the page but it's unintuitive to go back through the sign in flow to get there.

Ideally, the sign in flow wouldn't disrupt the user's browsing experience.

### Acceptance Criteria (how do I know when I’m done?)
- After successful sign in (which may require successful email confirmation), the website the user was on before sign in is restored

### Not included
- If the user does not complete sign in, the the user was on before sign in is restored (mcomella: because I'm not sure how feasible this is)

mcomella:
QA: after the user successfully signs in, the browser should update to the page the user was on before sign in: please verify this behavior. Note that the user will have to first close onboarding and then close the overlay to see the web content with the restored page.

abodea:
Verified as fixed in **`FFTV 4.5`** note that after a successfully login, I was redirected to the menu overview and the `Your tabs. Now ready for prime time` overlay prompt was displayed, and after closing it the menu overview was displayed.

PR #2898, opened 2019-09-25
Issue #2805: return user to page they started on after sign in

mcomella:
## Checklist

### DIFF
--- a/app/src/gecko/java/org/mozilla/tv/firefox/ext/EngineView.kt
+++ b/app/src/gecko/java/org/mozilla/tv/firefox/ext/EngineView.kt
@@ -10,6 +10,7 @@ import android.webkit.ValueCallback
 import android.widget.FrameLayout
 import androidx.annotation.VisibleForTesting
 import mozilla.components.concept.engine.EngineView
+import mozilla.components.support.base.log.logger.Logger
 import org.mozilla.geckoview.GeckoView
 import org.mozilla.geckoview.ScreenLength
 import org.mozilla.tv.firefox.ext.Js.CACHE_JS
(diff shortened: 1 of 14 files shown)


app/src/testSystem/java/org/mozilla/tv/firefox/helpers/MockWebBackForwardList.kt
```diff
@@ -0,0 +1,58 @@
...
+package org.mozilla.tv.firefox.helpers
+
+import android.graphics.Bitmap
+import android.webkit.WebBackForwardList
+import android.webkit.WebHistoryItem
+
+/**
+ * An implementation of the [WebBackForwardList] abstract class for testing.
+ */
+class MockWebBackForwardList(
```
Baron-Severin: SUPER optional nit: these terms are fairly fuzzy, and are frequently used interchangeably, but at least following the definition I'm aware of, this would be a 'fake' rather than a 'mock'.  https://martinfowler.com/articles/mocksArentStubs.html

app/src/gecko/java/org/mozilla/tv/firefox/ext/EngineView.kt
```diff
@@ -177,6 +180,10 @@ object WebHistory {
 val EngineView.backForwardList: WebHistory
     get() = WebHistory
 
+fun EngineView.maybeGoBackBeforeFxaSignIn() {
+    logger.warn("maybeGoBackBeforeFxaSignIn not implemented")
```
Baron-Severin: I would replace this with a crash.  If we ever move to GV, we'll want to know about this sooner rather than later.  We can always comment it out at that time if it's considered low priority.

mcomella: I was on the fence about this so sounds good to me.

app/src/testSystem/java/org/mozilla/tv/firefox/helpers/MockWebBackForwardList.kt
mcomella: I'm part of the problem 😭

PR #2906, opened 2019-10-02
Don't push LAT releases to amazon

mitchhentges:
This avoids submitting LAT apps to Amazon (the API doesn't support LAT releases, and existing automation behaviour is to create an "upcoming version"), and adjusts email content to link to the unsigned apk.

## Checklist

- [x] This is tested

### DIFF
--- a/taskcluster/ci/email/kind.yml
+++ b/taskcluster/ci/email/kind.yml
@@ -15,13 +15,19 @@ jobs:
     worker-type: succeed
     run-on-tasks-for: [github-release]
     dependencies:
+      build: build-release
       sign: 'Sign for Github'
       push: 'Push to Amazon'
     worker:
       to-address: firefox-tv@mozilla.com
-      content: |
-        Automation for this release is ready. Please:
-        * Download the APK and attach it to the [Github release](https://github.com/mozilla-mobile/firefox-tv/releases/tag/{tag})
-        * [Deploy the new release on Amazon](https://developer.amazon.com/apps-and-games/console/app/amzn1.devportal.mobileapp.7f334089688646ef8953d041021029c9/release/amzn1.devportal.apprelease.4ca3990c43f34101bf5729543343747a/general/detail)
+      content:
+        by-release-type:
+          production: |
+            Automation for this release is ready. Please:
+            * Download [the signed APK](<sign/public/build/target.apk>) and attach it to the [Github release](https://github.com/mozilla-mobile/firefox-tv/releases/tag/{tag})
+            * [Deploy the new release on Amazon](https://developer.amazon.com/apps-and-games/console/app/amzn1.devportal.mobileapp.7f334089688646ef8953d041021029c9/release/amzn1.devportal.apprelease.4ca3990c43f34101bf5729543343747a/general/detail)
+          lat: |
+            Automation for this release is ready. Please:
+            * Download [the signed APK](<sign/public/build/target.apk>) and attach it to the [Github release](https://github.com/mozilla-mobile/firefox-tv/releases/tag/{tag})
+            * Download [the unsigned APK](<build/public/build/target.apk>) and [deploy the new release on Amazon](https://developer.amazon.com/apps-and-games/console/app/amzn1.devportal.mobileapp.7f334089688646ef8953d041021029c9/release/amzn1.devportal.apprelease.4ca3990c43f34101bf5729543343747a/general/detail)
       subject: "Release {tag} is ready for deployment"
-      link-text: "{tag} APK"
--- a/taskcluster/ci/push-apk/kind.yml
+++ b/taskcluster/ci/push-apk/kind.yml
@@ -2,29 +2,27 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 ---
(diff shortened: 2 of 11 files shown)


taskcluster/firefox_tv_taskgraph/target_tasks.py
```diff
@@ -7,11 +7,27 @@
...
+        release_type = parameters["release_type"]
+
+        if release_type == "LAT":
+            def filter(task, params):
+                return tag_filter("lat", task, params)
+        else:
+            def filter(task, params):
+                return tag_filter("production", task, params)
+    else:
+        def filter(task, params):
```
mcomella: nit: I don't love that this name shadows a built-in function name.

mitchhentges: That's a good point, though I'm following the pattern from `fenix`.
I'm not super familiar with `taskgraph`, but so far it seems like the existing convention

taskcluster/ci/email/kind.yml
```diff
@@ -15,13 +15,19 @@ jobs:
     worker-type: succeed
     run-on-tasks-for: [github-release]
     dependencies:
+      build: build-release
```
JohanLorenzo: Nit: What does specifying this dependency add? As far as I understand `Sign for Github`, already depends on `build-release`, I don't think we need to depend on `build-release` explicitly.

https://github.com/mozilla-mobile/firefox-tv/blob/50e5e72e8e913750cd5d669c26002a8cd43a6678/taskcluster/ci/signing/kind.yml#L18

taskcluster/firefox_tv_taskgraph/email.py
```diff
@@ -18,23 +18,20 @@
 @transforms.add
 def email_task(config, tasks):
     tag = config.params.get("head_tag")
+    release_type = config.params.get("release_type")
 
     for task in tasks:
         to_address = task["worker"].pop("to-address")
-        content = task["worker"].pop("content")
+        content_chooser = task["worker"].pop("content")
+        content = content_chooser["by-release-type"].get(release_type, "")
```
JohanLorenzo: Nit: Use `resolve_keyed_by()`. It handles all cases for you. For instance: https://github.com/mozilla-mobile/fenix/blob/6a8f7224dc6c583f480a258ba7af796fa562e935/taskcluster/fenix_taskgraph/transforms/signing.py#L22

taskcluster/firefox_tv_taskgraph/target_tasks.py
JohanLorenzo: I've followed this pattern from mozilla-central too. I've never taken the time to understand why it's okay to shadow this built-in.

taskcluster/firefox_tv_taskgraph/target_tasks.py
```diff
@@ -7,11 +7,27 @@
...
+
+
 @_target_task('default')
 def target_tasks_default(full_task_graph, parameters, graph_config):
     """Target the tasks which have indicated they should be run on this project
     via the `run_on_projects` attributes."""
-    def filter(task, params):
-        return standard_filter(task, params)
+
+    if parameters["tasks_for"] == "github-release":
```
JohanLorenzo: Like Tom said, it's better to move this `if` statement at the beginning of the pipeline, in https://github.com/mozilla-mobile/fenix/pull/5564/files#diff-53e691d50a0b26d43d925c2150567e76R52. 

Can be done in a followup.

taskcluster/ci/signing/kind.yml
```diff
@@ -9,7 +9,7 @@ transforms:
   - taskgraph.transforms.task:transforms
 
 jobs:
-  push:
+  sign:
```
JohanLorenzo: Followup: For clarity, it'd be nice to use a `single_dep` loader. This way, you don't have to:

1. Define 1 job per build-type
2. Handle dependencies manually.

You just have a template that you apply to every task that is loaded in this kind.

You can easily filter out based on an attribute, for instance:  https://github.com/mozilla-mobile/fenix/blob/2896b36d14362598a7f8e62e311e54db17991a24/taskcluster/ci/push-apk/kind.yml#L15

Example: https://github.com/mozilla-mobile/fenix/blob/2896b36d14362598a7f8e62e311e54db17991a24/taskcluster/fenix_taskgraph/loader/single_dep.py (this loader should move up to taskgraph, I think)

taskcluster/ci/push-apk/kind.yml
```diff
@@ -10,6 +10,8 @@ transforms:
 
 jobs:
   push:
+    attributes:
+      release-type: production
```
JohanLorenzo: Followup: Same comment about `dep_loader`

taskcluster/ci/email/kind.yml
mitchhentges: Oh, but I have an artifact-reference to `<build/...>`. Doesn't that mean I need to depend on `build-release` explicitly?

JohanLorenzo: Oh, that's a good question. I'm not sure for this particular case. Anyway, let's keep it for now. `kind-dependencies` will make manual `dependencies` go away, anyway.

mitchhentges:
@mcomella this is ready to land whenever you're ready

Issue #2990, opened 2020-06-15
Update AC version to the latest (44.0.1)


psymoon:
### Vision statement / What / Requirements
The current FFTV engine support and telemetry/glean integration is not up to date.

### Impact
- https://github.com/mozilla-mobile/firefox-tv/issues/2989
- up to date engine support.

### Acceptance criteria
AC version is updated.

PR #2992, opened 2020-06-24
Closes #2989: Add deletion-request ping support for legacy_client_id

### LINKED ISSUE #2989: Update CCPA compilance for FFTV
https://docs.google.com/document/d/1s_E-hmmwg95pLTZf-dB9ZV8zkg_bGqayeTW20BI7fk4/edit

psymoon:
This is following the guideline from https://mozilla.github.io/glean/book/user/pings/deletion_request.html#the-deletion-request-ping.


app/build.gradle
```diff
@@ -125,6 +128,11 @@ configurations {
     geckoImplementation {}
 }
 
+// Generate markdown docs for the collected metrics.
+ext.gleanGenerateMarkdownDocs = true
+ext.gleanDocsDirectory = "$rootDir/docs/glean"
+apply from: 'https://github.com/mozilla-mobile/android-components/raw/v' + rootProject.ext.moz_components_version + '/components/service/glean/scripts/sdk_generator.gradle'
```
mdboom, line 134: This is no longer the supported way to use the Glean code generation plugin.  See here: 

https://mozilla.github.io/glean/book/user/adding-glean-to-your-project.html#adding-new-metrics

psymoon: @mdboom this is using legacy a-c/glean version (24.0.0), I get the following compilation error:
`Plugin with id 'org.mozilla.telemetry.glean-gradle-plugin' not found.`

mdboom: Ok.  Be aware that the `sdk_generator.gradle` script has no way to ensure that it's running the correct version of the Glean generator, so this problem of generating the wrong code could re-occur in the future.  If there's any way you could upgrade to a-c 27.0.0 or later, you would have access to the Glean Gradle plugin, which resolves that issue.

psymoon: @mdboom I think we will address this in https://github.com/mozilla-mobile/firefox-tv/issues/2990.
AFAIK, we are currently lacking QA support for big a-c upgrade. I think it's safer to upgrade a-c version to 24 for `deletion-request ping` support.

psymoon:

This data will be used to delete legacy telemetry data associated with the id when they opt out of telemetry.

This allows us to comply with user privacy regulations.

Existing legacy telemetry does not currently have the ability to send a deletion request ping and this was the easiest way to get that information

N/A

| Measurement Description                      | Data Collection Category | Tracking Bug #                                               |
| -------------------------------------------- | ------------------------ | ------------------------------------------------------------ |
| Legacy telemetry client id  | Category 1                   | [Issue 2989](https://github.com/mozilla-mobile/firefox-tv/issues/2989)

6) How long will this data be collected?

For the lifetime of the application, owned by the amazon-dev team.

* All users with telemetry enabled, on release channel.

There is an option in settings to stop sending usage data that will be used as the opt-out mechanism.

This will be analyzed with the normal Glean methods such as GUD and GLAM.

Internally with Mozilla, through normal Glean reporting methods.

app/build.gradle
psymoon: @mcomella what do you think about this ^?

mcomella: Since we're only expecting to build one build from this in the next few weeks, I'd think this is probably fine to move forward with this.

@mdboom Do you agree?

@psymoon If mdboom agrees, can you file an issue about "upgrade to a-c 27 and use glean plugin" and talk to athomasmoz to ensure it's at the top of the backlog? It should be the first thing we do before adding more code (or at least validate that it won't break). Ideally, we'll upgrade to the latest version, but if not, we should do that issue.

app/build.gradle
```diff
@@ -236,7 +248,7 @@ dependencies {
     implementation "com.google.android.gms:play-services-oss-licenses:$oss_licenses_version"
 }
 
-task validateReleaseConfiguration(type: ValidateAndroidAppReleaseConfiguration)
+// task validateReleaseConfiguration(type: ValidateAndroidAppReleaseConfiguration)
```
mcomella, line 251: nit: can you add a comment explaining why we do this?

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketVideoFetchScheduler.kt
```diff
@@ -25,15 +26,16 @@ private const val FETCH_UNIQUE_WORK_NAME = "PocketFetch"
  * Schedules background fetches of the Pocket video data.
  */
 class PocketVideoFetchScheduler(
-    private val isPocketEnabledByLocale: () -> Boolean
+    private val isPocketEnabledByLocale: () -> Boolean,
+    private val context: Context
```
mcomella, line 30: nit: super-nit: `context` seems to come as the first method argument in all Android functions that use it so I try to follow that

mcomella, line 30: nit: -> `context` -> `appContext`

app/src/main/java/org/mozilla/tv/firefox/utils/ServiceLocator.kt
```diff
@@ -93,7 +93,7 @@ open class ServiceLocator(val app: Application) {
     @Suppress("DEPRECATION") // We need PocketEndpointRaw until we move to a-c's impl.
     val pocketEndpointRaw by lazy { PocketEndpointRaw(appVersion, buildConfigDerivables.globalPocketVideoEndpoint) }
     val pocketVideoStore by lazy { PocketVideoStore(app, app.assets, pocketVideoJSONValidator) }
-    val pocketVideoFetchScheduler by lazy { PocketVideoFetchScheduler(isPocketEnabledByLocale) }
+    val pocketVideoFetchScheduler by lazy { PocketVideoFetchScheduler(isPocketEnabledByLocale, app.applicationContext) }
```
mcomella, line 96: nit: I think it's safe to pass in `app` as the context, as we do in other places, since the type is `Application` and not `Activity`

app/build.gradle
mdboom: That makes sense.  Thanks for checking.

psymoon: Filed https://github.com/mozilla-mobile/firefox-tv/issues/2993.

app/build.gradle
```diff
@@ -4,6 +4,9 @@
 
 import org.mozilla.gradle.tasks.ValidateAndroidAppReleaseConfiguration
 
+plugins {
+    id "com.jetbrains.python.envs" version "0.0.26"
```
mcomella, line 8: Why is this needed? Is this used by the glean script?

app/build.gradle
```diff
@@ -216,6 +224,10 @@ dependencies {
     androidTestImplementation 'com.squareup.okhttp3:mockwebserver:3.10.0'
     testImplementation 'com.squareup.okhttp3:mockwebserver:3.10.0'
 
+    // For the initial release of Glean 19, we require consumer applications to
+    // depend on a separate library for unit tests. This will be removed in future releases.
+    testImplementation "org.mozilla.telemetry:glean-forUnitTests:21.3.0"
```
mcomella, line 229: Should this be 24.0.0? Same as `ext.moz_components_version`?

app/build.gradle
mdboom: Yep.  Related docs, FYI: https://mozilla.github.io/glean/book/user/adding-glean-to-your-project.html#adding-new-metrics

app/build.gradle
mdboom: No, this is correct.  This is the version of Glean that ships with a-c 2.4.0.0.  This is the gotcha I was talking about that the Glean Gradle plugin solves (handling this mapping automatically).  Unfortunately, the Glean Gradle plugin isn't available prior to a-c 27.0.0.

app/src/main/java/org/mozilla/tv/firefox/pocket/PocketVideoFetchScheduler.kt
psymoon: This will be removed since we are deprecating pocket stuff anyway

psymoon: This will be removed since we are deprecating pocket stuff anyway

app/src/main/java/org/mozilla/tv/firefox/utils/ServiceLocator.kt
psymoon: This will be removed since we are deprecating pocket stuff anyway

mcomella:
> I didn't know how to verify the legacy client ID was set so I didn't verify that

I reverified using the glean debug ping viewer and saw a legacy client ID was set. However, I didn't verify it was the same one as on the device – I'm going to assume that is correct.

PR #3010, opened 2020-07-10
Bump the version to 4.6.1

psymoon:


CHANGELOG.md
```diff
@@ -9,7 +9,7 @@ It diverges in the following ways:
 
 ## Unreleased
 
-## 4.6 - 2019-11-25
+## 4.6.1 - 2019-11-25
```
mcomella, line 12: nit: you shouldn't replace 4.6. You should add a new release for `4.6.1`, probably setting the date to monday, and explain what happened in that release, like https://github.com/mozilla-mobile/firefox-tv/blob/master/CHANGELOG.md#35-b---2019-04-16

PR #3014, opened 2020-07-14
Closes #3013: Reset the legacy-id whenever collectionEnabled is on

### LINKED ISSUE #3013: deletion-request ping is missing legacy ID if app is started with data disabled
## Steps to reproduce
- Start app
- Disable data upload
- Force stop app
- Start app
- Trigger Glean Debug Activity uploads: `adb shell am start -n org.mozilla.tv.firefox/mozilla.telemetry.glean.debug.GleanDebugActivity --es tagPings connect`
- Enable data upload
- Disable data upload

### Expected behavior
Legacy client ID appears in deletion-request ping data on dashboard

### Actual behavior
Legacy client ID is not in deletion-request ping. For example:
image

### Device information
* Fire TV device: ?
* Latest affected Firefox version: up/master 014894a46c7e764c8e7baa2274e2662d635f0bce
* Turbo Mode: ?

psymoon:
## Checklist

### DIFF
--- a/app/src/main/java/org/mozilla/tv/firefox/FirefoxApplication.kt
+++ b/app/src/main/java/org/mozilla/tv/firefox/FirefoxApplication.kt
@@ -110,15 +110,17 @@ open class FirefoxApplication : LocaleAwareApplication() {
                 // sending startup data, or 2) sending even when the user has toggled
                 // off data collection
                 Glean.setUploadEnabled(collectionEnabled)
+                if (collectionEnabled) {
+                    LegacyIds.clientId.set(UUID.fromString(TelemetryIntegration.INSTANCE.clientId))
+                }
             }
         }
     }
 
     private fun initGlean() {
         setGleanUpload()
+        LegacyIds.clientId.set(UUID.fromString(TelemetryIntegration.INSTANCE.clientId))
         Glean.initialize(applicationContext, Configuration(channel = BuildConfig.BUILD_TYPE))
-        val legacyId = TelemetryIntegration.INSTANCE.clientId
-        LegacyIds.clientId.set(UUID.fromString(legacyId))
     }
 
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mozilla/tv/firefox/FirefoxApplication.kt
```diff
@@ -110,15 +110,17 @@ open class FirefoxApplication : LocaleAwareApplication() {
                 // sending startup data, or 2) sending even when the user has toggled
                 // off data collection
                 Glean.setUploadEnabled(collectionEnabled)
+                if (collectionEnabled) {
```
st3fan, line 113: Is this code here actually needed? The legacy client id is already set in `initGlean` - why is that not enough?

I don't think this code will actually do change anything. If collectionEnabled is true then the legacy id will be set to the exact same value as in `initGlean` already did? What do I miss?

psymoon: @st3fan so the problem is that when the `collectionEnabled` is set to false on app start, `LegacyId.clientId` is not set (which I think is by design because you don't want to send metric pings when `collectionEnabled` is false). So as of now, if `collectionEnabled` is false on app start, then setting the `collectionEnabled` to true won't guarantee `LegacyId.clientId` to be set until you restart the app.

The code block here is observed stream so whenever collectionEnabled is set to true, we make sure the `LegacyId.clientId` is set. Better way to do this is to have `SharedPreference` hold an one-time flag for `LegacyId.clientId` and have the condition like:
```
if (collectionEnabled && !legacyIdSet) {
```

There is an edge case where the `collectionEnabled` is false from the beginning and never set to true. This would theoretically have missing `LegacyId.clientId` as well.

PR #3046, opened 2021-02-23
Implement a generic in-app message banner

st3fan:
## Checklist


app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/NavigationOverlayFragment.kt
```diff
@@ -29,12 +32,9 @@ import io.reactivex.android.schedulers.AndroidSchedulers
...
 import io.reactivex.rxkotlin.addTo
-import kotlinx.android.synthetic.main.fragment_navigation_overlay_orig.channelsContainer
-import kotlinx.android.synthetic.main.fragment_navigation_overlay_orig.navUrlInput
-import kotlinx.android.synthetic.main.fragment_navigation_overlay_orig.settingsTileContainer
-import kotlinx.android.synthetic.main.fragment_navigation_overlay_top_nav.exitButton
-import kotlinx.android.synthetic.main.fragment_navigation_overlay_top_nav.fxaButton
-import kotlinx.android.synthetic.main.hint_bar.hintBarContainer
+import kotlinx.android.synthetic.main.fragment_navigation_overlay_orig.*
+import kotlinx.android.synthetic.main.fragment_navigation_overlay_top_nav.*
+import kotlinx.android.synthetic.main.hint_bar.*
```
st3fan, line 37: This is a bit odd but to satisfy the linter. I'm not sure how the code originally got through the linter since it was definitely failing on the wildcards.

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/NavigationOverlayFragment.kt
```diff
@@ -170,6 +170,18 @@ class NavigationOverlayFragment : Fragment() {
         ).apply {
             onCreateView(view)
         }
+
+        val bannerLayout: View = view.findViewById(R.id.bannerLayout)
+
+        val bannerMoreInfoButton: Button = bannerLayout.findViewById(R.id.bannerMoreInfoButton)
+        bannerMoreInfoButton.setOnClickListener {
+            (activity as MainActivity).onNonTextInputUrlEntered("https://www.mozilla.org")
```
st3fan, line 178: This hard coded URL needs to be replaced at a later point.

app/src/main/res/values/strings.xml
```diff
@@ -295,4 +295,9 @@
     <string name="update_prompt_primary_button">Update</string>
     <!-- Button text to dismiss the prompt to update the app. -->
     <string name="update_prompt_secondary_button">Not now</string>
+
+    <!-- Text for the banner. -->
+    <string name="banner_text"><![CDATA[<b>Important Message:</b> This is a banner with some content that will wrap over two lines. That was line one and this is line two. Something something blah la la blah.]]></string>
+    <string name="banner_more_info_button">More Info</string>
+    <string name="banner_dismiss_button">Dismiss</string>
```
st3fan, line 302: This text is not final. For in-app messages we will pull the text from a different source probably. This is just for validation right now.

app/src/main/java/org/mozilla/tv/firefox/navigationoverlay/NavigationOverlayFragment.kt
mcomella: We can land it as is but it'd be ideal to define it as a constant so we don't forget to update it in both places.

mcomella (approved):
Looks good to me but we should consider accessibility as a follow-up. I added a few more thoughts but I'm not very concerned about them.

app/src/main/res/layout/fragment_browser.xml
```diff
@@ -14,6 +14,52 @@
...
 
+        <LinearLayout
+            android:id="@+id/bannerLayout"
+            android:layout_width="match_parent"
+            android:layout_height="wrap_content"
+            android:orientation="horizontal"
+            android:background="#FEBD4F"
+            android:gravity="center">
+
+          <TextView
```
mcomella, line 25: Ah, a few localization concerns:
- should we specify a max line count (or height) for this view? In a localized language, the text may overflow the banner, in the worst case covering the entire screen
- Similarly, in localized languages the dismiss and more info buttons could grow very large horizontally, further pushing this content down (or possibly hiding it entirely). Should we specify a max width for these buttons?

mcomella (commented):
A few thoughts are localization that actually are more concerning.

